PR #27, opened 2015-04-22
Fix another NodeIterator removing step

Joris-van-der-Wel:
This change makes it more consistent with the other rules. Also,
it is what the w3c test suite expects and how the browsers behave.

Here's a quick snippit to test:
```javascript
document.body.innerHTML = "<p>Efghijkl</p><p>Mnopqrst</p>";
var it = document.createNodeIterator(document);
while (it.nextNode());

var removed = document.body.children[1];
removed.parentNode.removeChild(removed);

console.info(it.referenceNode === document.body.children[0].lastChild);
console.info(it.pointerBeforeReferenceNode === false);
```

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8279,9 +8279,9 @@ Each {{NodeIterator}} object has these
   </ol>
 
  <li>Set the {{NodeIterator/referenceNode}} attribute
- to first <a>node</a>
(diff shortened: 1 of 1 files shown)


annevk:
So if I want to integrate this and preserve authorship I need your email address...

domenic:
I always use .patch URLs for that

PR #36, opened 2015-05-22
Add [TreatNullAs=EmptyString] to Attr.nodeValue/textContent

foolip:
This came up in an attempt to align Blink with the spec:
https://codereview.chromium.org/1146393003/

By adding [TreatNullAs=EmptyString], the spec would match current
Firefox and IE behavior. It probably doesn't matter, but is a quick
path to interop and less risky.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6011,8 +6011,8 @@ interface Attr {
   readonly attribute DOMString localName;
   readonly attribute DOMString name;
            attribute DOMString value;
-           attribute DOMString nodeValue; // legacy alias of .value
-           attribute DOMString textContent; // legacy alias of .value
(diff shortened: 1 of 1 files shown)


annevk:
We discussed this before. `textContent` is nullable and therefore cannot have this annotation. (It needs to return null.)

foolip:
Actually neither Attr.nodeValue nor Attr.textContent are nullable, but Node.nodeValue and Node.textContent are. If the Attr ones are made nullable again, it would have to be defined what getAttribute() returns after setting them to null.

foolip:
Ping?

annevk:
Closed by https://github.com/whatwg/dom/commit/df9901c05ea6a966925cc2ec110c9e8779ca4526

Issue #42, opened 2015-06-10
Question about references after Bikeshed conversion


ArkadiuszMichalski:
This is intentional that for every constructors/methods/algos you add all arguments to id in definition? 

setAttribute(name, value) 
https://dom.spec.whatwg.org/#dom-element-setattributename-value
but was 
http://dom.spec.whatwg.org/#dom-element-setattributenode

Omg now all references not work, the same is for URL spec but not for Fetch and Encoding :(

domenic:
To fix this you can use Bikeshed's `oldids=""` feature. It is sad that they changed.

tabatkins:
I can make this smarter, but changes will of course change *other* specs. But I doubt people really care, so I'm open to changing.

The problem is that if you have multiple overrides of a method, I need to have disambiguating IDs, so I took the conservative route and produced guaranteed-unambiguous IDs.  I can instead look for collisions and generate simple IDs when there aren't any.

ArkadiuszMichalski:
Hmm this looks ugly, first argument is connected to name without any separator:
https://dom.spec.whatwg.org/#dom-element-getelementsbytagnamensnamespace-localname 

But problem is that any pages that use references for this spec now not work, not only other specs but of course any pages, like MDN, etc. It would be good if changing ids format does not happen often.

domenic:
Yes, I think we should definitely restore the old IDs. The question is whether we do that by adding `oldids=""` for everything or get Bikeshed to do the work for us :)

tabatkins:
Get Bikeshed to do the work for you.

annevk:
If you have overloads you should still only have one method definition and one ID I think.

tabatkins:
No?  You have two method definitions, like:

```
interface Foo {
  DOMString bar(DOMString stringBaz);
  long bar(long numericBaz);
};
```

ArkadiuszMichalski:
Maybe add some less invasive method for colisions, like foo-bar,  foo-bar-2, foo-bar-3 or other format, but don't destroy all referrences what we have now. I don't remeber that see any colision in DOM, URL or Notification sepc so maybe for first def we can have old format but for next colision you can use this new style?

tabatkins:
Yes, I've already agreed that I need to fix non-overload IDs in Bikeshed.

annevk:
[...]

PR #46, opened 2015-06-25
`document.importNode` and `document.adoptNode` should throw exceptions when acting upon ShadowRoots

samgiles:
The `importNode` algorithm relies on a `clone` of `node` however, the Shadow
DOM spec stipulates that cloning a `ShadowRoot` should raise a
`DataCloneError`. Consequently, importing a `ShadowRoot` should
be unsupported.

Additionally, it doesn't make sense for a document to `adopt` a `ShadowRoot` -
the shadow host and `ShadowRoot` should not have different documents.  In this
instance a `HierarchyRequestError` seems to make sense.

See https://github.com/w3c/webcomponents/issues/125

Using `bikeshed`:

Note:  I ran `bikeshed update && bikeshed spec` however the updated `dom.html` had spurious other updates to anchors unrelated to this change - first time using `bikeshed`, have I done something wrong here?

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -102,6 +102,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: effective script origin
             text: origin alias; url: #concept-origin-alias
             text: Unicode serialization of an origin
+urlPrefix: https://w3c.github.io/webcomponents/spec/shadow/
+	type: dfn; urlPrefix: #dfn-
+		text: shadow root
 url: http://www.w3.org/TR/DOM-Parsing/#widl-Range-createContextualFragment-DocumentFragment-DOMString-fragment
     type: method; text: createContextualFragment(); for: Range;
 type: interface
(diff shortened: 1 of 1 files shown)


annevk:
You ran into #42 I think. I'll get to this next week.

samgiles:
Indeed that was the issue.

dom.bs
```diff
@@ -102,6 +102,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: effective script origin
             text: origin alias; url: #concept-origin-alias
             text: Unicode serialization of an origin
+urlPrefix: http://w3c.github.io/webcomponents/spec/shadow/
```
annevk: This should use HTTPS.

annevk:
It seems the reason it does not make sense to adopt a `ShadowRoot` is because it is always attached to a host.

We should probably inline that cloning throws as well, but that can be done separately.

annevk:
Thanks for doing this by the way!

samgiles:
No problem at all! :)

> We should probably inline that cloning throws as well, but that can be done separately.

Do you mean cloning a `ShadowRoot`? This is covered in the `ShadowRoot` spec.

annevk:
> Do you mean cloning a `ShadowRoot`? This is covered in the Shadow DOM spec.

Yes, but at some point we need to eliminate the monkey patching of DOM.

PR #47, opened 2015-06-30
Less ambiguous step description for Node.before

frewsxcv:
Before:

If viablePreviousSibling is non-null, set it to viablePreviousSibling’s
next sibling, and to parent’s first child otherwise.

After:

If viablePreviousSibling is non-null, set node to viablePreviousSibling’s
next sibling. Otherwise, set node to parent’s first child.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2120,9 +2120,9 @@ invoked, must run these steps:
  <li><p>Let <var>node</var> be the result of
  <a>converting <var>nodes</var> into a node</a>.
 
- <li><p>If <var>viablePreviousSibling</var> is non-null, set it to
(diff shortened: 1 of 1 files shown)


annevk:
What exactly is ambiguous? The commit looks wrong, _node_ should not be changed.

frewsxcv:
> node should not be changed.

What do you mean by this?

frewsxcv:
> If viablePreviousSibling is non-null, set it to viablePreviousSibling’s
next sibling

What does 'it' refer to here?

annevk:
_viablePreviousSibling_.

frewsxcv:
Why does the previous sibling's next sibling get set to itself?

PR #60, opened 2015-07-29
Queuing of mutation records looks inconsistent

nox:
To [replace](https://dom.spec.whatwg.org/#concept-node-replace) a *child* with *node* within a parent, the mutation record is queued after the removal of *child* and insertion of *node*:
<ol start=11>
<li><strong>Remove</strong> child from its parent with the suppress observers flag set.
<li>Let nodes be node’s children if node is a DocumentFragment node, and a list containing solely node otherwise.
<li><strong>Insert</strong> node into parent before reference child with the suppress observers flag set.
<li><strong>Queue a mutation record</strong> of "childList" for target parent with addedNodes nodes, removedNodes a list solely containing child, nextSibling reference child, and previousSibling previousSibling.
</ol>

That behaviour is the same to [replace all](https://dom.spec.whatwg.org/#concept-node-replace-all) a node within a parent:
<ol start=4>
<li><strong>Remove</strong> all parent’s children, in tree order, with the suppress observers flag set.
<li>If node is not null, <strong>insert</strong> node into parent before null with the suppress observers flag set.
<li><strong>Queue a mutation record</strong> of "childList" for parent with addedNodes addedNodes and removedNodes removedNodes.
</ol>

But for example, to [insert](https://dom.spec.whatwg.org/#concept-node-replace-all) a node into a parent before a child, mutation records are queued before the actual mutations:
<ol start=4>
<li>If node is a DocumentFragment node, <strong>queue a mutation record</strong> of "childList" for node with removedNodes nodes.
<li>If node is a DocumentFragment node, <strong>remove</strong> its children with the suppress observers flag set.
</ol>

Shouldn't all mutation records be queued after the tree was mutated? I understand that because micro tasks are used it doesn't matter much, but the ordering inconsistency makes it look like it actually matters.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1483,30 +1483,19 @@ into a <var>parent</var> before a <var>child</var>, with an optional
  a {{DocumentFragment}} <a>node</a>, and a
  list containing solely <var>node</var> otherwise.
 
- <li>
-  If <var>node</var> is a {{DocumentFragment}}
-  <a>node</a>, <a>queue a mutation record</a>
-  of "<code>childList</code>" for <var>node</var> with removedNodes
-  <var>nodes</var>.
-
-  Note: This step intentionally does not pay attention to the
-  <i>suppress observers flag</i>.
-
  <li>If <var>node</var> is a {{DocumentFragment}}
  <a>node</a>,
  <a>remove</a> its
  <a>children</a> with the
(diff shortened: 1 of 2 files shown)


annevk:
Thank you for your PR. I ended up removing a redundant newline before committing and did not want to bother you with that. Hence the distinct commit.

PR #70, opened 2015-09-04
Run latest bikeshed

RByers:
I just ran `curl https://api.csswg.org/bikeshed/ -F file=@dom.bs > dom.html` without making any changes to the .bs file.  I figure you probably want this landed independently from any actual spec changes, right?

### DIFF
--- a/dom.html
+++ b/dom.html
@@ -67,9 +67,9 @@
  </head>
  <body class="h-entry status-LS">
   <div class="head">
-   <p data-fill-with="logo"><a class="logo" href="https://whatwg.org/"> <img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-dom.svg"> </a></p>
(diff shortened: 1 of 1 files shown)


annevk:
Yeah, this is #65 and per http://krijnhoetmer.nl/irc-logs/whatwg/20150904#l-678 there's other regressions. I recommend just patching `dom.bs` for now.

RByers:
The spacing issue is now fixed (and it's pretty annoying IMHO).  From a quick glance at the [new output](https://rawgit.com/RByers/dom/bikeshed-update/dom.html) it looks like a net improvement to me.  Worth landing?

annevk:
Did you look at the link I posted? I don't want to suddenly reference UI Events for classes that are defined by DOM.

RByers:
Oh sorry, I missed that.  Ok, I'll leave this alone and submit patches only against the bs file for now (but maybe host a version of the html with my changes elsewhere in case people want to see it).

annevk:
That sounds great!

Issue #41, opened 2015-06-10
Attributes need case-insensitive matching on getting


annevk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1165851#c23 for what is (likely, hopefully someone confirms my analysis) wrong with the current specification.

annevk:
Note, the way this seems to be solved is simply lowercasing the argument to `createAttribute()`. That's certainly much more attractive.

Ms2ger:
https://github.com/w3c/web-platform-tests/pull/2198
https://github.com/w3c/web-platform-tests/pull/2203

PR #75, opened 2015-09-17
Make createAttribute() lowercase the input in HTML documents

foolip:
Fixes https://github.com/whatwg/dom/issues/41

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4670,12 +4670,12 @@ method, when invoked, must run these steps:
  production in XML, <a>throw</a> an
  {{InvalidCharacterError}} exception.
 
+ <li>If the <a>context object</a> is an <a>HTML document</a>, let
+ <var>localName</var> be <a>converted to ASCII lowercase</a>.
+
  <li>Return a new <a>attribute</a> whose
  <a for=Attr>local name</a> is <var>localName</var>.
 </ol>
-
-<p class="note no-backref">This method does not have its input
-<a>converted to ASCII lowercase</a>.
 
 The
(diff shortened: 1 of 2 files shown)


foolip:
PTAL. I'm not sure if the argument name should be changed now that it isn't necessarily the verbatim localName, I just left it alone.

annevk:
Oh I completely missed this. So Gecko now implements this simple behavior and Chrome does too?

foolip:
No, it hasn't been implemented yet, but based on the use counters I added in Blink and the telemetry in Gecko, we seem to have agreed that this is the approach we should try. See https://bugzilla.mozilla.org/show_bug.cgi?id=1176313

foolip:
If you want to wait with the spec change until it's been shipped, that's fine, it's just not the usual order.

foolip:
@bzbarsky, are you now attempting to implement what the spec says plus the change in this PR in Gecko, or what is the whole "package" or changes being attempted at the same time?

bzbarsky:
@foolip I've landed the change in this PR in Gecko nightlies.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1205364

I'm also attempting to reland the rest of the "follow the spec" changes; that's being tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1176313 and I expect to check it in sometime today.

foolip:
That sounds great, @bzbarsky! If those changes make it successfully to Firefox stable, I'd like to attempt the same spec sync in Blink.

PR #82, opened 2015-09-22
Add EventListenerOptions and passive event listeners

RByers:
My first crack at incorporating the ideas and discussion from my [EventListenerOptions proposal](https://github.com/RByers/EventListenerOptions) into the spec.  The result can be previewed [here](https://rawgit.com/RByers/dom/event-listener-options/dom.html).

One thing I'm still not super happy with is the somewhat confusing defaults for `mayCancel`.  On the one hand we [agreed](https://github.com/RByers/EventListenerOptions/issues/17) that we wanted the performance behavior (`mayCancel:false`) the be the default when using the new API form.  But we of course can't [change the existing behavior](https://github.com/RByers/EventListenerOptions/issues/22).  Let's discuss that further [here](https://github.com/RByers/EventListenerOptions/issues/17) if desired.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -104,6 +104,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: effective script origin
             text: origin alias; url: #concept-origin-alias
             text: Unicode serialization of an origin; url: #unicode-serialisation-of-an-origin
+        urlPrefix: infrastructure.html
+            text: in parallel
 urlPrefix: https://w3c.github.io/webcomponents/spec/shadow/
 	type: dfn; urlPrefix: #dfn-
 		text: shadow root
@@ -613,7 +615,7 @@ Lets look at an example of how <a>events</a> work in a <a>tree</a>:
    function test(e) {
      debug(e.target, e.currentTarget, e.eventPhase)
    }
-   document.addEventListener("hey", test, true)
+   document.addEventListener("hey", test, {capture: true})
    document.body.addEventListener("hey", test)
    var ev = new Event("hey", {bubbles:true})
    document.getElementById("x").dispatchEvent(ev)
@@ -727,17 +729,13 @@ inherits from the {{Event}} interface.
  {{Event/preventDefault()}} method.
 
  <dt><code><var>event</var> . <a method for=Event lt="preventDefault()">preventDefault</a>()</code>
- <dd>If invoked when the
- {{Event/cancelable}} attribute value is true,
- signals to the operation that caused <var>event</var> to be
- <a>dispatched</a> that it needs to be
- canceled.
+ <dd>If invoked when the {{Event/cancelable}} attribute value is true, and while executing a
+ listener for the <var>event</var> with {{EventListenerOptions/passive}} set to false, signals to
+ the operation that caused <var>event</var> to be <a>dispatched</a> that it needs to be canceled.
 
  <dt><code><var>event</var> . {{Event/defaultPrevented}}</code>
- <dd>Returns true if
- {{Event/preventDefault()}} was invoked
- while the {{Event/cancelable}} attribute
- value is true, and false otherwise.
+ <dd>Returns true if {{Event/preventDefault()}} was invoked successfully to indicate cancellation,
+ and false otherwise.
 
  <dt><code><var>event</var> . {{Event/isTrusted}}</code>
  <dd>Returns true if <var>event</var> was
@@ -799,6 +797,7 @@ flags that are all initially unset:
  <li><dfn export for=Event>canceled flag</dfn>
  <li><dfn export for=Event>initialized flag</dfn>
  <li><dfn export for=Event>dispatch flag</dfn>
+ <li><dfn export for=Event>in passive listener flag</dfn>
 </ul>
 
 The
@@ -816,8 +815,13 @@ must return the values they were initialized to.
 
 The
 <dfn method for=Event>preventDefault()</dfn>
-method must set the <a>canceled flag</a> if the
-{{Event/cancelable}} attribute value is true.
+method must set the <a>canceled flag</a> if the {{Event/cancelable}} attribute value is true and
+the <a>in passive listener flag</a> is unset.
+
+<p class="note no-backref">
+ This means there are scenarios where invoking {{preventDefault()}} has no effect. User agents are
+ encouraged to log the precise cause in a developer console, to aid debugging.
+</p>
 
 The
 <dfn attribute for=Event>defaultPrevented</dfn>
@@ -972,13 +976,18 @@ for historical reasons.
 <pre class=idl>
 [Exposed=(Window,Worker)]
 interface EventTarget {
-  void addEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
-  void removeEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
+  void addEventListener(DOMString type, EventListener? callback, optional (EventListenerOptions or boolean) options);
+  void removeEventListener(DOMString type, EventListener? callback, optional (EventListenerOptions or boolean) options);
   boolean dispatchEvent(Event event);
 };
 
 callback interface EventListener {
   void handleEvent(Event event);
+};
+
+dictionary EventListenerOptions {
+  boolean capture;
+  boolean passive;
 };
 </pre>
 
@@ -991,34 +1000,45 @@ occurred. Each {{EventTarget}} has an associated list of
 <p>An <dfn export id=concept-event-listener>event listener</dfn> can be used to observe a specific
 <a>event</a>.
 
-<p>An <a>event listener</a> consists of a <b>type</b>, <b>callback</b>, and <b>capture</b>. An
-<a>event listener</a> also has an associated <b>removed flag</b>, which is initially unset.
+<p>An <a>event listener</a> consists of a <b>type</b>, <b>callback</b>, <b>capture</b> and
+<b>passive</b>. An <a>event listener</a> also has an associated <b>removed flag</b>, which is
+initially unset.
 
 <p class="note no-backref">The callback is named {{EventListener}} for historical reasons. As can be
 seen from the definition above, an <a>event listener</a> is a more broad concept.
 
 <dl class=domintro>
- <dt><code><var>target</var> . <a method lt="addEventListener()">addEventListener</a>(<var>type</var>, <var>callback</var> [, <var>capture</var> = false])</code>
+ <dt><code><var>target</var> . <a method lt="addEventListener()">addEventListener</a>(<var>type</var>, <var>callback</var> [, <var>options</var>])</code>
  <dd>
   Appends an <a>event listener</a> for <a>events</a> whose {{Event/type}} attribute value
   is <var>type</var>. The <var>callback</var> argument sets the <b>callback</b> that will
-  be invoked when the <a>event</a> is <a>dispatched</a>. When set to true,
-  the <var>capture</var> argument prevents <b>callback</b> from being invoked when
-  the <a>event</a>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}}.
-  When false, <b>callback</b> will not be invoked when <a>event</a>'s {{Event/eventPhase}}
-  attribute value is {{Event/CAPTURING_PHASE}}. Either way, <b>callback</b> will be
-  invoked if <a>event</a>'s {{Event/eventPhase}} attribute value is {{Event/AT_TARGET}}.
+  be invoked when the <a>event</a> is <a>dispatched</a>.
 
-  The <a>event listener</a> is appended to <var>target</var>'s list of
-  <a>event listeners</a> and is not appended if it is a duplicate, i.e., having the same
-  <b>type</b>, <b>callback</b>, and <b>capture</b> values.
+  The <var>options</var> argument sets listener-specific options. For compatibility this can be just
+  a boolean, in which case the method behaves exactly as if the value was specified as
+  <var>options</var>' <code>capture</code> member.
 
- <dt><code><var>target</var> . <a method lt="removeEventListener()">removeEventListener</a>(<var>type</var>, <var>callback</var> [, <var>capture</var> = false])</code>
+  When set to true, <var>options</var>' <code>capture</code> member prevents <b>callback</b> from
+  being invoked when the <a>event</a>'s {{Event/eventPhase}} attribute value is
+  {{Event/BUBBLING_PHASE}}. When false (or not present), <b>callback</b> will not be invoked when
+  <a>event</a>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}. Either way,
+  <b>callback</b> will be invoked if <a>event</a>'s {{Event/eventPhase}} attribute value is
+  {{Event/AT_TARGET}}.
+
+  When set to true, <var>options</var>' <code>passive</code> member indicates that the
+  <b>callback</b> will not cancel the event by invoking {{preventDefault()}}. This is used to enable
+  performance optimizations described in [[#observing-event-listeners]].
+
+  The <a>event listener</a> is appended to <var>target</var>'s list of <a>event listeners</a> and is
+  not appended if it is a duplicate, i.e., having the same <b>type</b>, <b>callback</b>,
+  <b>capture</b> and <b>passive</b> values.
+
+ <dt><code><var>target</var> . <a method lt="removeEventListener()">removeEventListener</a>(<var>type</var>, <var>callback</var> [, <var>options</var>])</code>
  <dd>Remove the <a>event listener</a>
  in <var>target</var>'s list of
  <a>event listeners</a> with the same
  <var>type</var>, <var>callback</var>, and
- <var>capture</var>.
+ <var>options</var>.
 
  <dt><code><var>target</var> . <a method lt="dispatchEvent()">dispatchEvent</a>(<var>event</var>)</code>
  <dd><a>Dispatches</a> a synthetic event <var>event</var> to <var>target</var> and returns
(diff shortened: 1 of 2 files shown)


annevk:
How does this address e.g., https://github.com/RByers/EventListenerOptions/issues/20? Also, it's very confusing to still have outstanding discussion elsewhere. Not really sure what to pay attention to now.

annevk:
Basically, the way this is written I don't see much room for optimizations, unless you're doing something that's not allowed.

RByers:
It's addressed by what [@smaug--- said](https://github.com/RByers/EventListenerOptions/issues/20#issuecomment-137698318).  I.e. whether or not a particular event is `cancelable` is not the purview of DOM, it's up to the individual specs.  So a handler being added 'too late' is the same as other scenarios where a non-cancelable event is received by a handler, so already covered by the DOM spec.  If we can get consensus here, then I'll add something to touch-events (w3c/touch-events/#6) that indicates that a TouchEvent may be uncancelable if, at the time it was first received, there are no mayCancel=true handlers for it.  WDYT?  Is there something more I should add here (perhaps expand on the note?).

Regarding discussion, I thought there were enough separate non-trivial issues (with non-trivial discussion history) that it was valuable to continue to use a separate issue tracker for them.  But if you prefer I can try to close them all and merge all discussion into this PR.  I don't have a preference on whether minor editorial discussion occurs here or in separate issues.  But I defer to you on all of this - what's easiest for you at this stage?

annevk:
Well, what I said in https://github.com/RByers/EventListenerOptions/issues/20#issuecomment-137699314 is that what touch events do today doesn't require any changes in DOM. What you are suggesting here is that they observe listeners, which is something we want to avoid.

I don't really care where we settle on the design, but I'd like to settle on the design first. (I asked for the processing model before because the design wasn't clear. Now the processing model is clear, but the design seems wrong.)

RByers:
Ok, I suggest we continue this conversation in RByers/EventListenerOptions#20 since it's (IMHO) a continuation of the same debate.

annevk:
@RByers what is the plan for this feature?

RByers:
@annevk Sorry for the delay, we're still working on it (I've just bee caught up in some other things).  @dtapuska is actively working on the blink implementation.  I expect to be able to make the spec work here my top priority in a couple weeks.

RByers:
Ok @annevk, I've finally reworked this pull request according to your suggestions in  RByers/EventListenerOptions#20.  WDYT?

dom.bs
```diff
@@ -990,34 +1004,43 @@ occurred. Each {{EventTarget}} has an associated list of
...
   is <var>type</var>. The <var>callback</var> argument sets the <b>callback</b> that will
-  be invoked when the <a>event</a> is <a>dispatched</a>. When set to true,
-  the <var>capture</var> argument prevents <b>callback</b> from being invoked when
+  be invoked when the <a>event</a> is <a>dispatched</a>.
+
+  The <var>options</var> argument sets listener-specific options.  For compatibility this can be
+  just a boolean, in which case it determines the value of the <b>capture</b> option.
+
+  When set to true,
+  the <var>capture</var> option prevents <b>callback</b> from being invoked when
```
annevk: `<var>capture</var>` is no longer a variable so this seems wrong.

dom.bs
```diff
@@ -990,34 +1004,43 @@ occurred. Each {{EventTarget}} has an associated list of
...
   the <a>event</a>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}}.
   When false, <b>callback</b> will not be invoked when <a>event</a>'s {{Event/eventPhase}}
   attribute value is {{Event/CAPTURING_PHASE}}. Either way, <b>callback</b> will be
   invoked if <a>event</a>'s {{Event/eventPhase}} attribute value is {{Event/AT_TARGET}}.
 
+  When set to true, the <var>passive</var> option indicates that the <b>callback</b>
```
annevk: Same problem here.

dom.bs
```diff
@@ -1026,24 +1049,40 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
...
+<dfn method for=EventTarget><code>addEventListener(<var>type</var>, <var>callback</var>, <var>options</var>)</code></dfn>
 method, when invoked, must run these steps:
 
 <ol>
  <li><p>If <var>callback</var> is null, terminate these steps.
 
+ <li>If <var>options</var> is of type boolean, let <var>capture</var> be
+ <var>options</var> and let <var>passive</var> be false.  Otherwise let
+ <var>capture</var> and <var>passive</var> be the corresponding values in the
+ <var>options</var> {{EventListenerOptions}} dictionary.
```
annevk: 1. Given we need this, why have default values in IDL?
2. I would prefer initializing variables separately and then setting them based on the argument. I.e., use "let" once per variable name.

dom.bs
```diff
@@ -1026,24 +1049,40 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
...
-and <b>capture</b> is <var>capture</var>, set that <a>event listener</a>'s <b>removed flag</b> and
-remove it from the associated list of <a>event listeners</a>.
+<dfn method for=EventTarget><code>removeEventListener(<var>type</var>, <var>callback</var>, <var>options</var>)</code></dfn>
+method, when invoked, must, run these steps
+
+<ol>
+ <li>If <var>options</var> is of type boolean, let <var>capture</var> be
+ <var>options</var> and let <var>passive</var> be false.  Otherwise let
+ <var>capture</var> and <var>passive</var> be the corresponding values in the
+ <var>options</var> {{EventListenerOptions}} dictionary.
```
annevk: Same applies here.

dom.bs
```diff
@@ -1057,6 +1096,25 @@ invoked, must run these steps:
  <li><p><a>Dispatch</a> the <var>event</var> and return the value that returns.
 </ol>
 
+<h3 id=observing-event-listeners>Observing event listeners</h3>
+In general, developers do not expect the presence of an <a>event listener</a> to be
```
annevk: Use a newline after `<h3>` and an additional newline before the `<h3>`.

dom.bs
```diff
@@ -1057,6 +1096,25 @@ invoked, must run these steps:
...
+observable.  The impact of an <a>event listener</a> is determined by its <b>callback</b>.
+That is, a developer adding a no-op <a>event listener</a> would not expect it to have
+any side effects.
+
+Unfortunately, some event APIs have been designed such that implementing them
+efficiently requires observing <a>event listeners</a>.  For example, sensor APIs which
+enable an underlying device sensor, and touch APIs which can be used to block
```
annevk: Implementing them efficiently is not really sufficient motivation. Not executing code that cannot be observed is always fine. What is problematic is when it can be observed that adding and removing listeners has an effect. That is already the case with touch events as I understand it and therefore we are doing this, so we should use that as motivation, no?

dom.bs
```diff
@@ -1057,6 +1096,25 @@ invoked, must run these steps:
...
+asynchronous scrolling.  In some cases this problem can be mitigated by specifying
+the event to be {{Event/cancelable}} only when there is at least one
+non-{{EventListenerOptions/passive}} listener.  For example, non-{{EventListenerOptions/passive}}
+{{TouchEvent}} listeners must block scrolling, but if all listeners are {{EventListenerOptions/passive}} then
+scrolling can be allowed to start <a>in parallel</a> by making the {{TouchEvent}}
+uncancelable (so that calls to {{Event/preventDefault()}} are ignored).
```
annevk: We should make a note here that this "passive" hint is being observed by the code that dispatches the event since any listeners added during dispatch that are not "passive" will not be able to affect the event. In effect, what happens is that having solely passive observers at the time of action influences the event being dispatched. (The dispatcher code will set its passive flag.)

dom.bs
```diff
@@ -1138,9 +1196,15 @@ invoked, must run these steps:
    <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
    <a>event listener</a>).
 
+   <li>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s <a>passive flag</a>.
```
annevk: This seems wrong. Since before we said this would only happen if all listeners are passive. So this is in fact something that is done by those invoking dispatch.

dom.bs
RByers: How about `the capture member of <var>options</var>`?  Or do you prefer simply `the capture option`?

dom.bs
RByers: > Given we need this, why have default values in IDL?

Because we want to support cases like `addEventListener(t, h, {passive:true})` and `addEventListener(t, h, {capture:true})`.

> I would prefer initializing variables separately and then setting them based on the argument. I.e., use "let" once per variable name.

Ok.

dom.bs
RByers: Yes, agreed that the observability needs more mention here.  Part of the problem is that "observability" is a bit of a fuzzy concept when it comes to performance.  Is the scroll performance of a page "observable"?  Well if scrolling is blocked for 1s, or there's noticeable UI mis-synchronization (as can be seen [here](http://output.jsbin.com/qixakeh)) then I'd say that counts as "observable".  I'm tweaking this text to better reflect that.

dom.bs
RByers: Ok, I'm adding some more text along these lines.  Rather than say the "passive" hint is being observed, I think it's valuable to be specific that it's only the "absence of non-passive listeners" that's being observed.  Eg. the dispatching code does NOT make any distinction between the cases of mixed passive/non-passive listeners and non-passive only listeners.

See also the discussion below about the passive flag.

dom.bs
RByers: I thought this is what we agreed on [here](https://github.com/RByers/EventListenerOptions/issues/2#issuecomment-120755014) and summarized [here](https://github.com/RByers/EventListenerOptions/issues/2#issuecomment-125986241)?

I.e. calls to `preventDefault` from a passive listener never has any impact (since `passive` is a promise that the listener doesn't need to cancel).  But `Event.cancelable` is immutable and so can be changed only by code invoking dispatch based on the observation of ALL listeners.

RByers:
Thanks for the review!  I've updated the pull request based on your feedback.  I assume I should avoid squashing all the commits until the review is done and we're ready to merge.

dom.bs
RByers: Done

dom.bs
RByers: Done

dom.bs
RByers: Done

dom.bs
annevk: Ah okay. I think we should rename the flag on event then to "in passive listener flag" to make it clear what the scope is.

dom.bs
annevk: I think `<code>capture</code>` member is what is typically used.

dom.bs
annevk: That is not justification for having default values in the IDL though.

dom.bs
annevk: I thought we already changed the way we dispatch certain events (in a script-observable way, without timers) based on listeners registered. That's what I would count.

annevk:
Feel free to rebase and squash once you're confident it's in a good shape and I'll just do another pass. Might be easier.

dom.bs
RByers: Ah, I see what you're saying.  We could leave the default values out and update the spec text to say something like `If <var>options</var> has a <code>capture</code> member with value true, set <var>capture</var> to true'.  The way it's written now ("let capture and passive be the corresponding values in ..." seems to imply that those members will ALWAYS be present).

Personally I prefer the default values because it makes it more explicit that eg. {capture:false} and {} have identical behavior.  But if you think it's better to have the defaults be defined only by the algorithm, that's fine with me too.  I'm curious though, what's the advantage of that?

dom.bs
RByers: Ok.  I thought about this some more, and there is at least one subtle script-observable behavior change (at least in Chrome and Safari where I tested): if there are no touch listeners present and a listener is added mid-gesture, no events are sent for that finger.  I built a demo of this [here](http://output.jsbin.com/vuqeda).  You could argue however that this is a bug we could fix (though not sure we'd ever get Safari to change).

Do you think that's worth mentioning here?

dom.bs
RByers: Sounds good to me, thanks.

RByers:
Ok I think the only thing left is what exactly we want to cover in the "observing event listeners" section.  Let's keep [iterating on that](https://github.com/whatwg/dom/pull/82#discussion_r47803228), then once you're happy I'll rebase/squash (which I think will cause all the line-by-line discussions to disappear from GitHub :-( ).

dom.bs
annevk: Well, they have to be defined in the algorithm anyway. Since the dictionary is not guaranteed. So the IDL has no meaningful effect. I'd be okay with comments in the IDL saying what it effectively comes down to.

dom.bs
annevk: Yeah maybe. I must be something since I thought we had a very clear observable case to go from and therefore observing even more for this passive stuff was okay...

dom.bs
RByers: Ok, fair enough.  How's [this](https://github.com/RByers/dom/commit/3de3245316a84943b737a7db77a6d305a98b0ba8)?

Also my developer confusion argument is probably academic because any reasonable developer should assume that boolean options default to false anyway :-)

dom.bs
RByers: How about we move this discussion back to [the github issue](https://github.com/RByers/EventListenerOptions/issues/20) so we have a permanent record of the debate?  I'm sure it won't be the last time I have it <grin>.

dom.bs
```diff
@@ -727,16 +729,15 @@ inherits from the {{Event}} interface.
...
  signals to the operation that caused <var>event</var> to be
  <a>dispatched</a> that it needs to be
  canceled.
 
  <dt><code><var>event</var> . {{Event/defaultPrevented}}</code>
  <dd>Returns true if
- {{Event/preventDefault()}} was invoked
- while the {{Event/cancelable}} attribute
- value is true, and false otherwise.
+ {{Event/preventDefault()}} was used successfully to indicate cancellation.
```
annevk: (Adding more nitpicky comments. Feel free to ignore for now.) s/used/invoked/ We don't want to omit mentioning when it returns false. The original text did that correctly.

dom.bs
```diff
@@ -816,7 +818,14 @@ must return the values they were initialized to.
...
 <dfn method for=Event>preventDefault()</dfn>
 method must set the <a>canceled flag</a> if the
-{{Event/cancelable}} attribute value is true.
+{{Event/cancelable}} attribute value is true and
+the <a>in passive listener flag</a> is unset.
+
+<p class="note no-backref">
+ User agents are encouraged to generate a console warning or other debugging
+ aid to help authors identify places where calls to {{preventDefault()}}
+ have no effect.
```
annevk: @domenic started using "User agents are encouraged to log the precise cause in a developer console, to aid debugging." in HTML. I suggest we start from there.

dom.bs
```diff
@@ -1026,24 +1050,52 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
...
+ <li>Let <var>capture</var> and <var>passive</var> be false.
+
+ <li>If <var>options</var> is a boolean, set <var>capture</var> to
+ <var>options</var>.
+
+ <li>If <var>options</var> is a dictionary and <code>{{EventListenerOptions/capture}}</code> is
+ present in <var>options</var> with value true, then set <var>capture</var> to true.
+
+ <li>If <var>options</var> is a dictionary and <code>{{EventListenerOptions/passive}}</code> is
+ present in <var>options</var> with value true, then set <var>passive</var> to true.
```
annevk: Since these four steps are repeated below I feel like we should introduce an abstract operation that returns two values for capture and passive given a dictionary or boolean.

dom.bs
```diff
@@ -1058,6 +1110,32 @@ invoked, must run these steps:
 </ol>
 
 
+<h3 id=observing-event-listeners>Observing event listeners</h3>
+
+<p>In general, developers do not expect the presence of an <a>event listener</a> to be
+observable.  The impact of an <a>event listener</a> is determined by its <b>callback</b>.
```
annevk: Throughout, there should only be a single space after a period.

dom.bs
```diff
@@ -1058,6 +1110,32 @@ invoked, must run these steps:
...
+the event to be {{Event/cancelable}} only when there is at least one
+non-{{EventListenerOptions/passive}} listener.  For example, non-{{EventListenerOptions/passive}}
+{{TouchEvent}} listeners must block scrolling, but if all listeners are {{EventListenerOptions/passive}} then
+scrolling can be allowed to start <a>in parallel</a> by making the {{TouchEvent}}
+uncancelable (so that calls to {{Event/preventDefault()}} are ignored).  So code
+dispatching an event is able to observe the absence of non-{{EventListenerOptions/passive}}
+listeners, and use that to clear the {{Event/cancelable}} property of the event
+being dispatched.
+
+<p>Ideally, any new event types are defined such that they don't need this
```
annevk: s/new event types/new event APIs/
s/don't/do not/

dom.bs
RByers: Whoops, must have lost this by accident somewhere along the line.  Fixed.

dom.bs
RByers: Ok, how's [this](https://github.com/RByers/dom/commit/e3a2df3c7df30d68247d758c7f746ed76375b92c#diff-a467e681501e456ce8c1ef31425e1b41R825)?

dom.bs
RByers: I attempted that [here](https://github.com/RByers/dom/commit/e3a2df3c7df30d68247d758c7f746ed76375b92c#diff-a467e681501e456ce8c1ef31425e1b41R1052) and I agree it looks cleaner, thanks!  I couldn't find another example of returning multiple values, is what I've done precise enough?  Any suggestion for a term better than "normalize" (which could perhaps be confused with Node.normalize)?

dom.bs
RByers: Done.

dom.bs
RByers: Done.

dom.bs
annevk: "flatten" perhaps? What you did seems great. Returning multiple values in English feels quite natural and I've done it before without confusing anybody.

RByers: Thanks.  Sure "flatten" sounds better to me - done.

dom.bs
```diff
@@ -1139,9 +1216,15 @@ invoked, must run these steps:
...
+   <li>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s <a>in passive listener flag</a>.
+
    <li><p>Call <var>listener</var>'s <b>callback</b>'s {{EventListener/handleEvent()}}, with
    <var>event</var> as argument and <var>event</var>'s {{Event/currentTarget}} attribute value as
-   <a>callback this value</a>. If this throws any exception, <a>report the exception</a>.
+   <a>callback this value</a>.
+
+   <li>Clear <var>event</var>'s <a>in passive listener flag</a>.
+
+   <li>If the call to {{EventListener/handleEvent()}} threw any exception, <a>report the exception</a>.
```
annevk: Are you sure this is correct? Reporting the exception is synchronous. Seems like you could circumvent the passive restriction by storing the event somewhere and then throwing.

annevk:
(Sorry, I was about to merge, but then I spotted that. Not sure why I didn't catch it earlier.)

dom.bs
RByers: If reporting the exception is synchronous, then is it OK to just clear the passive listener flag afterwards (updated)?  That would definitely be preferable, I was just thinking that wouldn't be possible.

annevk: As far as I can tell that's fine. Where do you think it would break?

annevk:
The only other thing that looks a bit odd is the wrapping. I can clean that up I suppose.

dom.bs
RByers: I just wasn't thinking about it properly (thinking in terms of an algorithm with a 'throw' statement).  Thanks for catching this, I agree it's better and I don't anticipate any implementation issue.

RByers:
> The only other thing that looks a bit odd is the wrapping. I can clean that up I suppose.

Looks like the pattern is to use ~100 except in `<pre>` and `<code>` blocks, right?  I'm fixing that now.

RByers:
Ok, re-wrapped all lines I touched to 100 columns (except `<pre>` and `<code>` blocks).

annevk:
Landed as https://github.com/whatwg/dom/commit/253a21b8e78e37447c47983916a7cf39c4f6a3c5.

annevk:
Thank you!

PR #85, opened 2015-10-06
Fix #63: Ban insertNode() of Range's start node

### LINKED ISSUE #63: Consider banning insertNode() of the Range's start node
Test-case:

```html
<!doctype html>
foo
<script>
var range = document.createRange();
range.setStart(document.body.firstChild, 1);
var s = "Body children before: " + document.body.childNodes.length;
try {
  range.insertNode(range.startContainer);
  document.body.textContent = s + ", after: " + document.body.childNodes.length;
} catch (e) {
  document.body.textContent = "Exception";
}
</script>
```

Firefox outputs "Body children before: 2, after: 3", which matches the spec: there is no special handling of this case, and it winds up happily splitting the text node, then removing the first text node and re-inserting it back where it was, leaving the first portion no longer selected.  Chrome throws.  IE is the same as Firefox, but if you change the offset from 1 to 0 it outputs "Body children before: 2, after: 2", so it probably just aborts silently in that case.

IE/Firefox's behavior is just silly, and Chrome's seems a bit harsh.  I'm inclined to go with Chrome, out of the options available.  Failing that, I'd go with IE and at least special-case the scenario where we create an empty text node.  Firefox (which is the current spec) would be my last choice.

ayg:


ayg:
I updated in-place to address @annevk's comments on the commit.  (I don't know if this works nicely with Github.)

annevk:
Thank you @ayg. You might consider associating ayg@aryeh.name with your GitHbub account so you'll get accredited for the commits. (Nit for next time, we typically follow the ":" with a lowercase letter. Treating the first line of the commit as a title of sorts.)

PR #103, opened 2015-11-03
DOMTokenList hook to validate added token is in a supported set of keywords

yoavweiss:
This is a stab at adding the required hook in DOM so that `iframe@sandbox`, `link@rel` and hopefully future features would be able to use `DOMTokenList.add()` for feature detection.

Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=29061

/cc @annevk @domenic @zcorpan @Ms2ger

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8679,6 +8679,25 @@ interface DOMTokenList {
 <p>A {{DOMTokenList}} object also has an associated <a for="/">element</a> and an <a>attribute</a>'s
 <a for=Attr>local name</a>.
 
+<a lt="Other applicable specifications">Specifications</a> may define
+<dfn export for=Node id=concept-supported-tokens>supported tokens</dfn> for a {{DOMTokenList}}'s
+associated attribute.
+
+<p>A {{DOMTokenList}} object's
+<dfn export id=concept-domtokenlist-validation for="DOMString">validation steps</dfn> for a given
+<var>token</var> are:
+
+<ol>
+ <li><p>If the associated attribute does not define <a>supported tokens</a>, return true.
+
+ <li><p>Let <var>lowercase token</var> be a copy of <var>token</var>,
+ <a>converted to ASCII lowercase</a>.
+
+ <li><p>If <var>lowercase token</var> is present in <a>supported tokens</a>, return true.
+
+ <li><p>Return false.
+</ol>
+
 <p>A {{DOMTokenList}} object's
 <dfn export id=concept-dtl-update for="DOMTokenList">update steps</dfn> are to
 <a>set an attribute value</a> for the associated <a for="/">element</a> using associated
@@ -8702,10 +8721,11 @@ interface DOMTokenList {
 
  <dt><code><var>tokenlist</var> . <a for=DOMTokenList lt="add()">add(<var>tokens</var>&hellip;)</a></code>
  <dd>
-  <p>Adds all arguments passed, except those already present.
+  <p>Adds all valid arguments passed, except those already present.
   <p>Throws a {{SyntaxError}} exception if one if the arguments is the empty string.
   <p>Throws an {{InvalidCharacterError}} exception if one of the arguments contains any
   <a>ASCII whitespace</a>.
+  <p>Returns false if any invalid arguments passed, true otherwise.
 
  <dt><code><var>tokenlist</var> . <a for=DOMTokenList lt="remove()">remove(<var>tokens</var>&hellip;)</a></code>
  <dd>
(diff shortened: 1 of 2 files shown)


domenic:
I don't think this is sufficient, to only use the validation steps for `add()`'s return value. I think add and toggle should not add invalid values, and value's setter should validate before adding.

yoavweiss:
> I think add and toggle should not add invalid values, and value's setter should validate before adding.

The reason I didn't add that was @Ms2ger's concern that it would raise web compat issues. If this is not a concern, I can add it.

annevk:
@domenic do you also think we should canonicalize the tokens? I guess that might be too incompatible...

yoavweiss:
One more reasoning to avoid added invalid values is @annevk's comment on whatwg irc, saying that as currently specified, there's a good chance the list of supported keywords and actual support would get out of sync over time.

yoavweiss:
I've made the changes so that `add()` and `toggle()` will not add invalid tokens. I also added case insensitivity to the validation step. I did not add a validation step to the setter, per discussion on [irc](http://krijnhoetmer.nl/irc-logs/whatwg/20151104#l-272).

dom.bs
```diff
@@ -8679,6 +8679,18 @@ interface DOMTokenList {
 <p>A {{DOMTokenList}} object also has an associated <a for="/">element</a> and an <a>attribute</a>'s
 <a for=Attr>local name</a>.
 
+<a lt="Other applicable specifications">Specifications</a> may define
+<dfn export for=Node id=concept-supported-keywords>supported keywords</dfn> for a {{DOMTokenList}}'s associated attribute.
+
+<p>A {{DOMTokenList}} object's
+<dfn export id=concept-domtokenlist-validation for="DOMString">validation steps</dfn> for a given <var>token</var> are:
+<ol>
```
annevk: I would prefer some newlines here for readability. Similar to the other `<ol>`-formatted algorithms.

dom.bs
```diff
@@ -8679,6 +8679,18 @@ interface DOMTokenList {
 <p>A {{DOMTokenList}} object also has an associated <a for="/">element</a> and an <a>attribute</a>'s
 <a for=Attr>local name</a>.
 
+<a lt="Other applicable specifications">Specifications</a> may define
+<dfn export for=Node id=concept-supported-keywords>supported keywords</dfn> for a {{DOMTokenList}}'s associated attribute.
```
annevk: I think we should make that "supported tokens" or "valid  tokens", no?

dom.bs
yoavweiss: will add

dom.bs
yoavweiss: sure, I'll change to "supported tokens"

annevk:
Presumably @yoavweiss plans on implementing this in Chromium. @bzbarsky, do you think it's reasonable to add this kind of functionality to the web? It would be quite useful for feature testing.

bzbarsky:
This seems reasonable, I guess, and doesn't seem terribly burdensome to implement at first glance.  For feature detection you'd `add()` and see if it's there?

yoavweiss:
The feature detection story is basically:
```
    var link = document.createElement("link");
    if (link.relList && link.relList.add("preload")) {
        // supported, since add() returned true
    } else {
        // Not supported. One of three options:
        // * relList not implemented
        // * add() returned "undefined", since this spec change wasn't yet implemented
        // * add() returned "false" since "preload" is not a valid token
    }
```

The above example refers to `<link>`, but the story is similar for `<iframe>`.

yoavweiss:
ping

dom.bs
```diff
@@ -8810,6 +8832,7 @@ method, when invoked, must run these steps:
   <ol>
    <li><p>If <var>force</var> is passed and is false, return false.
 
+   <li><p>Otherwise, run <a>validation steps</a> with <var>token</var>. If the return value is false, return false.
```
annevk: This needs to be followed by a newline. Also, here and elsewhere text is not wrapped at 100 columns.

annevk:
Apart from that, I guess this looks good to merge. Would you be willing to squash the commits and rebase this?

yoavweiss:
Sure

yoavweiss:
Squashed, rebased and wrapped

annevk:
And please remove the dot from the first line of the commit message. It's considered to be a title. I should probably point to the HTML commit guidelines from the README of this repository.

Sorry for making you go through this, please let me know when you had enough and I'll finish up.

yoavweiss:
I can remove the dot, no problem :)

Regarding the [README](https://github.com/whatwg/dom/blob/master/README.md), it's not very descriptive :D Is there somewhere else with the commit guidelines?

annevk:
https://github.com/whatwg/html/blob/master/README.md is better. But basically I need to level up all these repositories to be more welcoming to contributors.

Ideally they can just share contribution guidelines, but per http://stackoverflow.com/questions/32296220/sharing-a-contributing-md-file-across-multiple-github-repos that seems not possible. So perhaps just providing pointers or having a WHATWG Wiki page explain things might be best.

yoavweiss:
OK, thanks for the links :)
Added myself to acknowledgements and fixed the commit message accordingly

annevk:
So yeah, I totally forgot about acks. Thanks whatwg/html! So yes, I'll merge this and then start with adding a link to whatwg/html's README for tips.

And thank you @yoavweiss for being patient and supplying this PR!

smaug----:
So if pages start to use
if (link.relList && link.relList.add("preload")) {
that breaks preload on browsers which don't have TokenList implementation updated.
(because .add returning void is false)

This feature feels odd "feature testing" and "actual feature usage" merged together.
Not much different to node.isSupported("foo") + using the feature, except that .add() forces one to actually
try to use the feature in order to know whether browser claims to support it.

Issue #106, opened 2015-11-12
DOMTokenList validation for object creation and attribute updates


annevk:
@yoavweiss I think we forgot to cover this since the text wasn't present yet in the `DOMTokenList` section but instead sorta left to each individual element/attribute combination.

Do you agree that we should update https://dom.spec.whatwg.org/#interface-domtokenlist to validate the tokens coming from attribute sets?

(See the two paragraphs before the domintro box.)

yoavweiss:
I agree that it should be validated, for the sake of consistency.

ArkadiuszMichalski:
Two small things:
1. In IDL `void add(DOMString... tokens);` << now return boolean
2. In validation steps algo (step 1. If the associated attribute’s...) << missing reference for "attribute’s"

yoavweiss:
Good catch, thanks!

PR #107, opened 2015-11-12
Add a validation step when a new DOMTokenList is created

yoavweiss:
Closes https://github.com/whatwg/dom/issues/106

@annevk - I added a validation step to the DOMTokenList creation. Can you take a look?

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8582,7 +8582,7 @@ interface DOMTokenList {
   readonly attribute unsigned long length;
   getter DOMString? item(unsigned long index);
   boolean contains(DOMString token);
-  void add(DOMString... tokens);
+  boolean add(DOMString... tokens);
   void remove(DOMString... tokens);
   boolean toggle(DOMString token, optional boolean force);
   void replace(DOMString token, DOMString newToken);
@@ -8606,8 +8606,8 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
 <var>token</var> are:
 
 <ol>
- <li><p>If the associated attribute's <a for=Attr>local name</a> does not define <a>supported
- tokens</a>, return true.
+ <li><p>If the associated <a>attribute</a>'s <a for=Attr>local name</a> does not define
+ <a>supported tokens</a>, return true.
 
  <li><p>Let <var>lowercase token</var> be a copy of <var>token</var>,
  <a>converted to ASCII lowercase</a>.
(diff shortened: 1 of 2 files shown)


dom.bs
```diff
@@ -8642,8 +8642,10 @@ are to:
 <a lt="named attribute"><var>associated attribute's local name</var> attribute</a> or
 associated <a for="/">element</a>'s
 <a lt="named attribute"><var>associated attribute's local name</var> attribute</a> is
-<a lt="attribute is set">set</a>, set <a>tokens</a> to the new <a for=Attr>value</a>,
-<a lt="ordered set parser">parsed</a>.
+<a lt="attribute is set">set</a>, let <var>temporary tokens</var> be the new <a for=Attr>value</a>,
```
annevk: I would prefer a "run these substeps" if we start creating variables. I think that would be a little easier to read.

yoavweiss: added

dom.bs
```diff
@@ -8642,8 +8642,16 @@ are to:
...
-<a lt="attribute is set">set</a>, set <a>tokens</a> to the new <a for=Attr>value</a>,
-<a lt="ordered set parser">parsed</a>.
+<a lt="attribute is set">set</a>, follow these substeps:
+
+<ol>
+ <li><p>Let <var>temporary tokens</var> be the new <a for=Attr>value</a>,
+ <a lt="ordered set parser">parsed</a>.
+
+ <li><p>For each <var>token</var> in <var>temporary tokens</var>, run
+<a>validation steps</a> with <var>token</var>. If the return value is true, append <var>token</var>
```
annevk: This and the next line are not indented properly.

dom.bs
```diff
@@ -8642,8 +8642,16 @@ are to:
 <a lt="named attribute"><var>associated attribute's local name</var> attribute</a> or
 associated <a for="/">element</a>'s
 <a lt="named attribute"><var>associated attribute's local name</var> attribute</a> is
-<a lt="attribute is set">set</a>, set <a>tokens</a> to the new <a for=Attr>value</a>,
-<a lt="ordered set parser">parsed</a>.
+<a lt="attribute is set">set</a>, follow these substeps:
```
annevk: s/follow/run/ is the language this specification uses.

yoavweiss:
Fixed the issues raised in comments

dom.bs
```diff
@@ -8606,7 +8606,7 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
 <var>token</var> are:
 
 <ol>
- <li><p>If the associated attribute's <a for=Attr>local name</a> does not define <a>supported
+ <li><p>If the associated <a>attribute</a>'s <a for=Attr>local name</a> does not define <a>supported
  tokens</a>, return true.
```
annevk: Last nit, no breaking inside of a single element. (HTML does it, but try to avoid it as it makes search & replace harder than it needs to be.)

With that fixed, if you rebase/squash, I'll land this.

yoavweiss:
Fixed and squashed

PR #109, opened 2015-11-12
Specify CharacterData interface instead of implementers

frewsxcv:


frewsxcv:
As you can tell by the commit message and description, I hastily committed this just to get initial reactions

annevk:
I think this is fine. @Ms2ger? It does seem that we could do this in more places. If we go ahead with this.

Ms2ger:
Don't feel strongly either way.

domenic:
I kind of liked having the switch only be over real node types, not including abstract base classes. But I don't feel strongly.

annevk:
True, we should either go all the way everywhere or just only use the "real" classes.

annevk:
I guess we'll leave this as is for now.

Issue #111, opened 2015-11-20
Let DOMTokenList.prototype.add() return false if there are no supported tokens


foolip:
This is to deal with an issue raised on blink-dev, discussed briefly on IRC:
https://groups.google.com/a/chromium.org/d/msg/blink-dev/xEoWkGEd_g4/ybkKf0GyAwAJ
http://krijnhoetmer.nl/irc-logs/whatwg/20151120#l-239

To summarize, it would be really bad if an attribute that currently isn't defined in terms of supported tokens is later "upgraded" to something like the rel attribute. Then it would be harder to reliably detect that a token is *really* supported, one might have to resort to e.g. `supported = !list.add('tokenthatwillneverbesupported') && list.add('tokentotest')` plus `list.remove('tokenthatwillneverbesupported')` as cleanup.

yoavweiss:
My problem with returning false if no supported tokens is present is that `add()` will always return false for things like `classList` and will return false for `<a rel>`'s `relList` on every value.

What would happen if one day we'd want to add supported tokens to `<a rel>`? There would be no way of telling if an attribute is really *not* supported. (whereas as defined today you can tell, even if in a cumbersome way)

foolip:
@yoavweiss on IRC:
> the algorithm needs to be redefined so that in that case, the token is added to DOMTokenList, even if we returned false

Oops, I had overlooked this bit, that in the cases where `add()` returns false due to validation failure, the token actually isn't added. Why is that so? It's still not an invariant that all of the tokens in the list will be supported, since one can simply set the attribute directly, or have nonsense in the markup.

Since the "that is not in tokens" bit skips the validation steps, it's also the case that `list.add('foo')` would return true even if 'foo' is not a valid token, if that token is already in the list. Kind of non-obvious, I think.

What would you think about leaving the behavior with regard to updating the attribute exactly as it was, and having the return value only be affected by token validation?

bzbarsky:
Have you considered just giving up on this boolean return value bit and moving to a tristate return value, all truthy, representing "supported", "not supported", and "no supported set" (in fact, they could just be those strings)?  Then a falsy return value would mean API not implemented yet (reducing the complexity of writing proper feature-detection in the face of downrev clients that don't implement the API yet), and a truthy one would tell you exactly what's going on with the support state.

yoavweiss:
I'm fine with that.

annevk:
[...]

PR #112, opened 2015-11-20
Let DOMTokenList's add() return false when there are no supported tokens

foolip:
Fixes https://github.com/whatwg/dom/issues/111

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8607,7 +8607,7 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
 
 <ol>
  <li><p>If the associated <a>attribute</a>'s <a for=Attr>local name</a> does not define
- <a>supported tokens</a>, return true.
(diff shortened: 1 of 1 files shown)


foolip:
Not patching `dom.html` because the latest Bikeshed also causes other spurious changes, so you do it :)

@yoavweiss, please review and say if you'd be OK with doing this in Blink.

yoavweiss:
I'd be OK with implementing assuming that this is the right way to go. I currently disagree about that, but let's discuss on the issue :)

yoavweiss:
I believe that as defined in this PR, it would also result with `add()` never actually adding tokens to DOMTokenList without "supported tokens"

annevk:
This is now #114 I assume?

foolip:
Yes, I'll close my issue and PR.

foolip:
Just this PR, actually.

PR #120, opened 2015-11-26
Fix #119: Merge DOMSettableTokenList into DOMTokenList

### LINKED ISSUE #119: Merge DOMSettableTokenList into DOMTokenList
Per discussion on IRC with @domenic and @zcorpan only HTML uses these classes. It seems like we could simplify things a bit by merging them into a single class and use `[PutForward]` everywhere.

The problem I could see is that libraries depend on changing the prototype of these classes in ways that are incompatible with such a simplification. What do folks think?

zcorpan:
Move DOMSettableTokenList#value to DOMTokenList and remove the
DOMSettableTokenList interface.

Issue: #119

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5041,7 +5041,7 @@ interface Element : Node {
 
            attribute DOMString id;
            attribute DOMString className;
-  [SameObject] readonly attribute DOMTokenList classList;
+  [SameObject, PutForwards=value] readonly attribute DOMTokenList classList;
 
   boolean hasAttributes();
   [SameObject] readonly attribute NamedNodeMap attributes;
@@ -8561,8 +8561,7 @@ These constants can be used for the
 
 <h2 id="sets">Sets</h2>
 
-<p class="note no-backref">Yes, the names {{DOMTokenList}} and {{DOMSettableTokenList}}
-are unfortunate legacy mishaps.
+<p class="note no-backref">Yes, the name {{DOMTokenList}} is an unfortunate legacy mishap.
 
 <h3 id="interface-domtokenlist">Interface {{DOMTokenList}}</h3>
 
@@ -8576,6 +8575,7 @@ interface DOMTokenList {
   boolean toggle(DOMString token, optional boolean force);
   void replace(DOMString token, DOMString newToken);
   boolean supports(DOMString token);
+           attribute DOMString value;
   stringifier;
   iterable&lt;DOMString>;
 };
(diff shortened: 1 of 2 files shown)


annevk:
@foolip @bzbarsky, any concerns?

foolip:
From the spec point-of-view this makes sense to me, one can already change the underlying value using `DOMTokenList.prototype.add()`, so also being able to set `value` ought not introduce any new problems in principle.

Implementation-wise, a potential problem is that `DOMTokenList` is actually a virtual base class, with concrete subclasses `ClassList` (not an IDL interface) and `DOMSettableTokenList`. Without actually trying to write the code, it's hard to say if merging these into a single internal representation would cause any trouble, like maybe perf regressions for `classList`.

bzbarsky:
> Implementation-wise, a potential problem is that DOMTokenList is actually a virtual base class

In Chrome. In Firefox it's not, for what it's worth.  Note that per spec there are currently `DOMTokenList`s that are neither `Element.classes` nor `DOMSettableTokenList`, though it looks like Chrome simply doesn't implement them.

I don't think there are any implementation issues here from Gecko's point of view.

What there _is_ is that the HTML spec needs to be updated in sync with this change, as do web platform tests.

zcorpan:
@bzbarsky the HTML PR is https://github.com/whatwg/html/pull/361

Not PR for wpt yet, but I could look into that. https://github.com/w3c/web-platform-tests/issues/2411

annevk:
I would like to hear from @foolip or someone else working on Blink before doing this. It seems totally like the thing I want us to be doing, but I don't want to upset anyone in the process.

foolip:
I asked @yoavweiss if he wanted to work on this and it seemed like *maybe* he would. Thoughts? Investigating this properly in Blink amounts to trying to implement the change, and I have a few things too many in the air right now.

yoavweiss:
I can try to implement it locally next week and see what happens.

annevk:
@yoavweiss any updates?

yoavweiss:
Not yet. Got postponed to post-holidays. Hope to get to it soon.

yoavweiss:
A blink patch is at https://codereview.chromium.org/1629403003/

zcorpan:
`classList` should have `[PutForwards=value]` in this PR; will fix.

zcorpan:
Rebased and added `PutForwards`. (Bikeshed picks the wrong `value` to xref in the IDL block, but it's not a fatal error, just a warning. https://github.com/tabatkins/bikeshed/issues/582)

annevk:
@zcorpan what's your preferred way of doing this? Should we just land this and see if browsers follow and hope there's no problem?

zcorpan:
Yeah that seems reasonable to me since @yoavweiss has implemented it for blink already, it seems like it should work out OK. I can file bugs for other browsers also.

annevk:
Thank you all!

zcorpan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1244328
https://bugs.webkit.org/show_bug.cgi?id=153677
https://connect.microsoft.com/IE/feedbackdetail/view/2299283/merge-domtokenlist-and-domsettabletokenlist

PR #121, opened 2015-11-26
Do not remove child from its parent when replacing a node by itself

nox:
If the replacing node is the replaced child, the node doesn't need and
shouldn't be adopted into the parent's document because it removes it from
the parent, which is already done in the very next step.

The adoption step is made conditional and not the removing one to not
artificially increase the number of mutation records being queued during
the replacing as a whole.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1600,6 +1600,14 @@ steps:
  <var>node</var> into <var>parent</var>'s
  <a>node document</a>.
 
+ <li>Let <var>removedNodes</var> be the empty list.
+
+ <li>If <var>child</var>'s parent is not null, set <var>removedNodes</var> to a list
+  solely containing <var>child</var> and remove <var>child</var> from its <var>parent</var>
+  with the <i>suppress observers flag</i> set.
+
+  <p class="note no-backref">The above can only be false if <var>child</var> is <var>node</var>.
+
  <li><a>Remove</a> <var>child</var>
  from its <var>parent</var> with the
  <i>suppress observers flag</i> set.
@@ -1614,7 +1622,7 @@ steps:
  the <i>suppress observers flag</i> set.
 
  <li><a>Queue a mutation record</a> of "<code>childList</code>" for target
- <var>parent</var> with addedNodes <var>nodes</var>, removedNodes a list solely containing
+ <var>parent</var> with addedNodes <var>nodes</var>, removedNodes <var>removedNodes</var>
  <var>child</var>, nextSibling <var>reference child</var>, and previousSibling
  <var>previousSibling</var>.
 
--- a/dom.html
+++ b/dom.html
@@ -982,11 +982,16 @@ steps:</p>
     <li>
      <p>Let <var>previousSibling</var> be <var>child</var>’s <a data-link-type="dfn" href="#concept-tree-previous-sibling">previous sibling</a>. </p>
     <li><a data-link-type="dfn" href="#concept-node-adopt">Adopt</a> <var>node</var> into <var>parent</var>’s <a data-link-type="dfn" href="#concept-node-document">node document</a>. 
+    <li>Let <var>removedNodes</var> be the empty list. 
+    <li>
+     If <var>child</var>’s parent is not null, set <var>removedNodes</var> to a list
+  solely containing <var>child</var> and remove <var>child</var> from its <var>parent</var> with the <i>suppress observers flag</i> set. 
+     <p class="note no-backref" role="note">The above can only be false if <var>child</var> is <var>node</var>. </p>
     <li><a data-link-type="dfn" href="#concept-node-remove">Remove</a> <var>child</var> from its <var>parent</var> with the <i>suppress observers flag</i> set. 
     <li>Let <var>nodes</var> be <var>node</var>’s <a data-link-type="dfn" href="#concept-tree-child">children</a> if <var>node</var> is a <code class="idl"><a data-link-type="idl" href="#documentfragment">DocumentFragment</a></code> <a data-link-type="dfn" href="#concept-node">node</a>, and a list containing solely <var>node</var> otherwise. 
     <li><a data-link-type="dfn" href="#concept-node-insert">Insert</a> <var>node</var> into <var>parent</var> before <var>reference child</var> with
  the <i>suppress observers flag</i> set. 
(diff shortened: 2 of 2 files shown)


nox:
[WebKit](https://github.com/WebKit/webkit/blob/master/Source/WebCore/dom/ContainerNode.cpp#L399-L400) queues no mutation record, and in fact skips all the sanity checks from steps 1-6 too. That doesn't seem correct.

[Gecko](http://mxr.mozilla.org/mozilla-central/source/dom/base/nsINode.cpp#1951) takes care to not queue a mutation record for the to-be-replaced child if it is the replacing node, but it still does it unconditionally for the replacing node anyway, queuing 2 records in total.

This patch instead conditionally adopts the replacing node only if it isn't the to-be-replaced child, thus queuing only one record.

nox:
I want this because I think it makes more sense overall, and because queuing only one mutation record means that the HTMLCollection instance bound to the parent's childNodes IDL attribute can easily be cached and updated by piggy-backing the mutation observers machinery.

annevk:
@smaug----, this is worth fixing in Gecko, no? @foolip?

foolip:
So if IIUC, we're talking about step 10 of the [replace](https://dom.spec.whatwg.org/#concept-node-replace) algorithm, used in `replaceChild()` and `replaceWith()`, right?

Is there the same problem in step 4 of [pre-insert](https://dom.spec.whatwg.org/#concept-node-pre-insert), which is used in a lot of places?

There appears to be inconsistencies here as well:
http://software.hixie.ch/utilities/js/live-dom-viewer/saved/3767

Gecko logs mutations, Blink does not.

annevk:
I don't see how pre-insert has this problem, since it explicitly does the node === child check.

foolip:
Where is that check? My example was `parent.insertBefore(a, b)` where that wouldn't actually change the order, but there's also `parent.insertBefore(a, a)` which I guess is what you're talking about?

nox:
@annevk `foo.replaceChild(document_type, document_type)` should fail if foo is an element.

nox:
@foolip There is no problem with pre-insert in the spec, AFAICT.

nox:
@annevk Never mind my previous comment, I thought you were saying the check in *insertBefore* means there is no problem.

@foolip The check @annevk is talking about is pre-insert's step 3.

foolip:
Step 3 is for a case like `parent.insertBefore(a, a)`, by turning it into `parent.insertBefore(a.nextSibling, a)`, but that wouldn't change the DOM.

What I'm looking for is the design principle that is upheld in **pre-insert** but not in **replace**. Should these APIs be no-ops when possible, or should they always remove and reinsert once?

nox:
I don't think they should be noops, ranges are supposed to change if you remove an element and put it back at the same place.

nox:
And I don't understand what you mean by a design principle upheld in pre-insert. Both pre-insert and replace change the DOM, a node gets removed and inserted in both.

foolip:
I'm just trying to precisely understand the problem. Is the number of mutation records the entirety of the problem, or is that just the side-effect?

annevk:
It's a side effect of a particular fix. There's several ways to fix the problem. Not doing any mutation at all however breaks the expectations for ranges so that is not considered an option.

foolip:
> the HTMLCollection instance bound to the parent's childNodes IDL attribute can easily be cached and updated by piggy-backing the mutation observers machinery

Is this actually true? The mutation observers are notified after the change in HTMLCollection should be observable.

nox:
That's why I said "mutation observers *machinery*", not the mutation observers themselves.

foolip:
Oh, I see you're a Servo developer. Without any context I assumed you were trying to build a polyfill or something.

Since it looks like existing implementations aren't precisely in sync with the spec, can you list what the mutation records would be per spec with and without this change? I see there's a "Queue a mutation record" step at the end, and wonder if there will still be duplicate records in the general case, where both the adopt step is run?

nox:
@foolip I wrote a test case there: <https://github.com/w3c/web-platform-tests/pull/2379>.

foolip:
Typo at the end of my previous comment. Should be "where both the adopt step and the remove step is run?"

annevk:
@foolip either you get a single record that lists the removed node and the added node. Or you get two records. The first with the removed node, the second with the added node (and no removed nodes). Those are the options we have. The first requires a simple change to step 10. The second requires changes to steps 11 and 14 and does not seem very straightforward to make.

foolip:
I haven't really used `MutationObserver` so that I could say which is better from the developer's point of view, but a single record seems OK. Does the proposed fix address all cases, or is it possible for *node*'s parent to be *parent* with something like `a.replaceWith(b)` where `a` and `b` are siblings? Any mutation records created by adopting in that case seem equally redundant.

nox:
Oh damn. Didn't think about siblings. I think we need to go @smaug----'s route finally. :(

nox:
@foolip We can't have a mutation record where the removed and added nodes weren't in the same position.

nox:
2 records need to be queued.

nox:
Ok so I was confused this afternoon and thought about this again. I still think queuing two records when child is node is wrong.

@foolip's case is not a problem with this patch. The two records being queued in the case of an arbitrary child being replaced by one of its siblings are: `[(removedNodes = node), (removedNodes = child, addedNodes = node)]`, so my fear of having a nonsensical record was wrong.

@smaug---- please explain why you think just `[(removedNodes = node, addedNodes = node)]` in the case where child is node is nonsensical.

nox:
@smaug---- That being said, here is what the step 11 would like with the behaviour you want:

> * If child’s parent is null, let removedNodes be the empty list. [Note: The above can only happen if child is node.]
> * Otherwise, remove child from its parent with the suppress observers flag set and let removedNodes be a list containing solely child.

I still consider my patch to be cleaner and a better representation of what happens to the DOM.

smaug----:
I listed two things on IRC
1) We should aim for consistency as long as possible. So if some step in an algorithm can be executed, it should be (by default) (step 10), and aim for adding special cases to steps which just can't be executed (step 11)
2) with your proposal we end up creating such mutation records which no other mutation creates, same node in removed and in added nodes

nox:
@smaug---- Ah, hadn't understood the second point. That's quite a good point. Will change patch.

nox:
@smaug---- Amended.

PR #122, opened 2015-11-26
Fix queuing of mutation records when removing a node from its parent

nox:
The old next sibling of the node needs to be stored somewhere before removing
the node from its parent.


annevk:
Can we use oldPreviousSibling's next sibling?

nox:
@annevk That makes things more complex. There could be no *oldPreviousSibling*, in which case we would need the parent's first child.

annevk:
Thank you!

PR #123, opened 2015-12-03
Remove feature detection from add() and add supports()

yoavweiss:
This follows up on the discussion we had in https://github.com/whatwg/dom/pull/114 regarding feature detection for DOMTokenLists:

* It removes references to "validation steps" from `add()`, `toggle()`, `replace()` and object creation.
* It adds `supports()` as discussed.
* When the associated attribute has no supported tokens, `TypeError` is thrown (by the "validation steps")

/cc @zcorpan @domenic @annevk

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8531,10 +8531,11 @@ interface DOMTokenList {
   readonly attribute unsigned long length;
   getter DOMString? item(unsigned long index);
   boolean contains(DOMString token);
-  boolean add(DOMString... tokens);
+  void add(DOMString... tokens);
   void remove(DOMString... tokens);
   boolean toggle(DOMString token, optional boolean force);
   void replace(DOMString token, DOMString newToken);
+  boolean supports(DOMString token);
   stringifier;
   iterable&lt;DOMString>;
 };
@@ -8556,7 +8557,7 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
 
 <ol>
  <li><p>If the associated <a>attribute</a>'s <a for=Attr>local name</a> does not define
- <a>supported tokens</a>, return true.
+ <a>supported tokens</a>, <a>throw</a> a <code>TypeError</code>.
 
  <li><p>Let <var>lowercase token</var> be a copy of <var>token</var>,
  <a>converted to ASCII lowercase</a>.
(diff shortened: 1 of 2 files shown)


zcorpan:
LGTM

domenic:
Approach LGTM, will leave any editorial concerns to @annevk.

yoavweiss:
@annevk - any comments?

dom.bs
```diff
@@ -8555,7 +8556,7 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
 
 <ol>
  <li><p>If the associated <a>attribute</a>'s <a for=Attr>local name</a> does not define
- <a>supported tokens</a>, return true.
+ <a>supported tokens</a>, throw {{TypeError}}.
```
annevk: This should be made consistent.

annevk:
So another thing I think we want to change is that "supported tokens" is an aspect of the `DOMTokenList` and not related to the attribute's local name. That requires changes to how "supported tokens" is defined, to be more in line with other associated slots. And it requires how that is referenced elsewhere.

yoavweiss:
@annevk - can we decouple the "supported tokens" redefinition from this PR? I'm happy to do that work, but it is not strictly related to this change, so perhaps shouldn't block it.

annevk:
Okay that sounds reasonable. Let me know if you want to squash or whether I should.

yoavweiss:
@annevk - rebased

annevk:
Thank you!

ArkadiuszMichalski:
@annevk missing checking empty string and ASCII whitespace in `token` argument is intentional? All other methods check this two things, even `contains()` which does not modify anything.

annevk:
We should probably either change `support()` or `contains()`, indeed. Leaning towards changing `contains()`. File an issue?

ArkadiuszMichalski:
Sure, https://github.com/whatwg/dom/issues/202

cdumez:
This new DOMTokenList.supports() reminds me of DOMImplementation.hasFeature() whose implementation gets so easily outdated and which is now a no-op in the latest DOM spec. I am personally not a fan.

phistuck:
@cdumez - the difference is that these (the ones discussed here) features are usually otherwise non feature-detectable, since they are not developer facing (from the side of the current document). `hasFeature` usually "detected" otherwise feature-detectable features.

PR #130, opened 2015-12-16
Fix replace algorithm (closes #129)

### LINKED ISSUE #129: Some fix for replace algo and question
https://dom.spec.whatwg.org/#concept-node-replace
12. If child’s parent is not null, set removedNodes to a list solely containing child and remove child from its parent with the suppress observers flag set. << first "parent" and "remove" should not be linked? Especially for "remove" because I don't know if you mean "remove algo" or direct term.

And have another question, why in step 13. we again remove `child` from its `parent`? We have it in step 12, or, if `child` is same as `node`, then removing from `parent` is done in step 10. via "adopt algo" (https://dom.spec.whatwg.org/#concept-node-adopt).

nox:
Cc @ArkadiuszMichalski

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1599,15 +1599,16 @@ steps:
  <li>Let <var>removedNodes</var> be the empty list.
 
  <li>
-  <p>If <var>child</var>'s parent is not null, set <var>removedNodes</var> to a list solely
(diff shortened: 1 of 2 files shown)


annevk:
Thank you! https://github.com/whatwg/dom/commit/cfaf3fa754fe53b1aa1dbe3f948a3272ae71b0e4

PR #135, opened 2015-12-21
fix minor typo

WesNetmo:


annevk:
Thank you, how would you like to be mentioned in the acknowledgments section?

WesNetmo:
thank you for asking, but i will wait for something more significant :D

annevk:
Up to you, but just so you know, we add folks to that section for less. 😊

Anyway, thank you!

WesNetmo:
rofl. well, if that's the case, my name is Giorgio Liscio :)
and thank _you_ for your work. each line of a spec is one more year for a frontend developer to live :D

Issue #116, opened 2015-11-23
setAttributeNode should not change order of existing attributes


foolip:
https://dom.spec.whatwg.org/#concept-element-attributes-set

Per spec, this will remove *oldAttr* and append the new *attr*. This matches Gecko, but not Blink or Edge in this test:
http://software.hixie.ch/utilities/js/live-dom-viewer/saved/3758

Here is the same test using `setAttribute`, where the attribute order does not change:
http://software.hixie.ch/utilities/js/live-dom-viewer/saved/3759

This could be fixed by having the concept of replacing attributes, in addition to append/remove.

CC @bzbarsky @Ms2ger

bzbarsky:
That's probably doable.  It adds more complexity into what is already a fairly complex codepath, unfortunately...

foolip:
Is the Gecko implementation pretty much step-by-step identical to the spec? If so I can see why it would complicate matters. In Blink there's first a search for the internal index of the attribute to replace, so the structure is quite different.

bzbarsky:
The Gecko implementation is pretty much step by step identical to the spec, with the additional complication that Attr nodes are not what's stored in the element's list.  They are reified as needed; the actual thing stored in the list is just the name and value.  So there is some additional complication in terms of keeping the Attr nodes and actual attributes in sync.  

I _think_ we could add a replace thing on top of all that stuff, though.

foolip:
Right, that's the same as in Blink, and I assume every browser engine ever.

I did try to change Blink to be a step-by-step copy of the spec's algorithm, and that was how I discovered this issue. I think it would work fine, but it seems faster to converge on what Blink/Edge/WebKit already do.

annevk:
So what kind of mutation record should replacing give? The same as remove/append does today? Do we really want to create special code paths just for `setAttributeNode`?

foolip:
Judging only by the code, I think it would do the same as when `setAttribute` or `setAttributeNS` changes an existing attribute.

That's not perfect. It would be weird that you can replace the Attr node, but for mutation observers it would look like the attribute has changed, which shows that Attr isn't actually the internal representation.

annevk:
I think that is fine personally, since mutation observers do not treat attributes as proper objects. However, I'm still wondering why we'd want to introduce this additional complexity for such an unused method. I guess because three out of four browsers already do it...

foolip:
[...]

PR #139, opened 2015-12-30
Fix the set an attribute algorithm

nox:
Only one mutation record should be queued, according to both WebKit and Gecko.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5151,6 +5151,26 @@ run these steps:
  <li>An <a>attribute is removed</a>.
 </ol>
 
+To <dfn export id=concept-element-attributes-replace lt="replace an attribute">replace</dfn> an
+<a>attribute</a> <var>oldAttr</var> by an <a>attribute</a> <var>newAttr</var>
+in an <a for="/">element</a> <var>element</var>, run these steps:
+
+<ol>
+ <li><p><a>Queue a mutation record</a> of "<code>attributes</code>" for <var>element</var>
+ with name <var>oldAttr</var>'s <a for=Attr>local name</a>,
+ namespace <var>oldAttr</var>'s <a for=Attr>namespace</a>,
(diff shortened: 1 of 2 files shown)


annevk:
How does this relate to #116? I guess if your assertion is correct that would mean `setAttributeNode()` could use this new path too.

nox:
I wasn't aware of #116, someone implemented setAttributeNode in Servo and people were wary of calling both the hooks for "an attribute is removed" and "an attribute is set".

dom.bs
```diff
@@ -5194,10 +5217,9 @@ To <dfn export id=concept-element-attributes-set>set an attribute</dfn> given an
 
  <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
 
- <li><p>If <var>oldAttr</var> is non-null, <a lt="remove an attribute">remove</a> it from
- <var>element</var>.
-
- <li><p><a lt="append an attribute">Append</a> <var>attr</var> to <var>element</var>.
+ <li><p>If <var>oldAttr</var> is non-null, <a lt="replace an attribute">replace</a> it
+  by <var>attr</var> in <var>element</var>. Otherwise <a lt="append an attribute">append</a>
+  <var>attr</var> to <var>element</var>.
```
annevk: The otherwise can still be its own step. "Otherwise, append _attr_ to _element_."

dom.bs
```diff
@@ -5151,6 +5151,29 @@ run these steps:
  <li>An <a>attribute is removed</a>.
 </ol>
 
+To <dfn export id=concept-element-attributes-replace lt="replace an attribute">replace</dfn> an
+<a>attribute</a> <var>old attribute</var> by an <a>attribute</a> <var>new attribute</var>
```
annevk: oldAttr / newAttr.

annevk:
We should still decide whether this is the behavior we want as it only affects legacy methods and complicates the specification.

PR #144, opened 2016-01-06
Added an example showing feature detection with supports()

yoavweiss:
In a [PR](https://github.com/w3c/preload/pull/39) for an example of DOMTokenList's supports() based feature detection, @igrigorik suggested that it might be better that the example will live in a more central location. Therefore, I think it would make sense to add it in DOM.

@annevk - can you take a look?


dom.bs
```diff
@@ -8705,6 +8705,28 @@ associated <a for="/">element</a>'s
   <p>Throws a <code>TypeError</code> exception if the associated attribute has no supported tokens defined.
 </dl>
 
+<p>The <a method for=DOMTokenList lt="supports()">supports()</a> method can be used to detect support for {{DOMTokenList}} tokens:
+<pre class=lang-javascript>
+var DOMTokenListSupports = function(tokenList, token) {
+      if (!tokenList || !tokenList.supports) {
```
annevk: 5 spaces indentation?

dom.bs
```diff
@@ -8705,6 +8705,28 @@ associated <a for="/">element</a>'s
...
+          return false;
+      }
+      try {
+          return tokenList.supports(token);
+      } catch (e) {
+          if (e instanceof TypeError) {
+            console.log("The DOMTokenList doesn't have a supported tokens list");
+          } else {
+            console.error("That shouldn't have happened");
+          }
```
annevk: Seems cleaner to not branch here and just log straight away.

dom.bs
```diff
@@ -8705,6 +8705,28 @@ associated <a for="/">element</a>'s
...
+      }
+};
+
+var linkSupportsPreload = tokenListSupports(document.createElement("link").relList, "preload"); // true
```
annevk: Perhaps it's better to leave the actual token here as a variable.

dom.bs
annevk: Also, `console` is not standardized. I guess that doesn't matter.

dom.html
```diff
@@ -704,10 +704,10 @@ <h3 class="heading settled" data-level="3.6" id="interface-eventtarget"><span cl
   is <var>type</var>. The <var>callback</var> argument sets the <b>callback</b> that will
   be invoked when the <a data-link-type="dfn" href="#concept-event">event</a> is <a data-link-type="dfn" href="#concept-event-dispatch">dispatched</a>. 
      <p>The <var>options</var> argument sets listener-specific options. For compatibility this can be just
-  a boolean, in which case the method behaves exactly as if the value was specified as <var>options</var>’ <code>capture</code> member.</p>
-     <p>When set to true, <var>options</var>’ <code>capture</code> member prevents <b>callback</b> from
+  a boolean, in which case the method behaves exactly as if the value was specified as <var>options</var>' <code>capture</code> member.</p>
+     <p>When set to true, <var>options</var>' <code>capture</code> member prevents <b>callback</b> from
```
annevk: Looks like you might need to update bikeshed?

yoavweiss: Just updated it before this patch

dom.bs
yoavweiss: Changing to 4.

dom.bs
yoavweiss: OK, dropping the TypeError check.

yoavweiss:
@annevk - updated the patch according to your comments.

Regarding Bikeshed, it looks up-to-date on my end.

dom.html
annevk: Did you run bikeshed update? Your output looks very wrong...

dom.bs
```diff
@@ -8705,6 +8705,24 @@ associated <a for="/">element</a>'s
...
+<p>The <a method for=DOMTokenList lt="supports()">supports()</a> method can be used to detect support for {{DOMTokenList}} tokens:
+<pre class=lang-javascript>
+var DOMTokenListSupports = function(tokenList, token) {
+    if (!tokenList || !tokenList.supports) {
+        return false;
+    }
+    try {
+        return tokenList.supports(token);
+    } catch (e) {
+          console.log("The DOMTokenList doesn't have a supported tokens list");
```
annevk: Inconsistent indenting.

dom.bs
```diff
@@ -8705,6 +8705,24 @@ associated <a for="/">element</a>'s
...
+    }
+};
+
+var linkSupportsPreload = tokenListSupports(document.createElement("link").relList, supportedToken); // true
```
annevk: When you make this generic, you should probably also make the rest generic and not have `// true`.

Also, the function name used here still doesn't match the one declared earlier on.

dom.bs
```diff
@@ -8705,6 +8705,24 @@ associated <a for="/">element</a>'s
   <p>Throws a <code>TypeError</code> exception if the associated attribute has no supported tokens defined.
 </dl>
 
+<p>The <a method for=DOMTokenList lt="supports()">supports()</a> method can be used to detect support for {{DOMTokenList}} tokens:
+<pre class=lang-javascript>
```
annevk: Needs a newline before `<pre>`

yoavweiss:
@annevk - Addressed your comments and reinstalled bikeshed to get things in order.

dom.bs
```diff
@@ -8705,6 +8705,25 @@ associated <a for="/">element</a>'s
   <p>Throws a <code>TypeError</code> exception if the associated attribute has no supported tokens defined.
 </dl>
 
+<p>The <a method for=DOMTokenList lt="supports()">supports()</a> method can be used to detect support for {{DOMTokenList}} tokens:
+
+<pre class=lang-javascript>
+var DOMTokenListSupports = function(tokenList, token) {
+    if (!tokenList || !tokenList.supports) {
+        return false;
```
annevk: Why return false here actually and not just undefined? Seems a bit weird.

yoavweiss: fair point. Will change

dom.bs
```diff
@@ -8705,6 +8705,25 @@ associated <a for="/">element</a>'s
   <p>Throws a <code>TypeError</code> exception if the associated attribute has no supported tokens defined.
 </dl>
 
+<p>The <a method for=DOMTokenList lt="supports()">supports()</a> method can be used to detect support for {{DOMTokenList}} tokens:
+
+<pre class=lang-javascript>
+var DOMTokenListSupports = function(tokenList, token) {
```
domenic: Minor tweak: `function domTokenListSupports(tokenList, token) {`

dom.bs
```diff
@@ -8705,6 +8705,25 @@ associated <a for="/">element</a>'s
...
+    if (!tokenList || !tokenList.supports) {
+        return;
+    }
+    try {
+        return tokenList.supports(token);
+    } catch (e) {
+        console.log("The DOMTokenList doesn't have a supported tokens list");
```
domenic: I'm not sure why we are catching this exception, and showing an example of people using `domTokenListSupports` on classList. We should not show examples of people doing wrong things, and definitely should not encourage them to write error-handling code for such programmer errors.

I'd remove the try/catch, and maybe if you still want to let them know that classList will throw, you should just add a comment at the end of that line like "will throw since classList has no supported tokens list"

dom.bs
```diff
@@ -8705,6 +8705,25 @@ associated <a for="/">element</a>'s
...
+    }
+};
+
+var linkSupportsToken = DOMTokenListSupports(document.createElement("link").relList, supportedToken);
+var classListSupportsFoo = DOMTokenListSupports(document.createElement("div").classList, "foo"); // undefined + console.log
+var linkSupportsFoo = DOMTokenListSupports(document.createElement("link").relList, "foo"); // false
```
domenic: I don't see a need for two examples for link. JavaScript developers know how to use variables.

yoavweiss:
@domenic - addressed your comments

dom.bs
```diff
@@ -8705,6 +8705,20 @@ associated <a for="/">element</a>'s
   <p>Throws a <code>TypeError</code> exception if the associated attribute has no supported tokens defined.
 </dl>
 
+<p>The <a method for=DOMTokenList lt="supports()">supports()</a> method can be used to detect support for {{DOMTokenList}} tokens:
+
+<pre class=lang-javascript>
+function DOMTokenListSupports(tokenList, token) {
```
domenic: lowercase function names that are not classes, so `domTokenListSupports`

dom.bs
```diff
@@ -8705,6 +8705,20 @@ associated <a for="/">element</a>'s
...
+
+<pre class=lang-javascript>
+function domTokenListSupports(tokenList, token) {
+    if (!tokenList || !tokenList.supports) {
+        return;
+    }
+    return tokenList.supports(token);
+};
+
+var linkSupportsToken = DOMTokenListSupports(document.createElement("link").relList, token);
```
domenic: the casing here does not match the declaration

domenic:
In general not sure about the pattern of including code for non-compliant browsers in the spec (i.e. browsers that don't implement .supports) but I'll leave that up to @annevk.

dom.bs
```diff
@@ -8724,6 +8724,20 @@ associated <a for="/">element</a>'s
...
 
+<p>The <a method for=DOMTokenList lt="supports()">supports()</a> method can be used to detect support for {{DOMTokenList}} tokens:
+
+<pre class=lang-javascript>
+function domTokenListSupports(tokenList, token) {
+    if (!tokenList || !tokenList.supports) {
+        return;
+    }
+    return tokenList.supports(token);
+};
```
domenic: @annevk's specs don't use semicolons. Need to fix lines 8732, 8734, 8735, 8737, 8738.

yoavweiss: removed

annevk:
@domenic makes a good point, I'm not sure about that either. It will be dated advice a couple years from now. I don't think we should be adding this.

yoavweiss:
I'm not sure I follow: does this refer to the entire example or to `if (!tokenList || !tokenList.supports) return` part?

annevk:
Everything that assumes you cannot invoke `supports()` directly.

domenic:
I think the example becomes

```js
// Will return true or false depending on whether the user agent implements the given token for <link rel>
document.createElement("link").relList.supports(token)

// Will always throw a TypeError since the class attribute has no supported tokens list
document.createElement("div").classList.supports(token)
```

annevk:
Also, if we go down that route I think it would be better if we put this inside the `<dl>` or directly after the normative `supports()` algorithm. It also needs to use `class=example`. Either on the `<pre>` or if that is preceded by a paragraph, through a surrounding `<div>`.

yoavweiss:
I wonder what the added value of that example would be. The point of the example from my perspective is to show developers how to use the feature for feature detection, which realistically means that they would need to detect support for `supports()` as well as the existence of the DOMTokenList their trying to test on. Otherwise the example is trivial.

Maybe we should just make sure that realistic examples are available in the realm of blog posts/explainers and leave it out of the spec?

annevk:
The example would indeed basically amount to, "this is how you use this method". Those can be useful to developers, since standards use IDL which is not quite what they are used to. I'm happy to leave it out though if you don't think it's of much value.

zcorpan:
I think we should leave this out. The spec already has this

> tokenlist . supports(token)
> Returns true if token is in the associated attribute’s supported tokens. Returns false otherwise.
>
> Throws a TypeError exception if the associated attribute has no supported tokens defined.

which has the same information as the example without legacy consideration.

yoavweiss:
Fair enough. Closing

PR #151, opened 2016-01-19
Fix #146: Only run adopting steps when changing documents

### LINKED ISSUE #146: Only run "adopting steps" if oldDocument !== document
https://dom.spec.whatwg.org/#concept-node-adopt-ext

> For each descendant in node’s inclusive descendants, in tree order, run the adopting steps with descendant and oldDocument\.

@jdm found in https://github.com/w3c/web-platform-tests/issues/2466 that this doesn't have the intended behavior for `img` -- it's intended to rerun "update the image data" when the `img` changes node document, but not when it's moved around in the same document (except when moved into or out of a `picture`).

https://html.spec.whatwg.org/multipage/embedded-content.html#relevant-mutations

@annevk suggested in \#whatwg

> perhaps the "adopting steps" should only run if the document actually changes (and otherwise you just get the removal notification)

This would fix `img` and wouldn't regress `template`, AFAICT.

Thoughts?

cc @yoavweiss @smaug---- @domenic

zcorpan:
This doesn't change the rest of the "adopt" steps.

dom.html not included.

Issue: #146

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4619,13 +4619,18 @@ a <var>document</var>, run these steps:
  <li><p>If <var>node</var>'s <a>parent</a> is not null, <a>remove</a> <var>node</var> from its
  <a>parent</a>.
 
- <li><p>For each <var>descendant</var> in <var>node</var>'s <a>inclusive descendants</a>, in
- <a>tree order</a>, set <var>descendant</var>'s <a>node document</a> to <var>document</var>.
- <!--AttrExodus as well as any associated {{Attr}} nodes-->
+ <li>
+  <p>If <var>document</var> is not the same as <var>oldDocument</var>:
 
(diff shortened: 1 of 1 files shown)


annevk:
I think it would be better to make step 3 and 4 substeps of a new step 3 that does that comparison and then runs the substeps, if necessary.

zcorpan:
Done.

annevk:
I cannot merge this until @tabatkins fixes bikeshed. It's become a big whack-a-mole game.

annevk:
Thank you, committed as https://github.com/whatwg/dom/commit/bb69d0b0b280bf283550f2c21cabd0e6aa1d4cfc.

annevk:
Oh, realized just now you created a local branch I could have rebased. Sorry.

PR #152, opened 2016-01-22
Lock a few IDs into place, so a recent Bikeshed update wont' change them.

tabatkins:
The only dom.html changes are some Bikeshed updates to the boilerplate markup, which have no visual effect on the spec.


annevk:
Thank you.

PR #155, opened 2016-02-02
Make addEventListener throw for service worker optimization

jungkees:
To optimize the service worker execution behavior, this change makes
addEventListener throw when it is called after the very first
evaluation of the service worker script.

SW issue: https://github.com/slightlyoff/ServiceWorker/issues/718

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1065,6 +1065,15 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
 method, when invoked, must run these steps:
 
 <ol>
+ <li>
+  <p>If the <a>context object</a>'s global object is a {{ServiceWorkerGlobalScope}} object and its
+  associated <a>service worker</a>'s <a>script resource</a>'s <a>has ever been evaluated flag</a> is
+  set, <a>throw</a> a <code>TypeError</code>. [[!SW]]
+
+  <p class="note no-backref">To optimize storing the event types allowed for the service worker and
+  to avoid non-deterministic changes to the event listeners, invocation of the method is allowed
+  only during the very first evaluation of the service worker script.
+
  <li><p>If <var>callback</var> is null, terminate these steps.
 
  <li><p>Let <var>capture</var> and <var>passive</var> be the result of <a>flattening</a>
@@ -1082,6 +1091,10 @@ method, when invoked, must run these steps:
 method, when invoked, must, run these steps
 
 <ol>
+ <li><p>If the <a>context object</a>'s global object is a {{ServiceWorkerGlobalScope}} object and
+ its associated <a>service worker</a>'s <a>script resource</a>'s <a>has ever been evaluated flag</a>
+ is set, <a>throw</a> a <code>TypeError</code>. [[!SW]]
+
  <li><p>Let <var>capture</var> and <var>passive</var> be the result of <a>flattening</a>
  <var>options</var>.
 


jungkees:
@annevk I meant to use the following xrefs:

- ServiceWorkerGlobalScope: https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#service-worker-global-scope-interface
- ServiceWorkerGlobalScope's associated service worker: https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#dfn-service-worker-global-scope-service-worker
- service worker's has ever been evaluated flag: https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#dfn-has-ever-been-evaluated-flag

jungkees:
Ah.. I have to change a bit. The _has ever been evaluated flag_ is not on _service worker_ but on its _script resource_. Will push the change.

dom.bs
```diff
@@ -1065,6 +1065,15 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
 method, when invoked, must run these steps:
 
 <ol>
+ <li>
+  <p>If the global object is a {{ServiceWorkerGlobalScope}} object and its associated
```
annevk: This should be context object's global object.

I guess doing it this way is okay. We don't need a generic solution yet. But once we do, you might have to update service workers again.

dom.bs
```diff
@@ -1065,6 +1065,15 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
...
+  <a>service worker</a>'s <a>script resource</a>'s <a>has ever been evaluated flag</a> is set,
+  <a>throw</a> an {{InvalidStateError}} exception. [[!SW]]
```
annevk: I prefer just using TypeError.

dom.bs
```diff
@@ -1065,6 +1065,15 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
...
+
+  <p class="note no-backref">To optimize the service worker execution behavior, the user agent
+  throws an {{InvalidStateError}} exception when the method is called after the very first
+  evaluation of the service worker script.
```
annevk: The optimization is storing the listeners. The main reason we are doing this is to make it clear you need to add them at a certain time for them to be meaningful.

annevk:
One other thought here, we should do this for `removeEventListener` too, no?

dom.bs
jungkees: Addressed.

dom.bs
jungkees: Addressed.

dom.bs
jungkees: I revised the note as such.

jungkees:
> One other thought here, we should do this for removeEventListener too, no?

Agreed. Addressed. The point is the event listeners should be deterministically set in the very first run. Removing one in the later runs where there's no chance to add it back seems not reasonable.

I pushed the changes.

dom.bs
```diff
@@ -1065,6 +1065,15 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
 method, when invoked, must run these steps:
 
 <ol>
+ <li>
+  <p>If the <a>context object</a>'s global object is a {{ServiceWorkerGlobalScope}} object and its
+  associated <a>service worker</a>'s <a>script resource</a>'s <a>has ever been evaluated flag</a> is
+  set, <a>throw</a> a <code>TypeError</code>. [[!SW]]
+
+  <p class="note no-backref">To optimize storing the event types allowed for the service worker and
+  to avoid non-deterministic additions of the event listeners, the user agent allows the invocation
```
annevk: s/additions of/changes to/
s/the user agent allows the invocation of the method/invocation of the method is allowed/

dom.bs
```diff
@@ -1065,6 +1065,15 @@ seen from the definition above, an <a>event listener</a> is a more broad concept
 method, when invoked, must run these steps:
 
 <ol>
+ <li>
+  <p>If the <a>context object</a>'s global object is a {{ServiceWorkerGlobalScope}} object and its
+  associated <a>service worker</a>'s <a>script resource</a>'s <a>has ever been evaluated flag</a> is
+  set, <a>throw</a> a <code>TypeError</code>. [[!SW]]
```
annevk: I'm a little bit afraid bikeshed won't pick up on these references if I try to build it. Do you happen to have bikeshed installed and could you try? If not, I'll figure it out.

annevk:
The commit message should be updated as well to account for the changes to `removeEventListener()`.

dom.bs
jungkees: Addressed.

dom.bs
jungkees: Your help would be appreciated!

annevk:
Thank you!

annevk:
(Had to make a few changes for bikeshed. While there I also changed a couple of things to use HTTPS.)

jungkees:
Thanks!

Issue #156, opened 2016-02-03
Spec dispatching prefixed animation and transition events


miketaylr:
cf. https://github.com/whatwg/compat/issues/24

Let me take a stab at a PR.

PR #158, opened 2016-02-06
Issue #156. Add legacy event type check to invoke algorithm.

miketaylr:
Not sure if I should link up the event type names to the transitions and animations specs, or just having them be strings is preferable.

See here for what Gecko does: https://mxr.mozilla.org/mozilla-central/source/dom/events/EventListenerManager.cpp#1237

And here for what Blink does (which is inherited from WebKit): https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/core/events/EventTarget.cpp&sq=package:chromium&l=351&dr=C&rcl=1452688339

r? @annevk

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1192,8 +1192,32 @@ for discussion).
    <li><p>If <var>event</var>'s <a>stop immediate propagation flag</a> is set, terminate the
    <a>invoke</a> algorithm.
 
-   <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s
-   <b>type</b>, terminate these substeps (and run them for the next <a>event listener</a>).
+   <li><p>Let <var>original event type</var> be <var>event</var>'s {{Event/type}} attribute value.
+
+   <li>
+    <p>If <var>original event type</var> is not <var>listener</var>'s <b>type</b>, run these
+    subsubsteps:
+
+    <ol>
+     <li>
+      <p>If <var>original event type</var> is an <a>ASCII case-insensitive</a> match for any of the
+      strings in the first column in the following table, set <var>event</var>'s {{Event/type}}
+      attribute value to the string in the second column on the same row as the matching string:
+
+      <table>
+       <thead>
+        <tr><th>Event Type<th>Legacy Event Type
+       <tbody>
+        <tr><td>"<code>animationend</code>"<td>"<code>webkitAnimationEnd</code>"
+        <tr><td>"<code>animationiteration</code>"<td>"<code>webkitAnimationIteration</code>"
+        <tr><td>"<code>animationstart</code>"<td>"<code>webkitAnimationStart</code>"
+        <tr><td>"<code>transitionend</code>"<td>"<code>webkitTransitionEnd</code>"
+      </table>
+
+     <li><p>If <var>event</var>'s {{Event/type}} attribute value is <var>listener</var>'s
+     <b>type</b>, continue, else terminate these substeps (and run them for the next
+     <a>event listener</a>).
+    </ol>
 
    <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}
    and <var>listener</var>'s <b>capture</b> is false, terminate these substeps (and run them for the
@@ -1211,6 +1235,10 @@ for discussion).
    <a>callback this value</a>. If this throws any exception, <a>report the exception</a>.
 
    <li><p>Clear <var>event</var>'s <a>in passive listener flag</a>.
+
+   <li>If <var>event</var>'s {{Event/type}} attribute value is not equal to
+   <var>original event type</var>, set <var>event</var>'s {{Event/type}} attribute value to
+   <var>original event type</var>.
   </ol>
 </ol>
 
--- a/dom.html
+++ b/dom.html
@@ -220,9 +220,9 @@
   "DOM Event Architecture" and "Basic Event Interfaces" chapters of <cite>DOM Level 3 Events</cite> <a data-link-type="biblio" href="#biblio-uievents-20031107">[uievents-20031107]</a> (specific types of events do not
   belong in the DOM Standard), and <cite>DOM Level 2 Traversal and Range</cite> <a data-link-type="biblio" href="#biblio-dom-level-2-traversal-range">[DOM-Level-2-Traversal-Range]</a>, and: </p>
      <ul class="brief">
-      <li>Aligning them with the JavaScript ecosystem where possible. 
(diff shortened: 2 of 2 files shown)


dom.html
```diff
@@ -2,7 +2,7 @@
  <head>
   <meta charset="utf-8">
   <title>DOM Standard</title>
-  <link href="https://www.whatwg.org/style/specification" rel="stylesheet">
+  <link href="https://resources.whatwg.org/standard.css" rel="stylesheet">
```
miketaylr: Uh, bikeshed made me do it (AFAICT, it's the same content...).

dom.bs
```diff
@@ -1217,7 +1217,32 @@ for discussion).
    <a>invoke</a> algorithm.
 
    <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s
-   <b>type</b>, terminate these substeps (and run them for the next <a>event listener</a>).
+   <b>type</b>, check for a legacy type attribute match by running these substeps:
+
+   <ol>
+    <li><p>If <var>event</var>'s {{Event/type}} attribute value is "animationend":
```
miketaylr: Maybe it's better to make this more DRY and say something like if value is one of "animationend", "animationiteration", "animationstart", "transitionend"...

Let <var>original event type</var> be "animationstart"
Change event type to "webkit" + <var>original event type</var>

Or something like that.

dom.bs
```diff
@@ -1217,7 +1217,32 @@ for discussion).
    <a>invoke</a> algorithm.
 
    <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s
-   <b>type</b>, terminate these substeps (and run them for the next <a>event listener</a>).
+   <b>type</b>, check for a legacy type attribute match by running these substeps:
```
annevk: Just say "run these substeps". No need to have an explanation here, though you could have it in a note after the steps.

dom.bs
```diff
@@ -1217,7 +1217,32 @@ for discussion).
...
+
+   <ol>
+    <li><p>If <var>event</var>'s {{Event/type}} attribute value is "animationend":
+    <ol>
+     <li><p>Let <var>original event type</var> be "animationend".
```
annevk: Introducing a variable in substeps seems suspect with respect to scoping. Best to introduce it earlier and just set it here.

dom.bs
```diff
@@ -1217,7 +1217,32 @@ for discussion).
...
+     <li><p>Change the <var>event</var>'s {{Event/type}} attribute value to "webkitanimationend".
```
annevk: It's just `<var>event</var>`, not `the <var>event</var>`.

dom.bs
```diff
@@ -1217,7 +1217,32 @@ for discussion).
...
+    </ol>
+    <li><p>If <var>event</var>'s {{Event/type}} attribute value is "animationiteration":
+    <ol>
+     <li><p>Let <var>original event type</var> be "animationiteration".
+     <li><p>Change the <var>event</var>'s {{Event/type}} attribute value to "webkitanimationiteration".
+    </ol>
+    <li><p>If <var>event</var>'s {{Event/type}} attribute value is "animationstart":
```
annevk: If the `<li>` has multiple children they need to start on their own line and be indented with a space. (Also applies earlier and later. Put the comment here for some breathing room.)

dom.bs
```diff
@@ -1217,7 +1217,32 @@ for discussion).
...
+    <ol>
+     <li><p>Let <var>original event type</var> be "animationstart".
+     <li><p>Change the <var>event</var>'s {{Event/type}} attribute value to "webkitanimationstart".
+    </ol>
+    <li><p>If <var>event</var>'s {{Event/type}} attribute value is "transitionend":
+    <ol>
+     <li><p>Let <var>original event type</var> be "transitionend".
+     <li><p>Change the <var>event</var>'s {{Event/type}} attribute value to "webkittransitionend".
```
annevk: Please put newlines between the `<li>`s. (Though not before the first and after the last `<li>` within an `<ol>`.)

dom.html
annevk: That seems fine. I guess I should update bikeshed.

miketaylr:
Thanks for the feedback!

miketaylr:
Want to have a look again @annevk? 

(I plan on squashing commits when/if this gets r+, btw)

annevk:
Also, I liked your suggestion of just making it a list so if we do this for more events we just have to add something to the list.

Hope that's alright.

miketaylr:
Sounds good, will make edits and we'll try again.

I also just realized the event names need to be camel-cased: `webkitAnimationEnd`, etc.

annevk:
Ah okay, in that case maybe using a mapping table is easier? If you need inspiration, e.g., https://dom.spec.whatwg.org/#dom-document-createevent has a mapping table.

miketaylr:
Perfect. I'll use something like that.

miketaylr:
OK - f89d534 introduces <var>original event type</var> earlier and  	6c52497 uses a table to map from standard to gross events types. (If this gets a thumbs up I'll fixup everything to a single commit)

miketaylr:
(just went ahead and squashed because why not)

dom.bs
```diff
@@ -1192,8 +1192,32 @@ for discussion).
...
+       <tbody>
+        <tr><td>"<code>animationend</code>"<td>"<code>webkitAnimationEnd</code>"
+        <tr><td>"<code>animationiteration</code>"<td>"<code>webkitAnimationIteration</code>"
+        <tr><td>"<code>animationstart</code>"<td>"<code>webkitAnimationStart</code>"
+        <tr><td>"<code>transitionend</code>"<td>"<code>webkitTransitionEnd</code>"
+      </table>
+
+     <li><p>If <var>event</var>'s {{Event/type}} attribute value is <var>listener</var>'s
+     <b>type</b>, continue, else terminate these substeps (and run them for the next
+     <a>event listener</a>).
```
annevk: This logic seems wrong. There's a minor problem with "continue" not being a thing, but you could fix that by inverting the if logic. The bigger problem is that if you terminate the substeps, event's type will still be set to the new value, which might be incorrect for the next event listener.

It seems that whenever you terminate these substeps (and this happens not just here, but also later), event's type needs to be set back to the original event type.

Am I missing something?

annevk: Another way of fixing this might be to match on all cells in the table and then pick the opposing column. If we do it that way that would require a note of sorts explaining the setup.

miketaylr:
I goofed up a rebase, so I'm gonna close here and open a new PR from a fresh branch. :hankey:

PR #168, opened 2016-02-17
Fixes #156. Add legacy event type check to invoke algorithm.

### LINKED ISSUE #156: Spec dispatching prefixed animation and transition events

miketaylr:
OK, let's try this again. 

(This approach doesn't use an internal slot on `Event`, but we could probably replace the `original event type` variable with that if you think it's worth it.)

The basic logic is this:

If the browser finds an unprefixed listener, set the `unprefixed listener flag` to true and do invoke as normal. 

At this point we don't care about unprefixed stuff. (My previous understanding was wrong, I thought we needed to check type for each listener -- we only need to do this until we find an unprefixed match (or if we don't find any match. See https://bug1236979.bmoattachments.org/attachment.cgi?id=8704268 for a test that covers this).

If `unprefixed listener flag` is unset, set `original event type`, run event type through the unprefixed -> prefixed event type table and do invoke just as before.

Final step is to set the event type back, and unset the `unprefixed listener flag` if either of those are set.

r? @annevk

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1193,33 +1193,72 @@ for discussion).
 
  <li><p>Initialize <var>event</var>'s {{Event/currentTarget}} attribute to <var>object</var>.
 
+ <li><p>Let <var>found</var> be the result of running <a>inner invoke</a> <var>object</var> with
+ <var>event</var>.
+
  <li>
-  <p>For each <var>listener</var> in <var>listeners</var>, whose <b>removed flag</b> is unset, run
-  these substeps:
+  <p>If <var>found</var> is false, run these substeps:
 
   <ol>
-   <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s
-   <b>type</b>, terminate these substeps (and run them for the next <a>event listener</a>).
+   <li><p>Let <var>originalEventType</var> be <var>event</var>'s {{Event/type}} attribute value.
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}
-   and <var>listener</var>'s <b>capture</b> is false, terminate these substeps (and run them for the
-   next <a>event listener</a>).
+   <li>
+    <p>If <var>event</var>'s {{Event/type}} attribute value is a match for any of the strings in the
+    first column in the following table, set <var>event</var>'s {{Event/type}} attribute value to
+    the string in the second column on the same row as the matching string, and terminate these
+    substeps otherwise.
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}} and
-   <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
-   <a>event listener</a>).
+    <table>
+     <thead>
+      <tr><th>Event type<th>Legacy event type
+     <tbody>
+      <tr><td>"<code>animationend</code>"<td>"<code>webkitAnimationEnd</code>"
+      <tr><td>"<code>animationiteration</code>"<td>"<code>webkitAnimationIteration</code>"
+      <tr><td>"<code>animationstart</code>"<td>"<code>webkitAnimationStart</code>"
+      <tr><td>"<code>transitionend</code>"<td>"<code>webkitTransitionEnd</code>"
+    </table>
 
-   <li><p>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s
-   <a>in passive listener flag</a>.
+    <li><p><a>Inner invoke</a> <var>object</var> with <var>event</var>.
 
-   <li><p>Call <var>listener</var>'s <b>callback</b>'s {{EventListener/handleEvent()}}, with
-   <var>event</var> as argument and <var>event</var>'s {{Event/currentTarget}} attribute value as
-   <a>callback this value</a>. If this throws any exception, <a>report the exception</a>.
-
-   <li><p>Clear <var>event</var>'s <a>in passive listener flag</a>.
+    <li><p>Set <var>event</var>'s {{Event/type}} attribute value to <var>originalEventType</var>.
   </ol>
 </ol>
 
+<p>To <dfn for="event listener" id=concept-event-listener-inner-invoke>inner invoke</dfn>
+an <var>object</var> with <var>event</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>found</var> be false.
+
+ <li>
+  <p>For each <var>listener</var> in <var>listeners</var>, whose <b>removed flag</b> is unset,
+ run these substeps:
+
+ <ol>
+  <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s <b>type</b>,  terminate these substeps (and run them for the next <a>event listener</a>).
+
+  <li><p>Set <var>found</var> to true.
+
+  <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}
+  and <var>listener</var>'s <b>capture</b> is false, terminate these substeps (and run them for the
+  next <a>event listener</a>).
+
+  <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}} and
+  <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
+  <a>event listener</a>).
+
+  <li><p>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s
+  <a>in passive listener flag</a>.
+
+  <li><p>Call <var>listener</var>'s <b>callback</b>'s {{EventListener/handleEvent()}}, with
+  <var>event</var> as argument and <var>event</var>'s {{Event/currentTarget}} attribute value as
+  <a>callback this value</a>. If this throws any exception, <a>report the exception</a>.
+
+  <li><p>Unset <var>event</var>'s <a>in passive listener flag</a>.
+ </ol>
+
+ <li><p>Return <var>found</var>.
+</ol>
 
 <h3 id=firing-events>Firing events</h3>
 
--- a/dom.html
+++ b/dom.html
@@ -69,7 +69,7 @@
   <div class="head">
    <p data-fill-with="logo"><a class="logo" href="https://whatwg.org/"> <img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-dom.svg"> </a> </p>
    <h1 class="p-name no-ref allcaps" id="title">DOM</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Living Standard — Last Updated <time class="dt-updated" datetime="2016-02-18">18 February 2016</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Living Standard — Last Updated <time class="dt-updated" datetime="2016-02-19">19 February 2016</time></span></h2>
    <div data-fill-with="spec-metadata">
     <dl>
      <dt>Participate:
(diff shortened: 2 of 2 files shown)


dom.bs
```diff
@@ -780,6 +780,7 @@ flags that are all initially unset:
  <li><dfn export for=Event id=initialized-flag>initialized flag</dfn>
  <li><dfn export for=Event id=dispatch-flag>dispatch flag</dfn>
  <li><dfn export for=Event id=in-passive-listener-flag>in passive listener flag</dfn>
+ <li><dfn export for=Event id=has-listener-flag>unprefixed listener flag</dfn>
```
annevk: Why does the id not match the name?

dom.bs
```diff
@@ -1194,6 +1195,8 @@ for discussion).
    <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s
    <b>type</b>, terminate these substeps (and run them for the next <a>event listener</a>).
 
+   <li><p>Set <var>event</var>'s <a>unprefixed listener flag</a> to true.
```
annevk: Flags are set or unset, they're not set to a value.

annevk:
I don't think this approach works. Because if you abort the steps and run them again event's type will have changed and you'll override the variable.

What you want to do is make the loop that is happening in step 5 of invoke reusable. First run it for the normal case. Then if nothing got invoked, run it again with the changed type. Does that make sense?

dom.bs
miketaylr: Just something I forgot to change when renaming.

miketaylr:
> What you want to do is make the loop that is happening in step 5 of invoke reusable. First run it for the normal case. Then if nothing got invoked, run it again with the changed type. Does that make sense?

Yeah, commit incoming.

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
...
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}
-   and <var>listener</var>'s <b>capture</b> is false, terminate these substeps (and run them for the
-   next <a>event listener</a>).
+  <li><p>Set <var>event</var>'s <a>found listener flag</a>.
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}} and
-   <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
-   <a>event listener</a>).
+  <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute the is {{Event/CAPTURING_PHASE}}
```
miketaylr: O_o, fix incoming.

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
...
+
+     <li><p><a>Find</a> listener type matches for <var>object</var>.
+
+     <li><p>Set <var>event</var>'s {{Event/type}} attribute value to <var>original event type</var>.
+   </ol>
+
+ <li><p>Unset <var>event</var>'s <a>found listener flag</a> if set.
+</ol>
+
+<p>To <dfn export for="event listener" id=concept-event-listener-find>find</dfn> listener type
```
miketaylr: Not really married to this "find listener type matches for object" name, suggested changes welcome.

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
 
  <li><p>Initialize <var>event</var>'s {{Event/currentTarget}} attribute to <var>object</var>.
 
+ <li><p><a>Find</a> listener type matches for <var>object</var>.
+
  <li>
-  <p>For each <var>listener</var> in <var>listeners</var>, whose <b>removed flag</b> is unset, run
+  <p>If <var>event</var>'s <a>found listener flag</a> is unset, run
```
annevk: This flag is checked as part of "invoke" only, and unset at the end of "invoke". Given that "invoke" is not invoked recursively, how does this work?

miketaylr: "Find listener type matches for object" runs before the flag check. 

If there's an unprefixed event type match for a listener's type -> set the `found listener` flag (that way we don't bother to do legacy type check)
If there's no event type match for a listener's type, the flag is never set, so we can change the event type to the legacy one before running "Find listener type matches for object" again.

(I don't think there's any point to unsetting the flag at step 7, if that's what you're getting at.)

miketaylr: > (I don't think there's any point to unsetting the flag at step 7, if that's what you're getting at.)

Oh wait, we need to unset it because of bubbling/capturing etc, no?

annevk: Sorry, I think your setup actually works. I had not looked carefully enough. Although which "step 7" do you mean here? Your doubting makes me doubt now. 😊

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
...
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}} and
-   <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
-   <a>event listener</a>).
+ <li><p>Unset <var>event</var>'s <a>found listener flag</a> if set.
+</ol>
 
-   <li><p>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s
-   <a>in passive listener flag</a>.
+<p>To <dfn export for="event listener" id=concept-event-listener-find>find</dfn> listener type
+matches for <var>object</var>, run these steps:
```
annevk: How about "inner invoke"? This is not just finding listeners, it's also invoking them.

dom.bs
```diff
@@ -780,6 +780,7 @@ flags that are all initially unset:
  <li><dfn export for=Event id=initialized-flag>initialized flag</dfn>
  <li><dfn export for=Event id=dispatch-flag>dispatch flag</dfn>
  <li><dfn export for=Event id=in-passive-listener-flag>in passive listener flag</dfn>
+ <li><dfn export for=Event id=found-listener-flag>found listener flag</dfn>
```
annevk: I was thinking, if "inner invoke" / "find" just returns a boolean, we don't need extra state on the event. I think that would be better.

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
...
-   next <a>event listener</a>).
+   <li>
+    <p>If <var>event</var>'s {{Event/type}} attribute value is an <a>ASCII case-insensitive</a>
+     match for any of the strings in the first column in the following table, set <var>event</var>'s
+     {{Event/type}} attribute value to the string in the second column on the same row as the
+     matching string, otherwise, terminate these substeps.
+
+    <table>
+     <thead>
+      <tr><th>Event Type<th>Legacy Event Type
```
annevk: Event type/Legacy event type

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
...
   <ol>
-   <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s
-   <b>type</b>, terminate these substeps (and run them for the next <a>event listener</a>).
+   <li><p>Set <var>original event type</var> to <var>event</var>'s {{Event/type}} attribute value.
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}
-   and <var>listener</var>'s <b>capture</b> is false, terminate these substeps (and run them for the
-   next <a>event listener</a>).
+   <li>
+    <p>If <var>event</var>'s {{Event/type}} attribute value is an <a>ASCII case-insensitive</a>
```
annevk: ASCII case-insensitive seems like copypasta.

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
...
+  <a>event listener</a>).
 
+  <li><p>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s
+  <a>in passive listener flag</a>.
+
+  <li><p>Call <var>listener</var>'s <b>callback</b>'s {{EventListener/handleEvent()}}, with
+  <var>event</var> as argument and <var>event</var>'s {{Event/currentTarget}} attribute value as
+  <a>callback this value</a>. If this throws any exception, <a>report the exception</a>.
+
+  <li><p>Clear <var>event</var>'s <a>in passive listener flag</a>.
```
annevk: While you're here, please s/Clear/Unset/.

dom.bs
```diff
@@ -1186,33 +1187,70 @@ for discussion).
...
+
+    <li><p><a>Find</a> listener type matches for <var>object</var>.
+
+    <li><p>Set <var>event</var>'s {{Event/type}} attribute value to <var>original event type</var>.
+  </ol>
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}} and
-   <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
-   <a>event listener</a>).
+ <li><p>Unset <var>event</var>'s <a>found listener flag</a> if set.
```
annevk: No need for "if set".

dom.bs
miketaylr: Ah, didn't realize you had https://dom.spec.whatwg.org/#strings already defined.

dom.bs
miketaylr: Heh, me doubting is just the normal state of affairs. 

I was referring to step 7 of "invoke": "Unset event’s found listener flag if set. ". But if we return a boolean like you suggested above, we can kill the flag.

dom.bs
annevk: Either way, event types are case-sensitive, no?

dom.bs
miketaylr: Like this better.

dom.bs
miketaylr: Gotcha, will do.

dom.bs
miketaylr: Yeah, I like this better too.

dom.bs
miketaylr: Oh. yes. >_<

So we don't need the case insensitive checking.

miketaylr: Yeah, "Unless otherwise stated, string comparisons are done in a case-sensitive manner." Cool.

miketaylr:
OK, made suggested tweaks (and thanks for all the suggestions thus far @annevk).

dom.bs
```diff
@@ -1186,33 +1186,70 @@ for discussion).
 
  <li><p>Initialize <var>event</var>'s {{Event/currentTarget}} attribute to <var>object</var>.
 
+ <li><p><a>Inner invoke</a> <var>object</var> with <var>event</var>.
```
annevk: "Let _found_ be the result of running inner invoke ..."

dom.bs
```diff
@@ -1186,33 +1186,70 @@ for discussion).
...
+     <thead>
+      <tr><th>Event type<th>Legacy event type
+     <tbody>
+      <tr><td>"<code>animationend</code>"<td>"<code>webkitAnimationEnd</code>"
+      <tr><td>"<code>animationiteration</code>"<td>"<code>webkitAnimationIteration</code>"
+      <tr><td>"<code>animationstart</code>"<td>"<code>webkitAnimationStart</code>"
+      <tr><td>"<code>transitionend</code>"<td>"<code>webkitTransitionEnd</code>"
+    </table>
+
+    <li><p><a>Inner invoke</a> <var>object</var> with <var>event</var>
```
annevk: This sentence needs to end with a dot.

dom.bs
```diff
@@ -1186,33 +1186,70 @@ for discussion).
...
+   <li><p>Set <var>original event type</var> to <var>event</var>'s {{Event/type}} attribute value.
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}
-   and <var>listener</var>'s <b>capture</b> is false, terminate these substeps (and run them for the
-   next <a>event listener</a>).
+   <li>
+    <p>If <var>event</var>'s {{Event/type}} attribute value is a match for any of the strings in the
+    first column in the following table, set <var>event</var>'s {{Event/type}} attribute value to
+    the string in the second column on the same row as the matching string, otherwise, terminate
+    these substeps.
```
annevk: "... matching string, and terminate these substeps otherwise."

dom.bs
```diff
@@ -1186,33 +1186,70 @@ for discussion).
...
+
+    <li><p>Set <var>event</var>'s {{Event/type}} attribute value to <var>original event type</var>.
+  </ol>
+</ol>
 
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}} and
-   <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
-   <a>event listener</a>).
+<p>To <dfn export for="event listener" id=concept-event-listener-inner-invoke>inner invoke</dfn>
+an <var>object</var> with <var>event</var>, run these steps:
```
annevk: No need to export this one.

dom.bs
```diff
@@ -1186,33 +1186,70 @@ for discussion).
...
  <li>
-  <p>For each <var>listener</var> in <var>listeners</var>, whose <b>removed flag</b> is unset, run
-  these substeps:
+  <p>If the return value of <a>inner invoke</a> <var>object</var> with <var>event</var> is false,
+  run these substeps:
 
   <ol>
-   <li><p>If <var>event</var>'s {{Event/type}} attribute value is not <var>listener</var>'s
-   <b>type</b>, terminate these substeps (and run them for the next <a>event listener</a>).
+   <li><p>Set <var>original event type</var> to <var>event</var>'s {{Event/type}} attribute value.
```
annevk: s/Set/Let/. s/original event type/originalEventType/.

dom.bs
```diff
@@ -1186,33 +1186,70 @@ for discussion).
...
-   <li><p>If <var>event</var>'s {{Event/eventPhase}} attribute value is {{Event/BUBBLING_PHASE}} and
-   <var>listener</var>'s <b>capture</b> is true, terminate these substeps (and run them for the next
-   <a>event listener</a>).
+<p>To <dfn export for="event listener" id=concept-event-listener-inner-invoke>inner invoke</dfn>
+an <var>object</var> with <var>event</var>, run these steps:
 
-   <li><p>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s
-   <a>in passive listener flag</a>.
+<ol>
+ <li><p>Let <var>listener match</var> be false.
```
annevk: listenerMatch

dom.bs
```diff
@@ -1186,33 +1186,70 @@ for discussion).
...
 
-   <li><p>Call <var>listener</var>'s <b>callback</b>'s {{EventListener/handleEvent()}}, with
-   <var>event</var> as argument and <var>event</var>'s {{Event/currentTarget}} attribute value as
-   <a>callback this value</a>. If this throws any exception, <a>report the exception</a>.
+ <li>
+  <p>For each <var>listener</var> in <var>listeners</var>, whose <b>removed flag</b> is unset,
+ run these substeps:
 
-   <li><p>Clear <var>event</var>'s <a>in passive listener flag</a>.
```
annevk: You dropped this line. We should still have "Unset event's in passive listener flag" after the listener is called.

annevk:
Starting to look good!

dom.bs
miketaylr: Much nicer, I was struggling with how to word this nicely.

dom.bs
miketaylr: Oh, my bad.

miketaylr:
I gotta `git push -f` to fix this merge conflict (just over the Last Updated date), so I'm just gonna squash everything while I'm here.

annevk:
Thank you @miketaylr. Landed as https://github.com/whatwg/dom/commit/a026ccf3efa8d34fb0cac580928f61d18b1e9f76. I added your name to the acknowledgments and ended up tweaking the whitespace a bit more. While ugly, the algorithm is still fairly readable due to your efforts.

miketaylr:
Thanks for the patient help with this @annevk. :balloon: :cake:

PR #173, opened 2016-02-23
clarify insert mutation algo step 6.1

therealglazou:
Step 6.1 of https://dom.spec.whatwg.org/#concept-node-insert is a bit unclear because it does not reference the **insert** and **append** mutation algorithms. This PR adds the links.


therealglazou:
invalid per discussion on IRC with anne and simon

PR #176, opened 2016-02-25
Fix typo in IDL comment for Attr's nodeName

domenic:


annevk:
Oh my, @zcorpan and I need to pay more attention to trivial changes.

zcorpan:
Note to self: avoid doing PRs while switching trains...

PR #178, opened 2016-02-26
Fix #148: Drop "keyevents" as an alias of "keyboardevent" for createEvent

### LINKED ISSUE #148: Drop "keyevents" as an alias of "keyboardevent" for createEvent
There's a test for this in wpt and a proposed fix in Blink:
http://w3c-test.org/dom/nodes/Document-createEvent.html
https://codereview.chromium.org/1569813002/

As discussed in that review, it looks like this alias is only supported by Gecko, and so we think it would be better to drop it from the spec. If it turns out to be unsafe to remove in Gecko, then the only way out is for all other engines to add it.

foolip:


foolip:
I'm afraid I didn't update `dom.html`, it seems like I'm almost never lucky enough to have a Bikeshed version (the latest) that doesn't result in a lot of other changes to the output.

PR #179, opened 2016-02-26
clarify EventTarget.dispatchEvent()

therealglazou:
https://dom.spec.whatwg.org/#dom-eventtarget-dispatchevent item 3 does not say where to dispatch the event while https://dom.spec.whatwg.org/#concept-event-dispatch is expecting at least a target.


annevk:
Thank you, committed as https://github.com/whatwg/dom/commit/59d283ab9c5d2e3dd823bca21eecaaaca402b181.

PR #180, opened 2016-02-28
Properly check event's stop immediate propagation flag in "inner invoke"

nox:


annevk:
How did you avoid the fatal error?

annevk:
We should probably point out that this fixes a regression caused by https://github.com/whatwg/dom/commit/1cce4fd5e39ac48c047d9de3d5efefc0a909db14.

nox:
> How did you avoid the fatal error?

I removed the wolf bits, then resetted the corresponding change in `dom.html`.

nox:
I added a reference to 1cce4fd5e39ac48c047d9de3d5efefc0a909db14 in the message.

annevk:
I see, I think I'll wait for @tabatkins to fix bikeshed to not result in a FATAL ERROR before landing this.

nox:
I used https://github.com/tabatkins/bikeshed/pull/611 and updated the spec with it, though I now realise this PR doesn't pass the tests. :(

annevk:
What tests?

nox:
I mean the Bikeshed one. See the [build log](https://travis-ci.org/tabatkins/bikeshed/builds/112569967).

annevk:
Now https://github.com/tabatkins/bikeshed/issues/609 is not making me happy. Going to postpone this a little longer.

PR #184, opened 2016-03-04
Fix #34: invoke insertion/removing steps for descendants

### LINKED ISSUE #34: insert/remove hooks need to be invoked for ancestor changes too
See https://lists.w3.org/Archives/Public/public-webapps/2015AprJun/0600.html and https://lists.w3.org/Archives/Public/public-webapps/2015AprJun/0572.html in particular.

annevk:


nox:
Let's rename `newNode` to `insertedNode` consistently. LGTM otherwise.

PR #188, opened 2016-03-14
Editorial: clone a range → clone the contents of a range

zcorpan:
See http://logs.glob.uno/?c=freenode%23whatwg&s=14+Mar+2016&e=14+Mar+2016#c988870 for confusion. In particular:

> yeah probably a confusion because of "clone" ; a "clone" operation that does not return the same object type... hmmm...

Also see https://github.com/tabatkins/bikeshed/issues/630 for the bogus change to UIEvents in the generated references.

Ack @therealglazou


annevk:
Thank you, I reverted the ID change. Did not seem worth it to break links.

PR #191, opened 2016-03-18
Specify for='' on the couple of bare <a>root</a> autolinks.

tabatkins:


annevk:
I updated bikeshed, but... I get a warning for `{{Event/Event()/eventInitDict!!argument}}`. I'd be happy to replace that with `<var>eventInitDict</var>` though, seems better.

However, also, CSSOM-VIEW suddenly gets a cousin CSSOM-VIEW-1 reference entry. That's less okay. Couldn't quite figure out why that was happening. Maybe a specref thing, @tobie?

tobie:
Seems to be a W3C thing. https://www.w3.org/TR/cssom-view/ --301--> https://www.w3.org/TR/cssom-view-1/ which Specref automatically reflects since https://github.com/tobie/specref/commit/9a0d8127440912e929d859199fbe428594d9fa53 (March, 17).

annevk:
I'd prefer just getting links to https://drafts.csswg.org/cssom-view/.

annevk:
But at the minimum no duplicates. But maybe that's part bikeshed then.

tobie:
Yeah, I would imagine the dups comes from having Specref for references on one side and Sheperd for cross-linking on on the other. But that might be an incorrect assumption.

annevk:
I have decided to not block on this and filed an issue upstream. Thank you @tabatkins for the fix. Landed as https://github.com/whatwg/dom/commit/150147aae81299bc384d163ffc7d257afb9757f0.

tabatkins:
Yeah, argument autolink handling is being a little bit broken right now (specifically, when you use a `for` value without arguments in the method signature). I'm fixing it as part of generally refactoring autolinking, so feel free to either ignore it or switch it to a plain `<var>` or something.

Will look into the duplicate CSSOM-VIEW stuff. More communication over on the Bikeshed bug.

PR #194, opened 2016-03-23
Fix #193: Shadow: Fix shadow-including root

### LINKED ISSUE #193: "shadow-including root" should be a recursive definition.
Regarding https://github.com/whatwg/dom/commit/7dae99cae42206a09869394e45844958a189937c#diff-a467e681501e456ce8c1ef31425e1b41R5181,

The definition looks incorrect.

That should be:
The shadow-including root of an object is its root's host's shadow-including root,
if the object's root is shadow root, and its root otherwise.

hayatoito:
Regarding #193.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5179,8 +5179,8 @@ or "<code>closed</code>").</p>
 encountered.
 
 <p>The <dfn export id=concept-shadow-including-root>shadow-including root</dfn> of an object is its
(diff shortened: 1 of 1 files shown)


annevk:
Thank you, committed as https://github.com/whatwg/dom/commit/76fbe554e5f444e66e88095970793b29035c4a85.

PR #204, opened 2016-04-06
Upstream Custom Elements

domenic:
Upstreaming from https://rawgit.com/w3c/webcomponents/15a203c8393aef0df7223ab1d43406aa11a7e71e/spec/custom/index.html.

I envision this being done in three stages:

- DOM pull request (i.e. this pull request), which references the above URL for certain concepts
- HTML pull request, which references DOM for certain concepts
- A small DOM pull request, which changes the links to point to HTML instead of the above URL.

Given that we'd like to land this soon, but our timezones are not always super-overlapping, feel free to tweak and merge while I'm asleep, or even just push fixup commits to this branch for me to review. That is, I'd rather we both do the work on this branch, and get this merged in sooner, instead of the traditional submitter-reviewer back and forth.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -23,6 +23,7 @@ Bikeshed problems:
 -->
 
 <script src=https://resources.whatwg.org/file-issue.js async></script>
+<script id=head src=https://resources.whatwg.org/dfn.js defer></script>
 
 <pre class='anchors'>
 urlPrefix: https://encoding.spec.whatwg.org/
@@ -95,6 +96,22 @@ urlPrefix: https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#
         text: service worker
         text: script resource
         text: has ever been evaluated flag
+urlPrefix: https://rawgit.com/w3c/webcomponents/15a203c8393aef0df7223ab1d43406aa11a7e71e/spec/custom/index.html#
+    type: selector; url: dfn-defined; text: :defined
+    type: dfn; urlPrefix: dfn-
+        text: look up a custom element definition; url: look-up-custom-element-definition
+        text: type extension
+        text: name; url: element-definition-name; for: custom element definition
+        text: local name; url: element-definition-local-name; for: custom element definition
+        text: constructor; url: element-definition-constructor; for: custom element definition
+        text: upgrade an element; url: upgrade-a-custom-element
+        text: enqueue a custom element upgrade reaction; url: enqueue-upgrade
+        text: enqueue a custom element callback reaction; url: enqueue-lifecycle-callback
+        text: valid custom element name
+        text: try to upgrade an element; url: try-upgrade
+urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
+    text: Construct; url: sec-construct; type: abstract-op
+    text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror; type: exception
 </pre>
 
 <pre class='link-defaults'>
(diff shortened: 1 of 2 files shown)


annevk:
This looks pretty good. Are we renaming "type extension" later on?

domenic:
Yeah, I figure we can leave that for later.

annevk:
Also, do you prefer landing this as four commits? I'll try get this merged tomorrow.

domenic:
I think it's up to you whether you think it'd be useful as four commits or one. Maybe they should be one, e.g. it's a bit weird to introduce "create an element" in the first commit and only start using it in the second and fourth commits. Certainly it'll look nicer as a single tweet with a nice headline like "Add support for custom elements" instead of these jargon-ish commit messages :).

rniwa:
Can we not merge this into DOM yet?  The spec is still changing quite rapidly, and there are multiple new features being proposed here and there.  I'd prefer having a separate document that contains the algorithms and concepts related to a custom element.

rniwa:
Apparently http://services.w3.org/htmldiff?doc1=https%3A%2F%2Fdom.spec.whatwg.org%2F&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fwhatwg%2Fdom%2Fcustom-elements%2Fdom.html tracks the diff only relevant for custom elements my comment is no longer relevant.

domenic:
Just for the record that particular URL will stop working after the merge, but this one will continue working: http://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fwhatwg%2Fdom%2Fcd643f8d142717580a9a913fa1fb63a27eb231de%2Fdom.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fwhatwg%2Fdom%2Fcustom-elements%2Fdom.html

I've added that link to https://github.com/whatwg/dom/wiki so that it's easy to remember where to find it, at least, instead of having to go back to this bug thread or IRC logs.

dom.bs
```diff
@@ -4602,15 +4679,24 @@ invoked, must run these steps:
...
- with no attributes,
- <a for=Element>namespace</a> set to the <a>HTML namespace</a>,
- <a for=Element>local name</a> set to <var>localName</var>, and
- <a>node document</a> set to the <a>context object</a>.
+ <li>Let <var>element</var> be the result of <a>creating an element</a> given the
+ <a>context object</a>, <var>localName</var>, null, the <a>HTML namespace</a>, <var>is</var>, and
+ with the <var>synchronous custom elements</var> flag set. Rethrow any exceptions.
+
+ <li>If <var>is</var> is non-null, then <a>set an attribute value</a> for <var>element</var> using
+ "<code>is</code>" and <var>is</var>.
```
annevk: This means you get an `attributeChangedCallback` for `is`, right? And queued mutation observers if you had set those up on yourself. That's what we want?

annevk:
For that diff to remain working we need to be careful not to remove the branch. @domenic I think this is fine, modulo that question, I'll let you do the honors.

sideshowbarker:
> For that diff to remain working we need to be careful not to remove the branch.

And for you to be careful to not remove it from your clone (and better yet, back it up somewhere)—so you can re-push to here it later, in case anybody does ever accidentally remove it.

sideshowbarker:
> For that diff to remain working we need to be careful not to remove the branch.

I also just now made the `custom-elements` branch protected in the same way master is—so that it can’t actually be deleted without accidentally.

sideshowbarker:
> I also just now made the `custom-elements` branch protected in the same way master is—so that it can’t actually be deleted without accidentally.

But if I did that prematurely please go into the settings and un-set it for now (if you need to still be able to force-push to it)—until such time as its in whatever final state you want it in.

dom.bs
domenic: Yeah, I am pretty sure it is. Originally I treated is="" very specially and tried to exclude it from attributeChangedCallback, mutation observers, etc. but it was a big pain for no real gain. This model is much simpler. See https://github.com/w3c/webcomponents/commit/5c4cc14e5f973d81ef9ab672b12951fef2d90471 commit message.

annevk:
\o/

domenic:
Woohoo, merged! I also updated https://github.com/whatwg/dom/wiki with some more resources that @rniwa and others might find helpful, in addition to the HTML diff.

PR #213, opened 2016-04-10
Change .createElement() namespace to match UAs

ayg:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=19431

See comments 12 and 20 there.  Gecko tried switching to the spec's
current wording and immediately had to back out due to extension
breakage.  It does not seem likely the current spec is implementable.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4689,8 +4689,12 @@ invoked, must run these steps:
  <li>If <var>is</var> is non-null and <var>definition</var> is null, then <a>throw</a> a
  {{NotFoundError}}.
 
+ <li>Let <var>namespace</var> be the <a>HTML namespace</a> if the <a>context
+ object</a>'s <a for=Document>content type</a> is "<code>text/html</code>" or
+ "<code>application/xhtml+xml</code>", and null otherwise.
+
  <li>Let <var>element</var> be the result of <a>creating an element</a> given the
- <a>context object</a>, <var>localName</var>, null, the <a>HTML namespace</a>, <var>is</var>, and
+ <a>context object</a>, <var>localName</var>, null, <var>namespace</var>, <var>is</var>, and
  with the <var>synchronous custom elements</var> flag set. Rethrow any exceptions.
 
  <li>If <var>is</var> is non-null, then <a>set an attribute value</a> for <var>element</var> using


annevk:
As far as I can tell `image/svg+xml` indeed gives null. That question did not seem answered in the bug.

@domenic since you touched this last, any concerns?

domenic:
This is bizarre, but I guess it's what every browser does. I don't think extension-related breakage is a good reason (extensions are vendor-specific and not part of the web), but if this is what everyone implements that's a good enough reason I guess.

Here is what Blink does: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/core/dom/Document.cpp&q=createElement&sq=package:chromium&l=716 The setting of "document class flags" is done pretty far away in the code base, but maybe .contentType is a good enough proxy.

annevk:
@foolip can you confirm for Blink?

foolip:
Yes, I checked yesterday and pondered whether `isXHTMLDocument() || isHTMLDocument()` is exactly the same as checking at `contentType`. I've poked around and can't find any exceptions, so it's probably close enough, but I could have missed something.

The `isXHTMLDocument() || isHTMLDocument()` test is used in a few other places that are worth comparing with:
 * The `document.title` setter, where [the spec](https://html.spec.whatwg.org/#document.title) instead just looks at the elements in the document.
 * For `execCommand` and related APIs, which throws an "execCommand is only supported on HTML documents" exception, but [the spec](https://w3c.github.io/editing/execCommand.html#execcommand()) instead talks about the `HTMLDocument` interface which is gone from the spec.
 * For `createContextualFragment`, in an optimization that isn't in [the spec](https://w3c.github.io/DOM-Parsing/#widl-Range-createContextualFragment-DocumentFragment-DOMString-fragment).

Having the concept of "HTML or XHTML document" might be useful.

ayg:
@annevk What question didn't seem answered in the bug?

annevk:
@ayg the one about `image/svg+xml`.

annevk:
@foolip yeah, if we change some of those APIs to use these concepts in the specifications we should consider introducing it I think. Something to keep in mind.

ayg:
@annevk http://w3c-test.org/dom/nodes/Document-createElement-namespace.html seems to test image/svg+xml, and comment 12 seems to indicate that browsers matched the proposed spec.

annevk:
Oh, I missed that it was tested, thanks. Anyway, I was already fine with this change 😊. I'll land it later today.

ayg:
It occurs to me that one peculiarity of using content type is that if you use createDocument() to create an (X)HTML document, it will have a content type of application/xml per spec, which maybe is not what we want?  Is there a better way to say "HTML or XHTML document"?  Chrome's createDocument() actually seems to set contentType based on the namespace, so it will behave as expected here.  Perhaps it would make sense to change createDocument() to match Chrome, and then this issue goes away.

foolip:
Good point @ayg, I remember now that this has come up before, but I'd forgotten it. Per spec "application/xml" is the default and it's never overridden, but Blink indeed does special case the XHTML and SVG namespaces. (So much about Document and its subinterfaces is subtly wrong per spec, so much to sort out...)

ArkadiuszMichalski:
Gecko goes even further and for `createDocument()` takes into account the passing DTD https://bugzilla.mozilla.org/show_bug.cgi?id=520969 (probably the only engine that does so).

annevk:
Landed as https://github.com/whatwg/dom/commit/c8ae9cbd46ca3175e07b7f205ebe375805013119. Thanks @ayg!

ArkadiuszMichalski:
Any infos what Edge or WebKit set for `contentType` when using createDocument()? IE11 not support this IDL attribute.
```js
console.log(document.implementation.createDocument("http://www.w3.org/1999/xhtml", "html", null).contentType);
console.log(document.implementation.createDocument("http://www.w3.org/2000/svg", "svg", null).contentType);
console.log(document.implementation.createDocument("http://www.w3.org/1998/Math/MathML", "math", null).contentType);
```
Firefox and Opera(Presto): `application/xml application/xml application/xml`
Chrome: `application/xhtml+xml image/svg+xml application/xml`
WebKit?
Edge?

zcorpan:
http://software.hixie.ch/utilities/js/live-dom-viewer/saved/4055

Gecko/Presto | Chromium/WebKit | Edge
----|----|----
log: application/xml <br>log: application/xml <br>log: application/xml | log: application/xhtml+xml <br>log: image/svg+xml <br>log: application/xml | ?

@domenic can you test Edge?

ArkadiuszMichalski:
Oh, I should post there, thx to redirect.

smaug----:
I don't know why the current spec would be web compatible. At least it is against what many implementations do
http://mozilla.pettay.fi/moztests/imagedoc.html

PR #216, opened 2016-04-12
Define "participate in a tree" instead of just "participate".

jyasskin:
This makes linking from other specs simpler, since they have to mention
trees in the human-readable text anyway. For example, https://webbluetoothcg.github.io/web-bluetooth/#bluetooth-tree.

That said, it's not a big problem if you reject this.

Demo at https://rawgit.com/jyasskin/dom/participate-linking/dom.html#concept-tree-participate.

@tabatkins, the UI of a link inside another definition is bad: the click handler on the `<dfn>`that shows the dfnPanel prevents clicks on the `<a>` from working.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -234,7 +234,7 @@ A <dfn export id=concept-tree>tree</dfn> is a finite hierarchical tree structure
 traversal of a <a>tree</a>.
 <!-- https://en.wikipedia.org/wiki/Depth-first_search -->
 
-An object that <dfn export for=tree id=concept-tree-participate lt="participate">participates</dfn> in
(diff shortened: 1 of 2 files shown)


tabatkins:
Filed <https://github.com/tabatkins/bikeshed/issues/664>, thanks.

annevk:
The way the DOM Standard itself does this is by linking tree as well. Bit torn on whether this is better.

annevk:
How about we just add additional `lt` values?

jyasskin:
Works for me; done.

annevk:
Thanks, landed as https://github.com/whatwg/dom/commit/9b8612a18d1348c3ca1f71ea1186f782107f6d1b.

PR #218, opened 2016-04-13
Set createDocument() content type from namespace

### LINKED ISSUE #217: createDocument() should use namespace to set content type
See https://github.com/whatwg/dom/pull/213#issuecomment-208916374 onwards for feedback from @ayg, @foolip, and @ArkadiuszMichalski on this.

ayg:
Fixes #217.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -449,6 +449,9 @@ The <dfn export>XML namespace</dfn> is
 The <dfn export>XMLNS namespace</dfn> is
 <code>http://www.w3.org/2000/xmlns/</code>.
 
+The <dfn export>SVG namespace</dfn> is
+<code>http://www.w3.org/2000/svg</code>.
+
 To <dfn export>validate</dfn> a <var>qualifiedName</var>, run these steps:
 
 <ol>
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -5089,6 +5095,23 @@ method, when invoked, must run these steps:
...
+
+ <dl class=switch>
+  <dt><a>HTML namespace</a>
+  <dd><code>application/xhtml+xml</code>
+
+  <dt><a>SVG namespace</a>
+  <dd><code>image/svg+xml</code>
+
+  <dt><a>MathML namespace</a>
+  <dd><code>application/mathml+xml</code>
```
zcorpan: Per https://github.com/whatwg/dom/issues/217#issuecomment-209438878 it seems this should be application/xml

ayg: Is MathML something we want to acknowledge in spec-land, or should we just ignore it as a failed experiment?  I guess if no more UAs are realistically planning to implement it, we should probably ignore it, yeah.

ayg: On second thought -- the HTML spec acknowledges SVG and MathML as things you can natively embed in HTML.  Given that, I think we should special-case MathML as well.  Alternatively, perhaps the <math> tag should be removed from HTML, unless anyone is ever going to use it in real life.  (I think Hixie added it on principle, not because there was significant implementer or author interest.)

zcorpan: MathML is supported in Gecko but Gecko uses `application/xml` here. I don't see any particular reason to use something different here; the main difference is for `createElement` but that doesn't special-case `application/mathml+xml` anyway, right?

annevk: I agree with @zcorpan, FWIW.

ayg:
Rewritten to address feedback.

zcorpan:
LGTM

ArkadiuszMichalski:
Are there any APIs which depend on `image/svg+xml` content type in Document? Or added due to Chromium/WebKit behaviour?

annevk:
I don't know of any APIs other than `createElement()` that use it so I guess it's the latter. Do you think it's problematic?

zcorpan:
`img` accepts `image/svg+xml` but rejects `application/xml` when loading a resource. You can't assign a document to an `img` but maybe if we add `srcObject` it could become relevant.

ArkadiuszMichalski:
@annevk No, it's not problem, just asked to eventualy learning something new.

annevk:
Landed as https://github.com/whatwg/dom/commit/537e579fac8d0a99544d798bb7d46660e59ffc78. Thanks @ayg!

Issue #227, opened 2016-04-18
Remove "svgevent" and "touchevent" from createEvent list?


zcorpan:
333b03cf6454e803d6433673c33f1292821dac0a added "touchevent" per Web compat research in https://www.w3.org/Bugs/Public/show_bug.cgi?id=25491#c7

https://github.com/whatwg/dom/pull/220 added "svgevent". Was it added because it was supported in Gecko and IE11? @ayg

According to the data in https://github.com/whatwg/html/issues/1054#issue-148590884 right now WebKit/Chromium/Gecko/Edge throw for `TouchEvent` and WebKit/Chromium/Edge throw for `SVGEvent`.

ayg:
At least based on Gecko source code, TouchEvent seems to be behind a pref: http://hg.mozilla.org/mozilla-central/file/1f16d3da9280/dom/events/EventDispatcher.cpp#l923  Maybe other engines also?

There was a problem in my testing methodology, because I grepped the output of http://w3c-test.org/dom/nodes/Document-createEvent.html but it previously omitted the trailing "s" from some of the plurals.  Chrome and Firefox support "SVGEvents", but only Firefox supports "SVGEvent".  New list:

```
AnimationEvent
BeforeUnloadEvent
CloseEvent
CompositionEvent
DeviceMotionEvent
DeviceOrientationEvent
DragEvent
ErrorEvent
FocusEvent
HashChangeEvent
IDBVersionChangeEvent
MutationEvent
MutationEvents
PageTransitionEvent
PopStateEvent
ProgressEvent
StorageEvent
SVGEvents
SVGZoomEvent
SVGZoomEvents
TextEvent
TrackEvent
TransitionEvent
WebGLContextEvent
WheelEvent
```

Exact methodology:

1) For each browser, copy-paste output of http://w3c-test.org/dom/nodes/Document-createEvent.html to a text file, grep for '^Fail.*NOT_SUPP', s/^[^"]*"//, s/".*//.  This gives a list of all the events that are not per the current version of the file.

2) cat chrome.txt firefox.txt ie.txt | sort | uniq -c | sort -nr | grep '^ *[23]' | sed 's/^ *[2-3] //' | sort

The only difference I see is SVGEvent -> SVGEvents.

zcorpan:
Thanks!

[...]

PR #228, opened 2016-04-18
SVGEvent is only Gecko, Blink also has SVGEvents

ayg:
As pointed out by zcorpan in #227.


annevk:
(I'll handle merging this since I'll need to generate `dom.html` before that.)

zcorpan:
LGTM.

http://software.hixie.ch/utilities/js/live-dom-viewer/saved/4077

`SVGEvents` works in WebKit/Chromium/Gecko (have not tested Edge).

PR #229, opened 2016-04-18
Shadow: define slotchange event

annevk:
Fixes https://github.com/w3c/webcomponents/issues/288.

This defines the slotchange event in response to remove/insert
operations, changes to a slot’s name attribute, and changes to an
element’s slot attribute.

This also introduces the assigned nodes concept for slots, and assigned
slot concept for slotables.

Slotables now also have a name, rather than a “get name”.

The slotchange event dispatches just after mutation observers.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2895,6 +2895,7 @@ To <dfn export>notify mutation observers</dfn>, run these steps:
  <li><p>For each <a>slot</a> <var>slot</var> in <var>signalList</var>, in order,
  <a>fire an event</a> named <code>slotchange</code>, with its {{Event/bubbles}} attribute set to
  true, at <var>slot</var>.
+ <!-- XXX scoped (change this while updating events for shadow trees) -->
 </ol>
 
 <hr>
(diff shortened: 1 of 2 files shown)


annevk:
Things I'm not happy with:

* The pattern "Let x be y" and then "If x is non-null, signal for x, assign for x" could be a thing
* "assign slotables for a tree" does signal, but "assign slotables" does not

These also seem related. The call for "signal a slot change" should probably move into "assign slotables" somehow.

annevk:
I am now reasonably happy.

hayatoito:
Looks good to me. Awesome work!
Let me provide implementor feedback later, hoping we can find a bug.

One nitpick:

- When inserting a slot, we set a suppress signaling flag.
- When removing a slot, we do not set a suppress signaling flag

I guess that's unintentional.

BTW, you added me to dom repository, but it looks I can not merge this PR.

annevk:
I wasn't sure whether removing should suppress events, I will add that.

As for the repository, you should now be able to create branches on this repository and update them. As well as assign and modify issues, etc. The master branch is a bit more controlled since it affects our publication process. Hope that's acceptable.

hayatoito:
> I wasn't sure whether removing should suppress events, I will add that.

Neither nor I. I have no idea which is better. For now, suppressing sounds natural. I might give feedback later when I get something.

> As for the repository, you should now be able to create branches on this repository and update them. As well as assign and modify issues, etc. The master branch is a bit more controlled since it affects our publication process. Hope that's acceptable.

I am fine. Thank you for the explanation!

PR #230, opened 2016-04-18
Fix #227: Make createEvent("touchevent") sometimes throw

### LINKED ISSUE #227: Remove "svgevent" and "touchevent" from createEvent list?

zcorpan:
Browsers typically disable touch events on non-touch devices, and
there exists Web content that detects this difference using
createEvent.

----

Also see https://github.com/w3c/touch-events/issues/64 for a possible spec hook for this.

I'm not sure about the wording, happy to tweak it.

I updated bikeshed and got some unrelated differences in the output, FYI.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4982,6 +4982,13 @@ invoked, must run these steps:
     <tr><td>"<code>wheelevent</code>"<td>{{WheelEvent}}<td>[[!UIEVENTS]]
   </table>
 
+ <li>
+  If the user agent does not support the <var>constructor</var> interface, or has it disabled
+  in its current configuration, let <var>constructor</var> be null.
+
+  <p class=note>Typically user agents have the {{TouchEvent}} interface disabled in some
+  configurations.
+
  <li>If <var>constructor</var> is null, <a>throw</a> a
  {{NotSupportedError}}.
 
--- a/dom.bs
+++ b/dom.bs
@@ -4928,17 +4928,16 @@ method, when invoked, must run these steps:
 
 <hr>
 
-The <dfn method for=Document><code>createEvent(<var>interface</var>)</code></dfn> method, when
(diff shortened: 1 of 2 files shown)


dom.bs
```diff
@@ -4982,6 +4982,13 @@ invoked, must run these steps:
     <tr><td>"<code>wheelevent</code>"<td>{{WheelEvent}}<td>[[!UIEVENTS]]
   </table>
 
+ <li>
+  If the user agent does not support the <var>constructor</var> interface, or has it disabled
+  in its current configuration, let <var>constructor</var> be null.
+
+  <p class=note>Typically user agents have the {{TouchEvent}} interface disabled in some
+  configurations.
```
annevk: Wouldn't this just follow from `new TouchEvent` throwing?

zcorpan: Yes but that would result in a `TypeError` whereas WebKit/Chromium/Gecko (at last) throw `NotSupportedError`.
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=4078

I suppose `TypeError` is Web compatible enough but are implementors OK with changing to `TypeError`?

annevk: I see. I think it would be better then to where we talk about the initial value of constructor we check for that being undefined and if it's undefined we throw a NotSupportedError and mention that this can be the case for TouchEvent in a note.

annevk:
Thanks!

PR #236, opened 2016-04-21
Remove 'passive' from the event listener key

RByers:
Fixes WICG/EventListenerOptions#27

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -949,10 +949,10 @@ callback interface EventListener {
 
 dictionary EventListenerOptions {
   boolean capture;
-  boolean passive;
 };
 
 dictionary AddEventListenerOptions : EventListenerOptions {
+  boolean passive;
   boolean once;
 };
 </pre>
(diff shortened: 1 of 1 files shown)


RByers:
Looks like the verison of bikeshed being served by api.csswg.org is resulting in some accidental changes (some linking seems broken).  So I undid the html changes and includes only the BS changes here.

annevk:
Landed as https://github.com/whatwg/dom/commit/a13a3c7fc14732691b50c51316b16de0915a61f2. I had to make a couple of additional changes since you "forgot" that passive moved to a new dictionary (that might be some of the broken linkage you saw).

Glad it's all good now (with passive).

RByers:
Oh shoot, sorry I missed that - thanks for fixing!  The broken linking was elsewhere, but I probably missed the warnings in the stream of other unrelated warnings :-(

annevk:
Yeah, bunch of events are not indexed. I decided to live with it for now, sorry it made things harder.

PR #240, opened 2016-04-28
Shadow: define event retargeting for shadow trees

### LINKED ISSUE #237: Figure out how listeners should fire AT_TARGET when doing retargeting
There are few different options when listeners on non-deep target could fire.

(1) Gecko has traditionally fired listeners this way for XBL and such:
capturing listeners, capturing listener on non-deep-target, capturing listeners in shadow content capturing and bubble listeners on shadow-target, bubble listeners in shadow content, bubble listeners on non-deep-target, bubble listener 
That does differentiate capturing and bubble listeners at target phase when the .currentTarget there is a shadow host. But it keeps the algorithms rather easily understandable. There just is internal phase for the event and listeners are called based on that.
FWIW, Gecko uses this also for input elements and such where the implementation of the element is effectively a native shadow DOM thingie  (native anonymous content in Gecko terms)

(2) Other option is to call both capturing and bubble listeners during capture phase before entering shadow content. That gives non-shadow listeners priority over the listeners in shadow DOM.

(3) Same as (2) but call the listeners at target during bubble phase, when the listeners in shadow DOM has been called already.

I don't like (2).
(1) gives kind of best of both setups (2) and (3) depending on whether light DOM uses capturing or bubble listeners, but it makes the existence of shadow dom sort of visible, since the listeners at target aren't called in the order they were registered, but in the order they were registered for capture or non-capture phase. 
I think I could live with (3) too.

annevk:
Includes support for scoped events that a number of HTML features
require. The invocation order handles CAPTURE first, then TARGET and
BUBBLE. Retargeting happens during the BUBBLE phase as developers
mostly use BUBBLE listeners and therefore it would be unexpected if
it was interleaved with the CAPTURE phase.

A split approach where capture listeners for the targets are invoked
during CAPTURE and bubble listeners are invoked during BUBBLE was
also considered, but eventually dismissed as it would make shadow
hosts observable.

Fixes #237 and fixes https://github.com/w3c/webcomponents/issues/485.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -627,23 +627,28 @@ interface Event {
   readonly attribute boolean cancelable;
   void preventDefault();
   readonly attribute boolean defaultPrevented;
+  readonly attribute boolean scoped;
 
   [Unforgeable] readonly attribute boolean isTrusted;
   readonly attribute DOMTimeStamp timeStamp;
 
-  void initEvent(DOMString type, boolean bubbles, boolean cancelable);
+  void initEvent(DOMString type, boolean bubbles, boolean cancelable); // historical
 };
 
 dictionary EventInit {
   boolean bubbles = false;
   boolean cancelable = false;
+  boolean scoped = false;
 };
 </pre>
 
-An <dfn export id=concept-event>event</dfn> allows for signaling that
-something has occurred, e.g. that an image has completed downloading. It is
-represented by the {{Event}} interface or an interface that
-inherits from the {{Event}} interface.
+<p>An {{Event}} object is simply named an <dfn export id=concept-event>event</dfn>. It allows for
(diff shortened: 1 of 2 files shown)


annevk:
@hayatoito @smaug----, @rniwa, @travisleithead, review appreciated!

hayatoito:
Looks good to me. Awesome work! Very well designed algorithm! As far as I traced the algorithm in my head, it matches Shadow DOM spec's intention.
Thank you!

annevk:
Thank you @hayatoito, that is very kind. It builds on well over a decade of research on this topic of course, many years of which have been driven by you and your team. Hopefully this iteration will stick around for a while.

PR #244, opened 2016-05-09
Rename scoped to composed and invert its meaning

annevk:
See https://github.com/whatwg/html/issues/1160 for context.


hayatoito:
LGTM

PR #247, opened 2016-05-10
Shadow: define composedPath()

### LINKED ISSUE #242: Event.prototype.deepPath as an attribute breaks TinyMCE 4.3.10
This is probably an evangelism but wordpress.com breaks due to its use of deepPath as an attribute.  Specifically, they have a code that retrieves `event.deepPath[0]`.

annevk:
Fixes #242. Also reverts
https://github.com/w3c/webcomponents/issues/432 as it appears to no
longer be needed. If we ever have an event path that starts with a
non-node and ends with a node we'll have to add that complexity back.

(The reason it is named composedPath() is because deepPath() and
various other names were no longer viable after Chrome's initial
experiment. See the various issues.)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -758,6 +758,10 @@ steps:
  <li><p>Return <var>composedPath</var>.
 </ol>
 
+<p class="note no-backref">This algorithm assumes that when the <var>target</var> argument to the
+<a>dispatch</a> algorithm is not a <a>node</a>, none of the tuples in <var>event</var> argument's
+eventual <a for=Event>path</a> will contain a <a>node</a> either.
+
 The <dfn attribute for=Event><code>eventPhase</code></dfn>
 attribute must return the value it was initialized to, which must be one of
 the following:
--- a/dom.bs
+++ b/dom.bs
@@ -673,9 +673,10 @@ list of tuples, each of which consists of an <b>item</b> (an {{EventTarget}} obj
  invoked.
 
  <dt><code><var>event</var> . {{Event/composedPath()}}</code>
- <dd>Returns <var>event</var>'s path, which includes objects on which listeners will be invoked,
- except for any <a>nodes</a> in <a>shadow trees</a> of which the <a for=/>shadow root</a>'s
- <a for=ShadowRoot>mode</a> is "<code>closed</code>".
+ <dd>Returns the <b>item</b> objects of <var>event</var>'s <a for=Event>path</a> (objects on which
+ listeners will be invoked), except for any <a>nodes</a> in <a>shadow trees</a> of which the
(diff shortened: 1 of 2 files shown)


dom.bs
```diff
@@ -733,6 +739,25 @@ attributes must return the values they were initialized to. When an
...
+
+ <li><p>Let <var>currentTarget</var> be <a>context object</a>'s <code for=Event>currentTarget</code>
+ attribute value.
+
+ <li>
+  <p>For each <var>tuple</var> in <a>context object</a>'s <a for=Event>path</a>:
+
+  <ol><li><p>If <var>currentTarget</var> is a <a>node</a> and <var>tuple</var>'s <b>item</b> is an
+  <a>unclosed node</a> of <var>currentTarget</var>, or <var>currentTarget</var> is <em>not</em> a
+  <a>node</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.</p></li></ol>
```
rniwa: This is not quite right if non-node target resides outside a shadow tree.  It's possible for hit region, etc... to exist outside a shadow and the event path to enter a shadow tree via a slot.  So we need to add *item* only if it's *unclosed* to the first `Node` that appears after such a non-Node target.

annevk: Hit regions will change to only be nodes. I don't think there's any other such feature. That is why I reopened that bug you raised and said that we could simplify this.

rniwa: Oh I see. Perhaps we should leave a note here clarifying that path can only contain Nodes or non-Node targets?

annevk: Well, path can contain both due to the `Window` object (exposed to JavaScript as the `WindowProxy` object), but that is only ever at the end. I will at a node that the ongoing assumption is that if the target is a non-node, the eventual path will also not contain nodes. I think that holds.

annevk:
I've added a note clarifying the underlying assumption.

annevk:
@hayatoito would you like to sign off?

dom.bs
```diff
@@ -671,6 +672,11 @@ list of tuples, each of which consists of an <b>item</b> (an {{EventTarget}} obj
  <dd>Returns the object whose <a>event listener</a>'s <b>callback</b> is currently being
  invoked.
 
+ <dt><code><var>event</var> . {{Event/composedPath()}}</code>
+ <dd>Returns <var>event</var>'s path, which includes objects on which listeners will be invoked,
+ except for any <a>nodes</a> in <a>shadow trees</a> of which the <a for=/>shadow root</a>'s
```
hayatoito: This sentence might not be accurate, however, I think it's okay because the normative behavior is exactly defined in the paragraph of the steps.

hayatoito:
Looks good to me. Thanks.

Regarding Node vs non-Nodes, I am still unsure what is the best way to define. The current definition might be good enough, as a starting point.

PR #248, opened 2016-05-10
Rename rootNode to getRootNode() and add composed argument

### LINKED ISSUE #241: Node.prototype.rootNode is not Web compatible
We've got at least two P1 bug reports for websites being broken after adding `rootNode` on `Node`.  We need to rename this attribute.

annevk:
Fixes #241. Unfortunately a variety of good short names is incompatible
with the web.

Furthermore, the argument was made that a method better exposes the
fact that this is not always a O(1) operation and can be used to return
the shadow-including root when given an argument.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3386,6 +3386,7 @@ interface Node : EventTarget {
 
   readonly attribute boolean isConnected;
   readonly attribute Document? ownerDocument;
+  Node getRootNode(optional GetRootNodeOptions options);
   readonly attribute Node? parentNode;
   readonly attribute Element? parentElement;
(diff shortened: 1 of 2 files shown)


annevk:
@rniwa @smaug---- @hayatoito @travisleithead, ideally we only rename this once. Does this look okay?

hayatoito:
Yeah, looks okay to me. Let's try with this name.

annevk:
Turns out not everyone agreed this was okay. We'll be removing the feature instead.

annevk:
And we're back to everyone agreeing again. Hurray.

PR #252, opened 2016-05-20
Use USVString for URLs and origins in IDL

zcorpan:
Change node.baseURI, document.URL, document.documentURI and
document.origin to use USVString.

----

Note: genned with updated bikeshed


annevk:
Thank you!

PR #253, opened 2016-05-24
Don't set custom element state redundantly

domenic:
As of https://github.com/whatwg/html/pull/1309, this is no longer necessary, as it is taken care of in the Construct(C) step.

Ideally don't merge this until that PR is merged.


dominiccooney:
This LGTM.

Am I right in understanding that I can review the .bs files and ignore the HTML files in these PRs, because presumably the HTML file is generated by a tool?

domenic:
Yep, exactly

domenic:
@annevk I rebased this and the requisite HTML PR is merged so it should be good to go.

annevk:
I should not have attempted to fix the date. That took so much longer than anticipated...

PR #256, opened 2016-05-31
Shadow: Update the definition of an unclosed node

hayatoito:
Fixes https://github.com/w3c/webcomponents/issues/486.


hayatoito:
Ops. It should be https://github.com/w3c/webcomponents/issues/484, instead of https://github.com/w3c/webcomponents/issues/486. Let me close this PR and send another one.

Issue #61, opened 2015-08-03
Patch HTML's base URL stuff to be in terms of DOM


annevk:
Per https://www.w3.org/Bugs/Public/show_bug.cgi?id=20976#c6 base URL will become a concept of documents.

Then we need a way to set the base URL of a document from another specification. When set, this changes the URL of the document and then for each inclusive descendant, in tree order, notifies the node of a base URL change, if the node registered for base URL changes. This allows e.g., `<a>` to correct its URL.

We also change adopt to run these base URL change notifications. It's unclear whether they should run before the adoption notification or after, or whether the adoption notification by itself is sufficient. I suspect we want to have both separately so you don't forget any.

@ajklein @coonsta @smaug---- does this make sense to you?

ajklein:
For which elements is notification of a base URL change necessary? Can't elements lazily look at the document's current base URL when returning complete URLs? This is how `<a>` works in Blink, for example. Are there cases where a change of base URL triggers some behavior? I tried this with an `<img>` and saw no re-fetch.

annevk:
That is an excellent question. We just need to update https://html.spec.whatwg.org/multipage/infrastructure.html#dynamic-changes-to-base-urls then to make it clear the `href` attribute gets parsed on getting. It's not defined as such now I think.

We do have to make it extremely clear then though at which point the underlying URL gets copied, e.g. for navigate or fetching.

annevk:
I will leave this open until HTML is patched.

domenic:
You also need to update the URL spec so that every time a URLUtils getter is called, the internal url is re-parsed with respect to a fresh "get the base" call.

ajklein:
I'm confused, why do we need to change the spec here? Is it somehow observable when the base URL update happens?

domenic:
@ajklein if I understand correctly the spec is wrong with regard to the test case from https://github.com/whatwg/url/issues/54. Browsers show the following behavior:

```js
console.log(document.querySelector("a").href); // http://example.com/foo.html
document.querySelector("base").href = "http://otherexample.org";
console.log(document.querySelector("a").href); // http://otherexample.org/foo.html
```

however the URL spec's URLUtils spec does not update `href` with the new base upon getting. (See: [href getter](https://url.spec.whatwg.org/#dom-urlutils-href), or any other getter in URLUtils.)

[...]

PR #258, opened 2016-05-31
Use HTML's 'document base URL' concept.

Ms2ger:
This definition is more correct right now, and using a getter rather than a
mutable field makes it easier to figure out what is returned.

CC <https://github.com/whatwg/html/issues/421>.


annevk:
Hmm, but we need some kind of callback for when it changes: https://github.com/whatwg/dom/issues/61. I guess defining that in HTML is not too bad, maybe a little weird?

domenic:
Note that the conclusion of that thread was that the callback is only necessary for CSS; everything else is done lazily.

annevk:
Also for custom elements that want that kind of behavior, no? Although I guess we could decide not to expose that primitive.

domenic:
Yeah, I don't think it's a primitive really. Custom elements will probably want to do the same thing `a` and `area` do, and compute lazily. CSS could theoretically even be specced that way, although understandably implementations prefer a callback approach.

annevk:
How can you compute this lazily? It's immediately visible.

domenic:
When querying :visited, or when rendering, you figure out the resolved absolute URL.

annevk:
So whenever you render you have to go through all links in the document? Rather than only that rare time someone updates the base URL? I think I know why browsers have their current setup, and why theoretically custom elements might want that too...

domenic:
I mean, yes, whenever you render a link, you could resolve its URL lazily. With appropriate caching that seems like a totally reasonable strategy. It's certainly a reasonable model for the spec.

I don't think there are any real-world custom element use cases here.

annevk:
Merged since as @Ms2ger indicated, this is more correct than the status quo.

PR #267, opened 2016-06-07
UAs create elements in HTML namespace when context object is an HTML …

### LINKED ISSUE #264: .createElement's namespace handling doesn't seem to match what UAs do
https://github.com/whatwg/dom/pull/213 made it rely only on contentType but that isn't what UAs
(Gecko, Blink, Edge at least) do. http://mozilla.pettay.fi/moztests/imagedoc.html
Perhaps contentType should be used only in case Document is non-HTMLDocument.
ImageDocument inherits HTMLDocument, so it is HTMLDocument and should create html elements by default.
I didn't test svg handling.

smaug----:
…document, fixes #264

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4883,7 +4883,7 @@ invoked, must run these steps:
 
  <li>Let <var>namespace</var> be the <a>HTML namespace</a>, if the <a>context object</a>'s
  <a for=Document>content type</a> is "<code>text/html</code>" or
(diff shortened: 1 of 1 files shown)


annevk:
Landed as https://github.com/whatwg/dom/commit/fb1c9863b16c39ae880becdac90e1156a8d61cde. Thank you!

(I also updated the non-normative description of `createElement()` and dropped the "text/html" content type since that can only happen for HTML documents.)

Issue #238, opened 2016-04-22
"In a document" confusion


domenic:
So I guess this has been discussed to death elsewhere, but naively putting on my web developer hat, something that's in a shadow tree is still "in a document". Maybe in English: the element is in the shadow tree, the shadow tree is in the document, therefore the element is in the document.

Have we considered "in a document's light DOM" or similar?

It seems like regardless of what we do we're going to have to audit every use of "in a document" and figure out whether it should become "in a document's light DOM" or "in a shadow-including document".

annevk:
I don't think we really discussed this much actually. @dglazkov, @smaug----, @hayatoito, @rniwa? Now would be a good time to decide on this, since I'm working on several patches to HTML that would be affected by this.

"In a light document" (to disambiguate) and "connected" (after the developer-facing API)?

smaug----:
"light document" sounds a bit odd, given all the different meanings of "light". One could think data documents being "light", since they are less heavy than documents bound to a browsing context.
And there is no shadow document. 

To me in a document clearly means light dom only, since there is parentNode chain from node to document in that case. That is what tree structures usually mean in computer science.
But I guess I could live with "in a document's light DOM".

rniwa:
I'd say "directly in a document" or "in a document without crossing shadow boundary" or "in a document tree".

annevk:
The DOM Standard does already introduce "light tree" as an opposite to "shadow tree", but given this feedback I'm inclined to just stick with "in a document" and "in a shadow-including document" meaning what they mean today.

domenic:
> To me in a document clearly means light dom only, since there is parentNode chain from node to document in that case. That is what tree structures usually mean in computer science.

Is that CS tree concept usually denoted using the word "in", or is there perhaps a more precise phrase?

I do think "in a document tree" (and "in a shadow-including document tree" or similar) helps a bit over the simple "in a document". It makes me aware that there's a concept known as "document tree" and that I should be thinking in terms of tree structures.

I like "in a light document"/"in a document's light DOM" and "connected" the most but I guess others don't.

smaug----:
"in a light document" doesn't quite make sense, but the the others are fine.

(Though, very subjective view, I prefer Gecko's naming: In composed document, in uncomposed document, but I think others have said they don't like those.)

hayatoito:
> I'm inclined to just stick with "in a document" and "in a shadow-including document" meaning what they mean today.

+1

[...]

PR #273, opened 2016-06-24
Shadow: Define "in the same shadow-including document"

hayatoito:
The definition is intended to be used in https://github.com/w3c/pointerlock/pull/4.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1541,6 +1541,10 @@ itself.</p>
 <p>An <a for=/>element</a> is <dfn export>in a shadow-including document</dfn> if its
 <a>shadow-including root</a> is a <a>document</a>.
 
+<p>An object <var>A</var> and an object <var>B</var> are <dfn export>in the same shadow-including
+document</dfn> if both <var>A</var> and <var>B</var> are <a>in a shadow-including document</a> and
+<var>A</var> and <var>B</var> share the same <a>shadow-including root</a>.
+
 <h5 id=shadow-tree-slots>Slots</h5>
 
 <p>A <a>shadow tree</a> contains zero or more <a for=/>elements</a> that are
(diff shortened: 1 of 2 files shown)


hayatoito:
It looks bikeshed updated a lot of places in dom.html. Is it okay?

dom.bs
```diff
@@ -1541,6 +1541,10 @@ itself.</p>
 <p>An <a for=/>element</a> is <dfn export>in a shadow-including document</dfn> if its
 <a>shadow-including root</a> is a <a>document</a>.
 
+<p>An object <var>A</var> and an object <var>B</var> are <dfn export>in the same shadow-including
+document</dfn> if both <var>A</var> and <var>B</var> are <a>in a shadow-including document</a> and
+<var>A</var> and <var>B</var> share the same <a>shadow-including root</a>.
```
annevk: Apparently "in a shadow-including document" is only for elements. Is that wrong or should this also be for elements only? Maybe "node" for both makes more sense?

dom.html
```diff
@@ -6269,11 +6272,6 @@ <h3 class="no-num no-ref heading settled" id="index-defined-here"><span class="c
   <h3 class="no-num no-ref heading settled" id="index-defined-elsewhere"><span class="content">Terms defined by reference</span><a class="self-link" href="#index-defined-elsewhere"></a></h3>
   <ul class="index">
    <li>
-    <a data-link-type="biblio">[ECMASCRIPT]</a> defines the following terms:
```
domenic: @tabatkins any idea why this xref disappeared?

dom.bs
domenic: +1 node for both

domenic:
> It looks bikeshed updated a lot of places in dom.html. Is it okay?

It looks to me like it's mostly updates to the syntax highlighting, but I commented on one troublesome reference removal. But it doesn't seem worth preventing merge. As long as you used the latest Bikeshed (either via the HTTP API or by pulling the latest from master and running `bikeshed update`) then these changes should not be a problem.

dom.html
tabatkins: Yeah, it's a bug on my side. Got it reported yesterday, will be fixing it shortly.

hayatoito:
Thank you for the review.

It looks that the relevant discussion is on going here: https://github.com/whatwg/dom/issues/238.
I would like to wait for the conclusion there, then update this PR accordingly.

hayatoito:
Let me close this PR since it is no longer valid, as per https://github.com/whatwg/dom/issues/238.

Issue #277, opened 2016-07-05
Add relatedTarget to event dispatch


annevk:
There's some `relatedTarget` logic in https://w3c.github.io/webcomponents/spec/shadow/ that needs to become part of the core event dispatch algorithm. Basically for all events that have a relatedTarget "internal slot".

@hayatoito is this something you might be interested in working on?

annevk:
Figuring out #276 is probably part of this.

hayatoito:
Yeah, let me work on this.

PR #281, opened 2016-07-14
Add relatedTarget to event dispatch

### LINKED ISSUE #276: Upstream retargeting algorithm
See https://w3c.github.io/webcomponents/spec/shadow/#dfn-retargeting-algorithm. It seems Pointer Lock might use it soon per https://github.com/w3c/pointerlock/pull/4 and it's needed by `activeElement` too.

I wonder if we should figure out if event dispatch can somehow use it after all.

hayatoito:
Fixes #276, #277

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -639,10 +639,18 @@ dictionary EventInit {
 <p>An {{Event}} object is simply named an <dfn export id=concept-event>event</dfn>. It allows for
 signaling that something has occurred, e.g., that an image has completed downloading.</p>
 
+<p>An <a>event</a> has an associated <dfn export for=Event>relatedTarget</dfn> (an
+{{EventTarget}} object). Unless stated otherwise it is null.</p>
+
+<p class="note">Other specifications may define the relatedTarget attribute which reflects an
+event's <a>relatedTarget</a>.
+[[UIEVENTS]]
+
 <p>An <a>event</a> has an associated <dfn export for=Event>path</dfn>. A <a for=Event>path</a> is a
-list of tuples, each of which consists of an <b>item</b> (an {{EventTarget}} object) and a
-<b>target</b> (null or an {{EventTarget}} object). A tuple is formatted as (<b>item</b>,
-<b>target</b>). A <a for=Event>path</a> is initially the empty list.</p>
+list of tuples, each of which consists of an <b>item</b> (an {{EventTarget}} object), <b>target</b>
+(null or an {{EventTarget}} object), and a <b>relatedTarget</b> (null or an {{EventTarget}}
+object). A tuple is formatted as (<b>item</b>, <b>target</b>, <b>relatedTarget</b>). A <a
+for=Event>path</a> is initially the empty list.</p>
 
 <dl class=domintro>
  <dt><code><var>event</var> = new <a constructor lt="Event()">Event</a>(<var>type</var> [, <var>eventInitDict</var>])</code>
@@ -1185,7 +1193,11 @@ for discussion).
   <!-- We should consider refactoring it to make it have less of a scope, since we don't really want
        folks to start using it for non-legacy scenarios. -->
 
- <li><p>Append (<var>target</var>, <var>targetOverride</var>) to <var>event</var>'s
+ <li>Let <var>relatedTarget</var> be the result of invoking <a>retargeting</a> <var>event</var>'s
+ <a>relatedTarget</a> against <var>target</var> if <var>event</var>'s <a>relatedTarget</a> is
+ non-null, and null otherwise.
+
+ <li><p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>'s
  <a for=Event>path</a>.
 
  <li><p>Let <var>parent</var> be the result of invoking <var>target</var>'s <a>get the parent</a>
@@ -1195,15 +1207,22 @@ for discussion).
   <p>While <var>parent</var> is non-null:</p>
 
   <ol>
+   <li><p>Let <var>relatedTarget</var> be the result of invoking <a>retargeting</a>
+   <var>event</var>'s <a>relatedTarget</a> against <var>parent</var> if <var>event</var>'s
+   <a>relatedTarget</a> is non-null, and null otherwise.
+
    <li><p>If <var>target</var>'s <a for=tree>root</a> is a
    <a>shadow-including inclusive ancestor</a> of <var>parent</var>, then append
-   (<var>parent</var>, null) to <var>event</var>'s <a for=Event>path</a>.
+   (<var>parent</var>, null, <var>relatedTarget</var>) to <var>event</var>'s <a for=Event>path</a>.
 
-   <li><p>Otherwise, set <var>target</var> to <var>parent</var> and append
-   (<var>parent</var>, <var>target</var>) to <var>event</var>'s <a for=Event>path</a>.
+   <li><p>Otherwise, if <var>parent</var> and <var>relatedTarget</var> are identical, then set
+   <var>parent</var> to null.
 
-   <li><p>Set <var>parent</var> to the result of invoking <var>parent</var>'s <a>get the parent</a>
-   with <var>event</var>.
+   <li><p>Otherwise, set <var>target</var> to <var>parent</var> and append (<var>parent</var>,
+   <var>target</var>, <var>relatedTarget</var>) to <var>event</var>'s <a for=Event>path</a>.
+
+   <li><p>If <var>parent</var> is non-null, then set <var>parent</var> to the result of invoking
+   <var>parent</var>'s <a>get the parent</a> with <var>event</var>.
   </ol>
 
  <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/CAPTURING_PHASE}}.
@@ -1216,6 +1235,8 @@ for discussion).
    in <var>event</var>'s <a for=Event>path</a>, that is either <var>tuple</var> or preceding
    <var>tuple</var>, whose <b>target</b> is non-null.
 
+   <li><p>Set <var>event</var>'s <a>relatedTarget</a> to <var>tuple</var>'s <b>relatedTarget</b>.
+
    <li><p>If <var>tuple</var>'s <b>target</b> is null, then <a>invoke</a> <var>tuple</var>'s
    <b>item</b> with <var>event</var>.
   </ol>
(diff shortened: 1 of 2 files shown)


hayatoito:
I am assuming that it is not a good idea that DOM Standard has a dependency to UIEvents specification.
Therefore, I used *set the relatedTarget* algorithm so that UIEvents can override it to set UIEvent's relatedTarget attribute.

Is this a good idea?

hayatoito:
@annevk Could you have a chance to review this PR?

dom.bs
```diff
@@ -1195,15 +1207,22 @@ for discussion).
...
+
    <li><p>If <var>target</var>'s <a for=tree>root</a> is a
    <a>shadow-including inclusive ancestor</a> of <var>parent</var>, then append
-   (<var>parent</var>, null) to <var>event</var>'s <a for=Event>path</a>.
+   (<var>parent</var>, null, <var>relatedTarget</var>) to <var>event</var>'s <a for=Event>path</a>.
 
-   <li><p>Otherwise, set <var>target</var> to <var>parent</var> and append
-   (<var>parent</var>, <var>target</var>) to <var>event</var>'s <a for=Event>path</a>.
+   <li><p>Otherwise, if <var>parent</var> and <var>relatedTarget</var> are identical, let
+   <var>parent</var> be null.
```
annevk: This should set parent to null. Only use "let" when introducing a variable.

dom.bs
```diff
@@ -5523,6 +5548,22 @@ or <var>A</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
 <a for=ShadowRoot>mode</a> is "<code>open</code>" and <a for=DocumentFragment>host</a> is an
 <a>unclosed node</a> of <var>B</var>.
 
+<p>To <dfn export lt="retarget|retargeting">retarget</dfn> a <var>node</var> against a
+<var>base</var>, run these steps:</p>
+
+<ol>
+ <li>Let <var>target</var> be <var>node</var>.
+
+ <li>Repeat these subsubsteps:
```
annevk: substeps, not subsubsteps.

dom.bs
```diff
@@ -5523,6 +5548,22 @@ or <var>A</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
...
+   <ol>
+     <li>If <var>target</var>'s <a>root</a> is not a <a>shadow root</a>, or <var>target</var>'s
+     <a>root</a> is a <a>shadow-including inclusive ancestor<a> of <var>base</var>, return
+     <var>target</var>.
+
+     <li>Let <var>target</var> be the <var>target</var>'s <a>root</a>'s <a>host</a>.
```
annevk: This should use "set".

dom.bs
```diff
@@ -5523,6 +5548,22 @@ or <var>A</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
 <a for=ShadowRoot>mode</a> is "<code>open</code>" and <a for=DocumentFragment>host</a> is an
 <a>unclosed node</a> of <var>B</var>.
 
+<p>To <dfn export lt="retarget|retargeting">retarget</dfn> a <var>node</var> against a
+<var>base</var>, run these steps:</p>
+
+<ol>
+ <li>Let <var>target</var> be <var>node</var>.
```
annevk: We don't really need this step, right? And instead of "run these steps" above we could say "repeat these steps until they return a node"?

annevk:
I was away last week, sorry for the delay.

dom.bs
hayatoito: I got it. Done.

dom.bs
hayatoito: Done. Removed.

dom.bs
hayatoito: Done.

dom.bs
hayatoito: Done.

hayatoito:
No problem. Thank you for the review!
I updated the PR.

dom.bs
```diff
@@ -5523,6 +5548,19 @@ or <var>A</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
 <a for=ShadowRoot>mode</a> is "<code>open</code>" and <a for=DocumentFragment>host</a> is an
 <a>unclosed node</a> of <var>B</var>.
 
+<p>To <dfn export lt="retarget|retargeting">retarget</dfn> an object <var>A</var> against an object
```
hayatoito: Now, I am using "an object A" here because it is consistent to other definitions around here.

dom.bs
```diff
@@ -639,10 +639,18 @@ dictionary EventInit {
 <p>An {{Event}} object is simply named an <dfn export id=concept-event>event</dfn>. It allows for
 signaling that something has occurred, e.g., that an image has completed downloading.</p>
 
+<p>An <a>event</a> has an associated <dfn export for=Event>relatedTarget</dfn> (an
+{{EventTarget}} object). Unless stated otherwise it is null.</p>
+
+<p>An <a>event</a> has an associated <dfn export>set the relatedTarget</dfn> algorithm,
+which takes an {{EventTarget}} <var>relatedTarget</var>. Unless specified otherwise it does
+nothing.</p>
```
annevk: So this algorithm exists because UI events need it and they need to do something different between normal UI events and Touch events?

Also, how does this work for synthetic events? I guess the event class defines it?

It's probably worth having a note here that points to the various implementations of this algorithm.

dom.bs
```diff
@@ -1195,15 +1207,22 @@ for discussion).
...
+   <var>event</var>'s <a>relatedTarget</a> against <var>parent</var> if <var>event</var>'s
+   <a>relatedTarget</a> is non-null, and null otherwise.
+
    <li><p>If <var>target</var>'s <a for=tree>root</a> is a
    <a>shadow-including inclusive ancestor</a> of <var>parent</var>, then append
-   (<var>parent</var>, null) to <var>event</var>'s <a for=Event>path</a>.
+   (<var>parent</var>, null, <var>relatedTarget</var>) to <var>event</var>'s <a for=Event>path</a>.
+
+   <li><p>Otherwise, if <var>parent</var> and <var>relatedTarget</var> are identical, set
+   <var>parent</var> to null.
```
annevk: "then set _parent_ to null"

The convention is to use "then" after the comma following an if statement.

dom.bs
```diff
@@ -1195,15 +1207,22 @@ for discussion).
...
 
-   <li><p>Otherwise, set <var>target</var> to <var>parent</var> and append
-   (<var>parent</var>, <var>target</var>) to <var>event</var>'s <a for=Event>path</a>.
+   <li><p>Otherwise, set <var>target</var> to <var>parent</var> and append (<var>parent</var>,
+   <var>target</var>, <var>relatedTarget</var>) to <var>event</var>'s <a for=Event>path</a>.
 
-   <li><p>Set <var>parent</var> to the result of invoking <var>parent</var>'s <a>get the parent</a>
-   with <var>event</var>.
+   <li><p>If <var>parent</var> is non-null, set <var>parent</var> to the result of invoking
+   <var>parent</var>'s <a>get the parent</a> with <var>event</var>.
```
annevk: then*

dom.bs
```diff
@@ -5523,6 +5548,19 @@ or <var>A</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
 <a for=ShadowRoot>mode</a> is "<code>open</code>" and <a for=DocumentFragment>host</a> is an
 <a>unclosed node</a> of <var>B</var>.
 
+<p>To <dfn export lt="retarget|retargeting">retarget</dfn> an object <var>A</var> against an object
+<var>B</var>, run these steps until they return an object:</p>
+
+<ol>
+ <li>Repeat these substeps:
```
annevk: This needs `<p>` and it needs to be on its own line since the `<li>` contains two block-level children.

dom.bs
```diff
@@ -5523,6 +5548,19 @@ or <var>A</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
...
+   <ol>
```
annevk: This is indented one space too many.

dom.bs
```diff
@@ -5523,6 +5548,19 @@ or <var>A</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
...
+     <li>If <var>A</var>'s <a>root</a> is not a <a>shadow root</a>, or <var>A</var>'s <a>root</a> is
+     a <a>shadow-including inclusive ancestor<a> of <var>B</var>, return <var>A</var>.
```
annevk: then*

annevk:
This looks really good, thank you. Are we tracking the work that needs to be done on the UI event side somewhere?

(I think I'm also be okay with just including the references to the UI events directly. What you've done is probably a little cleaner though.)

dom.bs
hayatoito: Good question. This is inspired by "get the parent", however, I expect that there is only one implementation. UI Events and Touch Event would actually do the same thing. They just store a given argument to their relatedTarget attribute.

I just would like to avoid mentioning relatedTarget *attribute* in DOM Standard because Event interface does not have relatedTarget attribute. If it is okay to mention relatedTarget attribute in DOM Standard, I would remove "set the relatedTarget" abstraction.

For example, can we say the following?
> Set event’s relatedTarget attribute to *relatedTarget*, if event has a relatedTarget attribute. Otherwise, do nothing.

annevk: Can't we just manipulate the "relatedTarget" concept you introduced for events already and define the relatedTarget attribute as reflecting that concept?

hayatoito: Ah. I think we can.

- In DOM Standard, we use only "relatedTarget" concept.
- In UI Events, define relatedTarget attribute as reflecting "relatedTarget" concept.

That sounds reasonable. Let me try it.

dom.bs
hayatoito: Done.

dom.bs
hayatoito: Done.

dom.bs
hayatoito: Done.

dom.bs
hayatoito: Done.

dom.bs
hayatoito: Done.

hayatoito:
I have rewritten the commit:
- It no longer uses "set the relatedTarget" abstraction.
- Add a note to mention that UI Events specification may define the relatedTarget attribute which reflects an event's relatedTarget.

annevk:
Great, thanks for working on this!

PR #282, opened 2016-07-14
Set the stage for preventing double custom element constructor calls

domenic:
This is part of https://github.com/w3c/webcomponents/issues/533. It introduces a new custom element state of "failed", which HTML will use on failed upgrades or failed parser-initiated element creation, to ensure that if a constructor fails, the element will not be upgraded in the future (i.e. the constructor will not be called again for that element).


annevk:
The term "custom element constructor" is not exported apparently.

annevk:
We can add it to the other PR, merge that first, and then wait?

domenic:
Hmm I'd rather land this first and pick up the autolink in a future build. Otherwise HTML references a non-existent custom element state.

annevk:
Okay, please add it to the HTML PR then so we don't forget.

domenic:
Yep, already done :)

PR #288, opened 2016-08-03
Shadow: Add "retargeting steps" to the dispatch algorithm

### LINKED ISSUE #286: Upstream Touch Events retargeting
See http://w3c.github.io/webcomponents/spec/shadow/#touch-events-retargeting.

This is the last part of the "Events" section in Shadow DOM spec, which we have not upstreamed to the DOM Standard (and/or Touch Events spec).

hayatoito:
Fix #286. Add the minimum hook to an event dispatch so other
specifications, such as TouchEvents spec, can retarget the event's
attributes, before invoking an event.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -651,6 +651,11 @@ list of tuples, each of which consists of an <b>item</b> (an {{EventTarget}} obj
 object). A tuple is formatted as (<b>item</b>, <b>target</b>, <b>relatedTarget</b>). A <a
 for=Event>path</a> is initially the empty list.</p>
 
+<p><a lt="Other applicable specifications">Specifications</a> may define
+<dfn export for=Event>retargeting steps</dfn> for all or some <a>events</a>.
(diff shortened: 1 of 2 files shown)


hayatoito:
@annevk could you review this PR?

@RByers FYI.

annevk:
Do we only need this for touch events? Or do we expect this to become more widely used?

annevk:
I also don't really like using "attributes" since what we really want them to adjust are the internal slots and IDL still needs to become less vague on that whole subject. Maybe we can make it "retargeting steps" and define it in the same way we define other hooks like that (e.g., "removing steps")?

hayatoito:
This is only for touch events, as of now.

"retargeting steps" sounds good. Let me update the PR.

hayatoito:
I have updated the PR. Could you take a look?

annevk:
Thank you! (I recommend updating your copy of bikeshed (git pull; bikeshed update).)

PR #293, opened 2016-08-15
Pass the old and new document to adoptedCallback

### LINKED ISSUE #292: The adopted callback should be passed the old and new document
https://dom.spec.whatwg.org/#example-5ffcda00

> For each inclusiveDescendant in node’s shadow\-including inclusive descendants that is a custom element, enqueue a custom element callback reaction with inclusiveDescendant, callback name "adoptedCallback", and an empty argument list\.

This doesn't help you track which document(s) you've been in; consider the case where an adopted callback causes more follow-on adoption. The same could be said of the connected and disconnected callbacks.

Maybe it is OK to consult the document or parentNode properties, but then the same could be said of the attributeChangedCallback, which is provided the old and new attribute values.

domenic:
Fixes #292.


annevk:
You should update bikeshed sometime.

annevk:
Also, thanks!

domenic:
I'm using the web service, so either that is out of date, or yours is...

annevk:
The web service must be then.

PR #295, opened 2016-08-16
Reinstate CDATASection

### LINKED ISSUE #190: List subclasses explicitly or not
@hayatoito raised this in https://github.com/whatwg/dom/commit/b2ee400a6542dc76d41bed5d2edb8c85ba241bfc. Filing so I don't forget.

Currently we list Text, ProcessingInstruction, Comment, explicitly, rather than simply stating CharacterData. Should we do the same for DocumentFragment and ShadowRoot, or is that case different because DocumentFragment is not an abstract interface?

We should probably treat that case different, since otherwise we'd have to list HTMLMarqueeElement and friends too. However, if we change this, perhaps it makes sense to just use abstract interfaces when they make sense?

@domenic had opinions here too I believe.

annevk:
I’m thirty now so it’s time to face reality. This fixes
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27386. This also closes
#190 by deciding to not enumerate subclasses of concrete (non-abstract)
interfaces (ShadowRoot and CDATASection in particular).

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3400,7 +3400,7 @@ interface Node : EventTarget {
   const unsigned short ELEMENT_NODE = 1;
   const unsigned short ATTRIBUTE_NODE = 2; // historical
   const unsigned short TEXT_NODE = 3;
-  const unsigned short CDATA_SECTION_NODE = 4; // historical
+  const unsigned short CDATA_SECTION_NODE = 4;
   const unsigned short ENTITY_REFERENCE_NODE = 5; // historical
   const unsigned short ENTITY_NODE = 6; // historical
   const unsigned short PROCESSING_INSTRUCTION_NODE = 7;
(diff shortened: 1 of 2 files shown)


dom.bs
```diff
@@ -1488,7 +1488,7 @@ can be used to explore this matter in more detail.
 
 <h3 id=node-trees>Node tree</h3>
 
-<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{ShadowRoot}}, {{Element}}, {{Text}},
```
foolip: Mind splitting the ShadowRoot changes into a separate commit for landing? Same PR is fine, I'll merge the sum of it now.

annevk: Yeah I can do that, provided the ShadowRoot changes look fine.

dom.bs
```diff
@@ -4982,6 +4986,20 @@ invoked, must return a new {{Text}} <a>node</a> with its <a>data</a> set to <var
 <p class="note no-backref">No check is performed that <var>data</var> consists of
 characters that match the <code><a type>Char</a></code> production.
 
+<p>The <dfn method for=Document><code>createCDATASection(<var>data</var>)</code></dfn> method, when
```
foolip: This matches exactly what Blink does :)

dom.bs
foolip: It does to me.

foolip:
OK, I've reviewed and compared to Blink, all LGTM!

Matching Blink CL: https://codereview.chromium.org/2249083002

dom.bs
annevk: Also Gecko.

foolip: Yay interop!

PR #296, opened 2016-08-16
Specify global object references in more detail

### LINKED ISSUE #206: Made more formal for object global
https://dom.spec.whatwg.org/#dom-document-document
"The Document() constructor, when invoked, must return a new document whose origin is the origin of the global object’s associated document."
What is global object here? In DOM we don't have such definition so you mean this from HTML https://html.spec.whatwg.org/multipage/webappapis.html#global-object? If yes maybe use refference?

This term exist in some other places:
https://dom.spec.whatwg.org/#dom-eventtarget-addeventlistener
https://dom.spec.whatwg.org/#dom-eventtarget-removeeventlistener
https://dom.spec.whatwg.org/#dom-documentfragment-documentfragment
https://dom.spec.whatwg.org/#dom-text-text
https://dom.spec.whatwg.org/#dom-comment-comment
https://dom.spec.whatwg.org/#dom-range-range

annevk:
You cannot just say “global object” these days and expect to get away
with it. Fixes #206.


annevk:
This cannot land straight away. HTML needs to export some more definitions first.

PR #299, opened 2016-08-16
Make Attr inherit from Node again

### LINKED ISSUE #102: Make Attr a Node again
Attr's child text nodes have been successfully [removed](https://groups.google.com/a/chromium.org/d/msg/blink-dev/4Hg0l42QHH0/H59LyoiJlI4J) in Blink recently, aligning with Gecko and the spec on this point. The big remaining difference is that Attr is not a Node in the spec, and this is a change no browser engine (except Servo) has made yet. I doubt that it's really worth it for the web platform to divorce Attr from Node.

First and foremost, it's not clear that successfully making the change would be of much value. DocumentType is already a weird kind of Node, so any algorithms dealing with nodes already have to deal with a similar case. In implementations, this is a sunk cost.

A practical issue is that it's hard to measure the risk, as every attribute and method that Attr inherits from Node and EventTarget is a potential problem. Even if that were done, the fact that `instanceof Node` no longer holds true is a risk that cannot be measured.

Are any implementors interested in attempting the change, and if so when?

annevk:
Unfortunately AttrExodus never quite became the success we all wished
it would be and the benefits with Attr no longer taking children (yay)
are rather marginal anyway.

The biggest change is to the compareDocumentPosition() algorithm. Other
noteworthy changes are that all attribute creation now needs to set a
node document.

Fixes #102.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3820,7 +3820,6 @@ the empty string instead, and then do as described below, switching on <a>contex
   </ol>
 
  <dt>{{Attr}}
- <dd>
  <dd><p>Set an existing attribute value</a> with <a>context object</a> and new value.
 
  <dt>{{Text}}
--- a/dom.bs
+++ b/dom.bs
@@ -1490,7 +1490,7 @@ can be used to explore this matter in more detail.
 
 <h3 id=node-trees>Node tree</h3>
 
-<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Attr}}, {{Text}},
+<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Text}},
 {{ProcessingInstruction}}, and {{Comment}} objects (simply called
 <dfn export id=concept-node>nodes</dfn>) <a>participate</a> in a <a>tree</a>, simply named the
 <dfn export id=concept-node-tree>node tree</dfn>.
@@ -1514,7 +1514,6 @@ can be used to explore this matter in more detail.
  <dd><p>Zero or more nodes each of which is {{Element}}, {{Text}}, {{ProcessingInstruction}}, or
  {{Comment}}.
  <dt>{{DocumentType}}
- <dt>{{Attr}}
  <dt>{{Text}}
  <dt>{{ProcessingInstruction}}
  <dt>{{Comment}}
@@ -4330,8 +4329,16 @@ To <dfn export>locate a namespace</dfn> for a <var>node</var> using
 
  <dt>{{DocumentType}}
  <dt>{{DocumentFragment}}
- <dt>{{Attr}}
  <dd>Return null.
+
+ <dt>{{Attr}}
+ <dd>
+  <ol>
+   <li><p>If its <a for=Attr>element</a> is null, then return null.
+
+   <li><p>Return the result of running <a>locate a namespace</a> on its <a for=Attr>element</a>
+   using <var>prefix</var>.
+  </ol>
 
  <dt>Any other node
  <dd>
@@ -4367,8 +4374,11 @@ method must run these steps:
 
    <dt>{{DocumentType}}
    <dt>{{DocumentFragment}}
-   <dt>{{Attr}}
    <dd>Return null.
+
+   <dt>{{Attr}}
+   <dd><p>Return the result of <a>locating a namespace prefix</a> for its <a for=Attr>element</a>,
+   if its <a for=Attr>element</a> is non-null, and null otherwise.
 
    <dt>Any other node
    <dd>Return the result of
--- a/dom.bs
+++ b/dom.bs
@@ -1490,7 +1490,7 @@ can be used to explore this matter in more detail.
 
 <h3 id=node-trees>Node tree</h3>
 
-<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Text}},
+<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Attr}}, {{Text}},
 {{ProcessingInstruction}}, and {{Comment}} objects (simply called
 <dfn export id=concept-node>nodes</dfn>) <a>participate</a> in a <a>tree</a>, simply named the
 <dfn export id=concept-node-tree>node tree</dfn>.
@@ -1514,12 +1514,12 @@ can be used to explore this matter in more detail.
  <dd><p>Zero or more nodes each of which is {{Element}}, {{Text}}, {{ProcessingInstruction}}, or
  {{Comment}}.
  <dt>{{DocumentType}}
+ <dt>{{Attr}}
  <dt>{{Text}}
  <dt>{{ProcessingInstruction}}
  <dt>{{Comment}}
  <dd><p>None.
 </dl>
-<!--AttrExodus -->
 
 The <dfn export id=concept-node-length for=Node>length</dfn> of a
 <a>node</a> <var>node</var> depends on
@@ -3523,6 +3523,9 @@ that is a <a>document</a>.
    <dt>{{Element}}
    <dd>Its {{Element/tagName}} attribute value.
 
+   <dt>{{Attr}}
+   <dd>Its <a for=Attr>qualified name</a>.
+
    <dt>{{Text}}
    <dd>"<code>#text</code>".
 
@@ -3553,10 +3556,8 @@ the first matching statement, switching on the <a>context object</a>:
  <dt>{{Element}}
  <dd><dfn const for=Node>ELEMENT_NODE</dfn> (1)
 
- <!--AttrExodus
  <dt>{{Attr}}
- <dd><dfn const for=Node>ATTRIBUTE_NODE</dfn> (2, historical);
- -->
+ <dd><dfn const for=Node>ATTRIBUTE_NODE</dfn> (2);
 
  <dt>{{Text}}
  <dd><dfn const for=Node>TEXT_NODE</dfn> (3);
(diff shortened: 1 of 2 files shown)


annevk:
I think one big thing that is not yet considered here is what should happen with certain existing APIs, such as mutation observers' `observe()`. Should that throw a `TypeError` if node is an attribute?

Are there other such APIs we need to consider?

@cdumez @foolip @smaug---- @Ms2ger @nox review appreciated.

Ms2ger:
All APIs that take `Node` should throw when passed an `Attr`, IMO. Don't care much about which exception; some may historically have been defined as throwing `InvalidNodeTypeError` or `HierarchyRequestError` or something.

annevk:
Most APIs will already throw `HierarchyRequestError` indeed. `observe()` doesn't seem to throw in implementations, but doesn't do anything either since we only ever queue attribute mutation records for elements. @smaug---- @dominiccooney @travisleithead @cdumez should we make `observe()` throw when passed an `Attr` node or just let it be useless as it is today?

annevk:
More concretely, given that the current patch matches implementations, I recommend we go with this (ideally with review though) and do any "normative changes" in a follow up.

smaug----:
I guess not throwing is good here. We don't throw when one tries to observe attribute changes using document and subtree: false either, or we don't throw observing childList changes when target is a textnode etc. So there are already many possibly ways to observe something which will never happen.

dom.bs
```diff
@@ -1490,7 +1490,7 @@ can be used to explore this matter in more detail.
 
 <h3 id=node-trees>Node tree</h3>
 
-<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Text}},
+<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Attr}}, {{Text}},
```
foolip: Attr doesn't actually participate in any tree though, it can't be appended to anything nor have any children.

dom.bs
```diff
@@ -1514,12 +1514,12 @@ can be used to explore this matter in more detail.
  <dd><p>Zero or more nodes each of which is {{Element}}, {{Text}}, {{ProcessingInstruction}}, or
  {{Comment}}.
  <dt>{{DocumentType}}
+ <dt>{{Attr}}
```
foolip: Ditto.

dom.bs
```diff
@@ -4318,6 +4324,7 @@ To <dfn export>locate a namespace</dfn> for a <var>node</var> using
 
  <dt>{{DocumentType}}
  <dt>{{DocumentFragment}}
+ <dt>{{Attr}}
```
foolip: Blink actually the equivalent of "Return the result of locating a namespace prefix for its element, if that is not null, and null otherwise"

cdumez: WebKit also calls the "locate a namespace" on the Attr's owner element.

dom.bs
```diff
@@ -4355,6 +4361,7 @@ method must run these steps:
 
    <dt>{{DocumentType}}
    <dt>{{DocumentFragment}}
+   <dt>{{Attr}}
```
foolip: This one would be "Return the result of locating a namespace prefix for its element, if that is not null, and null otherwise."

dom.bs
annevk: For the purposes of event dispatch it needs to say it does. I guess I should add a note.

annevk: Actually, maybe it doesn't.

annevk:
Anything else?

dom.bs
```diff
@@ -3827,6 +3819,10 @@ the empty string instead, and then do as described below, switching on <a>contex
    <li><p><a>Replace all</a> with <var>node</var> within the <a>context object</a>.
   </ol>
 
+ <dt>{{Attr}}
+ <dd>
```
foolip: Is this extra `<dd>` intentional?

dom.bs
```diff
@@ -4097,14 +4085,9 @@ A <a>node</a> <var>A</var>
...
   </dl>
- <li>If <var>A</var> is an <a for="/">element</a>, each
- <a>attribute</a> in its
- <a for=Element>attribute list</a> has an
- <a>attribute</a> with the same
- <a for=Attr>namespace</a>,
- <a for=Attr>local name</a>, and
- <a for=Attr>value</a> in <var>B</var>'s
- <a for=Element>attribute list</a>.
+ <li>If <var>A</var> is an <a for="/">element</a>, each <a>attribute</a> in its
```
foolip: This was the case before this change as well, but this effectively checks that A's attributes are a subset of B's, but they need to be equivalent. As implemented in Blink, it basically does this subset check one in each direction A-B and B-A.

annevk: Since there is a length check as well, what situation are you thinking of?

foolip: Oh, I missed that up in "Its namespace, namespace prefix, local name, and its number of attributes in its attribute list." Never mind then :)

dom.bs
```diff
@@ -4175,57 +4157,80 @@ returns as mask:
  <li><dfn const for=Node>DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC</dfn> (32, 20 in hexadecimal).
 </ul>
 
-The <dfn method for=Node>compareDocumentPosition(<var>other</var>)</dfn>
-method must run these steps:
+<p>The <dfn method for=Node><code>compareDocumentPosition(<var>other</var>)</code></dfn> method,
```
foolip: Here or later, wrap `contains(other)` below in `<code>` as well so that it's also orange?

dom.bs
```diff
@@ -4175,57 +4157,80 @@ returns as mask:
...
 
 <ol>
- <li>Let <var>reference</var> be the <a>context object</a>.
- <li>If <var>other</var> and <var>reference</var> are the
- same object, return zero.
+ <li><p>If <a>context object</a> is <var>other</var>, then return zero.
+
+ <li><p>Let <var>node1</var> be <var>other</var> and <var>node2</var> be <a>context object</a>.
+
+ <li><p>Let <var>attr1</var> be null.
```
foolip: Wow, this algorithm was more complicated than I thought. Is this a reverse engineering of Gecko's algorithm? I think I can spot some differences to what Blink does, and am not sure if they matter.

dom.bs
```diff
@@ -4175,57 +4157,80 @@ returns as mask:
...
+       {{Node/DOCUMENT_POSITION_FOLLOWING}}.
+      </ol>
+    </ol>
+
+   <li><p>If <var>attr2</var>'s <a for=Attr>element</a> is non-null, then set <var>node2</var> to
+   <var>attr2</var>'s <a for=Attr>element</a>.
+  </ol>
+
+ <li>
+  <p>If <var>node1</var>'s <a for=tree>root</a> is not <var>node2</var>'s <a for=tree>root</a>, then
```
foolip: This is a stellar API, I think the web platform should have more just like this.

On a more serious note, I think that when you're comparing two detached Attr nodes, you'll actually end up way down in "Return DOCUMENT_POSITION_FOLLOWING" instead of this ridiculous branch like implementations seem to.

foolip:
Wow, compareDocumentPosition is so weird. @cdumez, care to review those bits for sanity?

In Blink (and WebKit still I presume), if both nodes attributes attached to different elements, things  behave as if the attribute nodes are children of their elements, where attributes are considered to be before the real child nodes:
https://software.hixie.ch/utilities/js/live-dom-viewer/saved/4402

dom.bs
```diff
@@ -6764,31 +6778,25 @@ null.
...
- <a for=Attr>element</a> is null, set
- <a>context object</a>'s <a for=Attr>value</a> to the
- given value.
+ <li>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
+ <a for=Attr>value</a> to <var>value</var>.
 
- <li>Otherwise, <a lt="change an attribute">change</a> the
- <a>context object</a> from <a>context object</a>'s
- <a for=Attr>element</a> to the given value.
+ <li>Otherwise, <a lt="change an attribute">change</a> <var>attribute</var> from
```
foolip: Internal dialog: "What does it mean to change the value *from* an element *to* something, the value wasn't an element to begin with?!"

This wasn't new in this PR, but maybe "To change an attribute attribute *of* an element element to value, run these steps" would result in more natural language at the call site?

foolip:
OK, I've gotten through all of the changes now, great job I must say 👍

Some follow-up will probably be needed as weird things are discovered, but that's OK.

dom.bs
```diff
@@ -6697,23 +6715,19 @@ method, when invoked, must run these steps:
 
 <pre class=idl>
 [Exposed=Window]
-interface Attr {
+interface Attr : Node {
   readonly attribute DOMString? namespaceURI;
```
cdumez: So this stays on Attr instead of going to Node?

dom.bs
```diff
@@ -6697,23 +6715,19 @@ method, when invoked, must run these steps:
...
   readonly attribute DOMString? prefix;
```
cdumez: So this stays on Attr instead of going to Node?

dom.bs
```diff
@@ -6697,23 +6715,19 @@ method, when invoked, must run these steps:
...
   readonly attribute DOMString localName;
```
cdumez: So this stays on Attr instead of going to Node?

cdumez: The reason I am asking is that WebKit has these 3 on Node, instead of duplicating them between Attr and Element.

cdumez: Actually, it seems WebKit is the odd one here. Both Blink and Firefox have these attributes on Attr / Element instead of Node. So I guess I'll try to align WebKit.

cdumez:
Nothing from me. Looks good.

dom.bs
annevk: Yeah, I figured since we migrated these successfully we should keep them here.

dom.bs
annevk: Yeah, the adjustments are based on Gecko.

dom.bs
annevk: #307

dom.bs
annevk: Well spotted! Let me attempt a fix.

dom.bs
annevk: I don't like the call site stating the types. And "an" makes it even worse I think since these are very specific instances. Perhaps at some point we can change the way this is done, or adopt the abstract operation syntax from ECMAScript (I think I'd favor that).

Issue #300, opened 2016-08-17
CDATASection misses


foolip:
We missed a few things in https://github.com/whatwg/dom/pull/295, at least:

1. `appendChild(cdata)` should not throw, but does in https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity
1. `nodeName` should be `#cdata-section`
1. `normalize()` shouldn't actually touch CDATASection nodes, per my ad-hoc testing in Chrome and Firefox

There are also some places in the spec that say "switching on" that aren't wrong as such, but it could be made clearer so that nobody thinks it's possible to actually implement using a `switch(node.nodeType)`.

foolip:
I found these by searching for "switch" and "{{Text}}" in the spec and scratching my head.

annevk:
We really need IDL to define platform object branding to make some of this clearer.

foolip:
But it seems `wholeText` does merge adjacent `CDATASection` nodes in at least Blink, Edge and Gecko, so it's not actually implemented using a [contiguous Text nodes](https://dom.spec.whatwg.org/#contiguous-text-nodes) concept shared with `normalize()`.

foolip:
I have a PR for some of this, one minute.

foolip:
I looked for cases where Blink looks at the (internal) `nodeType` and not taking the same code path for `TEXT_NODE` and `CDATA_SECTION_NODE` and found two more things:

1. Range::surroundContents would throw if a CDATASection node is partially contained. That's unlikely to be intentional.
1. Two "FIXME: What about kCdataSectionNode?" in the editing code that check `nodeType` seemingly specifically to exclude CDATASection.

annevk:
Should we just fix `normalize()` in implementations?

foolip:
That's definitely worth consider, but I suspect the current behavior on `normalize()` may be intentional, if you take it as an invariant that calling `normalize()` shouldn't change what is serialized.

So assuming that it's already perfectly interoperable I think just spec'ing that would be best.

annevk:
Yeah, we should probably just define what is implemented at this point.

annevk:
I think this is good enough now until we get proper branding.

foolip:
Yep, thanks for making the spec just as crazy as implementations :)

PR #301, opened 2016-08-17
Fix and clarify more around CDATASection

foolip:
* Make appendChild(cdata) and similar not throw.
 * Let nodeName be #cdata-section.
 * Include CDATASection as a node in an exhaustive enumeration.
 * Collapse cases of Text+PI+Comment to CharacterData where possible
   so that it's more clear that subtypes are included.
 * Say "depending on" instead of "switched" on.

Part of #300.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3526,6 +3526,9 @@ that is a <a>document</a>.
    <dt>{{Text}}
    <dd>"<code>#text</code>".
 
+   <dt>{{CDATASection}}
+   <dd>"<code>#cdata-section</code>".
+
    <dt>{{ProcessingInstruction}}
    <dd>Its <a for=ProcessingInstruction>target</a>.
 
@@ -3620,6 +3623,9 @@ the first matching statement, switching on the <a>context object</a>:
 
  <dt>{{Text}}
  <dd>"<code>#text</code>".
+
+ <dt>{{CDATASection}}
+ <dd>"<code>#cdata-section</code>".
 
  <dt>{{ProcessingInstruction}}
  <dd>Its <a for=ProcessingInstruction>target</a>.
--- a/dom.bs
+++ b/dom.bs
@@ -1841,7 +1841,7 @@ of a <var>node</var> into a <var>parent</var> before a
 
  <li>
   If <var>parent</var> is a
-  <a>document</a>, and any of the statements below, switched
+  <a>document</a>, and any of the statements below, depending
   on <var>node</var>, are true, <a>throw</a> a
   {{HierarchyRequestError}}.
 
@@ -2061,7 +2061,7 @@ steps:
 
  <li>
   If <var>parent</var> is a
-  <a>document</a>, and any of the statements below, switched
+  <a>document</a>, and any of the statements below, depending
   on <var>node</var>, are true, <a>throw</a> a
   {{HierarchyRequestError}}.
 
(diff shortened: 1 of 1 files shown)


foolip:
The `normalize()` stuff from #300 is not addressed here.

dom.bs
```diff
@@ -1488,7 +1488,7 @@ can be used to explore this matter in more detail.
 
 <h3 id=node-trees>Node tree</h3>
 
-<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Text}},
+<p>{{Document}}, {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{Text}}, {{CDATASection}},
```
annevk: The reason CDATASection is not mentioned here is because ShadowRoot isn't either. We cannot have it both ways.

dom.bs
```diff
@@ -1526,9 +1527,7 @@ The <dfn export id=concept-node-length for=Node>length</dfn> of a
  <dt>{{DocumentType}}
  <dd>Zero.
 
- <dt>{{Text}}
- <dt>{{ProcessingInstruction}}
- <dt>{{Comment}}
+ <dt>{{CharacterData}}
```
annevk: See https://github.com/whatwg/dom/commit/443023ad7e5a04d050e21a206aff1b5bd66ddc07 for why I decided not to do this.

annevk:
The only thing that seems like a real issue here is `"<code>#cdata-section</code>"`.

dom.bs
foolip: Hmm, so I suppose that this is then enumerating the least-derived interfaces that can be instantiated? Otherwise it'd collapse into just saying "Node" :)

foolip: OK, I guess it's "So the new policy is to list concrete classes (DocumentFragment) unless they’re a subclass of a concrete class (ShadowRoot) and not list abstract classes (CharacterData)."

annevk: We could do something else (e.g., also use abstract classes). But without https://github.com/heycam/webidl/issues/97 I'm not sure this can be improved much.

foolip:
I've split out nodeName into https://github.com/whatwg/dom/pull/302 but think some of what's left would be worthwhile, will rebase and comment inline.

foolip:
OK, so after looking at what remained again, I'm left with just https://github.com/whatwg/dom/pull/301/commits/eb9252f00e085b0b6ce5e45d8c188dc548536cb2

Feel free to merge or reject and move on :)

annevk:
I prefer the term "switch" from "switch case".

foolip:
I don't understand, "switch case" doesn't appear in the spec. Do you prefer "depending on the context object" or "switching on the context object"? Both are now used and I picked the one that I thought was less likely to be blindly translated to switch-case in C++.

annevk:
I prefer the latter, but perhaps you're right that the current setup doesn't work. In particular branching on type seems wrong.

foolip:
New issue, or do you have a plan?

annevk:
Branding in IDL is still the only thing that resolves this I think. New issue to track that on DOM side seems fine.

foolip:
I see some discussion of branding in https://github.com/heycam/webidl/issues/125, is there a larger tracking issue for whatever is involved?

annevk:
No, that's it. I think we could clarify DOM meanwhile by putting CDATASection checks before Text. Since Text checks are true for both, but CDATASection will only be true for CDATASection.

annevk:
(Are you okay with deleting this branch btw?)

foolip:
Branch deleted, all that's needed to reproduce is to change either "depending on the context object" or "switching on the context object" to the other for consistency.

foolip:
One relevant difference between CDATASection and ShadowRoot is that CDATASection has its own nodeType. So switching on nodeType actually works for ShadowRoot, but not for CDATASection. Since we'll presumably never add a new nodeType, actually spelling out whether CDATASection is included anywhere that Text is mentioned would remove the risk of confusion.

PR #306, opened 2016-08-18
Shadow: Rename "unclosed node"

### LINKED ISSUE #285: "unclosed node of node" reads confusingly
Usually in a construction like "ADJ node of node", we expect the adjective to be a relational one, relating the first node to the second in some way, like "ancestor" or "sibling". But "unclosed" is a *personal* adjective describing the state of the first node, with no relational semantics at all, so when this phrase is is used it looks like a typo.  You can't rearrange it into similar phrases, either: "A is an ancestor node of node B" can reverse to "B's ancestor nodes", but "B's unclosed nodes" doesn't read well to me.

I've got a few suggestions for alternative wording, but I'm not married to any of them:

* A is an unclosed relative of B
* A is potentially reachable from B

The first one is a minimal change, just adding a relationship word to make it read better.  I like the second one a lot better, tho - it expresses the intent of the relationship a lot better.

hayatoito:
Now we use "closed-shadow-hidden" instead. Fixes #285.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -757,9 +757,10 @@ steps:
  <li>
   <p>For each <var>tuple</var> in <a>context object</a>'s <a for=Event>path</a>:
 
-  <ol><li><p>If <var>currentTarget</var> is a <a>node</a> and <var>tuple</var>'s <b>item</b> is an
-  <a>unclosed node</a> of <var>currentTarget</var>, or <var>currentTarget</var> is <em>not</em> a
-  <a>node</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.</p></li></ol>
+  <ol><li><p>If <var>currentTarget</var> is a <a>node</a> and <var>tuple</var>'s <b>item</b> is not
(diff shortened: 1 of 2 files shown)


annevk:
You want to run "git pull" and "bikeshed update" in your local bikeshed directory. I can fix it for this PR however.

annevk:
Also, I think prefer "A node _A_ is closed-shadow-hidden from a node _B_". Let me push an update so you can review.

hayatoito:
Thanks. Hmm. I did "git pull && bikeshed update" before sending a pull request. :(

The output, dom.html, is same in either case of:
1. `curl https://api.csswg.org/bikeshed/ -f -F file=@dom.bs >! dom.html`
2. 'git pull && bikeshed update && bikeshed spec dom.bs'

Maybe something is wrong.

annevk:
That is weird, the same happened to @domenic. Any ideas @tabatkins?

Anyway, okay if I merge this?

hayatoito:
Yeah, the fix looks good to me. Thanks!

PR #310, opened 2016-08-22
CDATASection nodes don't like to be normalized away

annevk:
Exclusive Text nodes to the rescue. Fixes part 3 of #300.


annevk:
If you don't get to it today, please don't merge it tomorrow if it's all good. Would have to fix the date first.

dom.bs
```diff
@@ -3822,87 +3822,55 @@ the empty string instead, and then do as described below, switching on <a>contex
  <a>nodes</a>.
 </dl>
 
+<p>The <dfn export>contiguous exclusive {{Text}} nodes</dfn> of a <a>node</a> <var>node</var> are
```
foolip: Maybe put this and the "exclusive Text node" concept right after wholeText, or in some way together and close to the "contiguous Text nodes" concept? It should help make it obvious without scrolling around what the difference between these things are. (I was first going to suggest cross-linked notes, that'd be another way.)

dom.bs
```diff
@@ -3822,87 +3822,55 @@ the empty string instead, and then do as described below, switching on <a>contex
...
+<var>node</var>, <var>node</var>'s <a>previous sibling</a> <a>exclusive <code>Text</code> node</a>,
+if any, and its <a>contiguous exclusive <code>Text</code> nodes</a>, and <var>node</var>'s
+<a>next sibling</a> <a>exclusive <code>Text</code> node</a>, if any, and its
+<a>contiguous exclusive <code>Text</code> nodes</a>, avoiding any duplicates.
+
 <p>The <dfn method for=Node><code>normalize()</code></dfn> method, when invoked, must run these
-steps for each {{Text}} <a>node</a> <a>descendant</a> of <a>context object</a>:
+steps for each <a>exclusive <code>Text</code> node</a> <var>node</var> <a>descendant</a> of
```
foolip: I think "each descendant exclusive Text node *node*" would read slightly better.

dom.bs
```diff
@@ -3822,87 +3822,55 @@ the empty string instead, and then do as described below, switching on <a>contex
...
- {{CharacterData/length}} attribute value.
+ <li>If <var>length</var> is zero, then <a>remove</a> <var>node</var> and continue with the next
+ <a>exclusive <code>Text</code> node</a>, if any.
 
- <li>If <var>length</var> is zero,
- <a>remove</a> <var>node</var> and
- continue with the next {{Text}}
- <a>node</a>, if any.
+ <li>Let <var>data</var> be the concatenation of the <a>data</a> of <var>node</var>'s
+ <a>contiguous exclusive <code>Text</code> nodes</a> (excluding itself), in <a>tree order</a>.
```
foolip: The "excluding itself" bit looks unnecessary now. You could instead just define a concept of "following exclusive Text nodes" or similar.

dom.bs
```diff
@@ -3822,87 +3822,55 @@ the empty string instead, and then do as described below, switching on <a>contex
...
- <var>node</var>, offset <var>length</var>,
- count 0, and data <var>data</var>.
+ <li>Let <var>currentNode</var> be <var>node</var>'s <a>next sibling</a>.
 
- <li>Let <var>current node</var> be <var>node</var>'s
- <a>next sibling</a>.
+ <li>
+  <p>While <var>currentNode</var> is an <a>exclusive <code>Text</code> node</a>:
 
- <li>While <var>current node</var> is a {{Text}} node:
```
foolip: Reviewed using git diff --word-diff -w, nothing to see here.

foolip:
LGTM % nits

dom.bs
annevk: I don't follow the second sentence, but I agree that it's not needed here.

foolip: I meant you could have a concept equivalent to what "contiguous exclusive `Text` nodes (excluding itself)" does in this concept. Since the previous sibling can't be an exclusive `Text` node and the node itself is excluded, all that's left is "following exclusive `Text` nodes".

dom.bs
```diff
@@ -3821,7 +3815,7 @@ steps for each <a>exclusive <code>Text</code> node</a> <var>node</var> <a>descen
  <a>exclusive <code>Text</code> node</a>, if any.
 
  <li>Let <var>data</var> be the concatenation of the <a>data</a> of <var>node</var>'s
- <a>contiguous exclusive <code>Text</code> nodes</a> (excluding itself), in <a>tree order</a>.
```
annevk: But we can't really change the concept given the other callsites. Prolly best to just not change this.

annevk:
Take another look?  I guess if you have a different idea in mind it might be easiest to add a commit.

dom.bs
foolip: Hmm, the only other call site I see is "Remove node’s contiguous exclusive Text nodes (excluding itself), in tree order." That's also excluding node, it's removing exactly the same nodes whose data were concatenated earlier, anything else would be a bug.

foolip:
OK, I started poking at how to do it, noticed instead that there was a problem in the domintro, and now I think I've spent enough time on CDATASection nodes for a while. LGTM.

PR #312, opened 2016-08-23
Further Attr node fixes for compareDocumentPosition()

### LINKED ISSUE #309: compareDocumentPosition with Attr nodes still not quite right
Context:
https://github.com/whatwg/dom/pull/299/files/82df412250bd1321562cbe50574219860b976645#r75397386
https://github.com/whatwg/dom/pull/299#issuecomment-240875443

There was a partial fix in https://github.com/whatwg/dom/pull/299/commits/6ae89edd25191d59822f90f2b5558ab22593048e, but when two Attr nodes attached to elements in the same tree and when you're comparing an Attr node to an element in the same tree it shouldn't do the pointer comparison, but instead behave as if the Attr nodes are children of its element.

@cdumez

annevk:
This algorithm is impossible. Hopefully fixes #309.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4167,16 +4167,21 @@ when invoked, must run these steps:
   JavaScript implementations a cached <code class='lang-javascript'>Math.random()</code> value can
   be used.
 
- <li><p>If <var>node1</var> is an <a>ancestor</a> of <var>node2</var>, or <var>node1</var> is
- <var>node2</var> and <var>attr2</var> is non-null, then return the result of adding
- {{Node/DOCUMENT_POSITION_CONTAINS}} to {{Node/DOCUMENT_POSITION_PRECEDING}}.
+ <li><p>If <var>node1</var> is an <a>ancestor</a> of <var>node2</var> and <var>attr1</var> is null,
+ or <var>node1</var> is <var>node2</var> and <var>attr2</var> is non-null, then return the result of
+ adding {{Node/DOCUMENT_POSITION_CONTAINS}} to {{Node/DOCUMENT_POSITION_PRECEDING}}.
 
- <li><p>If <var>node1</var> is a <a>descendant</a> of <var>node2</var>, or <var>node1</var> is
- <var>node2</var> and <var>attr1</var> is non-null, then return the result of adding
- {{Node/DOCUMENT_POSITION_CONTAINED_BY}} to {{Node/DOCUMENT_POSITION_FOLLOWING}}.
+ <li><p>If <var>node1</var> is a <a>descendant</a> of <var>node2</var> and <var>attr1</var> is null,
+ or <var>node1</var> is <var>node2</var> and <var>attr1</var> is non-null, then return the result of
+ adding {{Node/DOCUMENT_POSITION_CONTAINED_BY}} to {{Node/DOCUMENT_POSITION_FOLLOWING}}.
 
- <li><p>If <var>node1</var> is <a>preceding</a> <var>node2</var>, then return
- {{Node/DOCUMENT_POSITION_PRECEDING}}.
+ <li>
+  <p>If <var>node1</var> is <a>preceding</a> <var>node2</var>, then return
+  {{Node/DOCUMENT_POSITION_PRECEDING}}.
+
+  <p class="note">Due to the way <a>attributes</a> are handled in this algorithm this results in a
+  <a>node</a>'s <a>attributes</a> counting as <a>preceding</a> that <a>node</a>'s <a>children</a>,
+  despite <a>attributes</a> not <a>participating</a> in a <a>tree</a>.
 
  <li><p>Return {{Node/DOCUMENT_POSITION_FOLLOWING}}.
 </ol>
--- a/dom.bs
+++ b/dom.bs
@@ -4140,16 +4140,16 @@ when invoked, must run these steps:
 
     <ol>
      <li>
-      <p>For each <a>attribute</a> <var>node2Attr</var> in <var>node2</var>'s
+      <p>For each <a>attribute</a> <var>attr</var> in <var>node2</var>'s
       <a for=Element>attribute list</a>:
 
       <ol>
-       <li><p>If <var>node2Attr</var> <a for=Node>equals</a> <var>attr1</var>, then return the
-       result of adding {{Node/DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC}} and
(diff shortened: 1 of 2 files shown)


annevk:
(Please don't merge.)

dom.bs
```diff
@@ -4136,51 +4136,41 @@ when invoked, must run these steps:
...
    <li>
-    <p>If <var>attr2</var>'s <a for=Attr>element</a> is <var>node1</var> and <var>attr1</var> is
-    non-null, then:
+    <p>If <var>node2</var> is <var>node1</var> and <var>attr1</var> is non-null, then:
 
     <ol>
      <li>
-      <p>For each <a>attribute</a> <var>node2Attribute</var> in <var>attr2</var>'s
-      <a for=Attr>element</a>'s <a for=Element>attribute list</a>:
+      <p>For each <a>attribute</a> <var>node2Attr</var> in <var>node2</var>'s
```
foolip: Segfault, here it's possible for node1 and node2 to both be null while attr1 is non-null. I think the check needs to be "If node1 and attr1 are non-null and node1 is node2" or similar, i.e. add a null-check for node2.

annevk: Done.

foolip: Reading this anew, I also wonder if it'd be more clear to call *node2Attr* simply *attr* or something without a 2 in it, since node1==node2 here.

annevk: Good call.

dom.html
```diff
@@ -2519,51 +2519,41 @@ <h3 class="heading settled" data-level="4.4" id="interface-node"><span class="se
...
   either <code class="idl"><a data-link-type="idl" href="#dom-node-document_position_preceding">DOCUMENT_POSITION_PRECEDING</a></code> or <code class="idl"><a data-link-type="idl" href="#dom-node-document_position_following">DOCUMENT_POSITION_FOLLOWING</a></code>, with the
   constraint that this is to be consistent, together. </p>
      <p class="note no-backref" role="note">Whether to return <code class="idl"><a data-link-type="idl" href="#dom-node-document_position_preceding">DOCUMENT_POSITION_PRECEDING</a></code> or <code class="idl"><a data-link-type="idl" href="#dom-node-document_position_following">DOCUMENT_POSITION_FOLLOWING</a></code> is typically implemented via pointer comparison. In
   JavaScript implementations a cached <code class="lang-javascript highlight"><span class="nb">Math</span><span class="p">.</span><span class="nx">random</span><span class="p">()</span> </code> value can
```
foolip: Looks like a Bikeshed bug on this line, an extra space before `</code>` that you can see in the rendered result. @tabatkins, known issue?

dom.bs
```diff
@@ -4176,11 +4167,13 @@ when invoked, must run these steps:
   JavaScript implementations a cached <code class='lang-javascript'>Math.random()</code> value can
   be used.
 
- <li><p>If <var>node1</var> is an <a>ancestor</a> of <var>node2</var>, then return the result of
- adding {{Node/DOCUMENT_POSITION_CONTAINS}} to {{Node/DOCUMENT_POSITION_PRECEDING}}.
+ <li><p>If <var>node1</var> is an <a>ancestor</a> of <var>node2</var>, or <var>node1</var> is
```
foolip: Oh my, this is still a problem when an attribute is compared to a descendant of its element. Because one of *node1* and *node2* will become the attribute's element, either this or the next step will match. Maybe "If *node1* is an ancestor of *node2* <ins>and *attr1* is null</ins>, or *node1* is *node2* and *attr2* is non-null" and similarly for the next step?

dom.bs
```diff
@@ -4176,11 +4167,13 @@ when invoked, must run these steps:
...
+ <var>node2</var> and <var>attr2</var> is non-null, then return the result of adding
+ {{Node/DOCUMENT_POSITION_CONTAINS}} to {{Node/DOCUMENT_POSITION_PRECEDING}}.
 
- <li><p>If <var>node1</var> is a <a>descendant</a> of <var>node2</var>, then return the result of
- adding {{Node/DOCUMENT_POSITION_CONTAINED_BY}} to {{Node/DOCUMENT_POSITION_FOLLOWING}}.
+ <li><p>If <var>node1</var> is a <a>descendant</a> of <var>node2</var>, or <var>node1</var> is
+ <var>node2</var> and <var>attr1</var> is non-null, then return the result of adding
+ {{Node/DOCUMENT_POSITION_CONTAINED_BY}} to {{Node/DOCUMENT_POSITION_FOLLOWING}}.
 
  <li><p>If <var>node1</var> is <a>preceding</a> <var>node2</var>, then return
```
foolip: A  note around here that explains that we can end up here for attributes and that they are considered to come before the real descendants would be nice, since it's going to be impossible to extract intention from the algorithm alone.

foolip:
OK, finished reviewing, including  fixes up to "node2attr -> attr". Things that are impossible just take longer.

dom.bs
annevk: How can we end up here with either node1 or node2 being null?

annevk: I think I understand what you mean.

annevk:
I think I fixed the last two non-bikeshed issues.

tabatkins:
Looking into the Bikeshed issue.

dom.bs
```diff
@@ -4176,14 +4167,21 @@ when invoked, must run these steps:
   JavaScript implementations a cached <code class='lang-javascript'>Math.random()</code> value can
   be used.
 
- <li><p>If <var>node1</var> is an <a>ancestor</a> of <var>node2</var>, then return the result of
+ <li><p>If <var>node1</var> is an <a>ancestor</a> of <var>node2</var> and <var>attr1</var> is null,
+ or <var>node1</var> is <var>node2</var> and <var>attr2</var> is non-null, then return the result of
  adding {{Node/DOCUMENT_POSITION_CONTAINS}} to {{Node/DOCUMENT_POSITION_PRECEDING}}.
 
- <li><p>If <var>node1</var> is a <a>descendant</a> of <var>node2</var>, then return the result of
+ <li><p>If <var>node1</var> is a <a>descendant</a> of <var>node2</var> and <var>attr1</var> is null,
```
foolip: Oops, this one needs to say "and *attr2* is null"

foolip:
Reviewed "review comments", found a typo, but otherwise this now looks great, I like the added note!

tabatkins:
And should be fixed.

foolip:
Thanks Tab!

PR #313, opened 2016-08-23
Define a document's origin instead of linking to HTML's generic "origin" definition.

jyasskin:
I was going to update this to use `for=/` to link to HTML's "origin", but it turns out that all the uses of "origin" here refer to a document's origin, which should be its own definition.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -66,7 +66,6 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: concept-document-bc
             text: concept-document-window
             text: unit of related similar-origin browsing contexts
-            text: origin; url: concept-origin
         urlPrefix: infrastructure.html
             text: in parallel
             text: document base URL
@@ -4541,7 +4540,7 @@ Unless stated otherwise, a <a>document</a>'s <a for=Document>encoding</a> is the
 <a for=Document>type</a> is "<code>xml</code>", and its
 <a for=Document>mode</a> is "<code>no-quirks</code>".
 
-Unless stated otherwise, a <a>document</a>'s <a>origin</a> is a new <a>opaque origin</a>. [[!HTML]]
+Unless stated otherwise, a <a>document</a>'s <dfn for=document>origin</dfn> is a new <a>opaque origin</a>. [[!HTML]]
 
 A <a>document</a> is said to be an <dfn export>XML document</dfn> if its
 <a for=Document>type</a> is "<code>xml</code>", and an <dfn export>HTML document</dfn>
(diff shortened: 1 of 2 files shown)


domenic:
Document's origin is defined in https://html.spec.whatwg.org/multipage/browsers.html#concept-origin, however. (Specifically in the section that says "For Document objects".)

jyasskin:
Woah, I missed that. Hm... I think this spec is still the original definition of the document's origin, since HTML says 

> If the Document was obtained in some other manner … The default behaviour as defined in the WHATWG DOM standard applies. 

That said, if you think HTML should provide the `<dfn for=document>`, that also makes sense to me. I'm just going for a consistent convention that when we write "A foo's bar", that we have a `<dfn for=foo>bar</dfn>`.

domenic:
Yeah, I'm not sure whether HTML or DOM is the best place to define "a document's origin"... Maybe @annevk can decide :).

annevk:
DOM should probably own the slot with HTML filling it in mostly. I still need to fix document creation in HTML. 😟

jyasskin:
@tabatkins For readability, this is a case where it'd be really nice for the definition panel to link to uses of the definition from HTML, since when reading about a document's origin, I want to be able to easily find all the places it's set.

domenic:
OK cool, so this is ready to merge then I think? I'll let @annevk do the merge, but IMO we should help get this all straightened out ASAP, and it's great that @jyasskin has done most of the work for us :).

annevk:
Very weird, I'm sure @jyasskin has up-to-date bikeshed, but when I generate locally I get again something which has much less `<span>` noise. I wonder what is up with that.

annevk:
Thanks, landed as https://github.com/whatwg/dom/commit/e4b7778c543f0674454fa4eccf323ffde3f691fb with tweaks.

domenic:
@annevk given that this is the third time, maybe you should double-check your Bikeshed is updated? :) I think it's newer Bikeshed that has span noise.

Possibly your Pygments is not up to date, even if your Bikeshed is? Not sure how to check that exactly.

annevk:
I thought the span noise was removed? I update pretty regularly and did check… @tabatkins, thoughts?

jyasskin:
This particular span noise is caused by an out-of-date pygments. It turned out that *my* pygments was out of date, and `pip install --upgrade pygments` made my bikeshed runs match yours. I've been avoiding the problem recently by having TravisCI use https://api.csswg.org/bikeshed/ to build specs for me: https://github.com/WebBluetoothCG/web-bluetooth/blob/master/deploy.sh

annevk:
@jyasskin I updated pygments and now I get the span noise by the way. So maybe with "yours" you didn't mean me or it's a bigger mystery.

PR #315, opened 2016-08-26
Editorial: integrate more explicitly with IDL for callbacks

### LINKED ISSUE #303: Callback functions/interfaces need to be invoked with WebIDL concepts
This is needed to explain how you end up at https://html.spec.whatwg.org/multipage/webappapis.html#clean-up-after-running-script during event dispatch, which @jakearchibald mentioned in https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/

@domenic added IDs in https://github.com/heycam/webidl/pull/146

annevk:


dom.bs
```diff
@@ -8715,10 +8706,9 @@ these steps:
  <!-- !((1 << (node.nodeType - 1)) & whatToShow) -->
  <li>If <a for=traversal>filter</a> is null,
  return {{NodeFilter/FILTER_ACCEPT}}.
- <li>Let <var>result</var> be the return value of calling
- <a for=traversal>filter</a>'s
- {{NodeFilter/acceptNode()}} with <var>node</var> as
- argument. Rethrow any exceptions.
+ <li>Let <var>result</var> be the return value of <a>call a user object's operation</a> with
+ <a for=traversal>filter</a>, "<code>acceptNode</code>", and a list of arguments consisting of
```
domenic: I'd keep this as `{{NodeFilter/acceptNode()}}` instead of `"<code>acceptNode</code>"`, seems nicer.

annevk: It needs to be a string though…

domenic: Yeah, but it can be a string that links nicely... Similar to how in https://dom.spec.whatwg.org/#dfnReturnLink-3 we use `<{html}>` as a string (local name) and it seems OK.

annevk: Don't the parenthesis ruin it?

domenic: Oh, you're totally right, nevermind then.

dom.bs
```diff
@@ -1364,9 +1357,10 @@ an <var>object</var> with <var>event</var>, run these steps:
    <li><p>If <var>listener</var>'s <b>passive</b> is true, set <var>event</var>'s
    <a>in passive listener flag</a>.
 
-   <li><p>Call <var>listener</var>'s <b>callback</b>'s {{EventListener/handleEvent()}}, with
-   <var>event</var> as argument and <var>event</var>'s {{Event/currentTarget}} attribute value as
-   <a>callback this value</a>. If this throws an exception, <a>report the exception</a>.
+   <li><p><a>Call a user object's operation</a> with <var>listener</var>'s <b>callback</b>,
+   "<code>handleEvent</code>", a list of arguments consisting of <var>event</var>, and
+   <var>event</var>'s {{Event/currentTarget}} attribute value as <a>callback this value</a>. If this
```
domenic: I think it should be "the" callback this value here and below

domenic:
LGTM with "the" before callback this value

PR #318, opened 2016-08-30
Drop "invoking" from "invoking retargeting"

TakayoshiKochi:
This was pointed out during pointerLock spec change review:
https://github.com/w3c/pointerlock/pull/8

I'd like to get this reviewed by a native speaker whether it looks less
unnatural.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1187,7 +1187,7 @@ for discussion).
   <!-- We should consider refactoring it to make it have less of a scope, since we don't really want
        folks to start using it for non-legacy scenarios. -->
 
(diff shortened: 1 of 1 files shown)


annevk:
This looks better to me. @domenic care to do the native speaker review?

domenic:
Yeah this is nicer.

I think either "invoking retargeting with (...arguments...)" or "retargeting (...arguments...)" would be OK, with the former being a bit more explicit and the latter being more natural.

annevk:
Thanks, landed as https://github.com/whatwg/dom/commit/de60f841360f5995be0b49eebe7cda1c32275be8.

TakayoshiKochi:
Thanks!

PR #321, opened 2016-08-31
Mark ATTRIBUTE_NODE and SHOW_ATTRIBUTE as no longer historical

foolip:
Follow-up to #102.


foolip:
The only thing you can do with TreeWalker and Attr nodes is to get the node itself back, but that's how it works :)

annevk:
Ta!

PR #327, opened 2016-09-13
Shadow: Make Event.composedPath() use closed-shadow-hidden when currentTarget is window

### LINKED ISSUE #326: event.composedPath() should not return every items in path if event.currentTarget is window
According to the current composedPath() definition, window.composedPath() would
return every items, even if the item's root is a closed shadow root, because
window is not a node.

hayatoito:
Fixes #326.


dom.bs
```diff
@@ -744,11 +744,16 @@ steps:
...
-
-  <ol><li><p>If <var>currentTarget</var> is a <a>node</a> and <var>tuple</var>'s <b>item</b> is not
-  <a>closed-shadow-hidden</a> from <var>currentTarget</var>, or <var>currentTarget</var> is not a
-  <a>node</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.</p></li></ol>
-
+  <ol>
+   <li><p>If <var>currentTarget</var> is a {{Window}}, then:
+    <ol><li><p>If <var>tuple</var>'s <b>item</b> is not a <a>node</a>, or <var>tuple</var>'s
+    <b>item</b> is not <a>closed-shadow-hidden</a> from <var>tuple</var>'s <b>item</b>'s <a>node
+    document</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.</ol>
```
annevk: Why node document? Would shadow-including root not be a little cleaner if that's what you're going for? (I guess there's maybe a difference too, if the node is not currently in a document, although that would not yield a different result I think.)

dom.bs
```diff
@@ -744,11 +744,16 @@ steps:
 
  <li>
   <p>For each <var>tuple</var> in <a>context object</a>'s <a for=Event>path</a>:
-
-  <ol><li><p>If <var>currentTarget</var> is a <a>node</a> and <var>tuple</var>'s <b>item</b> is not
-  <a>closed-shadow-hidden</a> from <var>currentTarget</var>, or <var>currentTarget</var> is not a
-  <a>node</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.</p></li></ol>
-
+  <ol>
+   <li><p>If <var>currentTarget</var> is a {{Window}}, then:
```
annevk: {{Window}} object

dom.bs
```diff
@@ -744,11 +744,16 @@ steps:
 
  <li>
   <p>For each <var>tuple</var> in <a>context object</a>'s <a for=Event>path</a>:
-
-  <ol><li><p>If <var>currentTarget</var> is a <a>node</a> and <var>tuple</var>'s <b>item</b> is not
-  <a>closed-shadow-hidden</a> from <var>currentTarget</var>, or <var>currentTarget</var> is not a
-  <a>node</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.</p></li></ol>
-
+  <ol>
```
annevk: Newline before `<ol>`

dom.bs
```diff
@@ -744,11 +744,16 @@ steps:
...
+   <li><p>If <var>currentTarget</var> is a {{Window}}, then:
+    <ol><li><p>If <var>tuple</var>'s <b>item</b> is not a <a>node</a>, or <var>tuple</var>'s
+    <b>item</b> is not <a>closed-shadow-hidden</a> from <var>tuple</var>'s <b>item</b>'s <a>node
+    document</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.</ol>
+
+   <li><p>Otherwise if <var>currentTarget</var> is a <a>node</a> and <var>tuple</var>'s <b>item</b>
+   is not <a>closed-shadow-hidden</a> from <var>currentTarget</var>, or <var>currentTarget</var> is
+   not a <a>node</a>, then append <var>tuple</var>'s <b>item</b> to <var>composedPath</var>.
+  </ol>
```
annevk: Newline after `</ol>`

dom.bs
hayatoito: You are correct! I forgot that we can use `shadow-including root` here.

annevk: Also missing `</p></li>`

dom.bs
annevk: Also needs a newline before the `<p>` since its not the only child of the `<li>`

hayatoito: Done.

dom.bs
hayatoito: Done.

dom.bs
hayatoito: Done.

hayatoito:
Thank you for the review. I have updated the PR again.

dom.bs
hayatoito: Done.

annevk:
Just to be sure, the note after https://dom.spec.whatwg.org/#dom-event-composedpath still makes sense, right?

hayatoito:
Yeah, I think this note remains true. Actually, I tried to update the note, but I found we do not have to update it.

annevk:
So the only annoyance with merging this is that due to https://github.com/tabatkins/bikeshed/issues/815 we'll introduce a linking regression.

annevk:
I guess I should just accept that for now.

Issue #325, opened 2016-09-02
Define how script-created click events are special during dispatch


foolip:
https://w3c.github.io/uievents/#trusted-events
https://w3c.github.io/uievents/#default-action

If one starts reading at https://dom.spec.whatwg.org/#dom-eventtarget-dispatchevent there is no way to discover that click events are special. Probably it should be handled in the dispatch algorithm itself, or perhaps there should be an explicit extension point that UI Events could use.

Related to https://github.com/w3c/uievents/issues/104 and https://github.com/w3c/uievents/issues/106

@travisleithead @garykac @dtapuska @RByers

annevk:
As part of this we should resolve https://www.w3.org/Bugs/Public/show_bug.cgi?id=12230. Legacy wins once again.

domenic:
I wonder if we should work on PRs to UI events to fix those two bugs you mention, @foolip? They are long-standing and pervasive problems with UI events that have caused a lot of confusion throughout the ecosystem. (Also https://github.com/w3c/uievents/issues/95 and https://github.com/w3c/uievents/issues/90 maybe...)

Curious what the editors of UI events think.

foolip:
I'd also like to hear input from UI Events editors. I'm also reminded of https://groups.google.com/a/chromium.org/d/msg/blink-dev/WEMnlVIbe70/hP7fhHP0AAAJ (from @phistuck) about what kind of event it has to be. It seems likely that allowing any kind of event will work with more content and also makes the layering simpler.

annevk:
I talked about this a while back with @smaug----. https://dom.spec.whatwg.org/#concept-event-dispatch uses "get the parent" to determine where to invoke listeners. Something similar is used at the end of dispatch. https://www.w3.org/Bugs/Public/show_bug.cgi?id=12230#c29 also explains this. That is what we should use here and we should only do it if the canceled flag is not set.

Then we can let specifications define "legacy post-dispatch steps" for objects that need them (only `<a>`?). (I don't want to call it "default action", not even prefixed with "legacy". I think that would continue to cause much confusion.)

Does anyone see problems with that?

annevk:
(It's an awful lot of machinery for such an edge case, but I don't really see a better way.)

garykac:
Sorry for not seeing this thread earlier. We definitely want to fix this in
UI Events, but there are some subtleties in the terminology that I want to
make sure I get right when making this change.

Also, since people will still search for "default action" (it's used
[...]

PR #336, opened 2016-10-03
Do not dispatch a trusted event when target and reletedTarget are identical

hayatoito:
…Target are identical

Change the behavior of event dispatch as follows:

1. Make an event dispatch happen until the root of the tree when target
and relatedTarget are identical in synthetic events.

2. Do not dispatch an event when retargeted-target and
retargeted-reletedTarget are identical, but target and relatedTarget are
not identical.

(2) is a bug fix. This sitution happens in UA events. For example,
if a mouse pointer moves from an element in a shadow tree to
its shadow host, a "mouseover" event should not be dispatched on the
shadow host. The current spec allows a non-empty event path in this case.

Fixes https://github.com/w3c/webcomponents/issues/577

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1201,8 +1201,8 @@ for discussion).
  <a>relatedTarget</a> against <var>target</var> if <var>event</var>'s <a>relatedTarget</a> is
  non-null, and null otherwise.
 
- <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true and <var>target</var> and
- <var>relatedTarget</var> are identical, terminate these steps.
+ <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true and <var>target</var> is
+ <var>relatedTarget</var>, then terminate these steps.
 
  <li><p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>'s
  <a for=Event>path</a>.
--- a/dom.bs
+++ b/dom.bs
@@ -1201,9 +1201,8 @@ for discussion).
  <a>relatedTarget</a> against <var>target</var> if <var>event</var>'s <a>relatedTarget</a> is
  non-null, and null otherwise.
 
- <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true, <var>target</var> is
- <var>relatedTarget</var>, and <var>target</var> is not <var>event</var>'s <a>relatedTarget</a>,
- then return true.
+ <li><p>If <var>target</var> is <var>relatedTarget</var> and <var>target</var> is not
+ <var>event</var>'s <a>relatedTarget</a>, then return true.
 
  <li><p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>'s
  <a for=Event>path</a>.
--- a/dom.bs
+++ b/dom.bs
@@ -1201,8 +1201,9 @@ for discussion).
  <a>relatedTarget</a> against <var>target</var> if <var>event</var>'s <a>relatedTarget</a> is
  non-null, and null otherwise.
 
- <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true and <var>target</var> is
- <var>relatedTarget</var>, then terminate these steps.
+ <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true, <var>target</var> is
+ <var>relatedTarget</var>, and <var>target</var> is not <var>event</var>'s <a>relatedTarget</a>,
+ then terminate these steps.
 
  <li><p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>'s
  <a for=Event>path</a>.
--- a/dom.bs
+++ b/dom.bs
@@ -1203,7 +1203,7 @@ for discussion).
 
  <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true, <var>target</var> is
  <var>relatedTarget</var>, and <var>target</var> is not <var>event</var>'s <a>relatedTarget</a>,
- then terminate these steps.
+ then return true.
 
  <li><p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>'s
  <a for=Event>path</a>.
--- a/dom.bs
+++ b/dom.bs
@@ -1201,6 +1201,9 @@ for discussion).
  <a>relatedTarget</a> against <var>target</var> if <var>event</var>'s <a>relatedTarget</a> is
  non-null, and null otherwise.
 
+ <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true and <var>target</var> and
+ <var>relatedTarget</var> are identical, terminate these steps.
+
  <li><p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>'s
  <a for=Event>path</a>.
 
--- a/dom.html
+++ b/dom.html
@@ -882,7 +882,7 @@ of non-<code class="idl"><a data-link-type="idl" href="#dom-addeventlisteneropti
  non-null, and null otherwise. 
     <li>
      <p>If <var>event</var>’s <code class="idl"><a data-link-type="idl" href="#dom-event-istrusted">isTrusted</a></code> attribute is true, <var>target</var> is <var>relatedTarget</var>, and <var>target</var> is not <var>event</var>’s <a data-link-type="dfn" href="#event-relatedtarget">relatedTarget</a>,
- then terminate these steps. </p>
+ then return true. </p>
     <li>
      <p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>’s <a data-link-type="dfn" href="#event-path">path</a>. </p>
     <li>
--- a/dom.html
+++ b/dom.html
@@ -881,7 +881,7 @@ of non-<code class="idl"><a data-link-type="idl" href="#dom-addeventlisteneropti
     <li>Let <var>relatedTarget</var> be the result of <a data-link-type="dfn" href="#retarget">retargeting</a> <var>event</var>’s <a data-link-type="dfn" href="#event-relatedtarget">relatedTarget</a> against <var>target</var> if <var>event</var>’s <a data-link-type="dfn" href="#event-relatedtarget">relatedTarget</a> is
  non-null, and null otherwise. 
     <li>
-     <p>If <var>event</var>’s <code class="idl"><a data-link-type="idl" href="#dom-event-istrusted">isTrusted</a></code> attribute is true and <var>target</var> and <var>relatedTarget</var> are identical, terminate these steps. </p>
+     <p>If <var>event</var>’s <code class="idl"><a data-link-type="idl" href="#dom-event-istrusted">isTrusted</a></code> attribute is true and <var>target</var> is <var>relatedTarget</var>, then terminate these steps. </p>
     <li>
      <p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>’s <a data-link-type="dfn" href="#event-path">path</a>. </p>
     <li>
--- a/dom.html
+++ b/dom.html
@@ -71,7 +71,7 @@ pre.idl.highlight { color: #708090; }
    <p data-fill-with="logo"><a class="logo" href="https://whatwg.org/"> <img alt="WHATWG" height="100" src="https://resources.whatwg.org/logo-dom.svg"> </a> </p>
    <hgroup>
     <h1 class="p-name no-ref allcaps" id="title">DOM</h1>
-    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Living Standard — Last Updated <time class="dt-updated" datetime="2016-10-14">14 October 2016</time></span></h2>
+    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Living Standard — Last Updated <time class="dt-updated" datetime="2016-10-17">17 October 2016</time></span></h2>
    </hgroup>
    <div data-fill-with="spec-metadata">
     <dl>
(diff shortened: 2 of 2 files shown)


rniwa:
> Do not dispatch an event when retargeted-target and
retargeted-reletedTarget are identical, but target and relatedTarget are
not identical.

This has the same issue as the one I pointed out in https://github.com/w3c/webcomponents/issues/577.  In general, you can't just compare *target* and *related target* or their retargeted counterparts.  You need to look at whether we're crossing the shadow tree, and if both *target* and *related target* are in shadow tree from which we're exiting.

For example, it's perfectly fine to dispatch an event with target & related target to be identical, and the event MUST be dispatched.  This is the existing behavior of event dispatching and we can't change that for backwards compatibility reasons.

hayatoito:
Yeah, I am fully aware of that. The point is *UA events*.  There is a case where we should not dispatch a UA event. Could you have a chance to see the example in https://github.com/whatwg/dom/pull/336#issue-180583922?

Maybe https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/dom/shadow/shadow-boundary-events.html?l=150 might be helpful to understand the case. In this case, we should not dispatch an mouseover event on the shadow host (shadowF).

That is the motivation of this change. If we are to dispatch a *synthetic* event which has the same condition (the same target (shadowF) and the same relatedTarget (shadowI)), and want to make sure that event is dispatched,  we have to check whether an event is a synthetic event or not. I do not want to introduce the divergence between synthetic events and UA events in order to address this very minor case.

rniwa:
This is not a thing the generic event dispatch algorithm should be dealing with, the same way `composed` flag shouldn't be inferred by the type of the event we're firing.  Instead, we need to update UI events spec so that we don't fire events when `relatedTarget` is inside a shadow tree of `target`.

hayatoito:
Hmm. Maybe that would be possible. However, is it really worth doing?
Users can fire an synthetic UI events. Thus, that would not solve the real issue here.

rniwa:
I don't understand. You said earlier that the only case you care about with regards to the related target being inside the shadow tree of the target is when UA fires a trusted event.  Specifying this in the UI events spec would specifically solves this issue. Note that UI events spec specifies when and how events gets fired by UA. e.g. mouseout wouldn't get fired constantly whenever mouseover happens.

hayatoito:
I do not think that's a good idea.
Rather, I would do adding one "if .. else .." statement in DOM Standard, instead of adding a lot of information to UI Events spec. That would make our lives much simpler, I think.

BTW, do you prefer 1? Is my understanding correct?

> 1. Introduce the divergence (between synthetic events and UA events) so that synthetic events are always dispatched in the first tree.
> 2. We do not want to introduce the divergence between synthetic events and UA events in any case!

rniwa:
> I do not think that's a good idea.

The last time this came up for `composed`, people have argued that we should specify this at where events are dispatched instead of treating all trusted events the same so we need to check with @domenic and @annevk what they think about this case.

hayatoito:
We do not have to do the same way as we do for `composed` this time because we can specify the condition easily in general event dispatch section in DOM Standard. For example, "a trusted event which has non-null a related target property" might be good enough and robust.

hayatoito:
Anyway, let me update the PR so that a synthetic event is always dispatched in the *first* tree, at least until the root of the tree. I am fine with this change.

hayatoito:
I have updated the PR, re-using this thread. Please ignore the first comment in this discussion. The PR would become much simpler.

dom.bs
```diff
@@ -1201,6 +1201,9 @@ for discussion).
  <a>relatedTarget</a> against <var>target</var> if <var>event</var>'s <a>relatedTarget</a> is
  non-null, and null otherwise.
 
+ <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true and <var>target</var> and
+ <var>relatedTarget</var> are identical, terminate these steps.
```
hayatoito: Is that okay to terminate these steps without a return value?
That would be no difference in practice, but I am not sure what is the best.

annevk: Is this actually how we would implement this in practice? It seems somewhat weird to use the `isTrusted` attribute. Also in part because I still don't quite know when it's set and when it's not. Does `element.click()` set it?

smaug----:
Could we please not use PRs for discussing spec issues.

smaug----:
FWIW, after reading comments here, I was about to propose the same stuff what hayato then proposes in https://github.com/whatwg/dom/pull/336#pullrequestreview-2648799 , I think.
My current thinking is that UIEvents should tell what kind of events should be fired, and it plays in a layer above actual event dispatch. Then if shadow DOM affects to event dispatch, that is something to deal with in DOM spec.
(in a way even .composed could be thought to be part of dispatch, not really about the caller of dispatch method, even though, currently .composed is in UI spec events)

Also, I would imagine https://github.com/whatwg/dom/pull/336#pullrequestreview-2648799 is what would be implemented, not some per-event-firing-location relatedTarget check.

dom.bs
hayatoito: DOM Standard defines isTrusted as follows:

> event . isTrusted
> Returns true if event was dispatched by the user agent, and false otherwise.

I am not sure how we could interpret "event was dispatched by the user agent" strictly.
I tend to agree @smaug.  A click event triggered by `element.click()` is dispatched by users, instead of UA.

smaug----: postMessage() dispatching untrusted events? That is a bug.
click is the special case similarly to dispatching MouseEvent click explicitly needing to trigger default handling.

hayatoito: Thanks. Instead of using `isTrusted` here, just saying "If event is dispatched by the user agent" would be better?

annevk:
@smaug---- I don't really understand what you're proposing. I'd think we want to make the specification match how this is actually implemented in practice. When implementation and specification diverge, there's often room for bugs.

@hayatoito we can't say "dispatched by the user agent" since that is too vague. It sounds like what @smaug---- is saying there is that `click()` is no different from dispatching a synthetic `MouseEvent` object whose name is `click`. If that is the case making sure that is defined correctly depends on #325 I think.

smaug----:
> @smaug---- I don't really understand what you're proposing.

The same thing what @hayatoito proposed

hayatoito:
Can we merge this PR before #325?
Since this PR only affects an event which has relatedTarget, the result of #325, `click()`, will not affect the behavior, I think.

annevk:
Apologies, I got confused and you are correct. I rebased the PR and applied some minor nit fixes. I have two questions:

1. Is @rniwa now okay with this approach?
2. We're sure no specification relies on the return value of dispatch (when relatedTarget is set somehow) and would get tripped up by it not returning anything here?

hayatoito:
Thank you!

> We're sure no specification relies on the return value of dispatch (when relatedTarget is set somehow) and would get tripped up by it not returning anything here?

I am no sure. Maybe we can return `true` here, instead of not returning anything. That might be better.

As far as I confirmed in Blink, Blink returns `DispatchEventResult::NotCanceled` in event dispatch if the event path is empty, and I have not seen any problem for that so far.

annevk:
Yeah, returning true is probably better, especially if that matches implementations. I can make that change before landing.

@rniwa could you please review one more time? If I don't hear anything by tomorrow I'm just going to assume you're okay with this. We can always revisit post-commit.

dom.bs
```diff
@@ -1201,6 +1201,9 @@ for discussion).
  <a>relatedTarget</a> against <var>target</var> if <var>event</var>'s <a>relatedTarget</a> is
  non-null, and null otherwise.
 
+ <li><p>If <var>event</var>'s {{Event/isTrusted}} attribute is true and <var>target</var> is
+ <var>relatedTarget</var>, then terminate these steps.
```
rniwa: No, I don't think we should be specifying this in the DOM specification. Also, this would affect every UI event dispatch beyond just ones inside shadow trees, and I'm not okay with that.

Please don't merge this.

hayatoito: > Also, this would affect every UI event dispatch beyond just ones inside shadow trees

UA's UI Event which happens in a document tree never meets this condition.
Could you have an example of an UA's event which happens in a document tree which meets this condition?

annevk:
@rniwa what do you suggest we do instead? And how would/do you implement this in WebKit?

rniwa:
> @rniwa what do you suggest we do instead? And how would/do you implement this in WebKit?

I don't think this is a thing that belongs to the DOM specification. It should be specified in https://w3c.github.io/uievents/ which already specifies other conditions under which an event should be dispatched or not dispatched.

Adding this text to the DOM means that any event regardless of whether they're UI event or not would not be dispatched by UA if it happens to have a related target and it's equal to the target.

annevk:
How would you implement this though? Also, the whole reason DOM has relatedTarget logic is because there is no clean layering.

rniwa:
> How would you implement this though? Also, the whole reason DOM has relatedTarget logic is because there is no clean layering.

I don't know how relevant how each implementation does this check since that'll be an implementation detail but [WebKit checks this condition](https://trac.webkit.org/browser/trunk/Source/WebCore/page/EventHandler.cpp#L2448) when figuring out when & how to fire mouse events.

hayatoito:
> Adding this text to the DOM means that any event regardless of whether they're UI event or not would not be dispatched by UA if it happens to have a related target and it's equal to the target.

The PR does not have an intention to distinguish UI events or not. That is not what we agreed at all.
This PR is just a bug fix which I introduced in upstreaming Shadow DOM into DOM Standard in the past, where whether having relatedTarget or not matters.

Could you file another bug if you have another proposal? We should not discuss this behavior in this PR because it is the new proposal which we have not heard.

rniwa:
> The PR does not have an intention to distinguish UI events or not. That is not what we agreed at all.

That's precisely to what I'm objecting.  As much as you can claim that you never agreed to such a change, I've never agreed to making this change either so there is no consensus here either way.

hayatoito:
Could you file another bug with a concrete proposal? We can discuss that later after merging this PR.
Without a concrete proposal, it is difficult to discuss in a productive way.

This PR is basically a part of upstreaming a behavior which has been defined in Shadow DOM specification.
If we want to introduce a new behavior which is not defined anywhere, we should file a bug for that to discuss.

rniwa:
> Could you file another bug with a concrete proposal? We can discuss that later after merging this PR. Without a concrete proposal, it is difficult to discuss in a productive way.

The problem is that I'm specifically opposing to make this merge so I don't think filing a new issue is the right approach.  I've stated my proposal multiple times already.  We should make this change in UI events instead.

Since you've already acknowledged my approach in the past (e.g. https://github.com/whatwg/dom/pull/336#issuecomment-251279251), I don't understand why you're now insisting that I don't have a concrete proposal.

> If we want to introduce a new behavior which is not defined anywhere, we should file a bug for that to discuss.

There is no user-perceived behavior change I'm proposing.  I'm simply opposing the way to spec the behavior on the grounds that the generic event dispatch algorithm shouldn't be the one being modified to account for some mouse-event related quirks that UA needs to avoid.  Namely, not merging this particular spec change, and spec'ing the behavior this merge would have made, are not mutually exclusive.

hayatoito:
Okay. I am glad to hear you are not proposing a user-perceived behavior change. :) (I thought that you are proposing a new behavior).

Thus, what we have not agreed is how we can update the spec.

I am +1 for the current PR's approach because it is good enough and simple, as of now,
at least, until we have a good idea how to update the UI Events in a clear layering.

How about "Ship and iterate"? (Merge this PR, and think how to update UI Events in a clear layering later because it might be difficult and time-consuming). Is that reasonable?

rniwa:
Here's an alternative statement I'll be fine with:

> If `event`'s `isTrusted` attribute is true and `target` is `relatedTarget` but `target` is not `event`'s `relatedTarget`, then terminate these steps.

So that the change specifically targets the case where the `relatedTarget` becomes identical to `target` after retargeting.

The part I'm opposing specifically is the part where we're making all events with `relatedTarget` not fire when target and related targets are the same because that's really an orthogonal change to the generic event dispatching algorithm regardless of whether it happens in the practice or not.

domenic:
IMO if there is no normative behavior change being discussed then the exact spec mechanisms being used should be up to the discretions of the editors.

rniwa:
> IMO if there is no normative behavior change being discussed then the exact spec mechanisms being used should be up to the discretions of the editors.

There is a behavior implication here that all events with related targets identical to target won't be fired. Given the UI events spec doesn't precisely specify whether related target is different from target(e.g. [mouseout](https://w3c.github.io/uievents/#mouseout)), it's permissible to have UA that fires `mouseout` with related target and target being identical today although it doesn't seem to happen in practice in common cases in major browsers.

hayatoito:
Okay. Let's change the sentence as follows, and land this PR.

>  If event's isTrusted attribute is true and target is relatedTarget but target is not event's relatedTarget, then return true and terminate these steps.

@annevk I appreciate if could you merge this PR with this change.

annevk:
@rniwa note that relatedTarget in DOM is not generic per se. Only if an event class decides to make use of that internal slot will they participate in the choices DOM makes for it. If you want to call something relatedTarget but have different semantics, you could in theory use your own internal slot for it. (This is all a little hairy since we haven't formalized internal slots yet and UI events is a hopelessly vague specification, but, it should all work...)

Anyway, I'll update the PR per the discussion and land it tonight / tomorrow.

annevk:
PR updated.

rniwa:
On my second thought, don't we want this behavior to non-trusted events as well?  When someone manually fire a `mouseout` event, don't we want that event to stop when it reaches a point where retargeted target and retargeted relatedTarget are identical?  (We can discuss this in a separate issue since this PR discussion is getting really long).

annevk:
The beginning of this PR thread does touch upon the synthetic event issue. And in https://github.com/whatwg/dom/pull/336#issuecomment-251063191 @hayatoito argued for not making them diverge, but then the thread went into a different direction.

I also realize I forgot to update the text to make it return true as mentioned in https://github.com/whatwg/dom/pull/336#issuecomment-253481618. Will do that now.

annevk:
I'm going to land this with the `isTrusted` guard removed since it seems both @hayatoito and @rniwa are okay with that (I checked with @rniwa on IRC).

annevk:
@hayatoito are you planning on writing tests for this?

hayatoito:
Ah, I see. Okay to remove `isTrusted` guard since we have "target is not event's relatedTarget" clause. That would make sense.
I can add tests for WTP. Thank you!

dSalieri:
> Ah, I see. Okay to remove `isTrusted` guard since we added "target is not event's relatedTarget" guard. That would change the behavior slightly, but at least it would not bring an incompatible change unless Shadow DOM is used.
> 
> I can add tests for WPT. Thank you!

@hayatoito Well, I've read your issue, I see you avoided in spec using `isTrusted` property, instead this you added "target is not event's relatedTarget". 

```JavaScript
/// Let "target" be the host of a shadow root in an another shadow root
const ev = new Event("click");
ev.relatedTarget = target;
target.addEventListener("click", (e) => console.log(e));
target.dispatchEvent(ev)
```
This example will invoke callback but it shouldn't because condition returns `true`.

I see that browsers don't call callbacks if it conform that condition if it is not synthetic event and call if it is synthetic event.
> If target is relatedTarget and target is not event’s [relatedTarget](https://web.archive.org/web/20161021131512/https://dom.spec.whatwg.org/#event-relatedtarget), then return true.

Could you give some comment about that?

bathos:
> In theory, it doesn't matter how I set relatedTarget, the condition shouldn't allow callback functions to be called. But for some reason it doesn't match.

The internally associated “relatedTarget” (not sure why it’s camel cased in the spec, I thought the norm for this was using regular spaces) isn’t the `relatedTarget` attribute of `UIEvent`. The UIEvent attribute (→ ES accessor property) exposes it for those UIEvents specifically, but it exists for all events as private state. Setting a data property with the same name doesn’t do anything.

dSalieri:
@bathos yeah I observed that `relatedTarget` is accessor property. But this UIEvent attribute exists for concrete `UIEvents` like `MouseEvent` or `FocusEvent`. Hence for them I can't do ordinary assignment because they are getters, I talk about `relatedTarget` properties. I only do initializing as I specify above. But what about `CustomEvent`? What if I want to emulate behaviour like a `MouseEvent` when my cursor point onto element of shadow root and then I do "move" to his host element that also placing in shadow root, I expect to see no one callback (callback won't be called).

Summarize: Inspite of that spec explicit says about `relatedTarget` as a property of event, there are differences. Your own created properties don't count in whatwg algos. Right?

bathos:
> Summarize: Inspite of that spec explicit says about relatedTarget as a property of event, there are differences. Your own created properties don't count in whatwg algos. Right?

Right. Despite sharing the same name as an attribute of UIEvent and a dictionary member that UIEvent constructors accept, when the spec says “event’s relatedTarget”, it isn’t referring to an ES property. It’s an abstract internal association, not something specific to the ECMAScript binding.

PR #341, opened 2016-10-10
Change createElement to report exceptions instead of rethrowing

domenic:
Such exceptions can occur due to custom element constructors or invariant checks. This closes https://github.com/w3c/webcomponents/issues/569 by moving the catch-and-return-HTMLUnknownElement from HTML to DOM; HTML will subsequently be updated to remove that logic.

/cc @rniwa @dominiccooney. HTML PR incoming; I am hoping @rniwa can upstream his test changes from https://bugs.webkit.org/attachment.cgi?id=290872&action=prettypatch

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5668,7 +5668,8 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 
   <ol>
    <li>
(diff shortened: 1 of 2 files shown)


rniwa:
r=me

PR #342, opened 2016-10-11
Define special dispatch behavior for click events

### LINKED ISSUE #325: Define how script-created click events are special during dispatch

annevk:
User agents have always had special behavior for synthetic click
events. This is discussed to great extent in
https://www.w3.org/Bugs/Public/show_bug.cgi?id=12230.

This change introduces the hooks to no longer make that magical, but an
understood and integral part of the processing model, ugly as it is.

This changes requires corresponding changes and cleanup in HTML.

Fixes #325.


annevk:
I drastically changed this commit to account for new findings discussed in the issue.

annevk:
The main outstanding thing here is passing the event to "activation behavior".

annevk:
Done. At this point I'd really appreciate review from @smaug----, @dtapuska, @foolip, and @domenic. Thanks!

annevk:
The two things to consider when reviewing:

* Should we start accounting now for the possibility that a shadow host has activation behavior? (Then we need to traverse the parent chain even for non-bubbling events, but only consider actual targets (there can be multiple with shadow trees).)
* Are the names okay for use in HTML?

annevk:
I created tests for this change too, Firefox fails a bit more, as discussed in the issue thread for this PR: https://github.com/w3c/web-platform-tests/pull/3974.

dom.bs
```diff
@@ -1028,6 +1028,18 @@ specified otherwise it returns null.
 <p class="note no-backref"><a>Nodes</a>, <a for=/>shadow roots</a>, and <a>documents</a> override
 the <a>get the parent</a> algorithm.
 
+<p>Each {{EventTarget}} object can have an associated
+<dfn export for=EventTarget>legacy pre-click-dispatch steps</dfn> algorithm,
+<dfn export for=EventTarget>legacy post-canceled-click-dispatch steps</dfn> algorithm, and
+<dfn export for=EventTarget>activation behavior</dfn> algorithm. The <a>activation behavior</a>
+algorithm is passed an <a>event</a>, as indicated in the <a>dispatch</a> algorithm.</p>
+
+<p class="note no-backref">An {{EventTarget}} object that has <a>legacy pre-click-dispatch steps</a>
```
domenic: This phrasing in the non-normative note seems a little strange. I think it's stating an observation about the world, but maybe it's meant to be more of a requirement?

Perhaps a better conceptual framework would be to define two special types of EventTargets:

- clickable event targets, which have activation behavior
- clickable event targets with interception, which additionally have legacy pre-click dispatch steps and legacy post-canceled-click-dispatch steps

(better names welcome). Then you say e.g. "every HTMLXElement is a clickable event target with the following activation behavior: ..."

dom.bs
```diff
@@ -1225,12 +1240,28 @@ for discussion).
...
-   <var>parent</var>'s <a>get the parent</a> with <var>event</var>.
+   <li>
+    <p>If <var>parent</var> is non-null, then:
+
+    <ol>
+     <li><p>If <var>event</var>'s {{Event/bubbles}} attribute is true, <var>activationTarget</var>
+     is null, and <var>parent</var> has <a>activation behavior</a>, then set
+     <var>activationTarget</var> to <var>parent</var>.
+     <!-- If an EventTarget object with activation behavior can ever have an associated shadow root
+          we need to complicate this somewhat. -->
```
domenic: So the only reason this isn't complicated yet is that we disallow attachShadow on elements with interesting activation behavior? That is kind of fragile, if so... hmm.

dom.bs
annevk: Yeah, I should rephrase it as saying that you can have activation behavior and that if you have that, you can additionally have both pre-click and post-canceled-click. That would be a lot more clear and not warrant a note to state the obvious.

dom.bs
annevk: The alternative is changing both substeps that append to event's path to have this kind of check. But if we never add activation behavior to shadow hosts, that would be needlessly complicated. But maybe I should do it anyway just to be forward compatible.

annevk:
1. I kept "activation behavior" as name because HTML already uses it. If someone else prefers "post-click-dispatch steps" though we'll go with that. It doesn't matter much until we decide to expose an API for it.
2. The "future-proof alternative in case shadow hosts ever get activation behavior" commit allows for seeing the difference between the future proof alternative that we need if shadow hosts ever become suitable activation targets.

dom.bs
```diff
@@ -1028,6 +1028,22 @@ specified otherwise it returns null.
 <p class="note no-backref"><a>Nodes</a>, <a for=/>shadow roots</a>, and <a>documents</a> override
 the <a>get the parent</a> algorithm.
 
+<p>Each {{EventTarget}} object can have an associated
+<dfn export for=EventTarget>activation behavior</dfn> algorithm. The <a>activation behavior</a>
```
foolip: Should the concept be dropped from HTML in favor of this? It becomes a bit weird to have activation behavior for both elements and event targets.

dom.bs
annevk: We could do that too. I wonder if impl might already use shadow roots here, eg with input, so maybe not accounting for it won't work anyway.

dom.bs
```diff
@@ -1204,6 +1220,9 @@ for discussion).
  <li><p>Append (<var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>) to <var>event</var>'s
  <a for=Event>path</a>.
 
+ <li><p>Let <var>activationTarget</var> be <var>target</var>, if <var>target</var> has
```
foolip: No comma before if. ("Let relatedTarget be the result of retargeting event’s relatedTarget against target if event’s relatedTarget is non-null, and null otherwise.")

dom.bs
```diff
@@ -1280,6 +1321,20 @@ for discussion).
 
  <li><p>Set <var>event</var>'s <a for=Event>path</a> to the empty list.
 
+ <li>
+  <p>If <var>event</var> is a {{MouseEvent}} object, <var>event</var>'s {{Event/type}} attribute is
```
foolip: Maybe factor out the condition into a variable? I'm pretty sure an implementation would, and it makes it very clear that there's no case where *activationTarget* has changed in the interim to consider.

dom.bs
```diff
@@ -1280,6 +1321,20 @@ for discussion).
...
+  "<code>click</code>", and <var>activationTarget</var> is non-null, then:
+
+  <ol>
+   <li><p>If <var>event</var>'s <a>canceled flag</a> is unset and <var>activationTarget</var> has
+   <a>activation behavior</a>, then run <var>activationTarget</var>'s <a>activation behavior</a>
```
foolip: Doesn't *activationTarget* always have activation behavior, given the conditions where it's assigned?

foolip:
@dtapuska, could you, or someone on your team, mind reviewing this in detail and comparing with how you hope Blink's implementation will end up. If you don't think it's possible or worthwhile to match this line-for-line, are there interesting things to test where the two models don't quite align?

dom.bs
annevk: Hmm, this is a pretty common pattern. I guess I can start removing commas whenever I see them, but it does make it a bit more readable I think.

dom.bs
annevk: It also ended up adding conditionals (doh). Will let you to be the judge.

dom.bs
annevk: Done per below.

dom.bs
foolip: "only set activationTarget when isActivationEvent is true" changes LGTM

dom.bs
foolip: Ah, thanks.

annevk:
@smaug---- approved this PR on IRC. @dtapuska could you give an ETA for a review?

dtapuska:
lgtm

PR #346, opened 2016-10-20
Editorial: rename activation-behavior-related algorithms

annevk:
Per discussion in https://github.com/whatwg/html/pull/1925.


annevk:
The only thing I'm not sure about are the hyphens. @sideshowbarker?

sideshowbarker:
> The only thing I'm not sure about are the hyphens.

The hyphens look right to me.

PR #347, opened 2016-10-20
Editorial: provide a shorthand for other specifications to create events

annevk:
This can be useful when you can’t use fire, but need to initialize
event attributes in a particular manner or keep the event around for
some reason.

Note that creating events without a constructor doesn’t work, as you
would not get the dictionary to usefully default attribute values.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -942,24 +942,52 @@ method must, when invoked, run these steps:
 
 <h3 id=constructing-events>Constructing events</h3>
 
-When a <dfn export for=Event id=concept-event-constructor>constructor</dfn> of the
-{{Event}} interface, or of an interface that inherits from the {{Event}} interface, is
-invoked, these steps must be run:
+<p>When a <dfn export for=Event id=concept-event-constructor>constructor</dfn> of the {{Event}}
+interface, or of an interface that inherits from the {{Event}} interface, is invoked, these steps
+must be run, given the arguments <var>type</var> and <var>eventInitDict</var>:
 
 <ol>
- <li><p>Create an <a>event</a> that uses the interface the constructor was invoked upon.
+ <li><p>Create a new object <var>event</var> using this interface.
 
- <li><p>Set its <a>initialized flag</a>.
+ <li><p>Set <var>event</var>'s <a>initialized flag</a>.
 
- <li><p>Initialize the {{Event/type}} attribute to the <var>type</var> argument.
+ <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>type</var>.
 
- <li><p>If there is an <var>eventInitDict</var> argument, then for each <a>dictionary member</a>
- present, find the attribute on <a>event</a> whose <a spec=webidl>identifier</a> matches the key of
- the <a>dictionary member</a> and then set the attribute to the value of that
+ <li><p>For each <a>dictionary member</a> present in <var>eventInitDict</var>, find the attribute on
+ <var>event</var> whose <a spec=webidl>identifier</a> matches the key of the <a>dictionary
+ member</a> and then set the attribute to the value of that <a>dictionary member</a>.
+
+ <li><p>Return <var>event</var>.
+</ol>
+
+<p>To <dfn export id=concept-event-create lt="creating an event|create an event">create an
+event</dfn> using <var>eventInterface</var>, which must be either {{Event}} or an interface that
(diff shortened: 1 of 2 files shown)


domenic:
Hmm this just kind of pushes around my uncomfortableness with "initial value" into DOM. I was hoping that you would factor out the initialize-using-the-constructor's-dictionary algorithm in https://dom.spec.whatwg.org/#constructing-events into something that is called from this new hook.

annevk:
I see. DOM already uses "initial value" for this. I think at some point for `document.createEvent()` we looked into doing it differently, but this seemed like the best option.

domenic:
Well, notably it only uses initial value for events :). So a single factoring job would fix the existing problem and introduce a nice new hook.

annevk:
I guess what I'm saying is that I don't know how and the last time I tried to figure it out we ended up with this.

domenic:
Ah OK. I can try to create a PR then.

annevk:
That sounds great. Feel free to add a commit to this branch if the general shape of this new entry-point looks fine.

domenic:
@annevk pushed a commit; let me know what you think.

dom.bs
```diff
@@ -1447,10 +1471,8 @@ initialized, and a <var>legacy target override flag</var>, run these steps:
  <li><p>If <var>eventConstructor</var> is not given, then let <var>eventConstructor</var> be
  {{Event}}.
 
- <li><p>Let <var>event</var> be the result of <a for=Event lt=constructor>invoking</a> the initial
- value of <var>eventConstructor</var> with the argument <var>e</var>.
-
- <li><p>Initialize <var>event</var>'s {{Event/isTrusted}} attribute to true.
+ <li><p>Let <var>event</var> be the result of <a>creating</a> <var>eventConstructor</var> with
+ <var>e</var>.
```
annevk: You removed the name argument so this no longer works.

dom.bs
```diff
@@ -942,25 +942,49 @@ method must, when invoked, run these steps:
...
+<ol>
+ <li><p>Create a new object <var>event</var> using the <var>eventInterface</var> interface.
+
+ <li><p>Set <var>event</var>'s <a>initialized flag</a>.
+
+ <li><p>For each <a>dictionary member</a> of the dictionary accepted by <var>eventInterface</var>'s
+ constructor (which will either be {{EventInit}} or a dictionary that inherits from it), find the
+ attribute on <var>event</var> whose <a spec=webidl>identifier</a> matches the key of the
+ <a>dictionary member</a> and then set the attribute to the default value of that <a>dictionary
+ member</a>.
```
annevk: @heycam we can set type separately quite easily (that's a constant for all events). Though if you have other ideas that's fine too. I'll let @domenic comment on your idea. I think I'm happy as long as you are all happy.

bzbarsky: I think heycam's idea is a lot clearer and less magical.  Explicitly instantiate the dictionary type involved (e.g. by doing https://heycam.github.io/webidl/#es-to-dictionary with `V` set to `null` and the relevant dictionary type being used.  Then explicitly call the guts of the relevant interface's constructor.

domenic:
Gave this another pass using the idea of converting undefined to the dictionary type and then using that. Let me know what you think.

dom.bs
```diff
@@ -942,25 +942,54 @@ method must, when invoked, run these steps:
...
- <li><p>Set its <a>initialized flag</a>.
+ <li><p>Set <var>event</var>'s <a>initialized flag</a>.
 
- <li><p>Initialize the {{Event/type}} attribute to the <var>type</var> argument.
+ <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>type</var>.
 
- <li><p>If there is an <var>eventInitDict</var> argument, then for each <a>dictionary member</a>
- present, find the attribute on <a>event</a> whose <a spec=webidl>identifier</a> matches the key of
- the <a>dictionary member</a> and then set the attribute to the value of that
+ <li><p>If <var>eventInitDict</var> is given, then for each <a>dictionary member</a> present in
```
domenic: I'm not actually sure myself... mind guiding me through the spec in that regard? In particular I think optional arguments are not always given.

annevk:
@heycam does this look okay now?

annevk:
(Looks okay to me.)

heycam:
Looks good to me!

PR #349, opened 2016-10-24
Add a realm argument to "create an event"

domenic:
Needed for https://github.com/whatwg/html/issues/1959

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -75,6 +75,7 @@ urlPrefix: https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#
         text: has ever been evaluated flag; for: service worker
 urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
     text: Construct; url: sec-construct; type: abstract-op
+    text: Realm; url: realm; type: dfn
 </pre>
 
 <pre class='link-defaults'>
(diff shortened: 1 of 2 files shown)


domenic:
I wonder if "fire an event" should pass along _target_'s relevant Realm. That would need some testing though (in theory testing every use of "fire an event").

annevk:
IDL should really have some low-level operation to create classes that everything else is layered upon. This is just a mess.

I think _target_'s relevant Realm would always be correct. Would be annoying to require a _Realm_ argument everywhere.

dom.bs
```diff
@@ -963,10 +964,15 @@ must be run, given the arguments <var>type</var> and <var>eventInitDict</var>:
...
-it, run these steps:</p>
+it, and optionally given a <a>Realm</a> <var>realm</var>, run these steps:</p>
 
 <ol>
- <li><p>Create a new object <var>event</var> using <var>eventInterface</var>.
+ <li>
+  <p>Create a new object <var>event</var> using <var>eventInterface</var>. If <var>realm</var>
+  is given, use that Realm; otherwise, use the default behavior defined in Web IDL.
+
+  <p class="note">As of the time of this writing Web IDL does not yet define any default behavior;
```
annevk: Should we make this class="XXX"?

domenic:
> I think target's relevant Realm would always be correct. Would be annoying to require a Realm argument everywhere. 

I ran some tests and this is not interoperable -_-.

- Chrome uses target's relevant Realm
- Gecko might use target's relevant Realm, but since it does the currently-nonstandard-but-oft-debated thing where adopting an element changes its relevant Realm, it's not interoperable with Chrome
- Edge appears to not use the target's relevant Realm, but instead maybe the realm of the document in which target finds itself or something similar
- Haven't tested WebKit yet.

Let's leave that for a separate issue.

PR #350, opened 2016-10-24
Set up the architecture for commit/branch snapshots

domenic:
@annevk I think the rest of the steps are up to you. You need to:
- Ensure the DEPLOY_USER username is correct in deploy.sh. If you'd rather keep the actual username secret we can talk about how to encrypt that.
- Generate a SSH key on the dom.spec.whatwg.org server and ensure it's trusted by the server (i.e. can be used to log in, by putting the public key in the appropriate place)
- Encrypt the resulting private key, and check the resulting file into this branch as `deploy_key.enc`. [These instructions](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd#get-encrypted-credentials), minus the part about "add that deploy key to your repository", should do the trick.
- Enable Travis for whatwg/dom when everything is ready. We can do this together to make sure things are working.
- Disable the existing deploy mechanisms.

It'd be a good idea to backup the current server copy of DOM so we can easily swap it back if something goes wrong while we're trying to land this. (That is, it'd be sad if DOM was down for even 10 minutes tomorrow morning.)

This setup assumes that there is a folder dom.spec.whatwg.org inside your home directory on the server, which is served to the web. If not we can tweak the script a bit.

We can also do somewhat of a dry-run by changing some variables, e.g. if you change WEB_ROOT to "dom.spec.whatwg.org/test/" then in theory it should spit out index.html and friends inside the /test subdirectory.


annevk:
Thank you! I'm going to wait with this until I figure out if DreamHost can move the domain back to the main WHATWG account. If not, I'll create a dedicated user within my account for this domain.

annevk:
DEPLOY_USER is correct.

I created a key and protected it by passphrase (it's on the server in the usual location). However, I don't have the travis command line utility and it's not immediately clear to me where I can find it.

dontcallmedom:
`gem install travis`?
I think you can also get the equivalent feature with `npm install travis-encrypt`

domenic:
I think a passphrase will prevent automated deployment. That would mean you'd need a human to type the passphrase every time you use that key, right?

annevk:
@dontcallmedom that seems to require sudo, and then when I use sudo I get `net-http-persistent requires Ruby version ~> 2.1.`. I finally got something with `npm install -g travis-encrypt`. Now onto the next step.

annevk:
(I used sudo for that too by the way, instead of fixing permissions as npm instructs you to do. Oops. I'm so bad at this.)

annevk:
Sigh, `travis-encrypt` is bogus as it doesn't work with keys.

annevk:
I tried following the instructions at https://github.com/travis-ci/travis.rb with the `gem install travis -v 1.8.2 --no-rdoc --no-ri` command but that yields exactly the same results as without the additional arguments. I'm hereby giving up.

domenic:
Since you moved servers, I can probably do it. You could also try upgrading your ruby somehow.

annevk:
This now depends on #361.

annevk:
Probably need to introduce a Makefile again because bikeshed now requires this invocation

```bikeshed spec dom.bs dom.html --md-Text-Macro="SNAPSHOT-LINK <local watch copy>"```

and over HTTP

```-F md-Text-Macro="SNAPSHOT-LINK <local watch copy>"```

needs to be added.

Also need to update the README for that.

annevk:
I think this is now ready to go. The various scripts appear to be working. I might have time later today @domenic when you wake up. Ping me when you have 30-60min available (I suspect it'll take less) and we'll see if we can make it happen.

PR #353, opened 2016-10-27
Stop re-signaling a slotchange event

hayatoito:
Fixes https://github.com/w3c/webcomponents/issues/571.


annevk:
This looks correct to me. Would be great if @rniwa could confirm.

rniwa:
r=me too.

PR #354, opened 2016-10-27
Use the relevant realm of the event target when firing an event

### LINKED ISSUE #352: What realm should be used for creating events from "fire an event"
Continuation of a discussion that started in #349. /cc @bzbarsky @Ms2ger.

A test at http://w3c-test.org/submissions/4088/html/semantics/forms/resetting-a-form/reset-form-event-realm.html reveals the following, assuming that reset()'s "fire an event" is representative in general:

- Chrome and Safari use the target's relevant Realm
- Firefox may use the target's relevant Realm; however, since it does the currently-nonstandard-but-oft-debated thing where adopting an element changes its relevant Realm, it's not interoperable with Chrome/Safari
- Edge coincidentally matches Firefox, but not for the same reasons. That is, it uses frames[1]'s realm in that test, but it does not have Gecko's adoption semantics, at least as far as I can tell by testing the prototypes (they were not swizzled). . Maybe it is going by the realm of the target's document?

I think maybe we should standardize on using target's relevant Realm, and then the fact that Gecko is not interoperable just ends up being part of the larger issue where Gecko has different adoption semantics. Thoughts welcome.

domenic:
Closes #352.

Some tests at https://github.com/w3c/web-platform-tests/pull/4088, but they are not exhaustive. I don't feel up to the task of writing a test for every instance of "fire" in the spec.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1478,7 +1478,7 @@ initialized, and a <var>legacy target override flag</var>, run these steps:
  {{Event}}.
 
  <li><p>Let <var>event</var> be the result of <a>creating an event</a> given
- <var>eventConstructor</var>.
+ <var>eventConstructor</var>, in the <a>relevant Realm</a> of <var>target</var>.
 
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>e</var>.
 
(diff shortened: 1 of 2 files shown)


annevk:
One interesting aspect there is that it sounds like Gecko does this for "dispatch", not "fire" necessarily. This means we'll have to be careful to do the correct thing for "dispatch" whenever that is used. I guess that's fine?

annevk:
Also, this and your test expose https://www.w3.org/Bugs/Public/show_bug.cgi?id=20567. Would it be better to use the node document's relevant Realm if the target is a node?

annevk:
I see, sounds reasonable then. @bzbarsky, can you confirm to be sure?

annevk:
Okay, though note that not everything goes through "fire an event" so more complicated invocations that go through "create an event" and then "dispatch" will still have to account for the Realm somehow.

I'll land this.

bzbarsky:
I think the right way to handle that is to require that a target is passed to "create an event" and use its Realm.

annevk:
I guess we could open an issue for that, but that would mean "dispatch" would also need to change to stop taking a target as argument?

bzbarsky:
Why?

annevk:
@bzbarsky I was thinking as a way to ensure that the targets actually match, but dispatch needs to keep that argument for other reasons of course.

PR #359, opened 2016-10-29
Reinstate active flag for iterators

ayg:
This reverts ccf51e85, and fixes
https://www.w3.org/Bugs/Public/show_bug.cgi?id=25412.  It brings the
spec back in line with Firefox instead of other browsers.  Boris argues
convincingly that the behavior without the active flag is much harder to
spec and test properly, and nobody has a use-case for allowing recursion
here, so it's easier all around to converge on Firefox behavior and just
ban recursion.  He doesn't want Firefox to allow recursion unless
someone writes good tests for it, which nobody wants to.  Other UAs were
asked for comment on the bug and didn't support or oppose.

@bzbarsky @travisleithead @domenic @cdumez

***
[Preview](https://s3.amazonaws.com/pr-preview/ayg/dom/active-flag.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/61f40b0...ayg:2c0e1fe.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8625,6 +8625,10 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
 to filter and traverse <a>node</a>
 <a>trees</a>.
 
+Each {{NodeIterator}} and {{TreeWalker}} object has an associated <dfn export
+id=concept-traversal-active for=traversal>active flag</dfn> to avoid
+recursive invocations.  It is initially unset.
+
 Each {{NodeIterator}} and {{TreeWalker}} object also
 has an associated <dfn export id=concept-traversal-root for=traversal>root</dfn>
 <a>node</a>,
@@ -8635,6 +8639,7 @@ To <dfn export id=concept-node-filter for=Node>filter</dfn> <var>node</var> run
 these steps:
 
 <ol>
+ <li>If the <a for=traversal>active flag</a> is set, throw an {{InvalidStateError}}.
  <li>Let <var>n</var> be <var>node</var>'s
  {{Node/nodeType}} attribute value minus 1.
  <li>If the <var>n</var><sup>th</sup> bit (where 0 is the least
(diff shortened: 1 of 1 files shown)


annevk:
@dominiccooney @foolip what do you think?

dom.bs
```diff
@@ -8785,6 +8785,10 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
 to filter and traverse <a>node</a>
 <a>trees</a>.
 
+Each {{NodeIterator}} and {{TreeWalker}} object has an associated <dfn export
+title=concept-traversal-active for=traversal>active flag</dfn> to avoid
```
annevk: s/title/id/ to preserve the old value.

foolip:
I guess there already exists some tests for this, or how do we know that only Gecko has the flag now?

foolip:
Didn't have to look too far, https://www.w3.org/Bugs/Public/show_bug.cgi?id=25412 has a Live DOM case used to determine current behavior.

I don't have much of an opinion on what behavior makes sense. Is there a recursion check for other cases where a script can call into C++ which synchronously invokes a callback? For dispatchEvent it looks like the limit varies quite a bit: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/4628

Looks like an explicit test for the Gecko behavior should be easier than a test for its absence, does a test already exist in the Gecko tree that you could upstream to web-platform-tests?

bzbarsky:
> Is there a recursion check for other cases where a script can call into C++ which synchronously invokes a callback?

Well, there's the document.write recursion limitation various browsers do, though it allows some number of levels of nesting....

ayg:
I'm sure we had a wpt test at some point, but I don't see it.  It would be easy to write a basic one.

ayg:
w3c/web-platform-tests#5651 tests this, so would anyone like to review?  Firefox passes the tests, and all other browsers presumably fail.  (I tested in Chrome and WebKit.)

annevk:
Tests and this patch LGTM. However, we don't really know what other browsers want to do. @RByers @cdumez @travisleithead?

@ayg another way forward here is to file bugs against the other browsers. That sometimes helps reaching the right people.

foolip:
I'll defer to @tkent-google for this, the failing test would likely show up on his radar first.

zcorpan:
document.write limit is https://github.com/whatwg/html/issues/1954

tkent-google:
Adding the flag looks reasonable.

ayg:
We have three out of four browsers in agreement here, and have working tests.  Only Edge hasn't replied.  Can we merge this?

annevk:
I went ahead and merged the tests. Anyone up for filing browser bugs? (Or have they already been filed?)

I can take care of that too I suppose once it's safe to merge things again.

annevk:
* https://bugs.webkit.org/show_bug.cgi?id=175312
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13137309/
* https://bugs.chromium.org/p/chromium/issues/detail?id=753271

I assumed Firefox is already covered.

PR #361, opened 2016-10-31
Editorial: resolve all novel bikeshed "link errors"

annevk:


annevk:
This also changes somewhat how certain terms are exposed to the world.

domenic:
As far as I can tell the output didn't change in any bad ways. I think you could have leveraged link-defaults instead of sprinkling for="" around but the source changes look good too. I guess these warnings really are a good thing, as it's good to be explicit about what e.g. `<a>index</a>` really is supposed to link to.

annevk:
The weird thing though is that it's not consistent. E.g., `<dfn for=tree>descendant</dfn>` does not seem to require a matching `<a for=tree>descendant</a>`. `<a>Descendant</a>` does not generate an error.

domenic:
I think that's because no other spec apparently defines `descendant`.

annevk:
So the moment one does I'd have to start updating DOM again? That's going to be fun.

domenic:
Yeah, we might be able to ask @tabatkins for a mode that warns when the `<a>`'s `for=""` doesn't match the `<dfn>`s `for=""`. I think the current mode is just geared around avoiding ambiguities in the current ecosystem, not around being future-proof.

annevk:
Hmm. I guess we can squash merge this meanwhile, but maybe best to wait until Wednesday when we can do the other PR to see if there are no new surprises.

tabatkins:
Yes, if you want future-proof turn on `Assume Explicit For` in your spec; then a missing `for` on an autolink is assumed to mean `for=/`, so you have to be explicit with every non-/ `for`.  You have to write more `for` values, but nothing can change under you.

PR #375, opened 2016-11-18
Editorial: make use of the Infra Standard

annevk:


annevk:
HTML still defines "ASCII case-insensitive" which makes it harder to land this (it would link to HTML for that term which is weird).

PR #385, opened 2016-12-15
Point to HTML for DOMStringList

zcorpan:
See https://github.com/w3c/IndexedDB/issues/28

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9579,6 +9579,8 @@ These are the changes made to the features described in
 {{DOMString}}, {{DOMException}}, and
 {{DOMTimeStamp}} are now defined in Web IDL.
 
+<p id=domstringlist>{{DOMStringList}} is now defined in HTML.
+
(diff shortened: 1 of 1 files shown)


domenic:
Note that per https://github.com/whatwg/html/issues/2179 we might want to move it back to HTML. Merging this seems fine in the meantime.

annevk:
You think DOMStringList should be in HTML? I was thinking DOM, but I don't really care. Anyway, I'd rather we make a plan and then update all the specifications together then point temporarily to Indexed DB.

domenic:
Since HTML is the one that would be using it, I was thinking HTML.

annevk:
Sounds good. I'm happy to work on that unless someone else volunteers. Shall I just add a section after "2.6.2 Collections" called "2.6.3 The DOMStringList interface"?

annevk:
https://github.com/whatwg/html/pull/2192 has a PR for that.

zcorpan:
LGTM

PR #406, opened 2017-02-01
Fallback to legacy type only when event is trusted

### LINKED ISSUE #404: Event not dispatched by the user agent should not fallback to legacy
Currently, in the [invoke](https://dom.spec.whatwg.org/#concept-event-listener-invoke) with event algorithm, when no listener is found and the event has a legacy prefixed event type, the event is invoked with the legacy event type.

I think this should check whether the original event is dispatched by the user agent. That says, I think in the following case:
```html
<script>
document.body.addEventListener('webkitTransitionEnd', evt => console.log(evt));
document.body.dispatchEvent(new TransitionEvent('transitionend'));
</script>
```
the listener shouldn't be triggered.

This proposal doesn't match what browsers currently do, but I suspect whether changing this could lead to web compatibility issues.

The main reason I think we should probably change is that, it would make it easier to write polyfill. Something like
```js
elem.addEventListener('webkitTransitionEnd', function(evt) => {
  elem.dispatchEvent(new TransitionEvent('transitionend'));
});
```
would work in both new browsers and old browsers.

I raise this issue because there is [a bug filed to Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1332699) about this for prefixed Fullscreen API that we are dispatching prefixed event when the script dispatches an unprefixed fullscreen event, which leads to an infinite recursive when there is no listener to unprefixed event.

Although fullscreen event isn't part of this spec (actually we don't spec any legacy event type for it currently), I think the same logic applies in both cases.

cc @smaug---- @annevk @miketaylr @foolip

upsuper:
Fixes #404.

Web platform test PR: w3c/web-platform-tests#4687

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1321,7 +1321,8 @@ for discussion).
  <var>event</var> and <var>listeners</var>.
 
  <li>
-  <p>If <var>found</var> is false and <var>event</var>'s {{Event/isTrusted}} attribute is true, run these substeps:
+  <p>If <var>found</var> is false and <var>event</var>'s {{Event/isTrusted}} attribute is true, then
+  run these substeps:
 
   <ol>
    <li><p>Let <var>originalEventType</var> be <var>event</var>'s {{Event/type}} attribute value.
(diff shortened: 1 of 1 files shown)


upsuper:
@annevk @miketaylr Can we move forward on this change as well as w3c/web-platform-tests#4687?

PR #419, opened 2017-03-07
Remove no-op range mutations from split a Text node

annevk:
Per https://github.com/w3c/web-platform-tests/issues/2499#issuecomment-172901580 this step is redundant.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/no-op-range-mutations/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/2358735...653fc60.html)


nox:
LGTM.

Issue #76, opened 2015-09-18
Should EventInit contain 'timestamp'?


RByers:
https://dom.spec.whatwg.org/#dictdef-eventinit

If we change (as planned) the semantics of `timestamp` to be the platform time (as a DOMHighResTimestamp), then should script be able to supply arbitrary values for it?

Relevant to blink change here: https://codereview.chromium.org/1352523002/#msg9

annevk:
Why?

RByers:
Sorry, let's back up a second, I may be jumping to conclusions based on partial discussions elsewhere.  Filed #80 to track the larger issue.

If we decide to do that, then `timestamp` is no longer an inherent property of the event, but data supplied by the thing that created the event, and as you've argued elsewhere (`TouchEvent.pageX` I think?) all such properties should be settable via the constructor.

Eg. imagine a PointerEvents polyfill that generates `pointermove` events from `mousemove` and `touchmove`.  If `timestamp` represents the time the input was first received by the OS, then the polyfill should be able to propagate the value from a `mousemove` into a generated `pointermove`.  There are other similar non-polyfill scenarios involving any synthetic input event.

annevk:
Interesting. I'm no longer sure whether `MouseEvent.prototype.offsetX` needs to be able to be set or should just be computed on creation time, but your `timestamp` scenario does somewhat argue for making it settable, or at least for creating an event based on an existing one.

RByers:
Thanks.  In addition to copying the timestamp from other events, there are at least some small scenarios where you really want to synthesize a timestamp.  Eg. Android and now iOS 9 align input to vsync through position/time interpolation.  It's reasonable for a library to want to do something similar taking, say, a 100hz input event source and converting it to a 60hz one phase-locked to `requestAnimationFrame`.  Doing that requires generating events with `timestamp` values that are interpolated (or perhaps even extrapolated) from the `timestamp`s of other events.

But I think we can wait until we've done the spec updates for #80 to decide exactly what we want to do here.

annevk:
Do we still want this?

RByers:
Seems like this is blocked on #23, right?  If the ultimate decision in #23 is that `timeStamp` should ALWAYS be the time the event object itself was created (i.e. pretty much useless), then there's probably no value in being able to set it by the constructor.

annevk:
I guess we can wait on that.

RByers:
[...]

PR #420, opened 2017-03-08
Update Event.timeStamp type to DOMHighResTimeStamp.

### LINKED ISSUE #23: High resolution timing for events
https://lists.w3.org/Archives/Public/public-web-perf/2014Jun/0013.html
https://bugzilla.mozilla.org/show_bug.cgi?id=1026804

majido:
Update Event.timeStamp type to DOMHighResTimeStamp.

This fixes #23 and the commit contains the following changes:
- The text asks for the high resolution time of the event occurrence time to be used when
available and otherwise event object creation time.
- Add note in legacy section describing the change in semantic.
- Add note encouraging user agents to make high res time available for UI Events.
- Add necessary spec reference for Bikeshed to pick up new definitions in hr-time spec.

***
[Preview](https://s3.amazonaws.com/pr-preview/majido/dom/high-res-timestamp.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/334428f...majido:3f9f333.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5177,7 +5177,7 @@ invoked, must run these steps:
 
  <li><p>Unset <var>event</var>'s <a>initialized flag</a>.
 
- <li><p>Return <var>event</var>.
+ <li><p>Return <var>event</var>.`
 </ol>
 
 <p class=note><a>Event</a> constructors ought to be used instead.
@@ -9850,6 +9850,7 @@ Kirill Topolyan,
 Koji Ishii,
 Lachlan Hunt,
 Lauren Wood,
+Majid Valipour,
 Malte Ubl,
 Manish Goregaokar,
 Manish Tripathi,
--- a/dom.bs
+++ b/dom.bs
@@ -633,8 +633,8 @@ algorithm below.
  false otherwise.
 
  <dt><code><var>event</var> . {{Event/timeStamp}}</code>
- <dd>Returns the <var>event</var>'s timestamp as the number of milliseconds measured relative
- from the <a>time origin</a>.
+ <dd>Returns the <var>event</var>'s timestamp as the number of milliseconds measured relative to
+ the <a>time origin</a>.
 </dl>
 
 The <dfn attribute for=Event><code>type</code></dfn> attribute must
@@ -767,12 +767,10 @@ the user agent to dispatch an <a>event</a> whose {{Event/isTrusted}} attribute i
 false.
 
 The <dfn attribute for=Event><code>timeStamp</code></dfn> attribute must return the value it was
-initialized to. When an <a>event</a> is created the attribute must be initialized with a
-{{DOMHighResTimeStamp}} representing the high resolution time in milliseconds that has passed
-since the <a>time origin</a>.
+initialized to.
 
 <p class=warning> User agents are strongly encouraged to set minimum resolution of the
-{{Event/timeStamp}} attribute to 5 microseconds following existing <a>clock resolution</a>
+{{Event/timeStamp}} attribute to 5 microseconds following the existing <a>clock resolution</a>
 recommendation. [[!HR-TIME]]
 
 <hr>
@@ -880,8 +878,8 @@ must be run, given the arguments <var>type</var> and <var>eventInitDict</var>:
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>type</var>.
 
  <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
- representing the <var>event</var>'s <dfn>construction time</dfn> which is the high resolution
- time from the <a>time origin</a> to the occurrence of the call to the <a>constructor</a>.
+ representing the high resolution time from the <a>time origin</a> to the occurrence of the call
+ to the <var>event</var>'s <a>constructor</a>.
 
  <li><p>For each <a>dictionary member</a> present in <var>eventInitDict</var>, find the attribute on
  <var>event</var> whose <a spec=webidl>identifier</a> matches the key of the
@@ -915,13 +913,15 @@ it, and optionally given a <a>Realm</a> <var>realm</var>, run these steps:</p>
  <a>dictionary member</a> and then set the attribute to the default value of that
  <a>dictionary member</a>.
 
- <li><p>If available, set <var>event</var>'s {{Event/timeStamp}} attribute to a
- {{DOMHighResTimeStamp}} representing the <var>event</var>'s <dfn>occurence time</dfn> which is
- the high resolution time from the <a>time origin</a> to the occurrence that the event is
- signaling, for example the moment when a particular user interaction occurred.
+ <li><p>Set <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
+ representing the high resolution time from the <a>time origin</a> to the occurrence that the
+ event is signaling.
 
- <p class=note>User agents are encouraged to make the high resolution occurrence time available
- for as many events as possible, in particular those related to user interaction.
+ <p class=note>For example, in macOs the occurrence time for input actions are available via the
+ timestamp property of corresponding <code>NSEvent</code> objects. So in this case, the
+ {{Event/timeStamp}} value should be equivalent to the <code>NSEvent</code>'s timestamp offset by
+ <a>time origin</a>, translated in milliseconds, and with its precision adjusted to meet the
+ recommended minimum <a>clock resolution</a>.
 
  <li><p>Initialize <var>event</var>'s {{Event/isTrusted}} attribute to true.
 
@@ -5173,7 +5173,6 @@ invoked, must run these steps:
  <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
  which is the high resolution time from the <a>time origin</a> to the occurrence of the call to
  {{Document/createEvent()}}.
-
 
  <li><p>Initialize <var>event</var>'s {{Event/isTrusted}} attribute to false.
 
--- a/dom.bs
+++ b/dom.bs
@@ -5177,7 +5177,7 @@ invoked, must run these steps:
 
  <li><p>Unset <var>event</var>'s <a>initialized flag</a>.
 
- <li><p>Return <var>event</var>.`
+ <li><p>Return <var>event</var>.
 </ol>
 
 <p class=note><a>Event</a> constructors ought to be used instead.
--- a/dom.bs
+++ b/dom.bs
@@ -917,7 +917,7 @@ it, and optionally given a <a>Realm</a> <var>realm</var>, run these steps:</p>
  representing the high resolution time from the <a>time origin</a> to the occurrence that the
  event is signaling.
 
- <p class=note>For example, in macOs the occurrence time for input actions are available via the
+ <p class=note>For example, in macOS the occurrence time for input actions are available via the
  timestamp property of corresponding <code>NSEvent</code> objects. So in this case, the
  {{Event/timeStamp}} value should be equivalent to the <code>NSEvent</code>'s timestamp offset by
  <a>time origin</a>, translated in milliseconds, and with its precision adjusted to meet the
@@ -5171,8 +5171,7 @@ invoked, must run these steps:
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to the empty string.
 
  <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
- which is the high resolution time from the <a>time origin</a> to the occurrence of the call to
- {{Document/createEvent()}}.
+ representing the high resolution time from the <a>time origin</a> to now.
 
  <li><p>Initialize <var>event</var>'s {{Event/isTrusted}} attribute to false.
 
--- a/dom.bs
+++ b/dom.bs
@@ -70,6 +70,9 @@ urlPrefix: https://w3c.github.io/ServiceWorker/#; spec: SERVICE-WORKERS
 urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
     text: Construct; url: sec-construct; type: abstract-op
     text: Realm; url: realm; type: dfn
+urlPrefix: https://w3c.github.io/hr-time/#
+    type:typedef; urlPrefix: dom-; text: DOMHighResTimeStamp
+    type:dfn; text: time origin
 </pre>
 
 <pre class='link-defaults'>
@@ -522,7 +525,7 @@ interface Event {
   readonly attribute boolean composed;
 
   [Unforgeable] readonly attribute boolean isTrusted;
-  readonly attribute DOMTimeStamp timeStamp;
+  readonly attribute DOMHighResTimeStamp timeStamp;
 
   void initEvent(DOMString type, optional boolean bubbles = false, optional boolean cancelable = false); // historical
 };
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -761,15 +764,14 @@ initialized to false.
...
-must return the value it was initialized to. When an
-<a>event</a> is created the attribute must be
-initialized to the number of milliseconds that have passed since
-00:00:00 UTC on 1 January 1970, ignoring leap seconds.
-<!-- leap seconds are ignored by JavaScript too -->
+The <dfn attribute for=Event><code>timeStamp</code></dfn> attribute must return the value it was
+initialized to. If available, the attribute should be initialized with the high resolution time of
+the occurrence that the event is signaling, for example the moment when a particular user
+interaction occurred. Otherwise it should be initialized with the time representing the creation
+time of the <var>event</var>.
```
domenic: Nit: "s/the time representing the creation time/the creation time"

dom.bs
```diff
@@ -761,15 +764,14 @@ initialized to false.
...
 
-<p class=XXX>This is highly likely to change and already does not reflect implementations well.
-Please see <a href=https://github.com/whatwg/dom/issues/23>dom #23</a> for more details.
+<p class=note>User agents are encouraged to make the high resolution time occurrence time
+available for as many events as possible in particular those related to user interaction.
```
domenic: Nit: comma after "as possible"

dom.bs
majido: Done.

dom.bs
majido: Done

dom.bs
```diff
@@ -761,15 +764,14 @@ initialized to false.
...
-must return the value it was initialized to. When an
-<a>event</a> is created the attribute must be
-initialized to the number of milliseconds that have passed since
-00:00:00 UTC on 1 January 1970, ignoring leap seconds.
-<!-- leap seconds are ignored by JavaScript too -->
+The <dfn attribute for=Event><code>timeStamp</code></dfn> attribute must return the value it was
+initialized to. If available, the attribute should be initialized with the high resolution time of
+the occurrence that the event is signaling, for example the moment when a particular user
+interaction occurred. Otherwise it should be initialized with the creation time of the
+<var>event</var>.
```
annevk: What does "creation time" mean?

dom.bs
```diff
@@ -761,15 +764,14 @@ initialized to false.
...
 false.
 
-The <dfn attribute for=Event><code>timeStamp</code></dfn> attribute
-must return the value it was initialized to. When an
-<a>event</a> is created the attribute must be
-initialized to the number of milliseconds that have passed since
-00:00:00 UTC on 1 January 1970, ignoring leap seconds.
-<!-- leap seconds are ignored by JavaScript too -->
+The <dfn attribute for=Event><code>timeStamp</code></dfn> attribute must return the value it was
+initialized to. If available, the attribute should be initialized with the high resolution time of
```
annevk: Can we enumerate when it's not?

annevk: It seems that for synthetic events this would never be available.

dom.bs
```diff
@@ -628,8 +631,8 @@ algorithm below.
  false otherwise.
 
  <dt><code><var>event</var> . {{Event/timeStamp}}</code>
- <dd>Returns the creation time of <var>event</var> as the number of milliseconds that
- passed since 00:00:00 UTC on 1 January 1970.
+ <dd>Returns the <var>event</var>'s timestamp as the number of milliseconds measured relative
+ from the <a>time origin</a>.
```
annevk: It seems to me that this term is also important for the normative definition, yet it doesn't appear there.

dom.bs
majido: What do you mean by synthetic events? If you are referring to those constructed in Javascript then that is probably correct and what for example Chrome implementation does. (BTW, I think issue #76 is proposing to make timestamp a constructor argument in which case this is not true anymore)

There may be other implementation related restrictions that makes the precise occurrence time  unavailable. For example on some platforms (e.g., older Windows) there is no cross-process monotonic clock which in browsers with multi-process architecture can mean we cannot use the original timestamp. Or sometimes the driver uses a different clock and timestamp is not reliable. In such cases we end up using the closest estimation we can get. I think falling back to "creation time" is acceptable here too.

dom.bs
majido: This is the same terminology that was used before. The way I read it was the time where the <var>event</var> instance was constructed which is what Chrome implementation was also using. 

Put it another way this is IMHO step 1 in section 2.4.  Perhaps it should link to that.

majido:
> Do we have tests already?

No I have not upstreamed the test yet. I was thinking to have this PR first and gain some confidence that it is accurate and then refer to it when upstreaming the test.

Also, hr-time spec has [recommendation](https://w3c.github.io/hr-time/#privacy-security) on limiting the minimum resolution of the clock exposed via performance.now() to prevent certain types of timing attacks. Given that synthetic events will use the same clock, they can be used to the same effect (e.g., repeatedly creating events and reading their timestamp). We have made sure to prevent this in Chrome. I think we should add some language to the spec (perhaps a note?) recommending the same thing. wdyt?

> the recommended minimum resolution of the Performance interface should be set to 5 microseconds.

dom.bs
annevk: The terminology before had specific instructions with respect to what the value should contain. "Creation time" by itself doesn't give me a number.

dom.bs
annevk: Yeah, a synthetic event is basically a "userland" event. It would be good if we were consistent in what value it would get there (and yes, maybe at some point we allow custom values there, but that would just end up being a branch there).

I'm not really sure how to best do it though. I only have a long term vision where user interaction events are defined in much more detail and they end up initializing this attribute properly (and by default it's just creation time, whatever that means).

annevk:
Yeah, a class=warning block that points there below the normative definition of the attribute seems good.

majido:
@annevk thanks for bearing with me on this.

So I just added a new commit which I think should address your feedback. Here are the changes I made:

Normative definition of timeStamp attribute now defines that it should be initialized to equivalent of performance.now() on creation. I also added warning about 5 microsends minimum resolution from hrtime spec and referenced that spec.

I also added details on how this attribute should be initialized in each algorithm where we create or construct an Event. 
 * On event creation, if available the occurrence time should be used to initialize the property.
 * On event construction or when `Document.createEvent` is used,  the time that the method is invoked should be used. **

** I actually feel that this may not be necessary since the attribute definition specifies the same but I followed the example of how `isTrusted` is spec'd. :confused:

dom.bs
majido: Done. I now have exact details on what to use for synthetic events (i.e., use construction time).

dom.bs
majido: Done. The wording now says ```When an <a>event</a> is created the attribute must be initialized with a {{DOMHighResTimeStamp}} representing the high resolution time in milliseconds that has passed since the <a>time origin</a>.``` 

This is similar to previous text but with new time type and origin. This initial value may change in event creation algorithm which now have the wording about occurrence time etc.

majido:
As I mentioned in #76, there may be potential for simplifying this if we add timestamp to EventInit with a default value. But perhaps that is best done in a separate patch.

majido:
@annevk  I also have a  [Blink patch]((https://codereview.chromium.org/2744553007/) ) to upstream parts of our tests for this change. I will land those once this patch is lgtm'd. Please take a look there an let me know if you have any comments on the tests as well.

annevk:
I think the issue I mention in #76 also applies to this patch.

dom.bs
```diff
@@ -628,8 +633,8 @@ algorithm below.
...
```
annevk: relative to, no?

dom.bs
```diff
@@ -761,15 +766,14 @@ initialized to false.
...
-The <dfn attribute for=Event><code>timeStamp</code></dfn> attribute
-must return the value it was initialized to. When an
-<a>event</a> is created the attribute must be
-initialized to the number of milliseconds that have passed since
-00:00:00 UTC on 1 January 1970, ignoring leap seconds.
-<!-- leap seconds are ignored by JavaScript too -->
+The <dfn attribute for=Event><code>timeStamp</code></dfn> attribute must return the value it was
+initialized to. When an <a>event</a> is created the attribute must be initialized with a
+{{DOMHighResTimeStamp}} representing the high resolution time in milliseconds that has passed
+since the <a>time origin</a>.
```
annevk: It seems we don't need to define what it is set to when the event is created as we already do that later separately, no?

dom.bs
```diff
@@ -761,15 +766,14 @@ initialized to false.
...
 
-<p class=XXX>This is highly likely to change and already does not reflect implementations well.
-Please see <a href=https://github.com/whatwg/dom/issues/23>dom #23</a> for more details.
+<p class=warning> User agents are strongly encouraged to set minimum resolution of the
+{{Event/timeStamp}} attribute to 5 microseconds following existing <a>clock resolution</a>
```
annevk: the existing*

Doesn't read well otherwise.

dom.bs
```diff
@@ -875,6 +879,10 @@ must be run, given the arguments <var>type</var> and <var>eventInitDict</var>:
 
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>type</var>.
 
+ <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
+ representing the <var>event</var>'s <dfn>construction time</dfn> which is the high resolution
+ time from the <a>time origin</a> to the occurrence of the call to the <a>constructor</a>.
```
annevk: What is the benefit of the new `<dfn>` here? Can we reword this without introducing a new `<dfn>`?

dom.bs
```diff
@@ -907,6 +915,14 @@ it, and optionally given a <a>Realm</a> <var>realm</var>, run these steps:</p>
  <a>dictionary member</a> and then set the attribute to the default value of that
  <a>dictionary member</a>.
 
+ <li><p>If available, set <var>event</var>'s {{Event/timeStamp}} attribute to a
+ {{DOMHighResTimeStamp}} representing the <var>event</var>'s <dfn>occurence time</dfn> which is
+ the high resolution time from the <a>time origin</a> to the occurrence that the event is
+ signaling, for example the moment when a particular user interaction occurred.
```
annevk: Same here.

annevk: Also, if what is available? Should we not just make this a requirement and not allow user agents not to have this?

dom.bs
```diff
@@ -5154,6 +5170,11 @@ invoked, must run these steps:
 
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to the empty string.
 
+ <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
+ which is the high resolution time from the <a>time origin</a> to the occurrence of the call to
+ {{Document/createEvent()}}.
+
+
```
annevk: One newline too many as far as I can tell.

annevk: It also seems ugly that we initialize it twice.

dom.bs
annevk: I think we should also have a note here that gives a far more detailed example. E.g., the name of a particular system call on some OS that gives you the time we'd expect to see exposed here.

majido: I removed "If available". I was thinking of cases where there is not any high resolution time available on the underlying event, e.g., some older Windows devices don't have a high resolution clock or the clock is not usable across processes. In such cases, Chromium uses the timestamp when event was received by renderer process as the closest thing we can have to the "occurrence time". 

I guess there is no need to have the spec special case this as the term "occurrence time" gives enough flexibility to user agents to handle such limitations. Also per your suggestion, I added an example (for macOs)  which should help guide how occurrence time should be interpreted by implementations.

dom.bs
majido: Done. 

As for duplicate initialization, I am not sure if I understand. The way I was reading the spec was that for non-synthetic events we do the `2.4 - concept-event-create ` that does not require invoking `2.4. - concept-event-constructor` which is only needed for synthetic events.

dom.bs
majido: Done.

dom.bs
majido: Done. Removed.

dom.bs
majido: Removed. To be honest, I found it odd too but was cargo culting the definition of `isTrusted` just above it.

dom.bs
majido: Done.

majido:
> Apologies for the delay in review. The main reason is that I've been away for 5+ weeks.

Not at all. Vacations are great :smile: and there is not particular rush for this. :+1:

dom.bs
```diff
@@ -915,13 +913,15 @@ it, and optionally given a <a>Realm</a> <var>realm</var>, run these steps:</p>
...
- {{DOMHighResTimeStamp}} representing the <var>event</var>'s <dfn>occurence time</dfn> which is
- the high resolution time from the <a>time origin</a> to the occurrence that the event is
- signaling, for example the moment when a particular user interaction occurred.
+ <li><p>Set <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
+ representing the high resolution time from the <a>time origin</a> to the occurrence that the
+ event is signaling.
 
- <p class=note>User agents are encouraged to make the high resolution occurrence time available
- for as many events as possible, in particular those related to user interaction.
+ <p class=note>For example, in macOs the occurrence time for input actions are available via the
```
annevk: macOS*

dom.bs
```diff
@@ -907,6 +913,16 @@ it, and optionally given a <a>Realm</a> <var>realm</var>, run these steps:</p>
  <a>dictionary member</a> and then set the attribute to the default value of that
  <a>dictionary member</a>.
 
+ <li><p>Set <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
+ representing the high resolution time from the <a>time origin</a> to the occurrence that the
+ event is signaling.
+
+ <p class=note>For example, in macOs the occurrence time for input actions are available via the
```
annevk: macOS

dom.bs
```diff
@@ -5154,6 +5170,10 @@ invoked, must run these steps:
 
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to the empty string.
 
+ <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
+ which is the high resolution time from the <a>time origin</a> to the occurrence of the call to
+ {{Document/createEvent()}}.
```
annevk: This was what I was talking about with respect to duplication, but I guess that's also following the isTrusted precedent. Probably okay. Can this become "from the time origin to now"? It's a little weird for the algorithm to refer to itself this way.

annevk:
Oh also, could you add an explicit pointer to the tests somewhere? I haven't been able to find them.

annevk:
Actually, it seems that for isTrusted we don't define anything in the constructor. So that's why it's slightly different with respect to that.

dom.bs
majido: Done.

dom.bs
majido: Replaced self-reference with "now".

majido:
Tests are here: https://codereview.chromium.org/2744553007/
I have not moved them to WPT yet as I was waiting for spec patch to land. Happy to address any comments you may have on them.

annevk:
Thanks, those look good to me.

Do you want to add your name to the Acknowledgments section as well? With that we can land this I think. I'll go ahead and file a bug against Edge so that's covered as well.

majido:
> Do you want to add your name to the Acknowledgments section as well? 

Absolutely. :medal_military:

majido:
I will go ahead an upstream the tests now.

dom.bs
```diff
@@ -5154,11 +5170,14 @@ invoked, must run these steps:
...
 
+ <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
+ representing the high resolution time from the <a>time origin</a> to now.
+
  <li><p>Initialize <var>event</var>'s {{Event/isTrusted}} attribute to false.
 
  <li><p>Unset <var>event</var>'s <a>initialized flag</a>.
 
- <li><p>Return <var>event</var>.
+ <li><p>Return <var>event</var>.`
```
annevk: Seems you added this by accident in the last commit.

majido: Ops! fixed.

annevk:
Thanks!

PR #423, opened 2017-03-16
Use a single exception for name validation

### LINKED ISSUE #319: Validating a qualified name does not match browsers in a corner case
``` html
<!DOCTYPE html>
<script>
try { document.createElementNS("a", "a:0") }
catch(e) { document.documentElement.textContent = e.name }
</script>
```

Outputs InvalidCharacterError in Chrome, Firefox, and Edge.  Per [spec](https://dom.spec.whatwg.org/#validate), it should throw NamespaceError, because "a:0" matches the Name production.  I think browsers should probably change here instead of the spec, because browsers' behavior doesn't make sense to me, but I'm filing a spec bug anyway because we seem to have interop against the spec.

annevk:
Tests: https://github.com/w3c/web-platform-tests/issues/5161.

Fixes #319.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/name-validation/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/334428f...24569d8.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -319,15 +319,9 @@ added.
 
 <h3 id=namespaces>Namespaces</h3>
 
-To <dfn export>validate</dfn> a <var>qualifiedName</var>, run these steps:
-
-<ol>
- <li><p>If <var>qualifiedName</var> does not match the <code><a type>Name</a></code> production,
- then <a>throw</a> an {{InvalidCharacterError}}.
-
- <li><p>If <var>qualifiedName</var> does not match the <code><a type>QName</a></code> production,
- then <a>throw</a> a {{NamespaceError}}.
-</ol>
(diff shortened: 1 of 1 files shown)


domenic:
I have confirmed that Safari Tech Preview has not updated to follow the (current) spec yet, at least for `createElementNS("a", "a:0")`, so I think this is landable with a clear conscience.

annevk:
So it's not that simple. The changed tests cause failures in all browsers. They all use NAMESPACE_ERR for some set of invalid names currently. So bz expressed willingness to change that so that name validation uses a single exception, but seeing the results we'd need explicit sign-off from @tkent-google and @cdumez too, at least.

domenic:
Oh. In that case maybe it is indeed better to wait to see if we can relax the validation rules, instead of changing twice :-/.

PR #426, opened 2017-03-17
Editorial: address Bikeshed linking errors

annevk:
***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/link-errors/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/334428f...7cd9c9c.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2602,7 +2602,7 @@ CharacterData implements ChildNode;
 </pre>
 
 <dl class=domintro>
- <dt><code><var>node</var> . {{before(nodes)}}</code>
+ <dt><code><var>node</var> . <a method for=ChildNode>before(<var>nodes</var>)</a></code>
  <dd>
   Inserts <var>nodes</var> just before <var>node</var>,
   while replacing strings in <var>nodes</var> with equivalent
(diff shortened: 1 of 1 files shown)


annevk:
@tabatkins enjoy your break, and when you're back, can you look at this and tell me why this works and the previous thing started failing at some point?

tabatkins:
Why it started failing is easy: your linking text is wrong, but it was only a little while ago that my code started correctly recognizing that it was wrong (due to the way widlparser worked).  In particular, `before(nodes)` is *not* a valid signature for the function; `before(...nodes)` is.  (That is, it *does not* take an argument named "nodes", it takes a *rest argument* named "nodes".)

However, I've got some somewhat-wrong auto-massager code that also does some somewhat-wrong fixup in some circumstances; in particular, if you have an autolink without an `lt`, and it's a functionish type, it'll set the `lt` to an argument-less signature automatically.  This is correct for CSS functions, which don't have overloads and often show up in autolinks with all sorts of nonsense for args (like `''repeat(3, 10px)''` in an example, which just wants to link to the `repeat()` function definition), but it's wrong for IDL methods in case there *are* overloads.  (The shorthands like `{{before(nodes)}}` *do* set the `lt` on the HTML they generate, so they don't hit the massager, thus them breaking.)

---

So, actions:

1. I need to fix the massager code to not fire on IDL methods; it'll cause confusing inconsistencies sometimes (like here).

2. You need to fix DOM to use the right signature, with `...nodes` as the arg.

PR #432, opened 2017-03-24
Editorial: define attribute list in terms of Infra's list

annevk:
***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/attribute-list-infra/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/3078a9d...53f20eb.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3999,12 +3999,8 @@ A <a>node</a> <var>A</var>
    <a>system ID</a>.
 
    <dt>{{Element}}
-   <dd>
-    Its <a for=Element>namespace</a>,
-    <a for=Element>namespace prefix</a>,
-    <a for=Element>local name</a>, and its
-    number of <a>attributes</a> in its
-    <a for=Element>attribute list</a>.
+   <dd>Its <a for=Element>namespace</a>, <a for=Element>namespace prefix</a>,
+   <a for=Element>local name</a>, and its <a for=Element>attribute list</a>'s <a for=list>size</a>.
 
    <dt>{{Attr}}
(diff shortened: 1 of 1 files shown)


annevk:
See also https://github.com/whatwg/infra/issues/103. Could wait for that potentially, but not sure.

dom.bs
```diff
@@ -1270,7 +1270,7 @@ for discussion).
 <ol>
  <li><p>If <var>event</var>'s <a>stop propagation flag</a> is set, then return.
 
- <li><p>Let <var>listeners</var> be a new <a>list</a>.
+ <li><p>Let <var>listeners</var> be a new <a for=/>list</a>.
```
domenic: Can't you use link defaults so you don't have to do this?

annevk: I don't like link defaults. What was that other option we thought of enabling at some point for WHATWG?

domenic: Explicit for, which requires for="" always if the source has it.

I think link defaults is good and honestly we could enable link defaults to infra for all of WHATWG on things like list/string.

Anchors is the bad one.

annevk: You mean we could put link defaults at a higher level than the source for WHATWG in Bikeshed somehow? That might be nice.

annevk: (The problem with link defaults is that what we consider a shared ecosystem is in fact using contradictory terms after all. So ideally one of the sides in the dispute changes somehow.)

annevk:
Used link-defaults, but guess I'll wait to see if "replace" becomes a thing.

PR #434, opened 2017-04-05
Define a 'CancelationController' and 'CancelationSignal' interface.

mikewest:
Fetch is sketching out a cancellation mechanism in whatwg/fetch#447, and
other specifications are interested in cargo-culting their way to similar
behaviors.

To that end, this patch defines a simple interface that provides enough
control for the simple case of cancellation, while paving the way for
more complex consumers like Fetch to tack on arbitrary complexity.

***
[Preview](https://s3.amazonaws.com/pr-preview/mikewest/dom/promisecontroller.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/6ba16e7...mikewest:4273de9.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1478,7 +1478,8 @@ steps:
 <ol>
  <li>Let <var>signal</var> be this object's {{PromiseController/signal}}.
  <li>Set <var>signal</var>'s [=PromiseSignal/aborted flag=].
- <li>[=Fire an event=] named <code>abort</code> at <var>signal</var>.</li>
+ <li>[=Fire an event=] named <code>abort</code> at <var>signal</var>.
+ <li>Run <var>signal</var>'s [=PromiseSignal/abort steps=].
 </ol>
 
 <h4 id=interface-cancellationcontroller>Interface {{PromiseSignal}}</h4>
@@ -1498,6 +1499,15 @@ interface PromiseSignal : EventTarget {
 
 Each {{PromiseSignal}} has an <dfn for=PromiseSignal>aborted flag</dfn> which is unset unless
 otherwise specified.
+
+Each {{PromiseSignal}} has an <dfn for=PromiseSignal>abort steps</dfn> algorithm which is
+executed when its [=PromiseSignal/aborted flag=] is set. Unless otherwise specified, this
+algorithm is a no-op.
+
+<p class="note no-backref">The [=PromiseSignal/abort steps=] algorithm enables APIs with more
+complex requirements to react in a reasonable way to {{PromiseController/abort()}}. For example,
+a given API's [=PromiseSignal/aborted flag=] may need to be propagated to a cross-thread
+environment (like a Service Worker).
 
 The <dfn attribute for=PromiseSignal>aborted</dfn> attribute's getter must return true if the
 object's [=PromiseSignal/aborted flag=] is set, and false otherwise.
--- a/dom.bs
+++ b/dom.bs
@@ -55,6 +55,10 @@ urlPrefix: https://w3c.github.io/ServiceWorker/#; spec: SERVICE-WORKERS
 urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
     text: Construct; url: sec-construct; type: abstract-op
     text: Realm; url: realm; type: dfn
+
+<!-- Remove this once WebIDL takes https://github.com/heycam/webidl/pull/339 -->
+urlPrefix: https://heycam.github.io/webidl/; spec: WEBIDL
+    text: CancelationError; url: cancelationerror; type: exception
 </pre>
 
 <pre class=link-defaults>
@@ -1428,8 +1432,8 @@ corresponding {{CancelationSignal}} object. The API which wishes to support canc
 such a {{CancelationSignal}}, and use its state to determine how (not) to proceed.
 
 APIs that rely upon {{CancelationController}} are encouraged to respond to
-{{CancelationController/cancel()}} by rejecting any unsettled {{Promise}} with a
-<dfn exception>CancelationError</dfn> exception.
+{{CancelationController/cancel()}} by rejecting any unsettled {{Promise}} with a new
+{{DOMException}} with [=error name=] "{{CancelationError}}".
 
 <div class=note>
  For example, a hypothetical <code>doAmazingness({ ... })</code> method could accept a
--- a/dom.bs
+++ b/dom.bs
@@ -564,7 +564,7 @@ algorithm below.
  the operation that caused <var>event</var> to be <a>dispatched</a> that it needs to be canceled.
 
  <dt><code><var>event</var> . {{Event/defaultPrevented}}</code>
- <dd>Returns true if {{Event/preventDefault()}} was invoked successfully to indicate cancellation,
+ <dd>Returns true if {{Event/preventDefault()}} was invoked successfully to indicate cancelation,
  and false otherwise.
 
  <dt><code><var>event</var> . {{Event/composed}}</code>
@@ -1421,22 +1421,22 @@ can only be used to influence an ongoing one.
 
 <h2 id=canceling-ongoing-activities>Canceling Ongoing Activities</h3>
 
-Though {{Promise}} objects don't have any built-in cancellation mechanism, many APIs using these
-concepts require cancellation semantics. {{CancellationController}} is meant to support these
-requirements by providing an {{CancellationController/cancel()}} method that toggles the state of a
-corresponding {{CancellationSignal}} object. The API which wishes to support cancellation can accept
-such a {{CancellationSignal}}, and use its state to determine how (not) to proceed.
+Though {{Promise}} objects don't have any built-in cancelation mechanism, many APIs using these
+concepts require cancelation semantics. {{CancelationController}} is meant to support these
+requirements by providing an {{CancelationController/cancel()}} method that toggles the state of a
+corresponding {{CancelationSignal}} object. The API which wishes to support cancelation can accept
+such a {{CancelationSignal}}, and use its state to determine how (not) to proceed.
 
-APIs that rely upon {{CancellationController}} are encouraged to respond to
-{{CancellationController/cancel()}} by rejecting any unsettled {{Promise}} with a
-<dfn exception>CancellationError</dfn> exception.
+APIs that rely upon {{CancelationController}} are encouraged to respond to
+{{CancelationController/cancel()}} by rejecting any unsettled {{Promise}} with a
+<dfn exception>CancelationError</dfn> exception.
 
 <div class=note>
  For example, a hypothetical <code>doAmazingness({ ... })</code> method could accept a
- {{CancellationSignal}} object in order to support cancellation as follows:
+ {{CancelationSignal}} object in order to support cancelation as follows:
 
  <pre class=lang-javascript>
-  const controller = new CancellationController();
+  const controller = new CancelationController();
   const signal = controller.signal;
   doAmazingness({ ..., signal })
     .then(result => ...)
@@ -1447,77 +1447,77 @@ APIs that rely upon {{CancellationController}} are encouraged to respond to
   controller.cancel();
   </pre>
 
- APIs that require more granular control could extend both {{CancellationController}} and
- {{CancellationSignal}} according to their needs.
+ APIs that require more granular control could extend both {{CancelationController}} and
+ {{CancelationSignal}} according to their needs.
 </div>
 
-<h3 id=interface-cancellationcontroller>Interface {{CancellationController}}</h3>
+<h3 id=interface-cancelationcontroller>Interface {{CancelationController}}</h3>
 
 <pre class="idl">
 [Constructor(), Exposed=(Window,Worker)]
-interface CancellationController {
-  readonly attribute CancellationSignal signal;
+interface CancelationController {
+  readonly attribute CancelationSignal signal;
 
   void cancel();
 };
 </pre>
 <dl class=domintro>
- <dt><code><var>controller</var> = new <a constructor lt="CancellationController()">CancellationController</a>()</code>
- <dd>Returns a new <var>controller</var> whose {{CancellationController/signal}} is set to a newly
- created {{CancellationSignal}} object.
+ <dt><code><var>controller</var> = new <a constructor lt="CancelationController()">CancelationController</a>()</code>
+ <dd>Returns a new <var>controller</var> whose {{CancelationController/signal}} is set to a newly
+ created {{CancelationSignal}} object.
 
- <dt><code><var>controller</var> . </code>{{CancellationController/signal}}
- <dd>Returns the {{CancellationSignal}} object associated with this object.
+ <dt><code><var>controller</var> . </code>{{CancelationController/signal}}
+ <dd>Returns the {{CancelationSignal}} object associated with this object.
 
- <dt><code><var>controller</var> . </code>{{CancellationController/cancel()}}
- <dd>Invoking this method will set this object's {{CancellationSignal}}'s
- [=CancellationSignal/cancelled flag=], thereby signaling to any observers that the associated
- activity should be cancelled.
+ <dt><code><var>controller</var> . </code>{{CancelationController/cancel()}}
+ <dd>Invoking this method will set this object's {{CancelationSignal}}'s
+ [=CancelationSignal/canceled flag=], thereby signaling to any observers that the associated
+ activity should be canceled.
 </dl>
 
-The <dfn attribute for=CancellationController><code>signal</code></dfn> attribute must return the value
-to which it was initialized. When a {{CancellationController}} is created, the attribute must be
-initialized to a newly created {{CancellationSignal}} object.
+The <dfn attribute for=CancelationController><code>signal</code></dfn> attribute must return the value
+to which it was initialized. When a {{CancelationController}} is created, the attribute must be
+initialized to a newly created {{CancelationSignal}} object.
 
-The <dfn method for=CancellationController><code>cancel()</code></dfn> method, when invoked, must run
+The <dfn method for=CancelationController><code>cancel()</code></dfn> method, when invoked, must run
 these steps:
 
 <ol>
- <li>Let <var>signal</var> be this object's {{CancellationController/signal}}.
- <li>Set <var>signal</var>'s [=CancellationSignal/cancelled flag=].
- <li>Run <var>signal</var>'s [=CancellationSignal/cancellation steps=].
+ <li>Let <var>signal</var> be this object's {{CancelationController/signal}}.
+ <li>Set <var>signal</var>'s [=CancelationSignal/canceled flag=].
+ <li>Run <var>signal</var>'s [=CancelationSignal/cancelation steps=].
  <li>[=Fire an event=] named <code>cancel</code> at <var>signal</var>.
 </ol>
 
-<h4 id=interface-cancellationsignal>Interface {{CancellationSignal}}</h4>
+<h4 id=interface-cancelationsignal>Interface {{CancelationSignal}}</h4>
 
 <pre class="idl">
 [Exposed=(Window,Worker)]
-interface CancellationSignal : EventTarget {
-  readonly attribute boolean cancelled;
+interface CancelationSignal : EventTarget {
+  readonly attribute boolean canceled;
 
   attribute EventHandler oncancel;
 };
 </pre>
 <dl class=domintro>
- <dt><code><var>signal</var> . </code>{{CancellationSignal/cancelled}}
- <dd>Returns true if this {{CancellationSignal}} has been cancelled, and false otherwise.
+ <dt><code><var>signal</var> . </code>{{CancelationSignal/canceled}}
+ <dd>Returns true if this {{CancelationSignal}} has been canceled, and false otherwise.
 </dl>
 
-Each {{CancellationSignal}} has an <dfn for=CancellationSignal>cancelled flag</dfn> which is
+Each {{CancelationSignal}} has an <dfn for=CancelationSignal>canceled flag</dfn> which is
 unset unless otherwise specified.
 
-Each {{CancellationSignal}} has an <dfn for=CancellationSignal>cancellation steps</dfn> algorithm
-which is executed when its [=CancellationSignal/cancelled flag=] is set. Unless otherwise
+Each {{CancelationSignal}} has an <dfn for=CancelationSignal>cancelation steps</dfn> algorithm
+which is executed when its [=CancelationSignal/canceled flag=] is set. Unless otherwise
 specified, this algorithm is a no-op.
 
-<p class="note no-backref">The [=CancellationSignal/cancellation steps=] algorithm enables APIs with
-more complex requirements to react in a reasonable way to {{CancellationController/cancel()}}. For
-example, a given API's [=CancellationSignal/cancelled flag=] may need to be propagated to a
+<p class="note no-backref">The [=CancelationSignal/cancelation steps=] algorithm enables APIs with
+more complex requirements to react in a reasonable way to {{CancelationController/cancel()}}. For
+example, a given API's [=CancelationSignal/canceled flag=] may need to be propagated to a
 cross-thread environment (like a Service Worker).
 
-The <dfn attribute for=CancellationSignal>cancelled</dfn> attribute's getter must return true if the
-object's [=CancellationSignal/cancelled flag=] is set, and false otherwise.
+The <dfn attribute for=CancelationSignal>canceled</dfn> attribute's getter must return true if the
+object's [=CancelationSignal/canceled flag=] is set, and false otherwise.
 
 
 
--- a/dom.bs
+++ b/dom.bs
@@ -1419,6 +1419,88 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
 can only be used to influence an ongoing one.
 
 
+<h3 id=controlling-promises>Controlling {{Promise}}s</h3>
+
+<p>Though {{Promise}} objects don't have any built-in cancellation mechanism, many
+{{Promise}}-vending APIs require cancellation semantics. {{PromiseController}} is meant to support
+these requirements by providing an {{PromiseController/abort()}} method that toggles the state of
+a corresponding {{PromiseSignal}} object. The API which wishes to support cancelation can accept
+such a {{PromiseSignal}}, and use its state to determine how (not) to proceed.
+
+<div class=note>
+  For example, a hypothetical <code>DoAmazingness({ ... })</code> method could accept a
+  {{PromiseSignal}} object in order to support cancellation as follows:
+
+  <pre class=lang-javascript>
+    const controller = new PromiseController();
+    const signal = controller.signal;
+    DoAmazingness({ ..., signal }).then(result => ...);
+
+    ...
+
+    controller.abort();
+  </pre>
+
+  APIs that require more granular control could extend both {{PromiseController}} and
+  {{PromiseSignal}} according to their needs.
+</div>
+
+<h4 id=interface-promisecontroller>Interface {{PromiseController}}</h3>
+
+<pre class="idl">
+[Constructor(), Exposed=(Window,Worker)]
+interface PromiseController {
+  readonly attribute PromiseSignal signal;
+
+  void abort();
+};
+</pre>
+<dl class=domintro>
+ <dt><code><var>controller</var> = new <a constructor lt="PromiseController()">PromiseController</a>()</code>
+ <dd>Returns a new <var>controller</var> whose {{PromiseController/signal}} attribute value is set
+ to a newly created {{PromiseSignal}} object.
+
+ <dt><code><var>controller</var> . </code>{{PromiseController/signal}}
+ <dd>Returns the {{PromiseSignal}} object associated with this object.
+
+ <dt><code><var>controller</var> . </code>{{PromiseController/abort()}}
+ <dd>Invoking this method will set this object's {{PromiseSignal}}'s [=PromiseSignal/aborted flag=],
+ thereby signaling to any observers that the associated activity should be aborted.
+</dl>
+
+The <dfn attribute for=PromiseController>signal</dfn> attribute must return the value to which it
+was initialized. When a {{PromiseController}} is created, the attribute must be initialized to a
+newly created {{PromiseSignal}} object.
+
+The <dfn method for=PromiseController><code>abort()</code></dfn>, when invoked, must run these
+steps:
+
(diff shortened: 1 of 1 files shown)


mikewest:
I'm flailing a bit, @jakearchibald. I think this is more or less all we need for Credential Management, and seems like it paves the way for y'all to derive `FetchController` and `FetchSignal` as sketched out in https://github.com/whatwg/fetch/issues/447#issuecomment-281731850. WDYT?

@annevk: I have no idea if DOM is the right place for this, but it seemed to fit without too much squinting. I'd appreciate your feedback regardless of where it ends up.

FYI: @jyasskin, @kpaulh, @equalsJeffH, and @vijaybh as this is relevant to our ongoing CM API thread, as well as w3c/webauthn#380.

annevk:
Are we going to define a non-DOMException for the actual cancellation exception as well then so it could be factored out and move into the language if it ever comes that established?

When I talked rather informally with @ajklein we could only think of Fetch as really needing this, but if more folks want cancellation (some elaboration would be appreciated) we should maybe loop in some other folks for review as well.

Overall your PR looks fine, but I think we need some more abstraction for specification-consumers of the signal (that will then cancel/abort the actual operation). It's also not immediately obvious how this works where we expose the signal (a mirroring cross-thread instance of sorts I suppose) to the service worker (while the controller is on the main thread) which can then pass it along to new fetches (zero or more) which the controller will also end up affecting.

mikewest:
Thanks, @annevk!

> Are we going to define a non-DOMException for the actual cancellation exception as well then so it could be factored out and move into the language if it ever comes that established?

Sure. The Fetch proposal hand-waves an `AbortError`. Happy to fold that into this patch if you'd like.

> but if more folks want cancellation (some elaboration would be appreciated)

WebAuthn wants `navigator.credentials.get()` to be cancellable. @jyasskin and @equalsJeffH have more context there, I'm just trying to support them.

> we should maybe loop in some other folks for review as well.

Please CC the world. :)

> Overall your PR looks fine, but I think we need some more abstraction for specification-consumers of the signal (that will then cancel/abort the actual operation). 

More non-normative notes about usage and suggested spec text for the examples? Or more something else?

> It's also not immediately obvious how this works where we expose the signal (a mirroring cross-thread instance of sorts I suppose) to the service worker (while the controller is on the main thread) which can then pass it along to new fetches (zero or more) which the controller will also end up affecting.

It seems like that would be a Fetch-specific addition, wouldn't it? It's not clear to me that most APIs would need cross-process updates for a signal's state. I guess we can pull some sort of mirroring primitive up to DOM if that would be helpful, but it seems somewhat specific to Fetch's use case.

annevk:
I think in particular the signal object needs a hook ("abort steps" or some such) that get run when `abort()` is invoked. Their default implementation could be a no-op. Those steps might be enough to implement all the Fetch additions.

dom.bs
```diff
@@ -1478,7 +1478,8 @@ steps:
 <ol>
  <li>Let <var>signal</var> be this object's {{PromiseController/signal}}.
  <li>Set <var>signal</var>'s [=PromiseSignal/aborted flag=].
- <li>[=Fire an event=] named <code>abort</code> at <var>signal</var>.</li>
+ <li>[=Fire an event=] named <code>abort</code> at <var>signal</var>.
+ <li>Run <var>signal</var>'s [=PromiseSignal/abort steps=].
```
annevk: I suspect we probably want the event to dispatch after the abort steps have run as the event listeners can cause side effects the abort steps might not anticipate.

domenic:
Streams is also interested in using this. I will try to review soon.

dom.bs
mikewest: Inverted in https://github.com/whatwg/dom/pull/434/commits/8681c81ca74efa6a739eb499c95f2cece0f325f1.

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
 can only be used to influence an ongoing one.
 
 
+<h3 id=controlling-promises>Controlling {{Promise}}s</h3>
```
domenic: This probably doesn't belong under the "Events" section. I'm not sure if it belongs in DOM at all, but I don't know of a better place really, so who knows... new WHATWG spec time?

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+
+Though {{Promise}} objects don't have any built-in cancellation mechanism, many {{Promise}}-vending
```
domenic: I prefer "cancelation" with one "l" but as long as we're consistent we'll be good.

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+APIs require cancellation semantics. {{PromiseController}} is meant to support these requirements by
+providing an {{PromiseController/abort()}} method that toggles the state of a corresponding
+{{PromiseSignal}} object. The API which wishes to support cancellation can accept such a
+{{PromiseSignal}}, and use its state to determine how (not) to proceed.
+
+Upon {{PromiseController/abort()}}, the relevant {{Promise}} will reject with an
+<dfn exception>AbortError</dfn> [=simple exception=].
+
+<div class=note>
+ For example, a hypothetical <code>DoAmazingness({ ... })</code> method could accept a
```
domenic: Don't capitalize doAmazingness

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
 
+<h3 id=controlling-promises>Controlling {{Promise}}s</h3>
+
+Though {{Promise}} objects don't have any built-in cancellation mechanism, many {{Promise}}-vending
+APIs require cancellation semantics. {{PromiseController}} is meant to support these requirements by
+providing an {{PromiseController/abort()}} method that toggles the state of a corresponding
+{{PromiseSignal}} object. The API which wishes to support cancellation can accept such a
+{{PromiseSignal}}, and use its state to determine how (not) to proceed.
+
+Upon {{PromiseController/abort()}}, the relevant {{Promise}} will reject with an
```
domenic: This isn't always the case, right? It depends on what the accepting method will do. I think this needs to be phrased as a convention or something, since it's not enforced.

dom.bs
domenic: Is it cancel(l)ation or abortion? The API says abortion...

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+[Constructor(), Exposed=(Window,Worker)]
+interface PromiseController {
+  readonly attribute PromiseSignal signal;
+
+  void abort();
+};
+</pre>
+<dl class=domintro>
+ <dt><code><var>controller</var> = new <a constructor lt="PromiseController()">PromiseController</a>()</code>
+ <dd>Returns a new <var>controller</var> whose {{PromiseController/signal}} attribute value is set
```
domenic: I'd remove "attribute value" (in general mentioning "attributes" in dev-facing docs is just confusing).

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+
+ <dt><code><var>controller</var> . </code>{{PromiseController/signal}}
+ <dd>Returns the {{PromiseSignal}} object associated with this object.
+
+ <dt><code><var>controller</var> . </code>{{PromiseController/abort()}}
+ <dd>Invoking this method will set this object's {{PromiseSignal}}'s [=PromiseSignal/aborted flag=],
+ thereby signaling to any observers that the associated activity should be aborted.
+</dl>
+
+The <dfn attribute for=PromiseController>signal</dfn> attribute must return the value to which it
```
domenic: Needs `<code>`

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+was initialized. When a {{PromiseController}} is created, the attribute must be initialized to a
+newly created {{PromiseSignal}} object.
+
+The <dfn method for=PromiseController><code>abort()</code></dfn>, when invoked, must run these
```
domenic: Missing "method"

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+Each {{PromiseSignal}} has an <dfn for=PromiseSignal>abort steps</dfn> algorithm which is
+executed when its [=PromiseSignal/aborted flag=] is set. Unless otherwise specified, this
+algorithm is a no-op.
+
+<p class="note no-backref">The [=PromiseSignal/abort steps=] algorithm enables APIs with more
+complex requirements to react in a reasonable way to {{PromiseController/abort()}}. For example,
+a given API's [=PromiseSignal/aborted flag=] may need to be propagated to a cross-thread
+environment (like a Service Worker).
+
+The <dfn attribute for=PromiseSignal>aborted</dfn> attribute's getter must return true if the
```
domenic: Needs `<code>`

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+ <dt><code><var>signal</var> . </code>{{PromiseSignal/aborted}}
+ <dd>Returns true if this {{PromiseSignal}} has been aborted, and false otherwise.
+</dl>
+
+Each {{PromiseSignal}} has an <dfn for=PromiseSignal>aborted flag</dfn> which is unset unless
+otherwise specified.
+
+Each {{PromiseSignal}} has an <dfn for=PromiseSignal>abort steps</dfn> algorithm which is
+executed when its [=PromiseSignal/aborted flag=] is set. Unless otherwise specified, this
+algorithm is a no-op.
```
domenic: Hmm, so how would this be used? It may need an example. Is the idea that e.g. fetch() would accept a `signal`, then _mutate_ its abort steps to abort the fetch? What if you pass the same signal in to multiple fetches? Maybe it should be a list of abort steps, so that consumers append to the list?

dom.bs
```diff
@@ -1419,6 +1419,103 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+<h3 id=controlling-promises>Controlling {{Promise}}s</h3>
+
+Though {{Promise}} objects don't have any built-in cancellation mechanism, many {{Promise}}-vending
+APIs require cancellation semantics. {{PromiseController}} is meant to support these requirements by
+providing an {{PromiseController/abort()}} method that toggles the state of a corresponding
+{{PromiseSignal}} object. The API which wishes to support cancellation can accept such a
+{{PromiseSignal}}, and use its state to determine how (not) to proceed.
+
+Upon {{PromiseController/abort()}}, the relevant {{Promise}} will reject with an
+<dfn exception>AbortError</dfn> [=simple exception=].
```
domenic: You can't just define new simple exceptions, but I know you are aware, per your comments about hand-waving. Still, something to take care of before this gets merged.

dom.bs
domenic: This actually isn't really specific to promises. For example it could be used to cancel a stream:

```js
const readableStream = getStream({ signal });

// ... later ...
signal.abort(); // basically the same as readableStream.cancel()
```

annevk: 1. Is using IDL fine?
2. How big can this possibly get?

dom.bs
annevk: I was thinking that Fetch would subclass these and define custom steps there. Having the steps be part of the base class helps in that you don't have to mutate most other things. How Fetch handles multiple fetches and such could be done in such a setup with a single set of abort steps.

Now, I don't know for sure that Fetch would subclass these. @jakearchibald should probably weigh in.

dom.bs
domenic: IDL should be fine. I can't imagine this getting much bigger; new spec was mostly in jest.

dom.bs
domenic: Hmm, I see. Fetch would subclass, but streams (and possibly whatever @mikewest wants this for originally) would not.

dom.bs
annevk: That is to say, if this remains fairly small long term, I don't like the idea of it being on its own. In DOM seems somewhat reasonable given that events are signals of sorts too, but it's pushing things a bit too of course.

And if this needs to be a non-IDL thing, perhaps Streams is a good place? It's also fits somewhat and Fetch has a dependency on that anyway, as will many other things.

domenic: Note that it has a dependency on HTML via EventHandler, somewhat amusingly.

dom.bs
annevk: But maybe you're right that we should design this more so that it even works for all scenarios when not subclassed. I do think that whatever this get passed to needs to be able to hook into it, since we don't really have specification-only event listeners (I think that would be a bigger thing to add).

dom.bs
mikewest: I'll defer to y'all about where you'd like this to live. I have no strong opinion, so long as it lives somewhere.

mikewest: `CancellationController`? That seems reasonable as a base class for `FetchController` if @jakearchibald needs such a thing.

dom.bs
mikewest: Google says "cancellation", probably because that's how the word is spelled. :nerd_face:

mikewest: I'm going to change everything to `cancel()`/cancellation. I used `abort()` because that's what folks in https://github.com/whatwg/fetch/issues/447 were saying, but I'm having trouble getting past "abortion"'s other common usage while typing it out.

dom.bs
mikewest: Yup. It feels a bit strange for DOM to be making this assertion about other specifications, but I agree with @annevk's suggestion that we should define the exception here.

dom.bs
mikewest: I'm happy to phrase this however you'd like. I'm not sure I'm up for submitting something to TC39, but perhaps there's a middle ground? :)

dom.bs
mikewest: Ok.

dom.bs
mikewest: Sure.

dom.bs
mikewest: Though, it's not clear to me why you'd use this model for streams when they have a built in cancellation mechanism. Do you think it's worth changing the streams API in line with what we have to do here for Promises?

domenic:
Oh, I forgot, DOM has already set a precedent for one "l" in cancelation/canceled/cancelable, per EventInit's cancelable. (The spec also consistently uses one "l" in spec-exposed concepts.)

dom.bs
domenic: I'm not sure if we have a middle ground between "use DOMException" and "write an ES-style spec". I guess the ES-style spec could be small though, basically just saying that it fulfills the [_NativeError_ Object Structure](https://tc39.github.io/ecma262/#sec-nativeerror-object-structure) with name X.

mikewest:
*sigh*

Ok. This is `referer` all over again. :)

dom.bs
domenic: But also I'm not sure if avoiding DOMException is worthwhile since we already have a dependency on EventTarget.

dom.bs
domenic: The idea would be to allow people to pass in the signal to many APIs, streams being just one. So it's OK if an API has both a dedicated cancel mechanism and also accepts the platform-generic one, IMO.

annevk:
A single l in canceled is not a misspelling however.

dom.bs
annevk: I guess that is fair.

kpaulh:
I took a look with regards to using this for WebAuthn, and I think it well meets our use case. I agree with domenic that the cancelation/cancel nomenclature is clearer than promise/abort - as odd as one 'l' looks to me right now :)

dom.bs
```diff
@@ -1419,6 +1419,107 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+</pre>
+<dl class=domintro>
+ <dt><code><var>signal</var> . </code>{{CancellationSignal/cancelled}}
+ <dd>Returns true if this {{CancellationSignal}} has been cancelled, and false otherwise.
+</dl>
+
+Each {{CancellationSignal}} has an <dfn for=CancellationSignal>cancelled flag</dfn> which is
+unset unless otherwise specified.
+
+Each {{CancellationSignal}} has an <dfn for=CancellationSignal>cancellation steps</dfn> algorithm
```
jyasskin: This might need to be passed in through the creation of the CancelattionController. We'll see what works in prose once FetchController is specified using this.

rbuckton:
@domenic I know you were looking at this for TC39. It feels like this is something that should be a ECMAScript built-in that has a consistent API regardless of host environment (e.g. no dependency on EventTarget). Also, for the purposes of comparison, [this is the version](https://github.com/rbuckton/prex/blob/master/docs/cancellation.md) I've been discussing with @bterlson for the last year.

domenic:
@rbuckton to repeat what I said in IRC, we tried to create a host-agnostic, ergonomic version in TC39 and failed, so we are no longer interested in pursuing that venue or approach. EventTarget is here to stay. Other hosts can copy, or not; we're not here to solve their problems, but instead to solve the web's.

mikewest:
I thought about the "cancelation steps" bits this morning, and it's not clear to me that they make sense if we want the signal to be observable by multiple things at once. That is, we would either need the "list of cancellation steps" that @domenic mentioned earlier, or we need to expose an event that the consumer-side spec would listen for. Aesthetically, I prefer the latter, though I realize that it's not really a concept that exists at the moment.

dom.bs
```diff
@@ -1419,6 +1419,107 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+ APIs that require more granular control could extend both {{CancellationController}} and
+ {{CancellationSignal}} according to their needs.
+</div>
+
+<h3 id=interface-cancellationcontroller>Interface {{CancellationController}}</h3>
+
+<pre class="idl">
+[Constructor(), Exposed=(Window,Worker)]
+interface CancellationController {
+  readonly attribute CancellationSignal signal;
```
mikewest: Done in https://github.com/whatwg/dom/pull/434/commits/36028dd814d2144acfd4efa88b36b66742208d79

annevk:
An ordered set seems fine. Whoever adds those steps should also make sure they're safe with respect to other steps.

ljharb:
Per conversation with @annevk on IRC: I'd like to see a cross-realm way to unforgeably determine if the rejection reason is a cancellation or not - comparing to a string property, or using `Object.prototype.toString.call` output, is not sufficient, and neither is `instanceof`. In other words, for all the same reasons `Array.isArray` exists - which does a brand check that can't be faked or tampered with - it'd be ideal to have something like `CancelException.is()` or similar.

jakearchibald:
I'm guessing `err.name == "AbortError"` isn't sufficient?

ljharb:
@jakearchibald nope, since `{ name: "AbortError" }` or `function AbortError() {}` would be a false positive.

bterlson:
@Domenic fwiw, my perception is that one attempt at cancellation was tried and failed for reasons that are not entirely clear (at least to me). If there is good rationale for why this proposal is better than a host agnostic one, I'm glad to hear it, but I'll be very sad if the only reason is political. There is no doubt that the TC39 proposal that was ultimately rejected is way more involved than this proposal. I believe that a cancellation token approach very similar to this proposal has always been most likely to succeed in TC39 and, as far as I can tell, has not been tried. We should try it. What am I missing?

Anyway, thinking out loud:

I'm intrigued that this proposal is very similar to cancellation tokens as implemented in other systems (and as discussed by Ron and others within TC39)... but not exactly. Eg. it doesn't seem to have much support for compositing cancellation signals. Has this been considered?

IMO I'd like to see that the thrown thing on cancellation does not inherit from "Error". Having a separate top-level object makes checking in some cases easier, perhaps, and gives us a clean object hierarchy to extend later. It's also similar to the Ruby object hierarchy where certain classes of errors are a separate branch of an inheritance tree and are not caught by the normal "catch" block. Doing something similar might make it easier in the future to add similar language-level integration. At the very least it could make pattern matching more clear (using strawman syntax, please forgive me, and where Error here is an identifier pattern that is essentially doing an instanceof check):

```
catch (s) {
  match (s) {
    |> Error: throw s;
    |> CancelSignal: handleCancel();
}
```

ajklein:
I was supportive of @jakearchibald getting something done specifically for fetch (since that's an established API which has been waiting on cancellation for years). But if we're going to go down the path of generalizing cancellation, I agree with @bterlson that a minimal, token-based approach in ECMAScript seems like the better route here for the long-term health of the ecosystem.

What I most want to avoid is producing _two_ general but incompatible cancellation primitives, one used by web APIs and one used in ES.

ljharb:
@annevk @jakearchibald 
Here are the concerns I think are relevant to "detecting if a reason is a cancellation":
 1. hard to screw up accidentally / no collisions: this is why a simple "name" property or similar isn't a good idea; if it's possible to conflict, it's going to cause somebody a problem. Here are some options that satisfy this constraint:
    1. (same realm) `instanceof CancelError`
    1. (cross-realm) cross-realm `Symbol.asInstance` method that does `instanceof` and falls back to a brand check (or just does the brand check)
    1. (cross-realm) new well-known Symbol property, like `[Symbol.isCancelError]: true`
    1. (cross-realm) brand-checking method, like `CancelError.is()`
 1. Cross-realm: values are passed across realms, be they iframes, node's `vm` module, or in the upcoming Realms API. It is done infrequently but nonzero now, but its use will likely go up over time. Here are some options that satisfy this constraint:
    1. (can collide) a string property, like "name", set to a primitive value, like "CancelError".
    1. (no collisions) cross-realm `Symbol.asInstance` method that does `instanceof` and falls back to a brand check
    1. (no collisions) new well-known Symbol property, like `[Symbol.isCancelError]: true`
    1. (no collisions) brand-checking method, like `CancelError.is()`
 1. Unforgeable: this is the constraint that often gets pushback on TC39. There are two classes of concerns here: an "own" lie, or malicious tampering.
    - With "own lie", there's two issues: something claiming to be a CancelError when it's not; and something claiming an individual CancelError is not one, when it is. To be fair, these only really matter if a CancelError has special branded behavior elsewhere.
   - With "malicious tampering", there are also two issues: malicious code making all "somethings" (all `Number`s, all `String`s, all `Error`s, whatever) claim to be a CancelError when they're not; and malicious code making all CancelErrors claim to *not* be CancelErrors when they are. These matter in two cases: one, if there's special branded behavior elsewhere; two, when trying to protect library code against global interference/tampering (tampering done after library parse time, of course)
   - Here are some options that satisfy this constraint:
    1. brand-checking method, like `CancelError.is()`
    1. (cross-realm) cross-realm `Symbol.asInstance` method that does `instanceof` and falls back to a brand check (or just does the brand check)

Here's a matrix of these options and constraints:

… | No Collisions | Cross-Realm | Unforgeable
------------ | ------------- | ------------- | -------------
string property | ❌  | ✅  | ❌ 
`instanceof` | ✅ | ❌ | ❌ 
`Symbol.isCancelError` | ✅   | ✅  | ❌ 
`Symbol.asInstance` | ✅ | ✅ | ✅ 
`CancelError.is()` | ✅ | ✅ | ✅

bzbarsky:
For what it's worth, using words with multiple commonly accepted spellings in APIs is very error-prone.  Note how the title of this PR uses "Cancellation" but the PR itself mostly uses "cancelation"...

If you _are_ going to pick this name, as far as I can tell "cancellation" is far more popular (e.g. has 10x the number of Google search result hits and is the primary spelling in the dictionaries I have to hand right now), so probably you want the "ll" form.  But again, I think this name is best avoided.

annevk:
TC39: tokens is actually incompatible with what we want for Fetch. We need objects that we can extend. I'm happy with TC39 uplifting what we come up with here (and leaving events and maybe the exception host-defined), but I don't want to yet again wait for TC39. I'm also happy to take any feedback into account that makes such uplifting easier.

Naming: I'm happy to rename this AbortController / AbortSignal, especially if we can reuse "AbortError" as @domenic pointed out in the IDL PR. Any reason we didn't go with that?

Branding: I think this is something TC39 should solve for all objects. Once they do, we'll have it here too. See also https://github.com/heycam/webidl/issues/129.

Composing cancellation signals: @bterlson can you provide an example?

Error versus a new type: it seems if you just reversed the two classes in @bterlson's example and check for abort/cancel first you'd be fine, regardless of base class.

(Note that I haven't checked the above with @jakearchibald and @mikewest, who are mostly driving the work here.)

ljharb:
@annevk TC39 has solved it for all objects (except Error) - all objects except Error have a method somewhere that does a brand check (and either throws or returns a boolean). That's what I'm requesting HTML do too.

annevk:
@ljharb what's the cross-realm instanceof for a Map object? Is there a pattern we can follow for platform objects?

ljharb:
@annevk there's lots of them - `function isMap(value) { try { Map.prototype.has.call(value); return true; } catch (e) { return false; } }` works fine ([link](https://tc39.github.io/ecma262/#sec-map.prototype.has))

However, I'd encourage you to follow the pattern of `Array.isArray`, since having to catch exceptions is unergonomic and slow.

annevk:
@ljharb in that case it seems good if TC39 solved it for its all objects, including error, in that way (i.e., like `isArray()`) too. Just add a method that can be used for this.

ljharb:
@annevk as i said, it is solved, just not in a generic way - and the committee isn't interested (sadly) in solving it in a generic way. The method already exists - hosts can check for the presence of internal slots, and do whatever they want. They can even create their own internal slots and check for the presence of those.

jakearchibald:
Firstly, huge thanks to @mikewest for getting this together!

Naming: I'm happy with using "abort" rather than "cancel" if it gets us out of the naming issues. I personally find "cancellation" the more natural spelling, but I'd totally just deal with it (like I do with US vs UK spellings).

Error type: I don't really see the issue with using `AbortError`, especially if we change the naming as above.

Should the `signal` be transferrable? That would help explain how it interacts with streams and service workers. Although, since a signal can be given to multiple consumers, the APIs would need to create their own signal that mimics the signal they're given, and transfer that. I guess we can do some hand-waving here and deal with full transferring later.

@slightlyoff you didn't like the naming of "signal" right? Speak now or forever hold your peace.

jyasskin:
`CancelController` and `CancelError` might also be options.

@annevk: @rbuckton's https://github.com/rbuckton/prex/blob/master/docs/cancellation.md#new-cancellationtokensourcelinkedtokens has the `CancellationTokenSource` constructor take a list of `CancellationTokens` such that if any of those tokens is cancelled, the new source also gets cancelled. That's composition, right? @bterlson, is it necessary to build this into the API, or can users just call `token.register(() => source.cancel())`?

rbuckton:
@jyasskin: The semantics are a bit more complex, as it also handles registration of parts of the graph when the source is closed (reduces memory costs as callbacks can be garbage collected). It greatly simplifies a common pattern for moderate to advanced scenarios, but is only really useful if you have a shared and consistent cancellation API across many scenarios in an application.

It is also more consistent to have the API support linked tokens if `cancel` runs in a different turn and reports exceptions via a Promise as 'prex' does, as that adds additional complexity to have to write manually.

bzbarsky:
> Just add a method that can be used for this and can be extended by hosts.

Note that the behavior of instanceof can be extended by hosts, or indeed by script.  It's just that everyone objects to actually doing it, apparently.

annevk:
@jyasskin thanks for clarifying. That seems like something we could add later; the ability to construct a signal from multiple other signals.

@rbuckton what is the reason you went with delayed notification ("promise timing")?

@bzbarsky from https://twitter.com/sebmarkbage/status/851158743555358720 it seems there's at least some interest left to further explore this. The main argument I heard against is that if folks start using this it makes it harder to test things as mock objects could no longer be used.

rbuckton:
@annevk: originally it was not delayed, though I need to find the mail thread that convinced me to change it. A related thread about cancellation can be [found here](https://esdiscuss.org/topic/cancellation-architectural-observations).

rbuckton:
Actually, that is the thread. I specifically [switch to delayed notification](https://esdiscuss.org/topic/cancellation-architectural-observations#content-28) as a result of that discussion.

annevk:
@rbuckton thanks, that's an interesting scenario. Let me try to sketch it out in terms of the API we have:
```js
const controller = new AbortController,
      signal = controller.signal
signal.onabort = abortion
fetch("https://test:test/", { signal }).then(..., rejection)
controller.abort()
```
The `fetch()` call directly rejects, with `rejection` being invoked at the end-of-task. `abortion` however is invoked directly after `controller.abort()`, before `rejection`. A solution to that would be to `.then()` the promise generated by the method to which _signal_ is passed and only once that rejects with an "AbortError", flip the state and dispatch the `abort` event.

Over in https://github.com/whatwg/fetch/issues/447 folks wanted to be able to synchronously observe when `abort()` was invoked, though I suppose if that's actually a strong enough use case that could also be solved by the controller.

annevk:
The above solution doesn't work of course once you pass the signal to multiple fetches (well, you'd get some kind of Promise.race thingy I suppose, whoever flips the state first wins).

dom.bs
```diff
@@ -1419,6 +1423,114 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+concepts require cancelation semantics. {{CancelationController}} is meant to support these
+requirements by providing an {{CancelationController/cancel()}} method that toggles the state of a
+corresponding {{CancelationSignal}} object. The API which wishes to support cancelation can accept
+such a {{CancelationSignal}}, and use its state to determine how (not) to proceed.
+
+APIs that rely upon {{CancelationController}} are encouraged to respond to
+{{CancelationController/cancel()}} by rejecting any unsettled {{Promise}} with a new
+{{DOMException}} with [=error name=] "{{CancelationError}}".
+
+<div class=note>
```
domenic: `<div class=example id=...>` not `<div class=note>`

annevk:
I chatted with @domenic on IRC about the event and it was clarified to me that the primary purpose of the signal event is as a request to JavaScript-implemented signal consumers to please cancel yourself. The event is not necessarily useful for those using the `fetch()` API, so the scenario above is invalid.

We could clarify the purpose of the event by renaming it to "abortrequest" or some such.

jakearchibald:
I've branched this at https://github.com/whatwg/dom/pull/437 as @mikewest doesn't have time to continue working on it.

annevk:
I'm closing this per the above comment.

PR #435, opened 2017-04-07
Be more precise about removal and insertion

### LINKED ISSUE #174: atomic tree actions
The spec need atomic actions for insert append and  replace into a tree. Should describe what it means in terms of child nodes and references to/from previous next sibling, with potential update of the parent first and last child. Then these can be references from insert and replace mutation algorithms. For the time being, these atomic actions are undefined, at least in the spec.

annevk:
In particular, define them as operations on an object’s children, which
is now defined to be an ordered set.

Fixes #174. (Note that we don’t have to define how the various pointers
mentioned there are updated, since those are all ultimately based on
how parent/child are defined.)

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/tree-manipulation/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/6ba16e7...9a7e1ea.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -124,18 +124,18 @@ can be written that hooks into the provided extensibility hooks for
 
 <h3 id=trees>Trees</h3> <!-- Sorry reddit, this is not /r/trees -->
 
-A <dfn export id=concept-tree>tree</dfn> is a finite hierarchical tree structure. In
-<dfn export id=concept-tree-order>tree order</dfn> is preorder, depth-first
-traversal of a <a>tree</a>.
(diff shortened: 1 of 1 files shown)


domenic:
Actualy, when you're ready to merge this, can you merge https://github.com/whatwg/resources.whatwg.org/pull/52 first, and use this as a test that the changes don't break anything?

annevk:
I'd prefer you merge it then since I don't want to be on the hook for restoring the build system, if that's okay.

domenic:
Seems to have worked nicely, yay.

PR #436, opened 2017-04-10
Hide `nonce` content attribute values from non-script sources.

mikewest:
This patch extracts the `nonce` content attribute out to a generic
definition in DOM, rather than an HTMLScriptElement-specific definition
in HTML, and defines new behavior for insertion and cloning with the
intent of reducing the risk of side-channel leakage of the nonce's
value.

The nonce value is extracted from the content attribute when the element
is inserted into the DOM, and put into an internal property. The
content attribute's value is set to the empty string.

From then on, the property's value and the content attribute's value are
disconnected; alterations to one have no effect on the other, and
vice-versa.

The nonce's value is available to script via the `nonce` IDL attribute,
and so can be propagated just as today.

Addresses whatwg/html#2369.

***
[Preview](https://s3.amazonaws.com/pr-preview/mikewest/dom/nonce-hiding.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/4ba35d9...mikewest:78db5fb.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5776,8 +5776,8 @@ This and <a lt="other applicable specifications">other specifications</a> may de
 
 <ol>
  <li>
-  <p>If <var>inserted</var> is [=connected=], and it has a content attribute (<var>attr</var>) whose
-  value is not the empty string, then:</p>
+  <p>If <var>inserted</var> is [=connected=], and it has a content attribute (<var>attr</var>) named
+  [=Attr/local name=] "<code>nonce</code>" whose [=Attr/value=] is not the empty string, then:</p>
 
   <ol>
    <li>Let <var>nonce</var> be <var>attr</var>'s [=Attr/value=].</li>
--- a/dom.bs
+++ b/dom.bs
@@ -5776,8 +5776,9 @@ This and <a lt="other applicable specifications">other specifications</a> may de
 
 <ol>
  <li>
-  <p>If <var>inserted</var> is [=connected=], and it has a content attribute (<var>attr</var>) named
-  [=Attr/local name=] "<code>nonce</code>" whose [=Attr/value=] is not the empty string, then:</p>
+  <p>If <var>inserted</var> is [=connected=], and it has a content attribute (<var>attr</var>)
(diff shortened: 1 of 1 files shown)


mikewest:
Is this more or less what you were thinking, @annevk/@zcorpan?

mikewest:
(If so, I'll update the HTML patch to rely on this one. If not, we should chat a bit. :) )

dom.bs
```diff
@@ -5769,6 +5772,33 @@ This and <a lt="other applicable specifications">other specifications</a> may de
 <a for=/>elements</a>. The algorithm is passed <var>element</var>, <var>localName</var>,
 <var>oldValue</var>, <var>value</var>, and <var>namespace</var>.
 
+<p>The [=insertion steps=] for an {{Element}} (<var>inserted</var>) are as follows:
+
+<ol>
+ <li>
+  <p>If <var>inserted</var> is [=connected=], and it has a content attribute (<var>attr</var>) whose
```
annevk: Shouldn't this actually name the attribute in question?

mikewest: It should indeed. Fixed in https://github.com/whatwg/dom/pull/436/commits/3e4bffb3b35516c3fdfe55f4965bc869500208db.

dom.bs
```diff
@@ -5776,8 +5776,8 @@ This and <a lt="other applicable specifications">other specifications</a> may de
 
 <ol>
  <li>
-  <p>If <var>inserted</var> is [=connected=], and it has a content attribute (<var>attr</var>) whose
-  value is not the empty string, then:</p>
+  <p>If <var>inserted</var> is [=connected=], and it has a content attribute (<var>attr</var>) named
+  [=Attr/local name=] "<code>nonce</code>" whose [=Attr/value=] is not the empty string, then:</p>
```
annevk: You want to use https://dom.spec.whatwg.org/#concept-named-attribute since this is still not precise enough.

mikewest: Trying again in https://github.com/whatwg/dom/pull/436/commits/78db5fb120d10477b70f3f570ebea88def16d4bd. :)

PR #437, opened 2017-04-11
Abort controller

jakearchibald:
Branched from https://github.com/whatwg/dom/pull/434

***
[Preview](https://s3.amazonaws.com/pr-preview/jakearchibald/dom/abort-controller.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/8f8c1c3...jakearchibald:0b8cce3.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -13,7 +13,7 @@ No Editor: true
 !Tests: <a href=https://github.com/w3c/web-platform-tests/tree/master/dom>web-platform-tests dom/</a> (<a href=https://github.com/w3c/web-platform-tests/labels/dom>ongoing work</a>)
 !Translation (non-normative): <span title=Japanese><a href=https://triple-underscore.github.io/DOM4-ja.html lang=ja hreflang=ja rel=alternate>日本語</a></span>
 Logo: https://resources.whatwg.org/logo-dom.svg
-Abstract: DOM defines a platform-neutral model for events and node trees.
+Abstract: DOM defines a platform-neutral model for events, aborting activities, and node trees.
 Ignored Terms: EmptyString, Array, Document
 Boilerplate: omit feedback-header, omit conformance
 </pre>
@@ -557,7 +557,7 @@ algorithm below.
  the operation that caused <var>event</var> to be <a>dispatched</a> that it needs to be canceled.
 
  <dt><code><var>event</var> . {{Event/defaultPrevented}}</code>
- <dd>Returns true if {{Event/preventDefault()}} was invoked successfully to indicate cancellation,
+ <dd>Returns true if {{Event/preventDefault()}} was invoked successfully to indicate cancelation,
  and false otherwise.
 
  <dt><code><var>event</var> . {{Event/composed}}</code>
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -660,7 +660,7 @@ flags that are all initially unset:
 <ul>
  <li><dfn export for=Event id=stop-propagation-flag>stop propagation flag</dfn>
  <li><dfn export for=Event id=stop-immediate-propagation-flag>stop immediate propagation flag</dfn>
- <li><dfn export for=Event id=canceled-flag>canceled flag</dfn>
+ <li><dfn export for=Event id=canceled-flag>aborted flag</dfn>
```
annevk: This and similar changes to Events will need to be reverted. Simple search & replace is too simple unfortunately.

dom.bs
```diff
@@ -1419,6 +1419,134 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+ <dt><code><var>signal</var> . </code>{{AbortSignal/aborted}}
+ <dd>Returns true if this {{AbortSignal}} has been aborted, and false otherwise.
+</dl>
+
+Each {{AbortSignal}} has an <dfn for=AbortSignal>aborted flag</dfn> which is unset unless otherwise
+specified.
+
+Each {{AbortSignal}} has <dfn for=AbortSignal>abort callbacks</dfn>, which is an [=ordered set=] of
+algorithms which are to be executed when its [=AbortSignal/aborted flag=] is set. Unless otherwise
+specified, its value is the empty set.
```
annevk: Can we name this something different from callbacks. Callback already has a very specific meaning in IDL.

annevk:
Shall we rename the event to "abortrequest" to more clearly indicate it's a hook that needs to be implemented and not a hook that will tell you if something actually happened? See https://github.com/whatwg/dom/pull/434#issuecomment-292869591 onward for rationale.

dom.bs
jakearchibald: Oh ffs, sorry. I should have paid more attention to the patch.

jakearchibald:
Does "abortrequest" sound too much like "abort the network request"? What about "signalabort", along with "signalprioritychange" etc etc

annevk:
Given the choice, I'd favor sticking with "abort" and explaining what it means in a note or so (that it's a way for functions that accept signals to hook into the aborting mechanism).

dom.bs
```diff
@@ -1419,6 +1419,142 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+ <dt><code><var>controller</var> . </code>{{AbortController/signal}}
+ <dd>Returns the {{AbortSignal}} object associated with this object.
+
+ <dt><code><var>controller</var> . </code>{{AbortController/abort()}}
+ <dd>Invoking this method will set this object's {{AbortSignal}}'s [=AbortSignal/aborted flag=],
+ thereby signaling to any observers that the associated activity should be aborted.
+ </dl>
+
+The <dfn attribute for=AbortController><code>signal</code></dfn> attribute must return the value to
+which it was initialized. When a {{AbortController}} is created, the attribute must be initialized
```
jyasskin: "an AbortController"

dom.bs
```diff
@@ -1419,6 +1419,142 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+to a newly created {{AbortSignal}} object.
+
+The <dfn method for=AbortController><code>abort()</code></dfn> method, when invoked, must <a
+for=AbortSignal>signal abort</a> on this object's {{AbortController/signal}}.
```
jyasskin: You can probably write the link as `[=AbortSignal/signal abort=]`, although I haven't tested that with DOM's build system.

dom.bs
```diff
@@ -1419,6 +1419,142 @@ that gave folks all the wrong ideas. <a>Events</a> do not represent or cause act
...
+</div>
+
+<h3 id=interface-abortcontroller>Interface {{AbortController}}</h3>
+
+<pre class="idl">
+[Constructor(), Exposed=(Window,Worker)]
+interface AbortController {
+  [SameObject] readonly attribute AbortSignal signal;
+
+  void abort();
```
jyasskin: I agree with your choice to return `void` here. In particular, I expect that returning ["a Promise that is settled when cancellation has completed"](https://github.com/rbuckton/prex/blob/master/docs/cancellation.md#sourcecancel) will encourage developers to miss the fact that "an API listening to a signal may choose to ignore it" and so write bugs.

dom.bs
jakearchibald: @annevk what's your take on shortcuts like this? I use them in the service worker spec and find them easier to read/write, but then I lean heavily on the markdown stuff too. Do you have a preference for this spec?

annevk: I mostly use HTML (but also {{object}}) and I actually prefer explicit `<p>` starting tags too, but I figured I could add those once we're good to go.

I wouldn't particularly mind if you used such shortcuts though. I do mind newlines inside tags or phrasing-content elements.

dom.bs
rbuckton: @jyasskin The result of `cancel()` in prex is to observe any errors thrown in the registered callback (or here, the onabort event), and respond accordingly. In prex, registered callbacks are not evaluated immediately and as a result the only way to observe the exception is through a Promise. A Promise return wouldn't strictly be necessary if `cancel()` ran to completion synchronously.

annevk:
#438 is the corresponding issue.

annevk:
I addressed all nits I had I think. I also posted a design for serializing and deserializing `AbortSignal`  objects over in #438.

dom.bs
```diff
@@ -1413,6 +1413,152 @@ can only be used to influence an ongoing one.
...
+ <dt><code><var>controller</var> = new <a constructor lt="AbortController()">AbortController</a>()</code>
+ <dd>Returns a new <var>controller</var> whose {{AbortController/signal}} is set to a newly
+ created {{AbortSignal}} object.
+
+ <dt><code><var>controller</var> . </code>{{AbortController/signal}}
+ <dd>Returns the {{AbortSignal}} object associated with this object.
+
+ <dt><code><var>controller</var> . </code>{{AbortController/abort()}}
+ <dd>Invoking this method will set this object's {{AbortSignal}}'s [=AbortSignal/aborted flag=],
+ thereby signaling to any observers that the associated activity should be aborted.
```
domenic: Nit: I think this would be clearer as "... set this AbortSingal's aborted flag, and signal to any observers..." (instead of using "thereby")

dom.bs
```diff
@@ -1413,6 +1413,152 @@ can only be used to influence an ongoing one.
...
+</dl>
+
+<p>An {{AbortController}} object has an associated <dfn for=AbortController>signal</dfn> (an
+{{AbortSignal}} object). When an {{AbortController}} object is created, its
+<a for=AbortController>signal</a> must be set to a new {{AbortSignal}} object.
```
domenic: Do you want to define the constructor more explicitly, perhaps?

dom.bs
```diff
@@ -1413,6 +1413,152 @@ can only be used to influence an ongoing one.
...
+</ol>
+
+
+<h3 id=abortcontroller-api-integration>Using {{AbortController}} and {{AbortSignal}} objects in
+APIs</h3>
+
+<p>Any web platform API using promises that to represent operations that can be aborted must accept
+{{AbortSignal}} objects, use the [=AbortSignal/abort algorithms=] mechanism, and convey that the
+operation got aborted by rejecting the promise with the "{{AbortError}}" {{DOMException}}.
+
```
domenic: It might help to also suggest how the signal is accepted, so that we don't sometimes name it `signal` and sometimes `abortSignal`. Based on the example it sounds like `signal` is the way to go.

dom.bs
```diff
@@ -1413,6 +1413,152 @@ can only be used to influence an ongoing one.
...
+<p>To <dfn export for=AbortSignal>signal abort</dfn>, given a {{AbortSignal}} object
+<var>signal</var>, run these steps:
+
+<ol>
+ <li><p>If <var>signal</var>'s [=AbortSignal/aborted flag=] is set, then return.
+
+ <li><p>Set <var>signal</var>'s [=AbortSignal/aborted flag=].
+
+ <li><p><a for=set>For each</a> <var>algorithm</var> in <var>signal</var>'s
+ [=AbortSignal/abort algorithms=]: run <var>algorithm</var>.
```
domenic: Probably you want to clear the abort algorithms here

annevk: If we do that, we probably shouldn't allow adding new abort algorithms either if the aborted flag is set. Should we have a dedicated add algorithm folks can use?

dom.bs
annevk: Yeah, that sounds like a good idea. Currently it's not defined at all.

dom.bs
domenic: Hmm yeah, good point. We could do that, or we could go with a different strategy, of noting that abort algorithms will never fire after the aborted flag is set, so even though the spec doesn't clean them up explicitly, implementations probably should. I guess that's not as good though, once I typed it out.

annevk:
Addressed the remaining feedback and also created tests.

It seems we should land this together with Streams (maybe Streams a day later for Shepherd) and then file browser bugs for the combined feature. So each bug would point to this change and the change in Streams, and the tests for this change and the tests for the change in Streams.

Does that sound okay to everyone?

dom.bs
```diff
@@ -1564,9 +1564,17 @@ them. For instance, if the operation has already completed.
...
-<p>Any web platform API using promises that to represent operations that can be aborted must accept
-{{AbortSignal}} objects, use the [=AbortSignal/abort algorithms=] mechanism, and convey that the
-operation got aborted by rejecting the promise with the "{{AbortError}}" {{DOMException}}.
+<p>Any web platform API using promises to represent operations that can be aborted must adhere to
+the following:
+
+<ul class=brief>
+ <li>Accept {{AbortSignal}} objects through a <code>signal</code> dictionary member.
+ <li>Use the [=AbortSignal/abort algorithms=] mechanism.
+ <li>Convey that the operation got aborted by rejecting the promise with the "{{AbortError}}"
```
domenic: Nit: an "AbortError" DOMException, not the "AbortError" DOMException.

dom.bs
```diff
@@ -1534,6 +1534,16 @@ unless specified otherwise.
 <a for=/>set</a> of algorithms which are to be executed when its [=AbortSignal/aborted flag=] is
 set. Unless specified otherwise, its value is the empty set.
 
+<p>To <dfn for=AbortSignal>add</dfn> an algorithm <var>algorithm</var> to an {{AbortSignal}} object
```
domenic: `export` please

dom.bs
```diff
@@ -1554,8 +1564,16 @@ them. For instance, if the operation has already completed.
...
- [=AbortSignal/abort algorithms=]: run <var>algorithm</var>.
+ <li>
+  <p><a for=set>For each</a> <var>algorithm</var> in <var>signal</var>'s
+  [=AbortSignal/abort algorithms=]:
+
+  <ol>
+   <li><p>Run <var>algorithm</var>.
+
+   <li><p>Remove <var>algorithm</var> from <var>signal</var>'s
+   <a for=AbortSignal>abort algorithms</a>.
```
domenic: Mutating a list as you loop through it, although it's probably well-defined, can be confusing, and easy to implement correctly. I'd suggest just clearing the list after the loop.

domenic:
> It seems we should land this together with Streams (maybe Streams a day later for Shepherd) and then file browser bugs for the combined feature. So each bug would point to this change and the change in Streams, and the tests for this change and the tests for the change in Streams.
>
> Does that sound okay to everyone?

I think this will be more compelling if we could also land it together with aborting fetch(). However, I realize you are leaving for vacation soon so that might not work out timeline-wise.

Note also that only Blink has a pipeTo implementation. (Safari has one, but it is outdated and relies on an outdated WritableStream implementation, so effectively it doesn't really work...) So only Blink would be able to implement the streams-side change.

@yutakahirano @tyoshino @ricea for an idea on whether implementing this in order to cancel streams and/or fetch is a good near-term goal.

dom.bs
```diff
@@ -1413,6 +1413,184 @@ can only be used to influence an ongoing one.
...
+
+<dl class=domintro>
+ <dt><code><var>signal</var> . <a attribute for=AbortSignal>aborted</a></code>
+ <dd>Returns true if this {{AbortSignal}} has been aborted, and false otherwise.
+</dl>
+
+<p>An {{AbortSignal}} object has an associated <dfn for=AbortSignal>aborted flag</dfn>. It is unset
+unless specified otherwise.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>abort algorithms</dfn>, which is a
```
domenic: Hmm. This isn't exported, so I can't remove my algorithm from it. Maybe we should define "remove an algorithm from an AbortSignal" as well. Then it can stay un-exported.

annevk: If removing doesn't have any conditions I'll just export this instead.

annevk:
I can look into aborting fetch() to at least get an idea of what's needed, but I thought the plan was that we'd sort through the remainder of that in July, whereas we'd want to do this quickly to avoid blocking Streams.

dom.bs
domenic: I don't think it does, but it's kind of nice to say that the internal list of abort algorithms is an implementation detail, and other specs only interact by adding and removing. E.g. this way other specs cannot empty the list of algorithms, or reshuffle them.

domenic:
Yeah, that's fair; upon re-reviewing the threads I forgot how complicated the plans were for fetch.

dom.bs
annevk: In Fullscreen we want the other way. Maybe we should revisit that.

cc @foolip

foolip: @annevk, you mean https://github.com/whatwg/fullscreen/pull/79 + https://github.com/whatwg/html/pull/2650, right?

Whether to expose the data structure directly or wrap in a set of abstract operations I think won't have the same answer every time, but I suppose the more different specs that interact with a thing the more I'd lean towards abstract operations.

annevk: Yeah, I guess it's okay for it to be a little different.

annevk:
I looked into https://github.com/whatwg/fetch/pull/523 and while it's somewhat close if you don't stress the somewhat horrible "prose abort API" (terminate a fetch), there's a few things left to figure out, including service workers. That's not going to happen within a week. So it's either this and Streams or just let it all be for another 5-6 weeks.

dom.bs
```diff
@@ -1413,6 +1413,188 @@ can only be used to influence an ongoing one.
...
+<a for=/>set</a> of algorithms which are to be executed when its [=AbortSignal/aborted flag=] is
+set. Unless specified otherwise, its value is the empty set.
```
jakearchibald: I'm not too familiar with the spec language here, but this sounds like it encourages other specs to replace the set, but in reality we only want other things to add/remove item.

Should this say something like "The set is initially empty"?

jakearchibald: Following the above discussion, I guess other specs can't replace it.

dom.bs
jakearchibald: FWIW, I agree with abstract operations here. Prevents other specs setting "abort algorithms" etc etc.

jakearchibald:
I've added some example spec text in https://github.com/whatwg/dom/pull/437/commits/51b8ca159cc2ce4fc20fab962bbe70ca314ffee2 which could do with a review.

dom.bs
```diff
@@ -1413,6 +1413,212 @@ can only be used to influence an ongoing one.
...
+
+<div class=example id=aborting-ongoing-activities-spec-example>
+ The steps for a promise-returning method <code>doAmazingness(options)</code> could be as follows:
+
+ <ol>
+  <li>Let |p| be [=a new promise=].
+  <li>If |options|' <code>signal</code> member is present, then:
+   <ol>
+    <li>If |options|' <code>signal</code>'s [=AbortSignal/aborted flag=] is set, throw an
+    "{{AbortError}}" {{DOMException}}.
```
jakearchibald: Is it worth adding a note here to point out this will be turned into a rejection, or is that assumed knowledge?

jyasskin: I think it's not as widely appreciated as it should be and could use a note. Folks tend not to know about the "If op has a return type that is a promise type" part of https://heycam.github.io/webidl/#dfn-create-operation-function.

Or you could be less subtle: wait to create *p* until you need to store it or go *in parallel*, and "return [a promise rejected with](https://www.w3.org/2001/tag/doc/promises-guide#a-promise-rejected-with) ..." here.

domenic: IMO relying on automatically turning things into a rejection is a spec bug that Web IDL catches for you. We should go with @jyasskin's suggestion instead.

jakearchibald: Cheers! I've added it as a note

jakearchibald: @domenic ohhh really? In JS I consider:

```js
async function blah() {
  return Promise.reject(Error('broke'));
}
```

…an anti-pattern compared to:

```js
async function blah() {
  throw Error('broke');
}
```

I was applying the same logic to specs. Are specs just different, or am I wrong when it comes to JS too?

domenic: Specs are different; we need to be more explicit.

jakearchibald: Fair enough. Will change.

annevk:
I cleaned up your example text a little bit with a more preferred pattern (we'd rather return at the end, since that signals the end of an algorithm).

It's probably still good to wait with landing this until either Streams or Fetch is ready to go (and has tentative implementer commitments).

If everything is okay now I might squash the commits and update the commit message to point to the tests.

(Also, if we don't do anything Mike will be the author of the final commit. That's okay with me, but I'd also be fine with changing it to Jake. The commit message itself already acknowledges the joint effort.)

domenic:
Streams has Blink commitments at least.

jyasskin:
@kpaulh / @equalsJeffH if this lands, would you expect to get FF/Edge/Blink commitments for it in the first version of WebAuthn, or only in the second?

jakearchibald:
@annevk I don't mind who's name is on the commit.

annevk:
The commit is ready to go though further review appreciated. Firefox would likely first ship aborting `fetch()` as we don't have writable or transform streams. So that seems like the easiest MVP unless readable streams can use this as well.

annevk:
It seems that for `fetch()` we're leaning towards exposing a copy of the signal that is passed to `Request`. Should we add advice to that effect here as a general pattern for exposing signals?

jakearchibald:
Some additional info:

We expose signals on request objects so they're (eventually) visible within a service worker, and signal the page's intent to abort the request.

We use a copy because we can't use a reference once the request goes to a service worker. Also, we have plans to introduce a `FetchSignal` later which includes other fetch-specific signals, such as priority changes. By specifying a copy we avoid having request objects with signals that are sometimes null, and sometimes different types.

domenic:
From that description the pattern doesn't seem generally applicable to me.

kpaulh:
My expectation is that this would be part of the second version of WebAuthN. @equalsJeffH do you agree?

annevk:
@domenic or I will likely land this later today, given that writable streams has enough commitments to use it. (Cancelable `fetch()` is getting closer too, at least in terms of desired behavior and tests, but is not as far along yet.)

So if there's any concerns left now would be the time.

annevk:
Sigh. Because I edited the commit before landing it's now attributed to Jake and not Mike. Oh well, new objects everyone!

PR #446, opened 2017-04-24
Fix Web IDL extended attribute usage for recent changes

domenic:
In https://github.com/heycam/webidl/pull/286 we updated Web IDL so that
the extended attribute [TreatNullAs] now apply to types, not attributes.
This updates the spec's usage of [TreatNullAs] to conform to this.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/treatnullas-update/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/26faf9a...899def2.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6677,7 +6677,7 @@ string <var>value</var>, run these steps:
 <pre class=idl>
 [Exposed=Window]
 interface CharacterData : Node {
-  [TreatNullAs=EmptyString] attribute DOMString data;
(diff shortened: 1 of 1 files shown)


annevk:
My main thought was that it would be nicer to wait with this until you've renamed this extended attribute upstream as you said you would.

domenic:
I decided against doing that because if we're going to start renaming extended attributes, there's about 5 of them I'd like to rename at once.

PR #449, opened 2017-04-26
Make element and attribute names behave like the HTML parser

domenic:
@dominiccooney has started gathering data on this in Blink, so I wanted to help the process along by preparing a spec PR. We should not merge this yet.

The biggest open question, which we could use some help on, is what to do with the -NS variants, createElementNS and createAttributeNS/setAttributeNS.

One idea I thought of is that "validate and extract" should delegate to the HTML validate-and-canonicalize algorithms if the namespace passed in is the HTML namespace.

Another idea is that we should just leave them alone. But I think the same frameworks that are asking for this change for HTML elements probably also want it for SVG elements. (I'm surprised SVG elements care what namespace their attributes are in...)


annevk:
> The biggest open question, which we could use some help on, is what to do with the -NS variants, createElementNS and createAttributeNS/setAttributeNS.

Those are the variants you would actually use internally, as you don't want to use setAttribute() which is rather odd when it comes to namespaces. 

> (I'm surprised SVG elements care what namespace their attributes are in...)

HTML elements care too. Say more.

domenic:
> Those are the variants you would actually use internally, as you don't want to use setAttribute() which is rather odd when it comes to namespaces.

What do you mean "internally"? Looking at e.g. [Ember](https://github.com/glimmerjs/glimmer-vm/blob/7708feacbd3c7b72ec725283b365760a0e141fd4/packages/%40glimmer/runtime/lib/dom/helper.ts#L107) or [React](https://github.com/facebook/react/blob/b1768b5a48d1f82e4ef4150e0036c5f846d3758a/src/renderers/dom/shared/DOMPropertyOperations.js#L158-L167) it seems like everyone branches on whether they're supplied a namespace by their caller (presumably only for SVG cases) and uses setAttribute instead.

> HTML elements care too. Say more.

I would have thought that after all the trouble we went through to try to simplify Attrs, at least one of the changes that might have stuck was making their namespace APIs have no effect (but still be around for back-compat). I guess not.

annevk:
I mean browsers. Browsers definitely care about the namespace of an attribute. And the problem with `setAttribute()` is that it doesn't care, so you might be accidentally overwriting a value of a namespaced attribute, while you were just trying to set a non-namespaced attribute. (This is only a problem for frameworks of course, and only for those that care about usually theoretical scenarios.)

domenic:
I'm still a little unclear on what you mean by "care". They store it, so it can be reflected in other DOM APIs, sure. But e.g. if I set the `src=""` attribute on a script element using `setAttribute("src", "...")` (which sets a null namespace, instead of the HTML namespace) it seems like browsers still load the script file.

Basically my confusion is about whether anyone ever cares about setting the namespace of an attribute "correctly", such as when trying to set two attributes with the same name but different namespace. It seems like the fact frameworks provide methods that take a namespace means some people care. I am not sure why, given my above example of browsers not caring about the namespace for `<script src>`, although I suspect it has something to do with SVG.

annevk:
Why do you think it should be the HTML namespace? (Hint: if you use the HTML namespace for the "src" attribute, it won't load.)

domenic:
Ah, very interesting. Browsing https://developer.mozilla.org/en-US/docs/Web/SVG/Namespaces_Crash_Course, it appears that people use namespaces for `xlink:href` mostly.

This makes me lean toward not caring about the -NS methods for attributes at all as people should mostly only have to use them for `xlink:href`, which [is going away](https://twitter.com/dstorey/status/700425693524414467). They can just use the null namespace for all attributes most of the time, which means only using createAttribute/setAttribute.

I guess that still leaves createElementNS though, and we maybe should make the attribute-NS methods consistent with that. So, what should our answer be for createElementNS?

annevk:
Given that `createElement()` depends on context for creating HTML elements (and cannot create SVG elements) it seems unsuitable as well. So I think we definitely want to make `createElementNS()` work, if anything.

dominiccooney:
My $0.02, I think you do want createElementNS to use the HTML rules for local names in the HTML namespace. Why? It would be good if there's one line of code authors can reliably use to duplicate an element. I think that line has to be:

    document.createElementNS(
        element.namespaceURI,
        `${element.prefix}:${element.localName}`,
        magic(element));

(OK three lines. And a function, "magic", which somehow detects autonomous built-in elements and produces a dictionary with 'is' for them.)

Maybe validate and extract step 5 should specify that it splits on the *first* colon. Since `:foo` would not be a valid local name any more, validate and extract may want to map the empty string prefix to null so that the checks in step 6 work.

If you make the local name checks consistent with createElement's then that line of code should be sufficient for duplicating an element, whether it is HTML, SVG or vanilla.

domenic:
I've updated this with a revision that takes into account the -NS methods too.

dominiccooney:
I have posted some data from Chromium here: https://bugs.chromium.org/p/chromium/issues/detail?id=648179&desc=2#c19

domenic:
OK, so we finally have data!

- DOM valid, HTML parser invalid: 0.047 % page loads
- DOM invalid, HTML parser valid: 0.000003 % page loads

Unfortunately, it looks like simply aligning with the HTML parser is not going to be possible. @dominiccooney mentioned it might be possible to change the parser to allow more, but in general that's pretty high-cost and risky. Unless there's some unexpected enthusiasm for that, I'd rather set it aside.

What do people think of just expanding the valid-in-DOM category? Then we will still have two categories, but at least valid-in-DOM will be a superset of valid-in-HTML. I could work on updating this PR to illustrate that concretely, if people would like.

annevk:
I always thought that was the plan. With the syntax being less expressive. Matching the HTML exactly would never have worked due to casing depending on context and such.

strongholdmedia:
> What do people think of just expanding the valid-in-DOM category?
I personally think that you should say WHICH DOM.
There is almost as much cowboyism recently with custom rules and "solutions" lately (as in angular, etc), that makes me think the "creators" either forgot quirks mode and browser wars, or just entered the game after it was over (is it even, btw?)

dom.bs
```diff
@@ -266,42 +268,112 @@ against a <var>set</var>, run these steps:
...
+<ol>
+ <li><p>If <var>namespace</var> is the empty string, set it to null.
+
+ <li><p>Let <var>prefix</var> be null.
+
+ <li><p>Let <var>localName</var> be <var>qualifiedName</var>.
+
+ <li><p>If <var>qualifiedName</var> contains a "<code>:</code>" (U+003E), then split
+ <var>qualifiedName</var> on that first occurrence of "<code>:</code>" and set <var>prefix</var> to
+ the part before and <var>localName</var> to the part after.
```
snuggs: @domenic `before` and `after` seem a bit vague. Perhaps suffix with ... _`prefix` to the part before and `localName` to the part after the `:` respectively.

Just a thought. Sounded incomplete when reading out loud.

PR #452, opened 2017-04-28
DOMTokernList's toggle() cannot always update the attribute

### LINKED ISSUE #451: Web reality: DOMTokenList's toggle() cannot always update the corresponding attribute
@smaug---- mentioned https://bugzilla.mozilla.org/show_bug.cgi?id=1360230 in #444. Which is a little sad, but also fine to fix.

So we basically want to go back to only invoking the "update steps" when we mutate the set first. Unclear if we should do the same for `replace()`. Given this is a special case it probably warrants a note as well.

annevk:
The consistency change made in d396872f847d63b803393092eb81843e80d566b7
turned out to not be compatible with the web.

Fixes #451.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/toggle/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/5a532da...160d551.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9664,25 +9664,25 @@ method, when invoked, must run these steps:
  <li><p>If <var>token</var> contains any <a>ASCII whitespace</a>, then <a>throw</a> an
  {{InvalidCharacterError}}.
 
- <li><p>Let <var>result</var> be false.
-
  <li>
   <p>If <a>context object</a>'s <a>token set</a>[<var>token</var>] <a for=set>exists</a>, then:
 
   <ol>
    <li><p>If <var>force</var> is either not given or is false, then <a for=set>remove</a>
-   <var>token</var> from <a>context object</a>'s <a>token set</a>.
+   <var>token</var> from <a>context object</a>'s <a>token set</a>, run the <a>update steps</a> and
+   return false.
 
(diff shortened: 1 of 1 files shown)


ayg:
@annevk Anything holding this up?

annevk:
@ayg lack of review and a bit of uncertainty about test coverage.

ayg:
I don't see any tests for this either way.  I'll write some now.  The patch looks good, and it seems you should write a new one for `replace()` also.

ayg:
Typo in commit message: DOMTokernList.

ayg:
Tests submitted.

annevk:
So for replace we just want an early return after step 2 if context object's token set doesn't contain token? If we want to match existing implementations. I think if that's what you suggest I'll add that to this PR.

ayg:
Yes, exactly.

annevk:
Now blocked on https://github.com/tabatkins/bikeshed/issues/1066, but please regard this as a done deal.

```
DOMTokernList's toggle()/replace() cannot always update the attribute

The consistency change made in d396872f847d63b803393092eb81843e80d566b7
turns out to not be compatible with the web.

Tests: https://github.com/w3c/web-platform-tests/issues/6764.

Fixes #451.
```

PR #459, opened 2017-05-15
Don't let a slot do anything special if it isn't in a shadow tree

### LINKED ISSUE #447: Shadow: Specify when `slotchange` fires
Safari, Chrome and ShadyDOM currently exhibit different behavior when `slotchange` does and does not fire. As far as I understand the spec, it is not precisely spec’d, meaning all these current behaviors are technically spec compliant, but can be very frustrating for developers.

[Demo 1](https://jsbin.com/liveko/9/edit?html,output): Slots are created in the constructor
[Demo 2](https://jsbin.com/puzilim/3/edit?html,output): Slots are creatred in `connectedCallback`

**Which browser fires `slotchange` under what circumstances?**

| | Slots in constructor, parser creates element  | Slots in constructor, element gets upgraded | Slots in `connectedCallback`, parser creates element | Slots in `connectedCallback`, element gets upgraded |
|-|----|---|---|---|
|Chrome| Fire | No fire | Fire | No Fire |
|Safari| Fire | No fire | No Fire | No Fire |
|ShadyDOM| Fire | Fire | Fire | Fire |

(ShadyDOM was tested using Firefox).

While **my main concern is consistency across browsers**, I like ShadyDOM’s behavior best in terms of DX as it is consistent with how `attributeChangeCallback` in that it fires upon initialization. This would allow developers to have their update logic in the `slotchange` handler and they wouldn’t need to manually detect silently slotted elements in `connectedCallback()`.

/cc @hayatoito

hayatoito:
Fixes #447.

***
[Preview](https://s3.amazonaws.com/pr-preview/hayatoito/dom/slotchange-when-inserted.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/d9cd30d...hayatoito:e1db083.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1708,6 +1708,9 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
 <ol>
  <li><p>Let <var>result</var> be an empty list.</p></li>
 
+ <li><p>If <var>slot</var>'s <a for=tree>root</a> is not a <a for=/>shadow root</a>, then return
+ <var>result</var>.</p></li>
+
  <li><p>Let <var>slotables</var> be the result of <a>finding slotables</a> given
  <var>slot</var>.</p></li>
 
@@ -1719,7 +1722,8 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
 
   <ol>
    <li>
-    <p>If <var>node</var> is a <a>slot</a>, then:
+    <p>If <var>node</var> is a <a>slot</a> whose <a for=tree>root</a> is a <a for=/>shadow root</a>,
+    then:
 
     <ol>
      <li><p>Let <var>temporaryResult</var> be the result of <a>finding flattened slotables</a> given
@@ -1738,15 +1742,13 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
 
 <h5 id=assigning-slotables-and-slots>Assigning slotables and slots</h5>
 
-<p>To <dfn noexport>assign slotables</dfn>, for a <a>slot</a> <var>slot</var> with an optional
-<var>suppress signaling flag</var> (unset unless stated otherwise), run these steps:
+<p>To <dfn noexport>assign slotables</dfn> for a <a>slot</a> <var>slot</var>, run these steps:
 
 <ol>
  <li><p>Let <var>slotables</var> be the result of <a>finding slotables</a> for <var>slot</var>.
 
- <li><p>If <var>suppress signaling flag</var> is unset, and <var>slotables</var> and
- <var>slot</var>'s <a for=slot>assigned nodes</a> are not identical, then run
- <a>signal a slot change</a> for <var>slot</var>.
+ <li><p>If <var>slotables</var> and <var>slot</var>'s <a for=slot>assigned nodes</a> are not
+ identical, then run <a>signal a slot change</a> for <var>slot</var>.
 
  <li><p>Set <var>slot</var>'s <a for=slot>assigned nodes</a> to <var>slotables</var>.
 
@@ -1754,16 +1756,9 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
  <a>assigned slot</a> to <var>slot</var>.
 </ol>
 
-<p>To <dfn noexport>assign slotables for a tree</dfn>, given a <a>tree</a> <var>tree</var> and an
-optional set of <a>slots</a> <var>noSignalSlots</var> (empty unless stated otherwise), run these
-steps for each <a>slot</a> <var>slot</var> in <var>tree</var>, in <a>tree order</a>:
-
-<ol>
- <li><p>Let <var>suppress signaling flag</var> be set, if <var>slot</var> is in
- <var>noSignalSlots</var>, and unset otherwise.</p></li>
-
- <li><p>Run <a>assign slotables</a> for <var>slot</var> with <var>suppress signaling flag</var>.
-</ol>
+<p>To <dfn noexport>assign slotables for a tree</dfn>, given a <a>tree</a>
+<var>tree</var>, run <a>assign slotables</a> for each <a>slot</a> <var>slot</var> in
+<var>tree</var>, in <a>tree order</a>.
 
 <p>To <dfn noexport>assign a slot</dfn>, given a <a>slotable</a> <var>slotable</var>, run these
 steps:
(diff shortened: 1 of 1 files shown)


hayatoito:
@surma is taking care of the PR for the tests:  https://github.com/w3c/web-platform-tests/pull/5705

hayatoito:
It looks we have to update the existing test in WPT.
I am working on that at https://github.com/w3c/web-platform-tests/pull/5954.

annevk:
This change LGTM. We only need bugs against Chrome and WebKit I suspect, since nobody else is far enough along thus far.

hayatoito:
Yeah, regarding Chrome, I'll take care of it.

hayatoito:
I have updated the PR so that it can reflect the conclusion in #447.
@domenic , could you have a chance to review this?
cc: @annevk

dom.bs
```diff
@@ -1738,32 +1741,23 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
...
+<p>To <dfn noexport>assign slotables</dfn> for a <a>slot</a> <var>slot</var>, run these steps:
 
 <ol>
  <li><p>Let <var>slotables</var> be the result of <a>finding slotables</a> for <var>slot</var>.
 
- <li><p>If <var>suppress signaling flag</var> is unset, and <var>slotables</var> and
- <var>slot</var>'s <a for=slot>assigned nodes</a> are not identical, then run
- <a>signal a slot change</a> for <var>slot</var>.
+ <li><p>If <var>slotables</var> and <var>slot</var>'s <a for=slot>assigned
+ nodes</a> are not identical, then run <a>signal a slot change</a> for <var>slot</var>.
```
domenic: I notice that this instance of "signal a slot change" has fewer checks before it than the other two. Is that correct? (I don't understand the algorithms that well so I am just pattern-matching in the hopes of finding something useful :)

domenic: Nit: annevk prefers tags all on one line in his specs, so no linebreak in the middle of an `<a>stuff stuff</a>`. So the whole `<a for=slot>assigned nodes</a>` should get pushed to the next line.

hayatoito: Yes, that is correct and intentional.
- For assigned nodes' changes, the condition became relaxed.
- For other cases (aka fallback contents), the condition become more tight.

hayatoito: Thanks. Let me put it in the same line.

hayatoito:
@domenic Thank you for the review.
I have updated the PR, by squashing the commits.

I am now working on updating WPT. I'll ping @rniwa in PR for WPT too.

dom.bs
```diff
@@ -1719,7 +1722,8 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
 
   <ol>
    <li>
-    <p>If <var>node</var> is a <a>slot</a>, then:
+    <p>If <var>node</var> is a <a>slot</a> whose <a for=tree>root</a> is a <a for=/>shadow root</a>,
```
hayatoito: I also updated this line, which I forgot to fix in the previous PR.

hayatoito:
@rniwa, could you have a chance to take a look?

PR for WPT Tests is here: https://github.com/w3c/web-platform-tests/pull/5954

hayatoito:
FYI. I implemented this new behavior in Blink.

hayatoito:
@rniwa, friendly ping. I am assuming that this change sounds good to all of us.

rniwa:
The intention seems fine but did you check that we don't end up signaling the same slot multiple times per a single DOM change especially nested ones?

I haven't gotten around to fix WebKit's implementation but I intend to do it in near future.

hayatoito:
> The intention seems fine but did you check that we don't end up signaling the same slot multiple times per a single DOM change especially nested ones?

That shouldn't happen, as far as my understanding is correct. A slotchange is signaled at most once per a slot per a single DOM change for any tree structures.

rniwa:
I understand that's the intention but it's really hard for me to tell if that's what's gonna happen with this change or not.

In general, I find it really hard to read these spec diffs and figure out what's actually happening. It would be ideal if we could look at the spec with the change applied. Is there a way to do that?

hayatoito:
It looks this DOM repository adds a *Preview* or *Diff* link in the first comment in PR automatically.
I guess such a nice feature was recently introduced.

Though Diff link looks ugly for unknown reasons (Hmm, I remember that it showed a clean diff when I updated this PR), I think Preview link still works.

hayatoito:
@annevk  Thanks. I have rebased on master and force push.

annevk:
Great!

@rniwa you can review changes in a rendered way starting here (I think there's nothing before this point): https://s3.amazonaws.com/pr-preview/whatwg/dom/d9cd30d...hayatoito:32e6669.html#find-flattened-slotables.

rniwa:
That diff link was very useful! The change looks good to me.

r=me.

dom.bs
```diff
@@ -1738,32 +1742,23 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
...
+<p>To <dfn noexport>assign slotables</dfn> for a <a>slot</a> <var>slot</var>, run these steps:
 
 <ol>
  <li><p>Let <var>slotables</var> be the result of <a>finding slotables</a> for <var>slot</var>.
 
- <li><p>If <var>suppress signaling flag</var> is unset, and <var>slotables</var> and
- <var>slot</var>'s <a for=slot>assigned nodes</a> are not identical, then run
- <a>signal a slot change</a> for <var>slot</var>.
+ <li><p>If <var>slotables</var> and <var>slot</var>'s <a for=slot>assigned nodes</a> are not
+identical, then run <a>signal a slot change</a> for <var>slot</var>.
```
annevk: Could you add a space before identical? (I would do it, but I cannot modify this branch for reasons unclear.)

hayatoito: My bad. Done. I've updated the PR.

annevk:
Due to jetlag I forgot to complete the usual checklist:

* Tests haven't landed yet. @hayatoito will take care of that per https://github.com/w3c/web-platform-tests/pull/5705#issuecomment-305745169.
* Browser bugs haven't been filed. @rniwa and @hayatoito, I suspect you'll take care of that for your respective engines? Nobody else implements this yet so I'm guessing nothing is needed for them.

hayatoito:
Thanks for merging.
Regarding tests, I have updated existing tests for slotshange to reflect this change in another PR: https://github.com/w3c/web-platform-tests/pull/5954
I'll review https://github.com/w3c/web-platform-tests/pull/5705#issuecomment-305745169 too.

Regarding implementation, I could take care of Blink. I think @rniwa would take care of WebKit.

annevk:
Thanks, I'm going to assume all is in order then.

PR #464, opened 2017-05-31
Clarify insert-node's record's previousSibling

bwrrp:
This adds a previousSibling variable to the insert node algorithm in order to clarify that the previousSibling property of the mutation record generated by the insert node algorithm should be determined before inserting nodes. 

If implemented as described in the original text, the property would be determined after insertion, and will therefore always be the last of nodes.

***
[Preview](https://s3.amazonaws.com/pr-preview/bwrrp/dom/insertNode-MutationRecord-previousSibling.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/93db340...bwrrp:3927500.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1948,6 +1948,9 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
   <p class="note no-backref">This step intentionally does not pay attention to the
   <i>suppress observers flag</i>.
 
+ <li><p>Let <var>previousSibling</var> be <var>child</var>'s <a>previous sibling</a> or
+ <var>parent</var>'s <a>last child</a> if <var>child</var> is null.
+
  <li>
   <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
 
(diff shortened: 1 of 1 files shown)


nox:
This doesn't seem to work if *node* is *child*'s previous sibling.

bwrrp:
The "insert a node into a parent before a child" algorithm is invoked from the following algorithms:
 * "pre-insert a node into a parent",
 * "replace a child with node within a parent",
 * "replace all with a node within a parent"
 * "split a Text node"

In the first three, the "adopt a node into document" algorithm will have been run on *node*, removing it from any parent it may have had. In the last, *node* will be a new node, which does not have a parent. As a result, *node* can not be *child*'s previous sibling at this point of the "insert node" algorithm. I wouldn't mind adding a check for this case, but that would produce dead code in implementations that directly follow the spec.

nox:
Oh right! Thanks for the walkthrough, my brain was confused.

dom.bs
```diff
@@ -1948,6 +1948,9 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
   <p class="note no-backref">This step intentionally does not pay attention to the
   <i>suppress observers flag</i>.
 
+ <li>Let <var>previousSibling</var> be <var>child</var>'s <a>previous sibling</a> or
```
annevk: Needs to be `<li><p>`.

annevk:
@bwrrp can you add your name to the acknowledgments section as well? I'll land this later today.

bwrrp:
Done, I also rebased the branch onto the current master and squashed the commits into one.

annevk:
Excellent, thank you!

PR #466, opened 2017-06-15
Add child text content and its change steps

domenic:
Part of fixing https://github.com/whatwg/html/issues/2752. The "child text content" concept was previously defined in HTML, but is being moved here.

Probably shouldn't merge this until the corresponding HTML PR is ready and approved, in case there are some integration issues I missed out on, but I think this is reviewable as-is.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/child-text-content-changed/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/4512167...6a9b91f.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3857,6 +3857,10 @@ steps for each <a>descendant</a> <a>exclusive <code>Text</code> node</a> <var>no
  itself), in <a>tree order</a>.
 </ol>
 
+<p class="note">{{Node/normalize()}} does not need to run any
+<a>child text content change steps</a>, since although it messes with {{Text}} nodes extensively, it
+does so specifically in a way that preserves the <a>child text content</a>.
+
 <hr>
 
 <dl class=domintro>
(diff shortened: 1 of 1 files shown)


zcorpan:
I think the change in "insert" at least is wrong if *node*  is a `DocumentFragment`.

zcorpan:
Probably move the check into step 6?

zcorpan:
I think it would be good if @annevk reviewed this also.

annevk:
Does `normalize()` need to run this or not? If not, let's add a note. (Found by searching for "for each range".)

domenic:
> Does normalize() need to run this or not? If not, let's add a note.

It should not, unless there are edge cases of normalize() that I am not aware of. My understanding is normalize() is basically figuring out the minimal set of text nodes to represent the exact same child text content, so the child text content will not change.

I'll add a note to that effect.

annevk:
We should probably test that as normally even no-op changes trigger mutation records. And this is more than a no-op (although quite similar).

domenic:
Hmm, what kind of test do you mean? For the textarea case there would be no difference between running the steps, or not. Running the steps unnecessarily just updates the API value to be itself, which is unobservable.

annevk:
I guess if we don't have any endpoint where it can observed it does not matter.

Issue #468, opened 2017-06-23
Service worker check might be over-ambitious


domenic:
> If context object’s relevant global object is a ServiceWorkerGlobalScope object

Shouldn't this be checking if context object is a ServiceWorkerGlobalScope object? Otherwise it means every EventTarget created inside the service worker has this restriction.

annevk:
Duplicate of #371.

PR #469, opened 2017-06-23
Allow removing event listeners by group

### LINKED ISSUE #208: Removing event listeners through an identifier
Reportedly when writing components it would be convenient to be able to remove a whole swatch of event listeners through a common identifier, rather than having to enumerate all the specific types and callbacks.

I have discussed this a long time ago with @wycats I believe and more recently with @RByers. Now that `addEventListener()` supports a dictionary we could make this a reality. E.g.,

``` js
obj.addEventListener("click", ..., { class: "clicker" })
obj.addEventListener("focus", ..., { class: "clicker" })
obj.removeEventListener({class: "clicker" })
```

An alternative to overloading `obj.removeEventListener()` would be adding something like `obj.removeEventListenerClass("clicker")`, but that seems less extensible.

domenic:
Closes #208.

Tests: https://github.com/w3c/web-platform-tests/pull/6331

Chrome is tentatively interested in this, although the first step is getting a spec and tests up for review. Would love to hear if other implementers are interested--- @smaug----, @cdumez, @travisleithead?

I decided to go with an overload instead of a new method, so we can leave new method names for something potentially nicer like `on`/`off`. Overloading for this pattern is pretty common, e.g. it is what jQuery does. And introducing two similar-but-different method names like `removeEventListener` + `deleteEventListener` seemed like a bad idea.

I also added an overload `removeEventListener(type, { group })` which is not really necessary. We could get rid of that if we don't like it. But it fell out naturally.

This also preserves what I'm pretty sure is a bug, per #468, but at least it centralizes the probably-buggy check.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/event-namespaces/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/d9cd30d...9400bac.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -54,7 +54,6 @@ urlPrefix: https://w3c.github.io/ServiceWorker/#; spec: SERVICE-WORKERS
         text: has ever been evaluated flag; for: service worker
 urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
     text: Realm; url: realm; type: dfn
-    text: SameValueZero; url: sec-samevaluezero; type: abstract-op
 </pre>
 
 <pre class=link-defaults>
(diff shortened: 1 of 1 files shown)


smaug----:
( This is a pull request so try to not comment API proposal here. )

dom.bs
```diff
@@ -325,10 +325,23 @@ function imgFetched(ev) {
...
+<a>Event listeners</a> can be removed by utilizing the
+{{EventTarget/removeEventListener(type, callback, options)}} method, passing the same arguments.
+
+Once can also provide a <em>group</em> when adding an <a>event listener</a>. This can be any value,
+including a simple string. Then later, the same group value can be provided to
+{{EventTarget/removeEventListener(type, options)}} or {{EventTarget/removeEventListener(options)}},
+to remove all event listeners in that group:
+
+<pre class=lang-javascript>
+oven.addEventListener("turnon", () => { &hellip; }, { group: "cookies" })
```
Krinkle: `cookies` is a fun and lightweight example word, but I think in this context it has greater chances of confusing less-experienced readers due to being a very over-loaded term. Perhaps we can come up with another example here, or if nothing else, plain `"example"`.

domenic: Thanks. Will make it brownies instead during the next revision.

kenchris:
I was wondering whether supplying a signal like the AbortController/AbortSignal would make more sense instead of groups. Then you can abort multiple things with the same controller.

```
accelerometer.addEventListener('reading', listener, {signal:controller.signal});
controller.abort();

Rodriguez2014:
Y'all are jerks i got my wife an attorney. Because of you all bulling, harrassing, assulting, intimidating, and stalking my wife

annevk:
I suggest we close this now that you can use abort signals as grouping mechanism thanks to @benjamingr's work. Thoughts?

domenic:
Oh, right, this was mine originally. Yeah, signals is basically a better version of this; let's close!

Issue #410, opened 2017-02-15
Consider removing `document.origin`


mikewest:
We have `self.origin` in HTML, present inside workers and documents. It seems redundant and confusing to suggest that there's a different origin for documents than for the global object they host. Perhaps we can simply remove this in favor of the property on the global?

@bzbarsky seems to be on board with this notion. @cdumez, WDYT?

Edge hasn't implemented this yet, but perhaps @travisleithead has an opinion from a TAG/platformey perspective?

cdumez:
WebKit does not currently support self.origin but we will likely implement it in the future. Once we support self.origin, we can start deprecating document.origin in WebKit but I suspect it is going to be a while until we actually remove it (if ever, depending on how much it beaks existing content). That said, I do not have an objection if people want to drop document.origin from the spec, I just think it is a bit early given that self.origin is not well deployed yet.

Here are a few things that are not mentioned on this issue that would help decide:
1. Who currently supports document.origin? I seem to remember it is only Blink and WebKit but someone should confirm.
2. Who already ships self.origin? I suspect no one yet even though it is implemented in Firefox nightly iirc.

bzbarsky:
Correct on both counts, as far as I know: `document.origin` is only in Blink and WebKit, but not interoperable between the two in shipped versions.  `self.origin` is only in Firefox nightlies so far.

mikewest:
>  I just think it is a bit early given that self.origin is not well deployed yet.

I think we're just looking for feedback from other vendors at this point. If you're on board with the idea of migrating to `self.origin`, and can commit to the (hopefully minimal) work necessary to implement it at some point in the future, and send a console warning or something when you do, then I'm pretty sure we'll do the same in Blink. We're arguing about the impact of this a bit in https://groups.google.com/a/chromium.org/d/msg/blink-dev/CO52Bt15cuc/ya6u_devCQAJ if you'd like to skim through and add your thoughts.

To your questions:

1. Only Blink and WebKit (and, really, only WebKit nightly after your recent fix: you could, for instance, drop that entirely rather than keeping the broken implementation, right?)

2. Firefox nightly is it. We'll have a patch for Blink tomorrowish, but it won't hit stable for a few weeks.

cdumez:
Sure, I can commit to implementing self.origin in WebKit in the relatively short-term, given that other browsers are already implementing it and I like the idea.

I also do not object to dropping document.origin from the spec given that this is a Blink/WebKit thing only and there is a better alternative available.

[...]

PR #475, opened 2017-07-05
Origins serialize to ASCII these days

annevk:
***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/origin-ascii/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/8f8c1c3...308dac6.html)


annevk:
I should maybe add a test, but I also thought the idea was to get rid of `document.origin` in favor of `self.origin`. @mikewest @bzbarsky?

annevk:
I just found #410. I guess since Firefox doesn't implement this anyway I could make this change to prevent further build warnings and then later drop it or just drop it. Would be good to know though how likely it is that Chrome and Safari can drop it.

annevk:
I merged this since I didn't get any response and the warnings started to get in the way of things. Anything further should probably go into #410.

PR #476, opened 2017-07-14
Meta: add a service worker to make the spec viewable offline

domenic:
Uses https://github.com/whatwg/resources.whatwg.org/pull/53.

@annevk, if you could look over this and ideally https://github.com/whatwg/resources.whatwg.org/pull/53 as well, I'll merge them together once approved, and keep my finger hovering on the revert button in case anything goes wrong.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/service-worker/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/fba2caf...22611c0.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -10101,4 +10101,11 @@ Per <a rel="license" href="//creativecommons.org/publicdomain/zero/1.0/">CC0</a>
 the extent possible under law, the editors have waived all copyright and related or
 neighboring rights to this work.
 
+<script>
+"use strict";
+if ("serviceWorker" in navigator) {
+  navigator.serviceWorker.register("/service-worker.js");
+}
+</script>
+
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -10101,4 +10101,11 @@ Per <a rel="license" href="//creativecommons.org/publicdomain/zero/1.0/">CC0</a>
 the extent possible under law, the editors have waived all copyright and related or
 neighboring rights to this work.
```
annevk: Maybe add a closing `</p>` here so the `<script>` doesn't end up inside the `<p>`? It doesn't really matter, but it seems better...

domenic: It appears Bikeshed's automatic-paragraphing takes care of this for us, wooho.

domenic:
This seems to have worked nicely, woohoo!

PR #488, opened 2017-08-03
DOMTokenList shouldn't add empty attributes

### LINKED ISSUE #462: DOMTokenList.remove() can add an attribute
Specification: https://dom.spec.whatwg.org/#dom-domtokenlist-remove
WPT: http://w3c-test.org/dom/nodes/Element-classlist.html

Should ```remove()``` really add an empty attribute?

According to the current specification, the following code should show ```""``` (empty string).

```
var c = document.createElement('div');
c.classList.remove('a');
alert('"' + c.getAttribute('class') + '"');
```

Safari TP works so.  

The WPT test contains some sub-tests which don't conform to the specification. e.g.  "classList.remove("a") with attribute value null (HTML node)"
These sub-tests assume ```remove()``` never adds ```class``` attribute, and Firefox nightly, Chrome, and Edge don't add an empty attribute.

ayg:
Fixes #462.

***
[Preview](https://s3.amazonaws.com/pr-preview/ayg/dom/DOMTokenList-remove-noop.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/bcb0b3f...ayg:a06ca93.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9506,10 +9506,17 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
  <li><p>Return false.
 </ol>
 
-<p>A {{DOMTokenList}} object's <dfn id=concept-dtl-update for="DOMTokenList">update steps</dfn> are
-to <a>set an attribute value</a> for the associated <a for="/">element</a> using associated
-<a>attribute</a>'s <a for=Attr>local name</a> and the result of running the
-<a>ordered set serializer</a> for <a>token set</a>.
+<p>A {{DOMTokenList}} object's <dfn id=concept-dtl-update for="DOMTokenList">update steps</dfn> are:
+
+<ol>
+ <li><p>If the associated <a for="/">element</a> does not have an associated
+ <a>attribute</a>, and <a>token set</a> is empty, do nothing.
+
+ <li><p>Otherwise, <a>set an attribute value</a> for the associated
+ <a for="/">element</a> using associated <a>attribute</a>'s
+ <a for=Attr>local name</a> and the result of running the
+ <a>ordered set serializer</a> for <a>token set</a>.
+</ol>
 
 <p>A {{DOMTokenList}} object's <dfn id=concept-dtl-serialize for=DOMTokenList>serialize steps</dfn>
 are to return the result of running <a>get an attribute value</a> given the associated
--- a/dom.bs
+++ b/dom.bs
@@ -9509,13 +9509,12 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
 <p>A {{DOMTokenList}} object's <dfn id=concept-dtl-update for="DOMTokenList">update steps</dfn> are:
 
 <ol>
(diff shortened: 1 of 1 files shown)


ayg:
Not totally sure about the wording here.

dom.bs
```diff
@@ -9509,10 +9509,17 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
...
 
-<p>A {{DOMTokenList}} object's <dfn id=concept-dtl-update for="DOMTokenList">update steps</dfn> are
-to <a>set an attribute value</a> for the associated <a for="/">element</a> using associated
-<a>attribute</a>'s <a for=Attr>local name</a> and the result of running the
-<a>ordered set serializer</a> for <a>token set</a>.
+<p>A {{DOMTokenList}} object's <dfn id=concept-dtl-update for="DOMTokenList">update steps</dfn> are:
+
+<ol>
+ <li><p>If the associated <a for="/">element</a> does not have an associated
+ <a>attribute</a>, and <a>token set</a> is empty, do nothing.
```
annevk: s/, and/ and/
s/do nothing/then return/

dom.bs
```diff
@@ -9509,10 +9509,17 @@ associated <a>attribute</a>'s <a for=Attr>local name</a>.
...
+
+ <li><p>Otherwise, <a>set an attribute value</a> for the associated
+ <a for="/">element</a> using associated <a>attribute</a>'s
+ <a for=Attr>local name</a> and the result of running the
+ <a>ordered set serializer</a> for <a>token set</a>.
```
annevk: s/Otherwise, s/S/

annevk:
Oh I just noticed you're still discussing the details with @tkent-google. Hopefully this is still somewhat useful.

ayg:
This is already tested for both remove and force-toggle.  The tests didn't match the spec before now.

annevk:
So would it be good to partially revert 03657b3556018de1675c58f7fbc5e51ac2602dfb now that we have a check in "update steps" or is that still fine as-is?

Also, we should maybe add the note

    <p class="note no-backref">The <a>update steps</a> are not always run for {{DOMTokenList/toggle()}} for web compatibility.

here too? Or generalize it somehow?

ayg:
This check isn't just for web compat, it's also common sense.  We shouldn't add an empty attribute when there previously wasn't any attribute -- what's the point?  You could even argue that if the last token in the list was removed, you should remove the attribute, although I don't advocate going that far.

The checks in 03657b3 are still needed for when `toggle` force-removes a token that's not present but there are other tokens.  This check is still needed for when `remove` is used, because that does run the update steps when removing a token that's already not present.  In principle this check could be in `remove` instead of the update steps, because that's the only case where it will be hit, but it's logically more appropriate here, and more robust in the event that we add other ways to alter DOMTokenLists in the future.

So I think this can be taken now as-is with no additional changes.

annevk:
Thank you, appreciate the explanation.

PR #489, opened 2017-08-06
Remove more types from document.createEvent()

### LINKED ISSUE #362: document.createEvent() supports lots of events that most UAs do not support initializers for
The current whitelist of event names for document.createEvent() is based on events that at least two of Chrome, Firefox, Safari, Edge support.  However, when generating the list, I didn't test whether they support a corresponding init*Event() method on the interface.  Without that, support in createEvent() seems pointless.  Here's a new test that lists not only what interfaces are supported by document.createEvent(), but what init methods are found:

https://jsfiddle.net/s53w679y/1/

The result is that a lot of them are supported with an initializer only in Edge, and supported without an initializer in Chrome (and in a few cases Edge as well).  To wit, all of the following are unsupported in Firefox and Safari, and are supported in Chrome but without an initializer method:

* AnimationEvent
* CloseEvent
* ErrorEvent
* FocusEvent
* IDBVersionChangeEvent (Edge also has no initializer)
* PageTransitionEvent (Edge also has no initializer)
* ProgressEvent
* TrackEvent (Edge also has no initializer)
* TransitionEvent 
* WebGLContextEvent (Edge also has no initializer)

And similarly:

* PopStateEvent: Supported by Chrome and Firefox with no special initializer.  Not supported by Safari.  Supported by Edge with a special initializer.
* SVGZoomEvent: Supported by Firefox with no special initializer.  Not supported at all by other browsers.
* SVGZoomEvents: Supported by Firefox, Safari, and Edge with no special initializer.  Not supported by Chrome.

I suggest all of these be removed from the spec, unless we think pages really use events that they can't initialize.  Maybe they use the default values for all the properties?  In some cases there are still initializers like initUIEvent from base classes.

@domenic, do you know why Chrome supports so many event types from createEvent() even though it has no initializer?  Would you have a problem with not supporting these in createEvent() anymore?

@bzbarsky, do you know why Gecko supports SVGZoomEvent(s) from createEvent() even though it has no initializer?  

@khuey In <https://hg.mozilla.org/mozilla-central/rev/1f60138bce8d>, why did you leave PopStateEvent support in createEvent but removed the initializer?  Just to match Chrome?  It looks like <https://bugzilla.mozilla.org/show_bug.cgi?id=1031051#c33> asked to remove it, but you didn't.

A couple of other event types that are interesting:

* BeforeUnloadEvent: Supported by Chrome and Firefox with no special initializer.  Not supported by Safari or Edge.  But this is not supposed to have any special initializer per spec, so that's fine.
* WheelEvent: Supported in both Edge and Safari with an initializer, but they're named differently (initWheelEvent vs. initWebKitWheelEvent).  Chrome has only initMouseEvent.  Firefox doesn't support it at all.  I don't know if we need to keep this in the spec.

ayg:
Fixes #362.  All of the removed event types except WheelEvent are
unsupported or are supported but cannot be properly initialized in
Blink, Gecko, and WebKit.  WheelEvent is supported in WebKit with a
vendor-prefixed initializer, but WebKit is willing to try dropping it.

***
[Preview](https://s3.amazonaws.com/pr-preview/ayg/dom/createEvent.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/4a58d67...ayg:cf75365.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5299,39 +5299,29 @@ invoked, must run these steps:
    <thead>
     <tr><th>String<th>Interface<td>Notes
    <tbody>
-    <tr><td>"<code>animationevent</code>"<td>{{AnimationEvent}}<td>[[!CSS3-ANIMATIONS]]
     <tr><td>"<code>beforeunloadevent</code>"<td>{{BeforeUnloadEvent}}<td rowspan=2>[[!HTML]]
-    <tr><td>"<code>closeevent</code>"<td>{{CloseEvent}}
     <tr><td>"<code>compositionevent</code>"<td>{{CompositionEvent}}<td>[[!UIEVENTS]]
     <!-- textevent also maps to CompositionEvent -->
     <tr><td>"<code>customevent</code>"<td>{{CustomEvent}}<td>
     <tr><td>"<code>devicemotionevent</code>"<td>{{DeviceMotionEvent}}<td rowspan=2>[[!DEVICE-ORIENTATION]]
     <tr><td>"<code>deviceorientationevent</code>"<td>{{DeviceOrientationEvent}}
     <tr><td>"<code>dragevent</code>"<td>{{DragEvent}}<td rowspan=2>[[!HTML]]
-    <tr><td>"<code>errorevent</code>"<td>{{ErrorEvent}}
     <tr><td>"<code>event</code>"<td rowspan=2>{{Event}}<td rowspan=2>
     <tr><td>"<code>events</code>"
     <!-- htmlevents and svgevents also map to Event -->
     <tr><td>"<code>focusevent</code>"<td>{{FocusEvent}}<td>[[!UIEVENTS]]
     <tr><td>"<code>hashchangeevent</code>"<td>{{HashChangeEvent}}<td>[[!HTML]]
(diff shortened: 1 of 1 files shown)


annevk:
(FWIW, there's no longer a need for you to fork this repository. You should be able to create branches directly on whatwg/dom.)

ayg:
Is that preferred?  My GitHub workflow in general is to fork, and it's easier for me to not have a different workflow for different repos, but if there's some advantage I could do it.

annevk:
It's a little easier for me to push fixups to local branches, but that's about it. No strong preference, just thought it might be more convenient for you to have less repositories to track.

ayg:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1388069

ayg:
Actually, that bug is the opposite (adding a type that we don't yet support).  Removing the two we do support that were removed in this commit is here: https://bugzilla.mozilla.org/show_bug.cgi?id=1388119

(It turns out ErrorEvent support was added to Firefox six months ago to match the spec, but now I think we should remove it again given that Chrome is willing to drop it.)

PR #490, opened 2017-08-08
Update width from 76 to 99 in vim modeline

ayg:
According to annevk on #488, text is supposed to be wrapped to 100
columns, but the modeline says 76.  I made it 99 instead of 100 so that
diffs will be 100 columns and won't overflow if your terminal has a
width of 100, but it could be actually 100 too if we want.

***
[Preview](https://s3.amazonaws.com/pr-preview/ayg/dom/modeline.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/bcb0b3f...ayg:16da399.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -10110,4 +10110,4 @@ if ("serviceWorker" in navigator) {
 }
 </script>
 
-<!-- vim: set expandtab shiftwidth=1 tabstop=1 textwidth=76 -->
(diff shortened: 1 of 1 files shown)


annevk:
So for clarity, in my editor the first character of a line is placed on column 1 and the last on column 100. So the newline comes at 101 if that makes sense. It'd be good to match that as that is what https://github.com/whatwg/dom#formatting is trying to convey.

ayg:
That works for me too.  I updated the PR.

PR #492, opened 2017-08-14
Editorial: simplify EventListenerOptions processing

annevk:
@RByers what do you think?

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/EventListenerOptions/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/aaf90dc...dcd948d.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -894,8 +894,8 @@ for historical reasons.
 <pre class=idl>
 [Constructor, Exposed=(Window,Worker)]
 interface EventTarget {
-  void addEventListener(DOMString type, EventListener? callback, optional (AddEventListenerOptions or boolean) options = false);
-  void removeEventListener(DOMString type, EventListener? callback, optional (EventListenerOptions or boolean) options = false);
+  void addEventListener(DOMString type, EventListener? callback, optional (AddEventListenerOptions or boolean) options);
+  void removeEventListener(DOMString type, EventListener? callback, optional (EventListenerOptions or boolean) options);
   boolean dispatchEvent(Event event);
 };
 
--- a/dom.bs
+++ b/dom.bs
@@ -894,8 +894,8 @@ for historical reasons.
 <pre class=idl>
 [Constructor, Exposed=(Window,Worker)]
 interface EventTarget {
-  void addEventListener(DOMString type, EventListener? callback, optional (AddEventListenerOptions or boolean) options);
-  void removeEventListener(DOMString type, EventListener? callback, optional (EventListenerOptions or boolean) options);
+  void addEventListener(DOMString type, EventListener? callback, optional (AddEventListenerOptions or boolean) options = false);
+  void removeEventListener(DOMString type, EventListener? callback, optional (EventListenerOptions or boolean) options = false);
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -1010,14 +1010,9 @@ are not to be used for anything else. [[!HTML]]
...
- <li><p>Let <var>capture</var> be false.
+ <li><p>If <var>options</var> is a boolean, then return <var>options</var>.
 
- <li><p>If <var>options</var> is a boolean, set <var>capture</var> to <var>options</var>.
-
- <li><p>If <var>options</var> is a dictionary, then set <var>capture</var> to <var>options</var>'s
- <code>{{EventListenerOptions/capture}}</code>.
-
- <li><p>Return <var>capture</var>.
+ <li><p>Return <var>options</var>'s <code>{{EventListenerOptions/capture}}</code>.
```
RByers: This all looks nice (not sure why I thought the extra text was useful - sorry).

dom.bs
```diff
@@ -894,8 +894,8 @@ for historical reasons.
 <pre class=idl>
 [Constructor, Exposed=(Window,Worker)]
 interface EventTarget {
-  void addEventListener(DOMString type, EventListener? callback, optional (AddEventListenerOptions or boolean) options);
-  void removeEventListener(DOMString type, EventListener? callback, optional (EventListenerOptions or boolean) options);
+  void addEventListener(DOMString type, EventListener? callback, optional (AddEventListenerOptions or boolean) options = false);
```
RByers: Why make this change?

IIRC we discussed this on my original PR.  I could dig up the history but perhaps the argument against it was that it was observably identical to not specifying any default and instead relying on the dictionary defaults.  If behavior is indeed identical with or without this change, then I'd prefer we leave it out - the meaning when named `options` seems more explicit.  Also since we're now doing spec/impl IDL diffing, we'd want to get all the implementations to update which seems silly if there's no behavior difference whatsoever.

annevk: This way we don't have to set the default value in prose. For some reason we didn't spot this solution in the original PR and we had all the defaults in prose and therefore I argued to not have them in IDL: https://github.com/whatwg/dom/pull/82#discussion-diff-47618015.

annevk: Actually, maybe this isn't needed because omitting the argument means an implicit dictionary gets passed. I'm not a 100% sure that's also true if the dictionary is part of a union, but it'd make sense.

annevk: Ah yes, that's what happens: https://heycam.github.io/webidl/#es-union. Okay, I'll revert the IDL changes.

RByers: Yeah, exactly.  Our [implementation](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/events/EventTarget.cpp?type=cs&q=EventTarget::addEventListener&sq=package:chromium&l=318) also works that way too - getting the dictionary defaults automatically from the bindings generator, not explicitly in code (ignoring, for now, our special treatment of passive).

annevk:
Thanks!

PR #503, opened 2017-08-26
Meta: require Trusty as the Travis dist

foolip:
See issue in https://github.com/whatwg/resources.whatwg.org/pull/60#issuecomment-325165334

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: generic
 
 env:


foolip:
This didn't actually work. Maybe the .travis.yml is ignored in pull requests, instead using the original one?

foolip:
Oops, sorry, by restarting that build on the master branch, the deploy happened again, and https://dom.spec.whatwg.org/commit-snapshots/48c4b4aa0d1b4044b7be62d3a75aaab94058161f/ will forever have the wrong date.

foolip:
> This didn't actually work.

I am silly, I asked for the wrong dist in .travis.yml :)

This seems like it will fix the problem, but there are still some other issues with the new deploy that I'm looking into.

foolip:
It worked! I guess this can be removed in not too long when it becomes the default.

foolip:
Commenting here because this is the first issue and the one I'm pointing to. I can't I understand why certain repos are getting precise by default and need this override while others seemingly do not. I'm just assuming there's some logic to it, and that eventually the explicit choice of dist can be reverted.

PR #512, opened 2017-09-14
📝 Editorial: add missing 'if'

### LINKED ISSUE #508: "Otherwise, localName is associated attribute’s ..."
https://dom.spec.whatwg.org/commit-snapshots/c9544325095195c3afdf6284c29b26e84ce572aa/#interface-domtokenlist

> Otherwise, localName is associated attribute’s local name,

Missing word "if" before localName

snuggs:
📝 Fixes #508 Grammatical error with missing 'if'

***
[Preview](https://s3.amazonaws.com/pr-preview/snuggs/dom/patch-1.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/c954432...snuggs:19b85ee.html)


annevk:
Thanks! Would you like to add your name to the Acknowledgments section?

The commit message you want here is probably something like
```
Editorial: add missing 'if'

Fixes #508.
```
but I can correct that when landing.

We should probably centrally define the commit message rules for WHATWG since we tend to apply https://github.com/whatwg/meta/blob/master/TEAM.md#handling-pull-requests as well, which isn't very clear.

snuggs:
Correct @annevk and I agree fully! As you would recommend perhaps for another Issue (that I can take care of myself)? 

That said the commit is updated.

By the way...where is the "Acknowledgements" section?

capture d ecran 2017-09-14 a 12 17 38

annevk:
Acknowledgments is a section in the standard itself. The same file you're already patching.

A commit message rules issue would probably be best over at https://github.com/whatwg/meta/issues/new and then we can figure out where to host the rules and get all the standards to link them. Thanks!

snuggs:
@annevk Updated as per your request.

#TodayILearned ... if the needle is far enough down in the LOC haystack not even VIM search finds the word. I can't make this stuff up. Ever think of "partializing" the spec? (whatever that would mean here) Just a thought.

Is there anything else we need for this one? Or can I go have a celebratory 🍷 later?:-)

annevk:
You can go ahead and do that, cheers! 😊

annevk:
(To answer your question, I don't really like the idea of splitting up this file. It's rather convenient for a standard to be in a single one. There must be some way to configure Vim so it can deal with it better.)

annevk:
Haha, I missed you sneaked in an emoji. Nice.

snuggs:
@annevk i'll blame it on beginner's luck. :roll_eyes: ;-)

See you around the issues! ✋

Issue #511, opened 2017-09-14
Event.target should return null after dispatch is complete, when stopped at ShadowRoot


TakayoshiKochi:
https://dom.spec.whatwg.org/#dispatching-events doesn't specify any operation on `event.target` after dispatching is done (step 14 and after). This leaves `event.taget` as what it was when the event was last set in step 13.1.

This, in turn, means that when the event is stopped at shadow root, and the reference to the event is stored somewhere in the event handler, the last `event.target` would sort of leak after dispatch is complete.

Would like to reset `event.target` to `null` when the event stopped at shadow root.

annevk:
Good catch. That seems like the correct solution. Always doing it seems too risky.

cc @whatwg/components

TakayoshiKochi:
We noticed that we missed `event.relatedTarget` and should be nullified when it's possible to leak shadow tree internals.  Let me update #513.

PR #513, opened 2017-09-15
Set event.target to null when dispatching finishes at ShadowRoot

TakayoshiKochi:
Resolves whatwg/dom#511.

***
[Preview](https://s3.amazonaws.com/pr-preview/TakayoshiKochi/dom/squashing_target.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/601655d...TakayoshiKochi:22d1fb6.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1253,9 +1253,8 @@ for discussion).
 
  <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/NONE}}.
 
- <li><p>Set <var>event</var>'s {{Event/target}} attribute and <var>event</var>'s
- <a>relatedTarget</a> to null if {{Event/target}}'s <a for=tree>root</a> is a
- <a for=/>shadow root</a>.
+ <li><p>If {{Event/target}}'s <a for=tree>root</a> is a <a for=/>shadow root</a>, set
+ <var>event</var>'s {{Event/target}} attribute and <var>event</var>'s <a>relatedTarget</a> to null.
 
  <li><p>Set <var>event</var>'s {{Event/currentTarget}} attribute to null.
 
--- a/dom.bs
+++ b/dom.bs
@@ -1253,7 +1253,7 @@ for discussion).
 
  <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/NONE}}.
 
- <li><p>If {{Event/target}}'s <a for=tree>root</a> is a <a for=/>shadow root</a>, set
+ <li><p>If {{Event/target}}'s <a for=tree>root</a> is a <a for=/>shadow root</a>, then set
  <var>event</var>'s {{Event/target}} attribute and <var>event</var>'s <a>relatedTarget</a> to null.
 
  <li><p>Set <var>event</var>'s {{Event/currentTarget}} attribute to null.
--- a/dom.bs
+++ b/dom.bs
@@ -1256,6 +1256,9 @@ for discussion).
  <li><p>Set <var>event</var>'s {{Event/target}} attribute to null if {{Event/target}}'s root is a
  <a for=/>shadow root</a>.
 
+ <li><p>Set <var>event</var>'s <a>relatedTarget</a> to null if <a>relatedTarget</a>'s root is a
+ <a for=/>shadow root</a>.
+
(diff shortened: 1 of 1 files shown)


TakayoshiKochi:
@annevk @hayatoito PTAL

TakayoshiKochi:
This should be accompanied with a corresponding layout test - will be created as a separate PR for WPT repository.

dom.bs
```diff
@@ -1253,6 +1253,9 @@ for discussion).
 
  <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/NONE}}.
 
+ <li><p>Set <var>event</var>'s {{Event/target}} attribute to null if {{Event/target}}'s root is a
+ <a for=/>shadow root</a> and <var>event</var>'s <a>composed flag</a> is unset.
```
hayatoito: > and <var>event</var>'s <a>composed flag</a> is unset.

We should remove this condition.
We need to clear event.target even if composed flag is set because target and relatedTarget can be identical in the middle and event path can be *trimmed* for that case.

hayatoito:
> This should be accompanied with a corresponding layout test - will be created as a separate PR for WPT repository.

Yeah, I am happy to review PR for WPT too.

dom.bs
TakayoshiKochi: Good catch. So we can determine if the event is stopped at shadow root just by looking at whether target's root is a shadow root.

TakayoshiKochi:
@hayatoito @annevk PTAL for `relatedTarget` addition.

Also, the corresponding WPT pull request is at
https://github.com/w3c/web-platform-tests/pull/7481
(via https://chromium-review.googlesource.com/c/chromium/src/+/681894)

annevk:
@TakayoshiKochi I don't quite know how `relatedTarget` works in all cases. Can we end up with a situation where target is null and relatedTarget is not? Or vice versa? Should we perhaps set both to null if either condition is true?

annevk:
Also, is currentTarget affected?

dom.bs
```diff
@@ -1253,6 +1253,12 @@ for discussion).
 
  <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/NONE}}.
 
+ <li><p>Set <var>event</var>'s {{Event/target}} attribute to null if {{Event/target}}'s root is a
```
annevk: We should link "root" here and below as well.

TakayoshiKochi:
> @TakayoshiKochi I don't quite know how relatedTarget works in all cases. Can we end up with a situation where target is null and relatedTarget is not? Or vice versa? Should we perhaps set both to null if either condition is true?

With the status quo, if an event bubbling ends up at a shadow root, its `relatedTarget` can be either in outside shadow tree or some shadow tree. If the former, `target` will become null but `relatedTarget` remain in my current spec patch - which may be weird and does not sound useful, even though it does not leak shadow tree.

On the other hand, once `event.target` is outside of any shadow tree, its corresponding `relatedTarget` should be also in the tree as a result of retargeting.

So I'm in favor of your idea, to set both to null if either condition is true.

> Also, is currentTarget affected?

`currentTarget` is already covered in step 16 (in the original spec) so that it will be set to null.

annevk:
Given what you're saying there, we only have to check target, right? Since relatedTarget is either equal or exposes less.

TakayoshiKochi:
> Given what you're saying there, we only have to check target, right? Since relatedTarget is either equal or exposes less.

Absolutely.  Thanks for pointing out.

annevk:
Thanks, my last change request is to put the condition first. "If X, then set Y to Z." is generally how we write these steps.

Given there's tests I think this is ready to merge then. Thanks a lot!

dom.bs
```diff
@@ -1253,6 +1253,9 @@ for discussion).
 
  <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/NONE}}.
 
+ <li><p>If {{Event/target}}'s <a for=tree>root</a> is a <a for=/>shadow root</a>, set
```

TakayoshiKochi: Thanks, done.

Will add a test case there for relatedTarget is in document tree while target is in shadow tree and event stops at shadow root to make sure both are set to null.

hayatoito:
LGTM

PR #535, opened 2017-11-09
Compute everything needed for composedPath() during dispatch

### LINKED ISSUE #525: event.composedPath() may return different results if DOM is modified between two calls
If I read the spec right, composedPath() relies on the current tree structure to figure out what to return.
So, if a listener does then something like

```js
var cp = event.composedPath();
/*move nodes around*/
var cp2 = event.composedPath();
```

cp and cp2 may contain different nodes, which is surprising, since the internal path for the event hasn't changed.
In particular, relying on "shadow-including inclusive ancestor" in  https://dom.spec.whatwg.org/#concept-closed-shadow-hidden in https://dom.spec.whatwg.org/#dom-event-composedpath looks wrong

annevk:
Fixes #525.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/composedpath/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/fb6638f...9857343.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1167,7 +1167,8 @@ for discussion).
  <li><p>Let <var>activationTarget</var> be <var>target</var>, if <var>isActivationEvent</var> is
  true and <var>target</var> has <a for=EventTarget>activation behavior</a>, and null otherwise.
 
- <li><p>Let <var>slotable</var> be null.
+ <li><p>Let <var>slotable</var> be <var>target</var>, if <var>target</var> is
+ <a for=slotable>assigned</a>, and null otherwise.
 
  <li><p>Let <var>slot-in-closed-tree</var> be false.
 
--- a/dom.bs
+++ b/dom.bs
@@ -16,6 +16,7 @@ Logo: https://resources.whatwg.org/logo-dom.svg
 Abstract: DOM defines a platform-neutral model for events, aborting activities, and node trees.
 Ignored Terms: EmptyString, Array, Document
 Boilerplate: omit feedback-header, omit conformance
+Indent: 1
 </pre>
 
 <!--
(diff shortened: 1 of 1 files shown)


annevk:
@tabatkins I get
```
Error running preprocessor, returned code: 1.
\033[1;31mFATAL ERROR:\033[0m Line 588 isn't indented enough (needs 1 indent) to be valid Markdown:
"  &lt;/ol>"
\033[7;31m ✘ \033[0m Did not generate, due to fatal errors
```
but I didn't even touch that line in this commit (nor does it contain `</ol>`).

annevk:
@hayatoito @smaug---- while the Preview link in OP is broken, the Diff link does somehow work and shows you the changes rendered in HTML. Hope that helps.

dom.bs
```diff
@@ -1151,22 +1158,42 @@ for discussion).
...
+ <li><p><a>Append to an event path</a> with <var>event</var>, <var>target</var>,
+ <var>targetOverride</var>, <var>relatedTarget</var>, and false.
 
  <li><p>Let <var>isActivationEvent</var> be true, if <var>event</var> is a {{MouseEvent}} object and
  <var>event</var>'s {{Event/type}} attribute is "<code>click</code>", and false otherwise.
 
  <li><p>Let <var>activationTarget</var> be <var>target</var>, if <var>isActivationEvent</var> is
  true and <var>target</var> has <a for=EventTarget>activation behavior</a>, and null otherwise.
 
+ <li><p>Let <var>slotable</var> be null.
```
hayatoito: We should set slotable to target if target is assigned, as we are doing in 11.2, here too.

tabatkins:
Ugh, I have a fix for the line numbers pointing wrong on my laptop (it's caused by me badly handling comment-stripping).  It's trying to point to one of the lines in the early-600s, because you're using a 1-space indent but haven't told Bikeshed that (it defaults to assuming a 4-space indent for markdown purposes - this includes html).

tabatkins:
And fix is finished and pushed - it now correctly points out the problem is on line 631.  You need to use a consistent indentation level thruout your spec - either set `Indent: 1` in your metadata (and then correct your anchor/link-defaults blocks to use a 1-space indent) or switch to using a 4-space indent for your HTML. (Or always use tabs - 1 tab = 1 indent no matter what, and tabs in `pre` are automatically converted to 2-space indents in the output file.)

annevk:
@hayatoito @smaug----  are either of you planning on writing tests covering all the various cases discussed?

hayatoito:
Yeah, let me add WPT on this.

hayatoito:
The test was added at: https://github.com/w3c/web-platform-tests/pull/8385

annevk:
Thank you!

PR #537, opened 2017-11-21
Stop using unit of related similar-origin browsing contexts

annevk:
Use similar-origin window agent instead.

***
[Preview](https://dom.spec.whatwg.org/branch-snapshots/annevk/agents/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/dom/56ca18c...4709d45.html)

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2078,16 +2078,16 @@ steps:
 
 <h5 id=signaling-slot-change>Signaling slot change</h5>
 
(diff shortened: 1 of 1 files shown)


annevk:
This seems like the right thing to do, but I'm going to punt on this for now until we make more progress in HTML defining the lifetime of all these things.

Issue #278, opened 2016-07-06
Figure out what to do with XMLDocument


domenic:
Right now `createDocument` returns an `XMLDocument`. It has this note:

> This method creates an XMLDocument rather than a normal document. They are identical except for the addition of the load() method deployed content relies upon.

Since load() is gone as of https://github.com/whatwg/html/commit/523f7a8773d2ab8a1eb0da6510651e8c5d2a7531, this note needs to be revised at the very least.

Edit: it appears @annevk already removed the note in https://github.com/whatwg/dom/commit/1595762e49358c7da9f6d6f284a3a3ae46987256 :)

I see two options:
- Make XMLDocument another alias of Document, just like HTMLDocument is currently specced to be. Remove all references to XMLDocument from DOM, and have HTML add a sentence about the alias.
- Keep the current empty-derived-interface setup in DOM. Maybe add some disparaging comments about this being a legacy feature.

Personally, I lean toward making it an alias, unless any implementers object. I think that the Document = HTMLDocument = XMLDocument story is a pretty good one, and in Blink at least, we're pretty close to implementing it---there are only a very small number of things left on HTMLDocument, none of which look very problematic. So I'd prefer to continue on that course.

There may be more history here (e.g. implementer objections) of which I am not aware.

cdumez:
FYI, Safari 9 and earlier had XMLDocument as an alias to Document. However, Safari 10 (which is in public beta) has an XMLDocument interface. We added it (even though empty) in an attempt to align with Firefox and Chrome.

That said, given that it is empty, I would be happy to go back to making it an alias to Document in WebKit.

annevk:
I guess the main problem with that is the Firefox-only code paths. 😟 I wonder how prevalent those still are.

foolip:
If Safari had it as an alias for a long time, then it's almost certainly web compatible to make it so. I think we should try to end up in one of these two states:

1. Both HTMLDocument and XMLDocument are aliases of Document.
2. HTMLDocument and XMLDocument are both empty interfaces.

In other words, what to do with XMLDocument I think depends on https://github.com/whatwg/dom/issues/221

foolip:
> HTMLDocument and XMLDocument are both empty interfaces.

In this case, I'm also skeptical of keeping the Document constructor even though Gecko has it already. It would (I think?) be the only way of getting a Document instance that's neither an HTMLDocument nor an XMLDocument instance, which seems like an annoying corner. I didn't dig deeply though.

[...]

Issue #221, opened 2016-04-14
Un-merge Document and HTMLDocument


ayg:
Does anyone object?  It's been years, nobody has done it, it's [claimed](https://www.w3.org/Bugs/Public/show_bug.cgi?id=22960#c2) that IE did it and then un-did it.  Boris has [serious reservations](https://bugzilla.mozilla.org/show_bug.cgi?id=897815#c9) about web-compatibility.  Is there any actual reason to merge them?  Obviously, it makes sense to keep APIs on Document where reasonable, but I don't see why it needs to be wholesale.

ayg:
@bz

bzbarsky:
Note that my serious reservations are not based on data, but rather on lack of data.  I could be convinced that this is ok, with actual data and analysis...

foolip:
There has been some efforts to move things from `HTMLDocument` to `Document` in Blink, and it's been successful so far. https://bugs.chromium.org/p/chromium/issues/detail?id=238368

Of what [remains](https://chromium.googlesource.com/chromium/src/+/fa50402230d37afe53a6e2d15d36a8285e1081e0/third_party/WebKit/Source/core/html/HTMLDocument.idl) I think that only `document.all` is worth worrying about. It's already on Document internally, but the risk is that some code depends on `xmlDocument.all` not existing.

I don't have time to spend on this *right now*, but if it'll get reverted in the spec I'd like a chance to try out the final changes in Blink, so let me know if there's a deadline.

bzbarsky:
Right, moving particular APIs to Document might well make a lot of sense.  Even moving them all might make sense.  What doesn't make sense to me is doing the move in a single commit with no attempt to do per-property analysis.

ayg:
So how this: leave the spec as-is for now, start moving properties to Document on a per-case basis, and if we find any properties we don't want to try moving, change the spec to reintroduce HTMLDocument with those properties.  Sound good to everyone?

foolip:
Yep, that sounds good. I'm pretty optimistic about Blink being able to make `HTMLDocument` empty, but there could be engine-specific compat issues. Not sure how risky the final step of making `HTMLDocument` an alias is, we'll just have to try it.

ayg:
Okay, so let's all try that.

cdumez:
What's the benefit of merging HTMLDocument into Document? Do we really want various HTML-specific attributes / operations to be exposed on non-HTML documents?

[...]

PR #551, opened 2017-12-25
Specify [LegacyWindowAlias=HTMLDocument] on Document

TimothyGu:
In https://html.spec.whatwg.org/multipage/window-object.html#the-window-object there is:

> For historical reasons, [`Window`](https://html.spec.whatwg.org/multipage/window-object.html#window) objects must also have a writable, configurable, non-enumerable property named **`HTMLDocument`** whose value is the [`Document`](https://html.spec.whatwg.org/multipage/dom.html#document) [interface object](https://heycam.github.io/webidl/#dfn-interface-object).

I'd like to get rid of that using the still relatively new [[`LegacyWindowAlias`](https://heycam.github.io/webidl/#LegacyWindowAlias)] extended attribute. However, Web IDL forbids that extended attribute from being used on partial interface definitions, and thus I had to open the PR here rather than as HTML. As much as I hate putting more HTML-specific things in DOM, I think this is minor enough to be okay.

As soon as this gets merged I'll open a PR to HTML removing that paragraph.

/cc @tobie

***
<a href="https://whatpr.org/dom/551.html" title="Last updated on Dec 25, 2017, 12:19 PM GMT (aab303c)">Preview</a> | <a href="https://whatpr.org/dom/551/5222eee...aab303c.html" title="Last updated on Dec 25, 2017, 12:19 PM GMT (aab303c)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4715,7 +4715,8 @@ object may be returned as returned by an earlier call.
 
 <pre class=idl force="Document">
 [Constructor,
- Exposed=Window]
+ Exposed=Window,
+ LegacyWindowAlias=HTMLDocument]
 interface Document : Node {
   [SameObject] readonly attribute DOMImplementation implementation;
   readonly attribute USVString URL;


tobie:
LGTM

annevk:
So #221 and #278 are not resolved yet. It's unclear if any engine actually implements this, so I'm not sure it's a good idea to make this change as it might lead folks to think we made some kind of decision here.

domenic:
I think it'd still be a good idea to add this here as it's a non-normative change that makes things easier to understand while reading the spec. (And also easier for any implementations like Servo or jsdom that implement it exactly.)

We could separately add a warning about #221 and #278, but I'd like to see this merged if possible.

annevk:
When I talked it through with @foolip it sounds like the solution will be that we have `HTMLDocument` as a distinct empty interface (similar to `XMLDocument`) and HTML will need to carefully define when it gets allocated. `Document` will still define all the joint members. And I guess we get to keep `SVGDocument` too?

foolip:
Only Safari has SVGDocument (as an alias of XMLDocument) so we don't need that.

ExE-Boss:
Unfortunately, it looks like https://github.com/whatwg/html/issues/4792 might throw a spanner into this.

annevk:
Closing as it seems even more unlikely we'll end up using this at this point.

PR #566, opened 2018-02-02
Change handling of is="" in element creation and cloning

domenic:
Part of fixing https://github.com/whatwg/html/issues/3402. This changes two things:

* It fixes the cloning steps to consult the internal is value, instead of the is="" content attribute.
* It removes the step from createElement()/createElementNS() that would insert an is="" attribute in the created element, as we've now decided that is="" is a way of communicating to the parser, and not a general-purpose indicator of customized built-in-ness.

---

This is coupled with a HTML change I will be submitting shortly, to the serializer (and to several examples).

@tkent-google, will you be able to help with tests for this? I am happy to review.

***
<a href="https://whatpr.org/dom/566.html" title="Last updated on Feb 2, 2018, 5:56 PM GMT (d510371)">Preview</a> | <a href="https://whatpr.org/dom/566/676fcbe...d510371.html" title="Last updated on Feb 2, 2018, 5:56 PM GMT (d510371)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4180,9 +4180,9 @@ dom-Range-extractContents, dom-Range-cloneContents -->
   <ol>
    <li><p>Let <var>copy</var> be the result of <a>creating an element</a>, given
    <var>document</var>, <var>node</var>'s <a for=Element>local name</a>, <var>node</var>'s
-   <a for=Element>namespace</a>, <var>node</var>'s <a for=Element>namespace prefix</a>, and the
-   value of <var>node</var>'s <code>is</code> attribute if present (or null if not). The
-   <var>synchronous custom elements flag</var> should be unset.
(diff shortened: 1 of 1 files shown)


tkent-google:
> @tkent-google, will you be able to help with tests for this? I am happy to review.

Yes.  I'll add/update tests when I'll update Chromium implementation.

tkent-google:
> > @tkent-google, will you be able to help with tests for this? I am happy to review.
>
> Yes. I'll add/update tests when I'll update Chromium implementation.

Test PR: https://github.com/w3c/web-platform-tests/pull/9505

smaug----:
So if there is definition for custom element (non built-in), is value is set to null, but if there isn't definition yet, is value can be possibly set to non-null. That is weird.

smaug----:
When doing something like 'new CustomBuiltIn()', is value isn't set, so cloning the element doesn't work.

PR #569, opened 2018-02-09
Following an abort signal

jakearchibald:
This PR makes it easier for one abort signal to follow another. This should avoid gotcha cases like https://github.com/whatwg/fetch/issues/668.

***
<a href="https://whatpr.org/dom/569.html" title="Last updated on Feb 12, 2018, 8:34 AM GMT (9ddd54f)">Preview</a> | <a href="https://whatpr.org/dom/569/cebdfa1...9ddd54f.html" title="Last updated on Feb 12, 2018, 8:34 AM GMT (9ddd54f)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1651,8 +1651,9 @@ them. For instance, if the operation has already completed.
  <li><p>[=Fire an event=] named <code>abort</code> at <var>signal</var>.
 </ol>
 
-<p>A <var>followingSignal</var> (an {{AbortSignal}}) <dfn export for=AbortSignal>follows</dfn> a
-<var>parentSignal</var> (an {{AbortSignal}}) by running these steps:
+<p>A <var>followingSignal</var> (an {{AbortSignal}}) is made to
+<dfn export for=AbortSignal>follow</dfn> a <var>parentSignal</var> (an {{AbortSignal}}) by running
+these steps:
 
 <ol>
  <li><p>If <var>followingSignal</var>'s [=AbortSignal/aborted flag=] is set, then return.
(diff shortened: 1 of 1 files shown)


jakearchibald:
Unrelated, I also fixed a reference to "resolve".

dom.bs
```diff
@@ -1651,6 +1651,24 @@ them. For instance, if the operation has already completed.
  <li><p>[=Fire an event=] named <code>abort</code> at <var>signal</var>.
 </ol>
 
+<p>A <var>followingSignal</var> (an {{AbortSignal}}) <dfn export for=AbortSignal>follows</dfn> a
+<var>parentSignal</var> (an {{AbortSignal}}) by running these steps:
```
domenic: Nit: I think I'd rather make the algorithm have a name that isn't an adjective, but instead a verb. E.g. "A followingSignal is made to follow a parentSignal by..."

jakearchibald: Would the usage remain "signalA follows signalB" or should that change to "Make signalB follow signalB".

I struggled to find wording around the usage 😕.

domenic: Probably "make signalB follow signalA". I guess I'd do `lt="follow"` then to make things easier.

jakearchibald:
Nits addressed here & https://github.com/whatwg/fetch/pull/670

domenic:
LGTM. I'll merge while @annevk is on vacation since this is actively causing implementer confusion so it's best to stamp it out early. Hopefully he can do any additional review when he gets back.

PR #582, opened 2018-03-04
Change DOMTokenList's replace() to return boolean

### LINKED ISSUE #577: Return value for DOMTokenList's replace()
DOMTokenList's `replace()` method does not currently have a specified return value. While using it I've had code which takes different paths depending on whether a replacement occurred or not. To find out if this was the case I must currently use separate conditions with `contains()`.

Could `replace(oldToken, newToken)` return a boolean indicating true if oldToken was indeed present and got replaced by newToken? This would help with writing cleaner code, and there's some precedent in the `toggle(token)`-method which returns true or false depending on whether the call resulted in the addition or removal of the token.

Given the relatively recent introduction of `replace()`, there's probably a good chance that this change is web compatible.

Zirro:
This specifies a return value for DOMTokenList's `replace(token, newToken)` method, making it return true if token was replaced by newToken and false otherwise. This is useful for code which takes different paths depending on whether a replacement occurred, avoiding the need for an extra condition using `contains()`.

I have modified the [Element-classlist.html](https://github.com/w3c/web-platform-tests/blob/master/dom/nodes/Element-classlist.html) test to include [testing](https://gist.github.com/Zirro/b0fb873ccb86eb3bb6596833558abca4#file-element-classlist-html-L406) of this change, and will submit a pull request to the WPT repository if it is approved here.

Fixes #577.

***
<a href="https://whatpr.org/dom/582.html" title="Last updated on Mar 7, 2018, 6:10 PM GMT (fd0157c)">Preview</a> | <a href="https://whatpr.org/dom/582/c05ca60...fd0157c.html" title="Last updated on Mar 7, 2018, 6:10 PM GMT (fd0157c)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9527,7 +9527,7 @@ interface DOMTokenList {
   [CEReactions] void add(DOMString... tokens);
   [CEReactions] void remove(DOMString... tokens);
   [CEReactions] boolean toggle(DOMString token, optional boolean force);
-  [CEReactions] void replace(DOMString token, DOMString newToken);
+  [CEReactions] boolean replace(DOMString token, DOMString newToken);
   boolean supports(DOMString token);
   [CEReactions] stringifier attribute DOMString value;
   iterable&lt;DOMString>;
@@ -9646,6 +9646,7 @@ are to return the result of running <a>get an attribute value</a> given the asso
  <dt><code><var>tokenlist</var> . <a method for=DOMTokenList lt=replace()>replace(<var>token</var>, <var>newToken</var>)</a></code>
  <dd>
   <p>Replaces <var>token</var> with <var>newToken</var>.
+  <p>Returns true if <var>token</var> was replaced with <var>newToken</var>, and false otherwise.
   <p>Throws a "{{SyntaxError!!exception}}" {{DOMException}} if one of the arguments is the empty
   string.
   <p>Throws an "{{InvalidCharacterError!!exception}}" {{DOMException}} if one of the arguments
@@ -9771,12 +9772,14 @@ method, when invoked, must run these steps:
  <a>throw</a> an "{{InvalidCharacterError!!exception}}" {{DOMException}}.
 
  <li><p>If <a>context object</a>'s <a>token set</a> does not <a for=set>contain</a>
- <var>token</var>, then return.</p>
+ <var>token</var>, then return false.</p>
 
  <li><p><a for=set>Replace</a> <var>token</var> in <a>context object</a>'s <a>token set</a> with
  <var>newToken</var>.
 
  <li><p>Run the <a>update steps</a>.
+
+ <li><p>Return true.
 </ol>
 
 <p class="note no-backref">The <a>update steps</a> are not always run for {{DOMTokenList/replace()}}
@@ -10092,6 +10095,7 @@ Kirill Topolyan,
 Koji Ishii,
 Lachlan Hunt,
 Lauren Wood,
+Magne Andersson,
 Majid Valipour,
 Malte Ubl,
 Manish Goregaokar,


dom.bs
```diff
@@ -9771,12 +9772,14 @@ method, when invoked, must run these steps:
  <a>throw</a> an "{{InvalidCharacterError!!exception}}" {{DOMException}}.
 
  <li><p>If <a>context object</a>'s <a>token set</a> does not <a for=set>contain</a>
- <var>token</var>, then return.</p>
+ <var>token</var>, return false.</p>
```
annevk: Please keep "then".

Zirro: Okay, I opted to remove "then" after finding that the phrasing ", return [bool]" was more common than ", then return [bool]" in algorithms throughout the document. Would you like me to change it elsewhere too, in a separate pull request?

annevk: Yeah, if you want.

annevk:
This makes it return true when you replace an existing item with the same item. I think that's probably okay, since it does cause an actual mutation to take place, but I'm not sure if @Zirro fully considered it.

Zirro:
Thanks for the feedback. I've restored "then".

> This makes it return true when you replace an existing item with the same item.

I did consider this, but couldn't see a compelling scenario where another behaviour would be preferable. I'd like to hear in case someone does, though.

---

Should I create a pull request in the WPT repository for the updated tests at this stage, or wait for further feedback / implementer interest?

I'd also be interested in submitting patches for this change to the major browser engines (assuming I can get them to compile), though I'll wait until this has been merged into the spec first.

annevk:
I guess it'd be good to get explicit support, even if it's a small change. You can start filing browser bugs though and creating a web-platform-tests PR. It seems unlikely this'll be controversial.

@foolip @bzbarsky @dstorey @cdumez or is it? (The proposal is to make `DOMTokenList`'s `replace()` return a boolean indicating it replaced something or not.)

bzbarsky:
Adding

    checkReplace("a b", "d", "a", false, "a b");

to the new test does catch the Firefox bug.

foolip:
We should wait for @dstorey for a bit before merging though. @Zirro, do you want to write tests for this? This is the test: https://wpt.fyi/dom/nodes/Element-classlist.html

Zirro:
@foolip I've submitted a pull request to the WPT repository now: https://github.com/w3c/web-platform-tests/pull/9920

> Basically it would return true if and only if it actually set the attribute (after the initial canonicalization bit), right?

@bzbarsky I believe this to be the case since the earlier condition - `If context object’s token set does not contain token, then return false` - should rule out any paths where the attribute does not get set. I think this should be covered by the existing tests once they've been updated, but let me know if I've missed something regarding this.

bzbarsky:
> I think this should be covered by the existing tests once they've been updated

The case I mention in https://github.com/whatwg/dom/pull/582#issuecomment-371361615 doesn't seem to be covered: trying to replace a token that is not present with one that is.

Zirro:
Well spotted, @bzbarsky! It wasn't clear to me how your test differed from the others while I was preparing the commit this morning, but you are correct - there's currently no test covering the replacement of a non-existent token with one that exists. I've added it to the WPT pull request now.

dstorey:
We're ok with this change. 

We have just implemented replace() in EdgeHTML17, and it is too late in the release to make changes such as this, so it isn't something that would get fixed in Edge right away.

annevk:
@Zirro thanks for pushing this. The only thing remaining is filing bugs against browsers. Giving them a pointer to this PR and the web-platform-tests PR. Once that's done I'll land the tests and this. Let me know if you need assistance with that.

Zirro:
@annevk 👍 Done:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1444909
- https://bugs.webkit.org/show_bug.cgi?id=183567
- https://bugs.chromium.org/p/chromium/issues/detail?id=820988
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16344078

dstorey:
This has now been implemented in Edge. It will likely be included in EdgeHTML 18.

foolip:
The Chromium issue was also fixed, should be in Chrome 67. And, looking at https://master-dot-wptdashboard.appspot.com/results/dom/nodes/Element-classlist.html?label=experimental I noticed that it's also fixed in Firefox.

And https://bugs.webkit.org/show_bug.cgi?id=183567 is fixed, so that means we're all done. That was fast :)

bzbarsky:
Yeah, the Firefox fix will ship in Firefox 61 in late June.

Issue #580, opened 2018-02-28
dispatch algorithm doesn't seem to handle the case when target or relatedTarget is Window object


smaug----:

annevk:
@smaug---- could you please provide more context. What exactly is there to handle?

smaug----:
Things like ''Let relatedTarget be the result of retargeting event’s relatedTarget against target'
where retargeting itself doesn't seem to work with Window.
"If A’s root .." what is Window's root? I couldn't find any definition for that.

annevk:
Thanks. Sounds like https://dom.spec.whatwg.org/#retarget should check if _A_ is a node, first, and return _A_ if not.

And then similar checks in 11.1.3, 11.2, 11.4, and 18.

smaug----:
Are there cases when target is Window, but relatedTarget something in a shadow DOM? Or vice versa

annevk:
Hmm yeah, I'm not sure. Maybe someone from @whatwg/components knows. I vaguely recall a discussion about this before and I'm somewhat surprised it resurfaced.

The other thing I wonder about here is when someone dispatches a synthetic event. Do we assume that in that case relatedTarget the concept is not set, even though the relatedTarget attribute can return something controlled by the developer?

hayatoito:
I remember I have tried to fix compsoedPath() for Window at
https://github.com/whatwg/dom/pull/327, however it wouldn't be enough.

I agree that we have to fix the standard so that we can consider non-node EventTarget nicely.
Basic idea should be:
-  Non-node EventTarget event.target and event.relatedTarget should be visible from anywhere.
-  From Window, any node in a shadow tree shouldn't be visible.
-  From Window, any node in a document tree should be visible.

annevk:
I think I'll try to merge this into #585 since it affects that step too.

annevk:
Pushed a fix for this too.

Issue #414, opened 2017-02-20
"create an event" doesn't work for certain UI events


domenic:
https://dom.spec.whatwg.org/commit-snapshots/26cfadd185649915bb73d147c8a7a41335cb7ab0/#constructing-events

Consider KeyboardEvent. The problem is that there are many elements in [EventModifierInit](https://w3c.github.io/uievents/#dictdef-eventmodifierinit) (which KeyboardEventInit derives from) which are not attributes. So that's not great.

One way to try to fix this would be something like trying to invoke the constructor with the default value, instead of just setting the attributes? And then we assume that the override defined in https://w3c.github.io/uievents/#event-constructors is good enough?

annevk:
Can't they simply define their own constructor?

domenic:
You mean, define their own version of "create an event"? I haven't checked, but it seems probable that e.g. HTML uses "create an event" to create mouse events... we'd have to go update HTML and any other specs to use the new UI events "create an event" instead of the DOM "create an event", if that's what you're suggesting.

annevk:
Maybe we should just inline https://w3c.github.io/uievents/#event-constructors as a legacy exception?

domenic:
I like that idea. @garykac, thoughts?

garykac:
Do you mean moving this algorithm snippet:
```
if Mouse or Keyboard event
  foreach EventModifierInit argument
    ...
```
into the DOM spec (and removing section 3.6 from UIEvents), or having UIEvents expose an function that the DOM spec would call?

The advantage of the latter is that it will be easier for me to modify the parts related to the "key modifier names", which is something that I need to do.

annevk:
We could do the latter as well, provided it's somehow limited to these event classes. I mainly want to dissuade others from trying to copy that pattern, although I guess we should consult with some implementers to make sure this doesn't happen in more places before we commit.

domenic:
I guess the main idea is to make https://dom.spec.whatwg.org/commit-snapshots/26cfadd185649915bb73d147c8a7a41335cb7ab0/#constructing-events work for UI events too. We should be able to do so in a way that calls out to UI events so you can maintain the steps there, mostly.

So concretely I think the best thing would be to add to UI events an algorithm "initialize a mouse or keyboard event from a initialization dictionary" (better name welcome) which takes an event and a dictionary.

[...]

Issue #392, opened 2017-01-05
"If target’s root is a shadow-including inclusiv..."


tbodt:
https://dom.spec.whatwg.org/commit-snapshots/e8a9f67886e554d81cf85277f61c6b1cd0ea6b44/#dispatching-events

> If target’s root is a shadow\-including inclusive ancestor of parent, then:​

If parent is the window object, which it could be since document's get the parent algorithm normally returns the window object, then it doesn't make sense to talk about the shadow-including inclusive ancestor of parent. Window objects do not participate in a tree.

annevk:
Good point, it seems the previous step also suffers from that. Should be easy enough to exclude Window object from those steps somehow.

annevk:
Duplicating against a newer issue #580 that has more information. I'll make sure to add Theodore Dubois to the acknowledgments section however. Let me know if you don't want that.

PR #585, opened 2018-03-08
Correct setting of target and relatedTargets post-dispatch

annevk:
@whatwg/components this fixes the following:

* Adds "touch target list" as a concept hold by event, so touch events can be retargeted. (This will require some special handling for Touch objects, but I think it's workable.)
* It now resets target/relatedTarget/touch target list post-dispatch based on pre-dispatch conditions.
* It makes dispatch account for non-nodes.

(The one thing remaining here is having a good look at event constructors and making sure we can make that work for the relatedTargets design. I think that requires some kind of callback type thing.)

***
<a href="https://whatpr.org/dom/585.html" title="Last updated on Mar 27, 2018, 10:30 AM GMT (67e64b4)">Preview</a> | <a href="https://whatpr.org/dom/585/5be84cc...67e64b4.html" title="Last updated on Mar 27, 2018, 10:30 AM GMT (67e64b4)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1330,8 +1330,8 @@ for discussion).
 </ol>
 
 <p>To <dfn noexport id=concept-event-path-append>append to an event path</dfn>, given an
-<var>event</var>, <var>target</var>, <var>targetOverride</var>, <var>relatedTargets</var>, and a
-<var>slot-in-closed-tree</var>, run these steps:</p>
+<var>event</var>, <var>target</var>, <var>targetOverride</var>, <var>relatedTarget</var>,
+<var>touchTargets</var>, and a <var>slot-in-closed-tree</var>, run these steps:</p>
 
 <ol>
  <li><p>Let <var>root-of-closed-tree</var> be false.
@@ -1341,7 +1341,8 @@ for discussion).
 
  <li><p><a for=list>Append</a> a new <a for=/>struct</a> to <var>event</var>'s <a for=Event>path</a>
  whose <a for=Event/path>item</a> is <var>target</var>, <a for=Event/path>target</a> is
- <var>targetOverride</var>, <a for=Event/path>relatedTargets</a> is <var>relatedTargets</var>,
+ <var>targetOverride</var>, <a for=Event/path>relatedTarget</a> is <var>relatedTarget</var>,
+ <a for=Event/path>touch target list</a> is <var>touchTargets</var>,
  <a for=Event/path>root-of-closed-tree</a> is <var>root-of-closed-tree</var>, and
  <a for=Event/path>slot-in-closed-tree</a> is <var>slot-in-closed-tree</var>.
 </ol>
--- a/dom.bs
+++ b/dom.bs
@@ -1173,140 +1173,148 @@ for discussion).
  <a for=Event>relatedTarget</a> against <var>target</var>.
 
  <li>
-  <p>If <var>target</var> is <var>relatedTarget</var> and <var>target</var> is not
-  <var>event</var>'s <a for=Event>relatedTarget</a>, then:
+  <p>If <var>target</var> is not <var>relatedTarget</var> or <var>target</var> is <var>event</var>'s
+  <a for=Event>relatedTarget</a>, then:
 
   <ol>
-   <li><p><a>Clear an event</a> with <var>event</var> and <var>clearTargets</var>.
-
-   <li><p>Return true.
-  </ol>
-
- <li><p>Let <var>touchTargets</var> be a new <a for=/>list</a>.
-
- <li><p><a for=list>For each</a> <var>touchTarget</var> of <var>event</var>'s
- <a for=Event>touch target list</a>, <a for=list>append</a> the result of <a>retargeting</a>
- <var>touchTarget</var> against <var>target</var> to <var>touchTargets</var>.
-
- <li><p><a>Append to an event path</a> with <var>event</var>, <var>target</var>,
- <var>targetOverride</var>, <var>relatedTarget</var>, <var>touchTargets</var>, and false.
-
- <li><p>Let <var>isActivationEvent</var> be true, if <var>event</var> is a {{MouseEvent}} object and
- <var>event</var>'s {{Event/type}} attribute is "<code>click</code>", and false otherwise.
-
- <li><p>Let <var>activationTarget</var> be <var>target</var>, if <var>isActivationEvent</var> is
- true and <var>target</var> has <a for=EventTarget>activation behavior</a>, and null otherwise.
-
- <li><p>Let <var>slotable</var> be <var>target</var>, if <var>target</var> is
- <a for=slotable>assigned</a>, and null otherwise.
-
- <li><p>Let <var>slot-in-closed-tree</var> be false.
-
- <li><p>Let <var>parent</var> be the result of invoking <var>target</var>'s <a>get the parent</a>
- with <var>event</var>.
-
- <li>
-  <p>While <var>parent</var> is non-null:</p>
-
-  <ol>
-   <li>
-    <p>If <var>slotable</var> is non-null:
-
-    <ol>
-     <li><p>Assert: <var>parent</var> is a <a for=/>slot</a>.
-
-     <li><p>Set <var>slotable</var> to null.
-
-     <li><p>If <var>parent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose
-     <a for=ShadowRoot>mode</a> is "<code>closed</code>", then set <var>slot-in-closed-tree</var> to
-     true.
-    </ol>
-
-   <li><p>If <var>parent</var> is a <a>slotable</a> and is <a for=slotable>assigned</a>, then set
-   <var>slotable</var> to <var>parent</var>.
-
-   <li><p>Let <var>relatedTarget</var> be the result of <a>retargeting</a> <var>event</var>'s
-   <a for=Event>relatedTarget</a> against <var>parent</var>.
-
    <li><p>Let <var>touchTargets</var> be a new <a for=/>list</a>.
 
    <li><p><a for=list>For each</a> <var>touchTarget</var> of <var>event</var>'s
-   <a for=Event>touchTargets</a>, <a for=list>append</a> the result of <a>retargeting</a>
+   <a for=Event>touch target list</a>, <a for=list>append</a> the result of <a>retargeting</a>
    <var>touchTarget</var> against <var>target</var> to <var>touchTargets</var>.
 
+   <li><p><a>Append to an event path</a> with <var>event</var>, <var>target</var>,
+   <var>targetOverride</var>, <var>relatedTarget</var>, <var>touchTargets</var>, and false.
+
+   <li><p>Let <var>isActivationEvent</var> be true, if <var>event</var> is a {{MouseEvent}} object
+   and <var>event</var>'s {{Event/type}} attribute is "<code>click</code>", and false otherwise.
+
+   <li><p>Let <var>activationTarget</var> be <var>target</var>, if <var>isActivationEvent</var> is
+   true and <var>target</var> has <a for=EventTarget>activation behavior</a>, and null otherwise.
+
+   <li><p>Let <var>slotable</var> be <var>target</var>, if <var>target</var> is
+   <a for=slotable>assigned</a>, and null otherwise.
+
+   <li><p>Let <var>slot-in-closed-tree</var> be false.
+
+   <li><p>Let <var>parent</var> be the result of invoking <var>target</var>'s <a>get the parent</a>
+   with <var>event</var>.
+
    <li>
-    <p>If <var>parent</var> is a <a>node</a> and <var>target</var>'s <a for=tree>root</a> is a
-    <a>shadow-including inclusive ancestor</a> of <var>parent</var>, then:
+    <p>While <var>parent</var> is non-null:</p>
 
     <ol>
-     <li><p>If <var>isActivationEvent</var> is true, <var>event</var>'s {{Event/bubbles}} attribute
-     is true, <var>activationTarget</var> is null, and <var>parent</var> has
-     <a for=EventTarget>activation behavior</a>, then set <var>activationTarget</var> to
-     <var>parent</var>.
+     <li>
+      <p>If <var>slotable</var> is non-null:
 
-     <li><p><a>Append to an event path</a> with <var>event</var>, <var>parent</var>, null,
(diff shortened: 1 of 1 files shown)


hayatoito:
I would like to take a look, however, I will take a vacation for a few days. Let me review in the next week.

dom.bs
```diff
@@ -1138,15 +1137,22 @@ for discussion).
...
+
+ <li><p>Let <var>relatedTargets</var> be a new <a for=/>list</a>.
+
+ <li><p><a for=list>For each</a> <var>relatedTarget</var> of <var>event</var>'s
+ <a for=Event>relatedTargets</a>, <a for=list>append</a> the result of <a>retargeting</a>
+ <var>relatedTarget</var> against <var>target</var> to <var>relatedTargets</var>.
 
- <li><p>If <var>target</var> is <var>relatedTarget</var> and <var>target</var> is not
- <var>event</var>'s <a for=Event>relatedTarget</a>, then return true.
+ <li><p>If <var>relatedTargets</var> <a for=list>contains</a> <var>target</var> and
```
smaug----: I doubt this is quite right in Touch case. Need to think this case some more.
I have some such case in mind where one of the relatedTargets does contain target before retargeting and then retargeting makes also some other relatedTarget to be target.

dom.bs
```diff
@@ -1197,11 +1205,12 @@ for discussion).
...
 
      <li><p><a>Append to an event path</a> with <var>event</var>, <var>parent</var>, null,
-     <var>relatedTarget</var>, and <var>slot-in-closed-tree</var>.
+     <var>relatedTargets</var>, and <var>slot-in-closed-tree</var>.
     </ol>
 
-   <li><p>Otherwise, if <var>parent</var> and <var>relatedTarget</var> are identical, then set
-   <var>parent</var> to null.
+   <li><p>Otherwise, if <var>relatedTargets</var> <a for=list>contains</a> <var>parent</var>, then
+   set <var>parent</var> to null.
```
smaug----: Gut feeling is that also this has similar problem.

dom.bs
```diff
@@ -5751,9 +5757,21 @@ is an object or one of its <a>shadow-including ancestors</a>.
 <var>B</var>, repeat these steps until they return an object:</p>
 
 <ol>
- <li><p>If <var>A</var>'s <a for=tree>root</a> is not a <a for=/>shadow root</a>, or <var>A</var>'s
- <a for=tree>root</a> is a <a>shadow-including inclusive ancestor</a> of <var>B</var>, then return
- <var>A</var>.
+ <li><p>If <var>A</var> is a {{Window}} object, then set <var>A</var> to <var>A</var>'s
+ <a>associated <code>Document</code></a>.
```
smaug----: Hmm, retargeting Window gives Document? Shouldn't this just return Window

annevk: Hmm yeah, I guess we just want it to behave as if you passed document.

annevk: I guess I should just remove this step altogether then?

dom.bs
annevk: It seems https://github.com/w3c/web-platform-tests/blob/master/touch-events/touch-retargeting.html can be extended somehow for more test coverage for touch. It's not entirely clear to me how that works given that on desktop the events are not really exposed.

dom.bs
TakayoshiKochi: Because window is not an instance of a Node, it falls into the line 5767 (A is not a node)?
Sounds good  - but wondering what's the reasoning behind adding this line.

dom.bs
```diff
@@ -5751,9 +5757,21 @@ is an object or one of its <a>shadow-including ancestors</a>.
...
+
+ <li>
+  <p>If one of the following is true
+
+  <ul class=brief>
+   <li><var>A</var> is not a <a>node</a>
+   <li><var>B</var> is not a <a>node</a>
```
TakayoshiKochi: Basically the change for retargeting algorithm from the original is these 2 lines above, am I understanding correct?

dom.bs
annevk: I added this because of https://github.com/whatwg/dom/issues/580#issuecomment-369813057, but I didn't really think it through.

dom.bs
annevk: Yeah.

dom.bs
hayatoito: I think removing this step is not enough. 
e.g. If A is a node in a shadow tree and B is a Window, we should loop until A reaches the outer most shadow host.

The following should work, I think.

1. If one of the following is true

   - A is not a node
   - A’s root is not a shadow root
   - B is a node and A’s root is a shadow-including inclusive ancestor of B

   then return A.

2. Set A to A’s root’s host.

annevk: Thanks, that does look good.

dom.bs
hayatoito: Yup, I confirmed the Blink's implementation.
We need to distinguish *relatedTarget* and *relatedTargets* somehow because only *relatedTarget* (in the sense of before this PR) should be considered here.
"*relaterTagets* contains parent" causes a sort of *false positive*.

annevk: How should touch targets be adjusted though?

annevk: Since the feedback I got before was that whenever we have relatedTarget handling, we also need to consider touch target handling.

hayatoito: These steps always need some time to understand its intention exactly. :(

My intention is:
- All touch targets should be adjusted in the same way as relatedTarget should be adjusted. They both should be adjusted.
- However, there is one difference between relatedTarget and touch targets. Only relatedTarget should be used to decide to exit this while loop "11. While parent is non-null". This step sets *parent* to null, thus I think this is an attempt to exit the loop.
To decide to exit this loop or not, in other words, to decide whether an event should be stopped at a shadow root or not,  only relatedTarget should be used as its judgement. Touch targets shouldn't be used there.

Since TouchEvent doesn't look to have a relatedTarget to me, TouchEvent never set parent to null here. In other words, TouchEvent is never stopped at a shadow root even if its target and one of touch targets are same.

annevk: Okay, so we effectively need to have relatedTarget as an internal slot and touchTargets as an internal slot and most of the time we do the same thing for both, but not there and not at the other place @smaug---- pointed out (which is hidden now for some reason).

I can make that change. I'd love some help with tests for touch targets btw since I don't really have a good way to run them here.

annevk: (The alternative option is that we keep one slot but branch on event interface, but that seems a little icky, but let me know if you'd prefer that.)

hayatoito: I would love to help to write more tests, though I'm afraid I can write tests only for synthetic TouchEvents. I am not sure how to write a test for non-synthetic TouchEvents.

> (The alternative option is that we keep one slot but branch on event interface, but that seems a little icky, but let me know if you'd prefer that.)

I don't have any preference.

btw, I am wondering how these *relatedTargets* will be used in TouchEvent spec.
TouchEvents has 3 *relatedTargets* lists.

```
    readonly        attribute TouchList touches;
    readonly        attribute TouchList targetTouches;
    readonly        attribute TouchList changedTouches;
```

annevk: I was thinking that since those lists are all immutable, a flattened list backing them is fine, but that only works if retargeting is consistent regardless of which list.

hayatoito: I see. Your assumption is correct. Retargeting is consistent regardless of which list.

annevk:
The PR now separates relatedTarget and touch target list and only has the special cases for relatedTarget. I still need to define an event creation callback of sorts so we can define TouchTarget / FocusEvent et al properly. I was also thinking of incorporating #402 here since we'll have to test that case anyway.

dom.bs
```diff
@@ -1156,15 +1167,26 @@ for discussion).
...
+ <a for=Event>relatedTarget</a>, or an {{EventTarget}} object in <var>event</var>'s
+ <a for=Event>touch target list</a> is a <a>node</a> and its <a for=tree>root</a> is a
+ <a for=/>shadow root</a>, and false otherwise.
+
+ <li><p>Let <var>relatedTarget</var> be the result of <a>retargeting</a> <var>event</var>'s
+ <a for=Event>relatedTarget</a> against <var>target</var>.
 
  <li><p>If <var>target</var> is <var>relatedTarget</var> and <var>target</var> is not
  <var>event</var>'s <a for=Event>relatedTarget</a>, then return true.
+ <!-- XXX Since we return early, we might want to unset more flags here. See #402. -->
```
annevk: Should we also unset target et al here if clearTargetsPostDispatch is true? It seems that would be consistent... (If that's the case it seems we should turn this into a short algorithm that we invoke here and later on.)

hayatoito: I have never thought about this case seriously so far. In this early return, I assume that any event listener is never called.

In this case, we don't have to worry about any leaking of a node in a shadow tree via an event object. Users don't have any change to access an event object, other than the original event dispatcher.

btw, the consistency may matter. I'm fine to reset target, relatedTarget, and touch target list here too.

annevk: Yeah, I'm mostly wondering about it from a consistency perspective. I could see you'd have some custom dispatching code and later store the event object after dispatch in a global place and maybe you don't want things to be leaked there ever. Doesn't seem like a sound setup, but other than this case we do provide the guarantee that nothing would leak.

annevk: I've now made this change.

annevk:
I consider this ready apart from the event construction specification callback, which is being discussed in #414. The ongoing feedback here is much appreciated though, please keep it coming.

dom.bs
```diff
@@ -10070,6 +10112,7 @@ Tab Atkins,
 Takashi Sakamoto,
 Takayoshi Kochi,
 Theresa O'Connor,
+Theodore Dubois,
```
smaug----: Is this related to this pr?

annevk: Yes: https://github.com/whatwg/dom/issues/392.

dom.bs
```diff
@@ -1156,149 +1166,160 @@ for discussion).
...
+     <li><p>Set <var>slot-in-closed-tree</var> to false.
+    </ol>
 
-  <ol>
-   <li><p>Set <var>event</var>'s {{Event/target}} attribute to the <b>target</b> of the last tuple
-   in <var>event</var>'s <a for=Event>path</a>, that is either <var>tuple</var> or preceding
-   <var>tuple</var>, whose <b>target</b> is non-null.
+   <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/CAPTURING_PHASE}}.
 
-   <li><p>Set <var>event</var>'s <a for=Event>relatedTarget</a> to <var>tuple</var>'s
```
smaug----: So where is this retargeting happening now?

dom.bs
```diff
@@ -1338,6 +1360,18 @@ for discussion).
 <ol>
  <li><p>If <var>event</var>'s <a>stop propagation flag</a> is set, then return.
 
+ <li><p>Set <var>event</var>'s {{Event/target}} attribute to the <a for=Event/path>target</a> of the
+ last tuple in <var>event</var>'s <a for=Event>path</a>, that is either <var>tuple</var> or
```
smaug----: what does 'last' mean in this context? In which direction is that

dom.bs
```diff
@@ -1338,6 +1360,18 @@ for discussion).
...
+ preceding <var>tuple</var>, whose <a for=Event/path>target</a> is non-null.
+
+ <li><p>Set <var>event</var>'s <a for=Event>relatedTarget</a> to <var>tuple</var>'s
+ <a for=Event/path>relatedTarget</a>.
+
+ <li><p>Set <var>event</var>'s <a for=Event>touch target list</a> to <var>tuple</var>'s
```
smaug----: hmm, perhaps a bit confusing, but I guess should be enough for touch objects. They just need to have a pointer to the event, so that they can update their target form there (basically map their original target to retargeted target or so)

hayatoito:
Looks okay to me. Thank you for updating this!

annevk:
So thinking about this in light of the tests at `shadow-dom/event-post-dispatch.html` it seems this is still not entirely correct. With the patch is proposed, you can observe whether an event started out in a shadow tree or not.

I think _clearTargets_ should instead be based on event's path. You find the last tuple in event's path with a non-null target. And then if that tuple contains a node inside a shadow tree in one of the three target fields, you set _clearTargets_ to true.

annevk:
Hmm, and should https://whatpr.org/dom/585.html#concept-event-listener-invoke be updated to check the propagation flag later, so target et al still get updated? That would lead to problems too otherwise I think.

hayatoito:
Does this need another review? Please ping me if the PR is ready.

annevk:
@hayatoito I'd like input from you and @smaug---- on https://github.com/whatwg/dom/pull/585#issuecomment-374604473 and what a solution to that might look like. The other thing we need here are more tests I suppose.

smaug----:
I think that approach sounds good. So, in case all the targets after actual dispatch are in light DOM, targets wouldn't get cleared.

hayatoito:
> I think clearTargets should instead be based on event's path. You find the last tuple in event's path with a non-null target. And then if that tuple contains a node inside a shadow tree in one of the three target fields, you set clearTargets to true.

I see. This sounds a reasonable change. Nice catch.

annevk:
Tentative commit message:
```
Shadow: fix event dispatch

This makes a number of large changes to event dispatch:

* The algorithm can no longer return early with true. The observable
  aspect is that the event object is "uninitialized" and false might 
  be returned if the event object was already canceled.
* It accounts for Touch Events in shadow trees by adding the 
  "touch target list" which is somewhat similar to relatedTarget.
* Retargeting is now solely controlled by the event dispatch 
  algorithm. The hook for other specifications is removed.
* Retargeting now properly accounts for non-node objects.
* Dispatch in general accounts for non-node objects too now.
* If targets need to be unset to prevent exposure of a shadow tree
  this is now calculated prior to invoking listeners.

Initializing relatedTarget and "touch target list" is up to the event
classes that need it. #614 will add "event constructing steps" that
enable this.

Tests: https://github.com/w3c/web-platform-tests/pull/9919
(probably still needs work given today's changes)

This closes #402, fixes #561, fixes #562, fixes #580, and fixes #602.
```
So more review is welcome here (for the last two commits and this commit message in particular) and more tests.

Sorry it got so big, but I'm not sure this could be elegantly made smaller.

hayatoito:
I've reviewed it, and it looks okay to me. A really great work!

annevk:
Bugs:

* https://bugs.webkit.org/show_bug.cgi?id=184079
* https://bugs.chromium.org/p/chromium/issues/detail?id=826702
* https://bugzilla.mozilla.org/show_bug.cgi?id=1449560

(No bug for Edge since they haven't started yet as far as I'm aware.)

PR #589, opened 2018-03-12
Add AbstractRange and StaticRange

annevk:
This refactors large parts of the DOM Standard to reclassify Range
objects as "live ranges" rather than "ranges". The bits shared
between StaticRange and Range objects are put on a shared superclass
named AbstractRange.

This also introduces a "collapsed" definition and uses it throughout.

It does not contain everything from
https://w3c.github.io/staticrange/. I based this draft on the IDL
present in implementations, coupled with the desire for a superclass.
Extensions beyond this seem best addressed as follow-ups.

Tests: ...

***
<a href="https://whatpr.org/dom/589.html" title="Last updated on Mar 16, 2018, 7:34 AM GMT (71954ff)">Preview</a> | <a href="https://whatpr.org/dom/589/06dae65...71954ff.html" title="Last updated on Mar 16, 2018, 7:34 AM GMT (71954ff)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -936,7 +936,7 @@ object, an <a>event listener</a> is a broader concept as can be seen above.
 which takes an <a>event</a> <var>event</var>, and returns an {{EventTarget}} object. Unless
 specified otherwise it returns null.
 
-<p class="note no-backref"><a for=/>nodes</a>, <a for=/>shadow roots</a>, and <a>documents</a>
+<p class="note no-backref"><a for=/>Nodes</a>, <a for=/>shadow roots</a>, and <a>documents</a>
 override the <a>get the parent</a> algorithm.
 
 <p>Each {{EventTarget}} object can have an associated
@@ -3539,14 +3539,11 @@ be initialized to null.
 
 <h4 id=garbage-collection>Garbage collection</h4>
 
-<a for=/>nodes</a> have a strong reference to
-<a>registered observers</a> in their
-list of <a>registered observers</a>.
+<p><a for=/>Nodes</a> have a strong reference to <a>registered observers</a> in their list of
+<a>registered observers</a>.
 
-<a>Registered observers</a> in a
-<a>node</a>'s list of
-<a>registered observers</a> have a weak
-reference to the <a>node</a>.
+<p><a>Registered observers</a> in a <a for=/>node</a>'s list of <a>registered observers</a> have a
+weak reference to the <a for=/>node</a>.
 
 
 <h3 id=interface-node>Interface {{Node}}</h3>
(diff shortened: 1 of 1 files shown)


annevk:
* https://bugs.webkit.org/show_bug.cgi?id=183561 (change impl)
* https://bugs.chromium.org/p/chromium/issues/detail?id=820947 (change impl)
* https://bugzilla.mozilla.org/show_bug.cgi?id=1444847 (add impl)
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16341745/ (add impl)

annevk:
Given that there's no constructor I think the IDL test changes are sufficient. The other bits would need to be tested as part of https://w3c.github.io/input-events/index.html and other producers of these objects.

So this is ready to land after review.

dom.bs
```diff
@@ -3650,7 +3605,7 @@ be initialized to null.
 
 <h4 id=garbage-collection>Garbage collection</h4>
 
-<a>Nodes</a> have a strong reference to
+<a for=/>nodes</a> have a strong reference to
```
domenic: Should stay uppercase

dom.bs
```diff
@@ -936,8 +936,8 @@ fields above, an <a>event listener</a> is a broader concept.
 which takes an <a>event</a> <var>event</var>, and returns an {{EventTarget}} object. Unless
 specified otherwise it returns null.
 
-<p class="note no-backref"><a>Nodes</a>, <a for=/>shadow roots</a>, and <a>documents</a> override
-the <a>get the parent</a> algorithm.
+<p class="note no-backref"><a for=/>nodes</a>, <a for=/>shadow roots</a>, and <a>documents</a>
```
domenic: Should stay upppercase

dom.bs
```diff
@@ -7466,261 +7502,110 @@ interface Range {
...
+<var>range</var>'s <a for="live range">root</a>, and (<var>node</var>, 0) is
+<a for="boundary point">after</a> <var>range</var>'s <a for=range>start</a>, and
+(<var>node</var>, <var>node</var>'s <a>length</a>) is <a for="boundary point">before</a>
+<var>range</var>'s <a for=range>end</a>.
 
- <li>Return <a for=Range>before</a>.
-</ol>
+<p>A <a for=/>node</a> is
+<dfn export for="live range" id=partially-contained>partially contained</dfn> in a <a>live range</a>
+if its an <a for=tree>inclusive ancestor</a> of the <a>live range</a>'s <a for=range>start node</a>
```
domenic: "its" should be "it is" here

dom.bs
```diff
@@ -9844,7 +9711,7 @@ These are the changes made to the features described in
 
 {{Node}} now inherits from {{EventTarget}}.
 
-<a>Nodes</a> are implicitly
+<a for=/>nodes</a> are implicitly
```
domenic: Should still be uppercase

dom.bs
```diff
@@ -8896,29 +8763,29 @@ these steps:
 <ol>
  <li>Let <var>s</var> be the empty string.
```
domenic: This whole algorithm just says "start node" and "end node" a lot but should probably say "the context object's start node"/end node.

(Preexisting problem, though.)

annevk:
Yeah, I was thinking there should be more of an introduction at some point. But I'm not sure it makes much sense without the constructor as only input events can create these objects now and then I'd have to introduce those as well.

I have updated the note to be a little more specific.

dstorey:
I've just implemented this in Edge. It should be included in EdgeHTML 18

PR #596, opened 2018-03-13
Editorial: add "add an event listener" hook

annevk:
This hook ensures that any special casing in addEventListener() is shared with event handler attributes.

Helps with https://github.com/w3c/ServiceWorker/issues/1004.

***
<a href="https://whatpr.org/dom/596.html" title="Last updated on Mar 14, 2018, 8:35 AM GMT (4ebd185)">Preview</a> | <a href="https://whatpr.org/dom/596/8e754b5...4ebd185.html" title="Last updated on Mar 14, 2018, 8:35 AM GMT (4ebd185)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4947,8 +4947,8 @@ for the <a>context object</a>.
   <p>If <var>localName</var> does not match the <code><a type>Name</a></code> production an
   "{{InvalidCharacterError!!exception}}" {{DOMException}} will be thrown.
 
-  <p>When supplied, <var>options</var>'s {{ElementCreationOptions/is}} member can be used to create
-  a <a>customized built-in element</a>.
+  <p>When supplied, <var>options</var>'s {{ElementCreationOptions/is}} can be used to create a
+  <a>customized built-in element</a>.
 
  <dt><code><var>element</var> = <var>document</var> . <a method for=Document lt=createElementNS()>createElementNS(namespace, qualifiedName [, options])</a></code>
 
@@ -4982,8 +4982,8 @@ for the <a>context object</a>.
    is "<code>xmlns</code>".
   </ul>
 
-  <p>When supplied, <var>options</var>'s {{ElementCreationOptions/is}} member can be used to create
-  a <a>customized built-in element</a>.
+  <p>When supplied, <var>options</var>'s {{ElementCreationOptions/is}} can be used to create a
+  <a>customized built-in element</a>.
 
  <dt><code><var ignore>documentFragment</var> = <var>document</var> . {{createDocumentFragment()}}</code>
  <dd>Returns a {{DocumentFragment}}
--- a/dom.bs
+++ b/dom.bs
@@ -1038,13 +1038,13 @@ must return a new {{EventTarget}}.
 if this would be useful for your programs. For now, all author-created {{EventTarget}}s do not
 participate in a tree structure.</p>
 
-<p>The
-<dfn method for=EventTarget><code>addEventListener(<var>type</var>, <var>callback</var>, <var>options</var>)</code></dfn>
-method, when invoked, must run these steps:
+<p>To <dfn export>add an event listener</dfn> given an <code>EventTarget</code> object
+<var>eventTarget</var>, string <var>type</var>, callback <var>callback</var>, and
+<a for=/>dictionary</a> or boolean <var>options</var>, run these steps:
 
 <ol>
  <li>
-  <p>If <a>context object</a>'s <a>relevant global object</a> is a {{ServiceWorkerGlobalScope}}
+  <p>If <var>eventTarget</var>'s <a>relevant global object</a> is a {{ServiceWorkerGlobalScope}}
   object and its associated <a>service worker</a>'s <a for="service worker">script resource</a>'s
   <a for="service worker">has ever been evaluated flag</a> is set, then <a>throw</a> a
   <code>TypeError</code>.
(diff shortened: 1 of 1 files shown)


domenic:
Since event handlers don't have options, maybe this should accept a few booleans instead of a dictionary-or-boolean (with good defaults that make sense for event handlers)?

annevk:
I'd rather not make this an ever expanding set of arguments. I'd be okay with making the last argument optional to make the HTML caller slightly less ugly, but I'm not sure it's worth it.

domenic:
The issue is that for anyone that calls it except for the addEventListener() definition, they have to manually assemble an AddEventListenerOptions dict, which feels like it should be specific to the addEventListener API and to JS consumers of it.

annevk:
I guess you're considering a third caller we haven't decided on adding thus far. I guess if that might not use AddEventListenerOptions we should make that argument optional for now and not pass it from HTML. Then whenever this third caller comes around we can restructure this without having to change HTML.

domenic:
Yeah I'm kind of thinking of https://domenic.github.io/async-local-storage/#add-a-simple-event-listener

annevk:
Hmm, I suppose we could use that signature, requiring the caller to create an event listener concept wrapper.

dom.bs
```diff
@@ -328,7 +328,7 @@ method, passing the same arguments.
 {{Event}} interface (or a derived interface). In the example above
 <var ignore>ev</var> is the <a>event</a>. It is
 passed as argument to
-<a>event listener</a>'s <b>callback</b>
+<a>event listener</a>'s <a for="event listener">callback</a>
```
domenic: Yesss I've been wanting this for a while.

dom.bs
```diff
@@ -958,48 +958,48 @@ are not to be used for anything else. [[!HTML]]
...
-  be invoked when the <a>event</a> is <a>dispatched</a>.
+  <p>Appends an <a>event listener</a> for <a>events</a> whose {{Event/type}} attribute value is
+  <var>type</var>. The <var>callback</var> argument sets the <a for="event listener">callback</a>
+  that will be invoked when the <a>event</a> is <a>dispatched</a>.
 
-  The <var>options</var> argument sets listener-specific options. For compatibility this can be just
-  a boolean, in which case the method behaves exactly as if the value was specified as
+  <p>The <var>options</var> argument sets listener-specific options. For compatibility this can be a
+  boolean, in which case the method behaves exactly as if the value was specified as
   <var>options</var>' <code>capture</code> member.
```
domenic: While here you could make all these `<code>`s link to the dictionary member definitions, as is done in the algorithms below.

dom.bs
```diff
@@ -958,48 +958,48 @@ are not to be used for anything else. [[!HTML]]
...
 
-  When set to true, <var>options</var>' <code>capture</code> member prevents <b>callback</b> from
-  being invoked when the <a>event</a>'s {{Event/eventPhase}} attribute value is
-  {{Event/BUBBLING_PHASE}}. When false (or not present), <b>callback</b> will not be invoked when
-  <a>event</a>'s {{Event/eventPhase}} attribute value is {{Event/CAPTURING_PHASE}}. Either way,
-  <b>callback</b> will be invoked if <a>event</a>'s {{Event/eventPhase}} attribute value is
-  {{Event/AT_TARGET}}.
+  <p>When set to true, <var>options</var>' <code>capture</code> member prevents
```
domenic: While here you could consistify to use "options's" per https://wiki.whatwg.org/wiki/Style#Grammar

dom.bs
```diff
@@ -1038,13 +1038,12 @@ must return a new {{EventTarget}}.
 if this would be useful for your programs. For now, all author-created {{EventTarget}}s do not
 participate in a tree structure.</p>
 
-<p>The
-<dfn method for=EventTarget><code>addEventListener(<var>type</var>, <var>callback</var>, <var>options</var>)</code></dfn>
-method, when invoked, must run these steps:
+<p>To <dfn export>add an event listener</dfn> given an <code>EventTarget</code> object
```
domenic: `{{EventTarget}}` instead of just code

dom.bs
```diff
@@ -1054,34 +1053,53 @@ method, when invoked, must run these steps:
...
+ <li><p>If <var>eventTarget</var>'s <a>event listener list</a> does not <a for=list>contain</a> an
+ <a>event listener</a> whose <a for="event listener">type</a> is <var>listener</var>'s
+ <a for="event listener">type</a>, <a for="event listener">callback</a> is <var>listener</var>'s
+ <a for="event listener">callback</a>, and <a for="event listener">capture</a> is
+ <var>listener</var>'s <a for="event listener">capture</a>, then <a for=list>append</a>
+ <var>listener</var> to <var>eventTarget</var>'s <a>event listener list</a>.
+</ol>
+
+<p class=note>The <a>add an event listener</a> concept exists to ensure
+<a>event handler attributes</a> use the same code path. [[HTML]]
```
domenic: This link is not working

dom.bs
annevk: I thought there was still an exception when the word is plural?

domenic: We've had this discussion before :). `options` is not a class of values; it is a single JS value. So it's not really plural in the grammatical sense that would trigger leaving off the s, where `capture` belongs to a collection of options.

Stated another way, here "options" is short for "options variable", which is not plural.

A case where it would be plural is "the stock options' combined value was a billion dollars."

annevk:
I think I addressed the feedback, but I ended up touching some unrelated sections since it seemed good to make those suggested editorial changes consistently throughout.

Issue #600, opened 2018-03-14
"create an event" does not work with required members


annevk:
In particular as defined today step 3

> Let defaultEventInitDict be the result of converting the JavaScript value undefined to the dictionary type accepted by eventInterface’s constructor. (This dictionary type will either be EventInit or a dictionary that inherits from it.)

would throw, which is a problem as https://searchfox.org/mozilla-central/search?q=required&path=Event.webidl suggests there's various events out there that have a required dictionary member.

We can deal with it failing, but that does seem to suggest that if you have a required dictionary member, you need to supply defaults in some other way than IDL. Alternatively we need to allow passing these members in some way to "fire an event" et al which would be the second refactoring to that whole operation within a rather sad short time period.

marcoscaceres:
Why not let "create an event" just take a dictionary override:

> 1. let initDict be a new SomeKindOfEventInitDict. 
> 2. Set initDict.requiredThing be "the required thing". 
> 3. "Create an event" with `MyEventInteface` ... deal with optional realm ... and initDict, and set ev to the resulting event. 

And then change:

> If dictionary was not passed, let dictionary be the result of converting the JavaScript value undefined to the dictionary type accepted by eventInterface’s constructor. (This dictionary type will either be EventInit or a dictionary that inherits from it.)

So then it just use the passes dictionary?

annevk:
That would probably work, but then you can't use "fire an event" for those events. And if we wanted to enable that, which would make sense, we'd have to redefine how the IDL attributes are initialized and refactor all callers.

(Multiple optional arguments also gets a little ugly and might warrant some refactoring of the callers.)

rong14u:
Man im trying to turn on java script on in this google locked phone.  Can u help please lol

ABirHAsan123-beep:
#600

PR #603, opened 2018-03-15
Make the two ways of event creation share a path

annevk:
This introduces the "inner event creation steps" so we don't duplicate a bunch of logic.

***
<a href="https://whatpr.org/dom/603.html" title="Last updated on Mar 16, 2018, 7:13 AM GMT (cfb07fa)">Preview</a> | <a href="https://whatpr.org/dom/603/06dae65...cfb07fa.html" title="Last updated on Mar 16, 2018, 7:13 AM GMT (cfb07fa)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -809,19 +809,10 @@ interface, or of an interface that inherits from the {{Event}} interface, is inv
 must be run, given the arguments <var>type</var> and <var>eventInitDict</var>:
 
 <ol>
- <li><p>Create a new object <var>event</var> using this interface.
-
- <li><p>Set <var>event</var>'s <a>initialized flag</a>.
+ <li><p>Let <var>event</var> be the result of running the <a>inner event creation steps</a> with
+ this interface, null, now, and <var>eventInitDict</var>.
 
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>type</var>.
-
- <li><p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
- representing the high resolution time from the <a>time origin</a> to the occurrence of the call
- to the <var>event</var>'s <a>constructor</a>.
-
- <li><p>For each <a>dictionary member</a> present in <var>eventInitDict</var>, find the attribute on
- <var>event</var> whose <a spec=webidl>identifier</a> matches the key of the
- <a>dictionary member</a> and then set the attribute to the value of that <a>dictionary member</a>.
 
  <li><p>Return <var>event</var>.
 </ol>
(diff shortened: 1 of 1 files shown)


annevk:
The main motivation is the ability to add a specification callback here that can be used for custom initialization and also to point out #600 which is rather bad.

dom.bs
```diff
@@ -832,34 +823,27 @@ using <var>eventInterface</var>, which must be either {{Event}} or an interface
...
- <a>dictionary member</a>.
+  <p class="XXX">This does not work if members are required; see
+  <a href="https://github.com/whatwg/dom/issues/600">whatwg/dom#600</a>.
 
- <li><p>Set <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
- representing the high resolution time from the <a>time origin</a> to the occurrence that the
- event is signaling.
+ <li>
+  <p>Let <var>event</var> be the result of running the <a>inner event creation steps</a> with
+  <var>eventInterface</var>, <var>realm</var>, the occurrence that the event is signaling, and
```
domenic: "the time of" the occurence

dom.bs
```diff
@@ -832,34 +823,27 @@ using <var>eventInterface</var>, which must be either {{Event}} or an interface
...
- <p class=note>For example, in macOS the occurrence time for input actions are available via the
- timestamp property of corresponding <code>NSEvent</code> objects. So in this case, the
- {{Event/timeStamp}} value will be equivalent to the <code>NSEvent</code>'s timestamp offset by
- <a>time origin</a>, translated in milliseconds, and with its precision adjusted to meet the minimum
- <a>clock resolution</a>.
+  <p class=note>For example, in macOS the occurrence time for input actions are available via the
+  timestamp property of corresponding <code>NSEvent</code> objects. So in this case, the
+  {{Event/timeStamp}} value will be equivalent to the <code>NSEvent</code>'s timestamp offset by
+  <a>time origin</a>, translated in milliseconds, and with its precision adjusted to meet the
+  minimum <a>clock resolution</a>.
```
domenic: I'm not sure "translated in(to?) milliseconds, and with its precision adjusted to meet the minimum clock resolution" belongs here anymore, since you moved those specifics down into the inner event creation steps and are now just using an abstract concept of "time". Probably these clauses should be moved down too.

annevk: This in particular is about the occurrence, but I'll see about making it match things a little better.

dom.bs
```diff
@@ -832,34 +823,27 @@ using <var>eventInterface</var>, which must be either {{Event}} or an interface
...
+  <p>Let <var>event</var> be the result of running the <a>inner event creation steps</a> with
+  <var>eventInterface</var>, <var>realm</var>, the occurrence that the event is signaling, and
+  <var>dictionary</var>.
 
- <p class=note>For example, in macOS the occurrence time for input actions are available via the
- timestamp property of corresponding <code>NSEvent</code> objects. So in this case, the
- {{Event/timeStamp}} value will be equivalent to the <code>NSEvent</code>'s timestamp offset by
- <a>time origin</a>, translated in milliseconds, and with its precision adjusted to meet the minimum
- <a>clock resolution</a>.
+  <p class=note>For example, in macOS the occurrence time for input actions are available via the
```
garykac: ...occurrence time for input actions **is** available via the timestamp property of **the** corresponding NSEvent **object**.

(or  ...occurrence **times** for input actions are available via the timestamp property of **the** corresponding NSEvent objects.)

PR #605, opened 2018-03-15
Editorial: be more consistent with callbacks

annevk:
***
<a href="https://whatpr.org/dom/605.html" title="Last updated on Mar 16, 2018, 10:13 AM GMT (3e5cd1b)">Preview</a> | <a href="https://whatpr.org/dom/605/5f4249f...3e5cd1b.html" title="Last updated on Mar 16, 2018, 10:13 AM GMT (3e5cd1b)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1347,7 +1347,7 @@ for discussion).
  <li><p>Let <var>listeners</var> be a new <a for=/>list</a>.
 
  <li>
-  <p><a for=/>For each</a> <var>listener</var> of <var>object</var>'s
+  <p><a for=list>For each</a> <var>listener</var> of <var>object</var>'s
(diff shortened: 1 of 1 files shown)


annevk:
I wrote this to fix some Bikeshed errors, I noticed two issues:

1. There's no callback this value for traversal objects and I'm not sure if that's correct.
2. IDL has rather confusing algorithm signatures.

annevk:
Filed https://github.com/heycam/webidl/issues/534 on IDL.

annevk:
It does appear there's indeed no callback this value for traversal objects. I suspect that's tested. So I think this can land as-is, and potentially in the future we can clean up the IDL bits even more.

PR #609, opened 2018-03-16
Use Infra to define mutation observers

### LINKED ISSUE #132: Small correct for Mutation observers
https://dom.spec.whatwg.org/#mutation-observer-compound-microtask-queued-flag
Missing initial state for this flag, just add "(initially unset)", whe have this for other flags (in this spec and in other spec):
https://dom.spec.whatwg.org/#stop-propagation-flag
https://dom.spec.whatwg.org/#concept-event-listener
https://encoding.spec.whatwg.org/#textdecoder
https://encoding.spec.whatwg.org/#gbk-flag
https://url.spec.whatwg.org/#non-relative-flag

---

https://dom.spec.whatwg.org/#registered-observer
Why we don't have definition for "observer", "options" and "source" with some id? Instead using `<b>` in other place for this terms it would be more convenient to use references. 

---

https://dom.spec.whatwg.org/#interface-mutationrecord
`readonly attribute Node target;` << this IDL should not be mark by `[SameObject]`? It's not change, similarly what we have for root in NodeIterator/TreeWalker (https://dom.spec.whatwg.org/#interface-nodeiterator, https://dom.spec.whatwg.org/#interface-treewalker)

annevk:
Fixes #132.

The main thing I still need to look into is figuring out how transient registered observers are actually removed, as the current text doesn't state that in much detail.

***
<a href="https://whatpr.org/dom/609.html" title="Last updated on Mar 29, 2018, 10:32 AM GMT (be1760f)">Preview</a> | <a href="https://whatpr.org/dom/609/787748d...be1760f.html" title="Last updated on Mar 29, 2018, 10:32 AM GMT (be1760f)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2613,8 +2613,8 @@ with an optional <i>suppress observers flag</i>, run these steps:
  <a>registered observer list</a>.
 
  <li><p>If <i>suppress observers flag</i> is unset, then <a>queue a tree mutation record</a> for
- <var>parent</var> with « », « <var>node</var> », <var>oldNextSibling</var>, and
- <var>oldPreviousSibling</var>.
+ <var>parent</var> with « », « <var>node</var> », <var>oldPreviousSibling</var>, and
+ <var>oldNextSibling</var>.
 
  <li><p>If <var>node</var> is a {{Text}} node, then run the <a>child text content change steps</a>
  for <var>parent</var>.
@@ -3177,7 +3177,7 @@ objects), which is initially empty.
 
 <hr>
 
-<p>Each <a for=/>node</a> has <dfn noexport>registered observer list</dfn> (a <a for=/>list</a> of
+<p>Each <a for=/>node</a> has a <dfn noexport>registered observer list</dfn> (a <a for=/>list</a> of
 zero or more <a>registered observers</a>), which is initially empty.
 
 <p>A <dfn noexport id=registered-observer>registered observer</dfn> consists of an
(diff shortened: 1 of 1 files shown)


annevk:
@smaug---- when the specification does

> Remove all transient registered observers whose observer is mo. 

> Remove all transient registered observers whose source is registered.

is this some kind of global search? And the goal is to remove them from each node's registered observer list? Seems quite expensive.

annevk:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=22759#c9 by @travisleithead has the better model here (the first suggestion) that didn't make it into the specification. That would make how this works much more explicit.

annevk:
So, we could give each "registered observer" a "transient registered observer node list". When removing a node, in step 16 of https://whatpr.org/dom/609.html#concept-node-remove, aside from appending a new transient registered observer to _node_'s  registered observer list, we also append _node_ to _registered_'s transient registered observer node list.

Then we could rewrite step 7 of https://whatpr.org/dom/609.html#dom-mutationobserver-observe to use this new node list on _registered_ to remove the transient registered observers from those nodes.

However, we also remove transient registered observers in https://whatpr.org/dom/609.html#notify-mutation-observers based on the observer field. Should `MutationObserver` (the value of that observer field) also keep a list of nodes that have transient registered observers?

@ajklein if you could help with this further formalizing of mutation observers that'd be appreciated. If it's too much to page in I can relate.

smaug----:
> is this some kind of global search?

no. MutationObserver has the list of transient observers, so it can just go through the list.
In particular in Gecko there is MutationReceiver per observed target internally doing actually observing, and it has list of transient receivers.

ajklein:
I suspect that @smaug---- has this swapped in better than I do, have been out of DOM land for several years now...

annevk:
```irc
<smaug> annevk: so, in Gecko DOMMutationObserver is the thing representing MutationObserver object. It has list of MutationReceivers which actually observers the mutations. Each such receiver may the have transient receivers. Right before a mutationobserver calls its callback, it goes through its mutationreceivers, and ask them to clear transient receivers
I guess that should answer to the first one... then the second question 
<smaug> and same happens with observe() https://searchfox.org/mozilla-central/source/dom/base/nsDOMMutationObserver.cpp#743
<annevk> hmm, that doesn't really map to the spec cleanly
<smaug> how so
<annevk> smaug: just that the spec has different concepts
<smaug> trying to see where
<annevk> smaug: oh so mutationreceivers are the nodes?
MutationObserver does have a node list
in the spec
<smaug> well, attached to nodes
<annevk> okay
so we'd iterate over the nodes and remove the transient registered observers from their registered observer list
<smaug> each receiver has target, and that target is a node
right, something like that
<annevk> smaug: and in observe we'd look at the registered observer's observer's node list and do the same is what you're saying?
<smaug> yeah, https://searchfox.org/mozilla-central/source/dom/base/nsDOMMutationObserver.cpp#743
in that code we update the receiver and remove the transient stuff
```

annevk:
@ArkadiuszMichalski interested in reviewing this?

dom.bs
```diff
@@ -3636,6 +3638,8 @@ that is a <a>document</a>.
 <a>node</a>'s <a>assigned slot</a>, if <a>node</a> is <a>assigned</a>, and <a>node</a>'s
 <a for=tree>parent</a> otherwise.
 
+<p class="note no-backref">Each <a for=/>node</a> also has a <a>registered observer list</a>.
```
smaug----: Each node has also ?

dom.bs
```diff
@@ -3321,9 +3322,10 @@ method, when invoked, must run these steps:
   <a for="registered observer">observer</a> is the <a>context object</a>:
 
   <ol>
-   <li><p>Remove all <a>transient registered observers</a> whose <b>source</b> is
-   <var>registered</var>.
-   <!-- XXX remove from where? -->
+   <li><p><a for=list>For each</a> <var>node</var> of the <a>context object</a>'s
+   <a for=MutationObserver>node list</a>, <a for=list>remove</a> all
+   <a>transient registered observers</a> whose <a for="transient registered observers">source</a> is
+   <var>registered</var> from <var>node</var>'s <a>registered observer list</a>.
```
smaug----: Something weird with this sentence.

dom.bs
annevk: The precedent in the document is "also has". And I think that's correct.

dom.bs
annevk: Can you be more specific? I can't really find anything odd.

dom.bs
smaug----: But which one reads better?

annevk: "also has" reads more natural to me.

Issue #87, opened 2015-10-08
Behaviour for pointerBeforeReferenceNode when all nodes do not pass filtration


ArkadiuszMichalski:
Traverse algo doesn't fit to Firefox/Chrome behaviour when we skip all nodes and try invoke NodeIterator.nextNode(). Per algo true should be change to false, but this browsers still keep true.

<script>

	var iterator = document.createNodeIterator(document, 0);
	alert(iterator.pointerBeforeReferenceNode); // true
	iterator.nextNode();
	alert(iterator.pointerBeforeReferenceNode); // true

</script>

Others edge case may by interesting, like removing referenceNode where some other node must be as referenceNode (even if not acceptable by filter), so  NodeIterator.nextNode()/NodeIterator.previousNode() should only change pointerBeforeReferenceNode or not?

 I don't know what Edge does, but IE11  still doesn't support pointerBeforeReferenceNode and referenceNode attributes.

Of course browsers can follow with these new guidelines if it's not compatibility problem and bring some simplification, or just fix this algo to fit actual browsers behaviour.

And a few comments on this algorithm (all for 3.1 step):
- "If direction..." direction is passing argument, then can be treat as variable 
- iterator collection was defined so use reference. BTW, it's collection (rooted at root, whose filter matches any node), but for iterator sense root should not be included to this collection (as first element)?
- having three /if/ conditions writing in one line (separate by dot) is quite readable, especially when first cooperate with second, and third is opposite to first. Something like we have here http://www.w3.org/TR/dom/#concept-nodeiterator-traverse is better.

ArkadiuszMichalski:
nextNode()
https://dom.spec.whatwg.org/#dom-treewalker-nextnode
3.2 If a node is following node and is not... < first node is unclear, maybe write: 
3.2. If there is a node which/that is following node and is not...

annevk:
> BTW, it's collection (rooted at root, whose filter matches any node), but for iterator sense root should not be included to this collection (as first element)?

I don't quite understand this comment.

annevk:
@triple-underscore do you have a test for that?

triple-underscore:
>do you have a test for that?

Consider the following DOM and script:

```
<div id="root"><div id="n1"><div id="n2"><div id="n3"></div></div></div></div>

<script>
const walker = document.createTreeWalker(
  document.getElementById('root'),
  NodeFilter.SHOW_ELEMENT,
  (e) => {
    return e.id === 'n2' ?
[...]

PR #612, opened 2018-03-22
Fix TreeWalker's nextNode()

annevk:
The prose supposed to be equivalent to iterating over following siblings and siblings of ancestors did not skip over descendants.

See https://github.com/whatwg/dom/issues/87#issuecomment-374385254.

***
<a href="https://whatpr.org/dom/612.html" title="Last updated on Mar 22, 2018, 5:39 PM GMT (ae2fb6b)">Preview</a> | <a href="https://whatpr.org/dom/612/02f7143...ae2fb6b.html" title="Last updated on Mar 22, 2018, 5:39 PM GMT (ae2fb6b)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9172,8 +9172,9 @@ steps:
      {{TreeWalker/currentNode}} attribute to <var>node</var> and return <var>node</var>.
     </ol>
 
-   <li><p>If there is a <a>node</a> that is <a>following</a> <var>node</var> and is not
-   <a>following</a> <a for=traversal>root</a>, then set <var>node</var> to the first such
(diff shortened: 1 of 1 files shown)


annevk:
@triple-underscore thanks for reporting this. I didn't make it check for result being FILTER_REJECT is FILTER_SKIP is already handled by the earlier algorithm that goes through children.

annevk:
(I also worked out the prose that would just do the next sibling / parent traversal, but it ends up adding a significant number of steps and I'm not sure it's really any clearer.)

annevk:
Looking at https://www.w3.org/Bugs/Public/show_bug.cgi?id=28107 again I went with the longer algorithm after all since "following root" was also wrong.

PR #614, opened 2018-03-23
Define event constructing steps

### LINKED ISSUE #414: "create an event" doesn't work for certain UI events

annevk:
This hook can be used by complex Event subclasses, such as KeyboardEvent and TouchEvent, to initialize their internal state upon creation.

Fixes #414.

***
<a href="https://whatpr.org/dom/614.html" title="Last updated on Mar 23, 2018, 12:46 PM GMT (b8e54d8)">Preview</a> | <a href="https://whatpr.org/dom/614/5be84cc...b8e54d8.html" title="Last updated on Mar 23, 2018, 12:46 PM GMT (b8e54d8)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -792,6 +792,14 @@ method must, when invoked, run these steps:
 
 <h3 id=constructing-events>Constructing events</h3>
 
+<a lt="Other applicable specifications">Specifications</a> may define
+<dfn export id=concept-event-constructor-ext>event constructing steps</dfn> for all or some
+<a for=/>events</a>. The algorithm is passed an <var>event</var> as indicated in the
+<a>inner event creation steps</a>.
+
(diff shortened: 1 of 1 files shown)


annevk:
@garykac your input on this would be much appreciated as well, as this'll primarily affect UI events I think.

PR #615, opened 2018-03-23
NodeIterator's pre-removing steps need to account for node being root

### LINKED ISSUE #496: NodeIterator pre-removing steps seem incorrect
https://dom.spec.whatwg.org/commit-snapshots/cd3bcaeea40b4b9355ae7be3a5799813faa492fc/#interface-nodeiterator

Consider the scenario at https://jsbin.com/zugoyutice/edit?html,console,output , in particular the removeChild() call. (Derived from https://github.com/tmpvar/jsdom/blob/3743b423cb45657401eef3475a07fd99e015b2e2/test/to-port-to-wpts/node-iterator.js#L429 , for the record)

toBeRemovedNode = referenceNode = root = the span

1. toBeRemovedNode is an inclusive ancestor of referenceNode, so step 1 passes
2. pointerBeforeReferenceNode = true (since it starts that way)
   1. next = null; there is nothing that is not an inclusive descendant of toBeRemovedNode but is an inclusive descendant of root
   2. Do nothing
   3. **Set pointerBeforeReferenceNode to false**
3. **Set referenceNode to the body**

However the test shows that pointerBeforeReferenceNode and referenceNode do not change.

This appears to have regressed in https://github.com/whatwg/dom/commit/259775f14d55b21de16cbb0a3c625741fd54aa0b, as jsdom has an implementation based on the older removing-steps-based algorithm, and updating broke <del>a bunch</del><ins>one, after I fixed other bugs of ours :)</ins> of our (to-be-upstreamed) tests.

annevk:
Fixes #496.

***
<a href="https://whatpr.org/dom/615.html" title="Last updated on Mar 23, 2018, 12:55 PM GMT (d6d929d)">Preview</a> | <a href="https://whatpr.org/dom/615/5be84cc...d6d929d.html" title="Last updated on Mar 23, 2018, 12:55 PM GMT (d6d929d)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8761,7 +8761,8 @@ associated <a for=traversal>active flag</a>,
 
 <ol>
  <li><p>If <var>toBeRemovedNode</var> is not an <a for=tree>inclusive ancestor</a> of the
- <var>nodeIterator</var>'s {{NodeIterator/referenceNode}} attribute value, then return.
+ <var>nodeIterator</var>'s {{NodeIterator/referenceNode}} attribute value, or
+ <var>toBeRemovedNode</var> is <var>nodeIterator</var>'s <a for=traversal>root</a>, then return.
 
  <li>
   <p>If <var>nodeIterator</var>'s {{NodeIterator/pointerBeforeReferenceNode}} attribute value is


annevk:
It already tests this. It seems I regressed this in https://github.com/whatwg/dom/commit/4937f1a5e39c0914ad2f0a3a35558cf1678e8b89.

domenic:
Hmm, I could swear I was able to pass NodeIterator-removal.html with a version in jsdom that implemented the broken spec... I guess I can try to prioritize porting to-port-to-wpts/node-iterator.js wholesale, though.

PR #619, opened 2018-03-27
Editorial: remove object argument from inner invoke

### LINKED ISSUE #618: "inner invoke" does not need to take object
It's event's currentTarget so we can just use that.

annevk:
Fixes #618 (make sure to amend the commit message when landing to mention this)


annevk:
It seems this was added in a026ccf3efa8d34fb0cac580928f61d18b1e9f76 without much thought as to why it was needed.

PR #632, opened 2018-04-10
Meta: export event listener's fields

domenic:
***
<a href="https://whatpr.org/dom/632.html" title="Last updated on Apr 10, 2018, 11:43 PM GMT (7b1ee4e)">Preview</a> | <a href="https://whatpr.org/dom/632/c80e017...7b1ee4e.html" title="Last updated on Apr 10, 2018, 11:43 PM GMT (7b1ee4e)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -946,12 +946,12 @@ when something has occurred.
 <a>event</a> and consists of:
 
 <ul class="brief">
- <li><dfn for="event listener">type</dfn> (a string)
- <li><dfn for="event listener">callback</dfn> (null or an {{EventListener}} object)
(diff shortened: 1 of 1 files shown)


annevk:
Maybe don't export removed?

domenic:
You still need to set it when creating a new event listener to pass to "add an event listener", don't you?

Although, I ended up not using this for async-local-storage after all, so we can close this until there's a real consumer.

annevk:
No, because there's default values. You need to set the type and callback though.

I'd support exporting (minus removed) since HTML already uses type/callback, but I'm also fine with waiting for a Bikeshed consumer.

domenic:
Right, I forgot about those. I'll close for now...

PR #638, opened 2018-04-16
Editorial: reference "set an existing attribute value"

### LINKED ISSUE #637: "Set an existing attribute value" not cross-linking correctly
https://dom.spec.whatwg.org/commit-snapshots/c80e0176d9565d2832ae3d535414faff7f6e7dda/#interface-node

> Set an existing attribute value with context object and new value\. 

Both places this appears, it does not correctly cross-link to the "set an existing attribute value" algorithm.

annevk:
Fixes #637.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3976,7 +3976,7 @@ instead, and then do as described below, depending on the <a>context object</a>:
 
 <dl class=switch>
  <dt>{{Attr}}
- <dd><p>Set an existing attribute value</a> with <a>context object</a> and new value.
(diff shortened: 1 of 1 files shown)


annevk:
It's a little concerning to me that Bikeshed does not error on this. @tabatkins is that intentional?

gsnedders:
@tabatkins 

> I just use the spec-compliant html5lib parser (by @jgraham, @gsnedders, and you, @annevk!), which doesn't surface any parse errors, such as lone end-tags.

They're in `HTMLParser.errors` (a list).

Issue #371, opened 2016-11-09
Nerfing all events on all event targets in a serviceworker after initial script evaluation seems odd


bzbarsky:
As of today, https://dom.spec.whatwg.org/commit-snapshots/8eb68e48b3f613663425120edc5039d83b3a9de8/#dom-eventtarget-addeventlistener says:

> If context object’s relevant global object is a ServiceWorkerGlobalScope object and its associated service worker’s script resource’s has ever been evaluated flag is set, then throw a TypeError. [SERVICE-WORKERS] 

Why are we doing this for all event targets?  Why are we even doing it for all events?  This makes no sense.

bzbarsky:
Looks like this was introduced in https://github.com/whatwg/dom/pull/155

// cc @jungkees

bzbarsky:
I mean, there's the question of why we're doing it at all too, instead of handling this like upload events on XHR.

jungkees:
> there's the question of why we're doing it at all too

It had started by https://github.com/w3c/ServiceWorker/issues/718#issue-92573396 to deterministically detect whether no event handler's registered for a given event (to reduce unnecessary SW start-up delay in such cases.) Over a discussion, we'd setteled on the current behavior. (TL;DR see https://github.com/w3c/ServiceWorker/issues/718#issuecomment-173830892 and below.)

> Why are we doing this for all event targets?

I think this was a miss. This should be only for the events on ServiceWorkerGlobalScope not for all the EventTargets.

> Why are we even doing it for all events?

Given that it should be only for ServiceWorkerGlobalScope, SW lifecycle events plus all other functional events including fetch, push and future extesions are within the scope of this behavior.

> handling this like upload events on XHR.

Could you give me a pointer or explain a bit more on this?

annevk:
Upload events are based on a flag whose state is determined once fetching starts. I think service workers has that already in place. We check if there are any event listeners registered and store those someplace.

bzbarsky:
> Over a discussion, we'd setteled on the current behavior

I don't really see a discussion.  I see @jakearchibald asserting that this should be the behavior and no one questioning it, or thinking about how it ties in to other specifications.  For example, it introduced inconsistent behavior between `addEventListener("fetch", ...)` and setting `onfetch`...

> SW lifecycle events plus all other functional events including fetch, push and future extesions are within the scope of this behavior

OK, but you're doing this for _all_ events, including random made-up ones the service worker script itself decides to fire.  I don't see why this is at all reasonable.

[...]

PR #653, opened 2018-06-06
Change add/remove event listener behavior for service workers

jungkees:
This changes addEventListener() and removeEventListener() to not throw even
after the very first evalution of the service worker script. Instead, this
specifies user agents have to show a console warning that the asynchronously
added listener's event type will not affect service worker's behavior with the
functional event stored during the first evaluation.

This change referneces the funcional event concept defined in Service Workers
spec, and confines the EventTarge object of the specified behavior to
ServiceWorkerGlobalScope object instead of all the platform object in the global
object.

Related SW issue: https://github.com/w3c/ServiceWorker/issues/1004.
Related SW PR: https://github.com/w3c/ServiceWorker/pull/1322.
Fixes https://github.com/whatwg/dom/issues/371.

***
<a href="https://whatpr.org/dom/653.html" title="Last updated on Nov 4, 2019, 8:51 AM UTC (8946b24)">Preview</a> | <a href="https://whatpr.org/dom/653/57512fa...8946b24.html" title="Last updated on Nov 4, 2019, 8:51 AM UTC (8946b24)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1151,21 +1151,16 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
 
 <ol>
  <li>
-  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object, its associated
-  <a for="ServiceWorkerGlobalScope">service worker</a>'s <a for="service worker">script
-  resource</a>'s <a for="script resource">has ever been evaluated flag</a> is set, and
-  <var>listener</var>'s <a for="event listener">type</a> matches the {{Event/type}} attribute value
-  of any of the <a>service worker events</a>, then <a>report a warning to the console</a> with a
-  description explaining that the <a>service worker event</a> has to be added before the script
-  evaluation completes, otherwise the user agent might not start the <a>service worker</a> for the
-  <a>event</a>. [[!SERVICE-WORKERS]]
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object, its
+  <a for="ServiceWorkerGlobalScope">service worker</a>'s
+  <a for="service worker">script resource</a>'s
+  <a for="script resource">has ever been evaluated flag</a> is set, and <var>listener</var>'s
+  <a for="event listener">type</a> matches the {{Event/type}} attribute value of any of the
+  <a>service worker events</a>, then <a>report a warning to the console</a> that this might not give
+  the expected results. [[!SERVICE-WORKERS]]
 
-  <p class="note no-backref">To avoid unnecessary delays, Service Workers [[!SERVICE-WORKERS]]
-  permits skipping event dispatch when no <a>event listeners</a> for the <a>service worker event</a>
-  have been deterministically added in the <a>service worker</a>’s <a for="service worker">set of
-  event types to handle</a> during the very first script evaluation. The <a>event listener</a> in
-  this step will still be added, but the implementations have to inform developers about this
-  behavior.
+  <p class="note no-backref">Service workers are only started for event listeners added during the
+  first script evaluation.
 
  <li><p>If <var>listener</var>'s <a for="event listener">callback</a> is null, then return.
 
@@ -1195,31 +1190,35 @@ method, when invoked, must run these steps:
  <var>once</var>.
 </ol>
 
-<p>To <dfn>remove an event listener</dfn>, given an {{EventTarget}} object <var>eventTarget</var>
-and an <a>event listener</a> <var>listener</var>, set <var>listener</var>'s
-<a for="event listener">removed</a> to true and <a for=list>remove</a> <var>listener</var> from
-<var>eventTarget</var>'s <a for=EventTarget>event listener list</a>.
-<!-- Intentionally not exported. -->
+<p>To <dfn export>remove an event listener</dfn>, given an {{EventTarget}} object
+<var>eventTarget</var> and an <a>event listener</a> <var>listener</var>, run these steps:
+
+<ol>
+ <li><p>If the <a>context object</a> is a {{ServiceWorkerGlobalScope}} object and its
+ <a for="ServiceWorkerGlobalScope">service worker</a>'s
+ <a for="service worker">set of event types to handle</a> contains <var>type</var>, then
+ <a>report a warning to the console</a> that this might not give the expected results.
+ [[!SERVICE-WORKERS]]
+
+ <li><p>Set <var>listener</var>'s <a for="event listener">removed</a> to true and
+ <a for=list>remove</a> <var>listener</var> from <var>eventTarget</var>'s
+ <a for=EventTarget>event listener list</a>.
+</ol>
+
+<p class=note>HTML needs this to define event handlers. [[HTML]]
 
 <p>To <dfn export>remove all event listeners</dfn>, given an {{EventTarget}} object
 <var>eventTarget</var>, <a for=list>for each</a> <var>listener</var> of <var>eventTarget</var>'s
 <a for=EventTarget>event listener list</a>, <a>remove an event listener</a> with
 <var>eventTarget</var> and <var>listener</var>.
 
-<p class="note">HTML needs this to define <code>document.open()</code>. [[HTML]]
+<p class=note>HTML needs this to define <code>document.open()</code>. [[HTML]]
 
 <p>The
 <dfn method for=EventTarget><code>removeEventListener(<var>type</var>, <var>callback</var>, <var>options</var>)</code></dfn>
 method, when invoked, must run these steps:
 
 <ol>
- <li><p>If the <a>context object</a> is a {{ServiceWorkerGlobalScope}} object and its associated
- <a for="ServiceWorkerGlobalScope">service worker</a>'s <a for="service worker">set of event types
- to handle</a> contains <var>type</var>, then <a>report a warning to the console</a> with a
- description explaining that the user agent will still start the <a>service worker</a> for the
- <a>service worker event</a> whose {{Event/type}} attribute value matches <var>type</var>.
- [[!SERVICE-WORKERS]]
-
  <li><p>Let <var>capture</var> be the result of <a>flattening</a> <var>options</var>.
 
  <li><p>If the <a>context object</a>'s <a for=EventTarget>event listener list</a>
--- a/dom.bs
+++ b/dom.bs
@@ -1151,12 +1151,14 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
 
 <ol>
  <li>
-  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object and its associated
-  <a>service worker</a>'s <a for="service worker">script resource</a>'s
-  <a for="script resource">has ever been evaluated flag</a> is set, then
-  <a>report a warning to the console</a> with a description explaining that the
-  <a>service worker event</a> is to be added synchronously, otherwise the user agent may not start
-  the <a>service worker</a> for the <a>event</a>. [[!SERVICE-WORKERS]]
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object, its associated
+  <a for="ServiceWorkerGlobalScope">service worker</a>'s <a for="service worker">script
+  resource</a>'s <a for="script resource">has ever been evaluated flag</a> is set, and
+  <var>listener</var>'s <a for="event listener">type</a> matches the {{Event/type}} attribute value
+  of any of the <a>service worker events</a>, then <a>report a warning to the console</a> with a
+  description explaining that the <a>service worker event</a> has to be added before the script
+  evaluation completes, otherwise the user agent might not start the <a>service worker</a> for the
+  <a>event</a>. [[!SERVICE-WORKERS]]
 
   <p class="note no-backref">To avoid unnecessary delays, Service Workers [[!SERVICE-WORKERS]]
   permits skipping event dispatch when no <a>event listeners</a> for the <a>service worker event</a>
@@ -1211,11 +1213,11 @@ and an <a>event listener</a> <var>listener</var>, set <var>listener</var>'s
 method, when invoked, must run these steps:
 
 <ol>
- <li><p>If the <a>context object</a> is a {{ServiceWorkerGlobalScope}} object and its associated
- <a>service worker</a>'s <a for="service worker">script resource</a>'s
- <a for="script resource">has ever been evaluated flag</a> is set, then
- <a>report a warning to the console</a> with a description explaining that the user agent will still
- start the <a>service worker</a> for the <a>service worker event</a> that was synchronously added.
+ <li><p>If the <a>context object</a> is a {{ServiceWorkerGlobalScope}} object, its associated
+ <a for="ServiceWorkerGlobalScope">service worker</a>'s <a for="service worker">set of event types
+ to handle</a> contains <var>type</var>, then <a>report a warning to the console</a> with a
+ description explaining that the user agent will still start the <a>service worker</a> for the
+ <a>service worker event</a> whose {{Event/type}} attribute value matches <var>type</var>.
  [[!SERVICE-WORKERS]]
 
  <li><p>Let <var>capture</var> be the result of <a>flattening</a> <var>options</var>.
--- a/dom.bs
+++ b/dom.bs
@@ -1150,17 +1150,13 @@ participate in a tree structure.</p>
 and an <a>event listener</a> <var>listener</var>, run these steps:
 
 <ol>
- <li>
-  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object, its
-  <a for="ServiceWorkerGlobalScope">service worker</a>'s
-  <a for="service worker">script resource</a>'s
-  <a for="script resource">has ever been evaluated flag</a> is set, and <var>listener</var>'s
-  <a for="event listener">type</a> matches the {{Event/type}} attribute value of any of the
-  <a>service worker events</a>, then <a>report a warning to the console</a> that this might not give
-  the expected results. [[!SERVICE-WORKERS]]
-
-  <p class="note no-backref">Service workers are only started for event listeners added during the
-  first script evaluation.
+ <li><p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object, its
+ <a for="ServiceWorkerGlobalScope">service worker</a>'s
+ <a for="service worker">script resource</a>'s
+ <a for="script resource">has ever been evaluated flag</a> is set, and <var>listener</var>'s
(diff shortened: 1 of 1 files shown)


annevk:
@domfarolino what is your expected timeline for fixing https://github.com/whatwg/console/issues/57?

domfarolino:
Under two weeks to be safe. I can prioritize that, thanks for the ping.

jungkees:
@annevk, please take a look. I pushed a commit that uses the Console standard.

dom.bs
```diff
@@ -1062,15 +1062,19 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
-  <p>If <var>eventTarget</var>'s <a>relevant global object</a> is a {{ServiceWorkerGlobalScope}}
-  object and its associated <a>service worker</a>'s <a for="service worker">script resource</a>'s
-  <a for="script resource">has ever been evaluated flag</a> is set, then <a>throw</a> a
-  <code>TypeError</code>.
-  [[!SERVICE-WORKERS]]
-
-  <p class="note no-backref">To optimize storing the event types allowed for the service worker and
-  to avoid non-deterministic changes to the event listeners, invocation of the method is allowed
-  only during the very first evaluation of the service worker script.
+  <p>If <var>listener</var> observes a <a>functional event</a>, <var>eventTarget</var> is a
```
jakearchibald: Feels clearer to say "If listener's type is…" then link to the list https://w3c.github.io/ServiceWorker/#execution-context-events.

Right now "observes" is a little vague, and "functional event" doesn't cover the lifecycle events.

dom.bs
```diff
@@ -1062,15 +1062,19 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
+  {{ServiceWorkerGlobalScope}} object, and its associated <a>service worker</a>'s <a for="service
+  worker">script resource</a>'s <a for="script resource">has ever been evaluated flag</a> is set,
+  then the user agent must <a>report a warning to the console</a> with a description that explains
+  that the user agent stores the event <a for="event listener">type</a> of the <a>functional
+  event</a> for the <a>service worker</a> only during the very first evalution of the <a>service
```
jakearchibald: Should the explanation be developer-focused here?

If so:

…the user agent must report a warning to the console explaining that an event with a type of listener's type should be added synchronously, otherwise, then the service worker restarts, events of this type will be dispatched before the listener is added.

dom.bs
```diff
@@ -1062,15 +1062,19 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
+  worker</a> script, and the asynchronously added <a>functional event</a> will not start the
+  <a>service worker</a>. [[!SERVICE-WORKERS]]
+
+  <p class="note no-backref">To optimize storing the event types allowed for the <a>service
+  worker</a>, Service Workers specification does not store the asynchronously added event <a
+  for="event listener">type</a>. The <a>event listener</a> will still be added under this condition,
+  but the implmentations have to inform developers about this behavior.
```
jakearchibald: We might not need this note if the developer-facing explanation is clear enough.

dom.bs
```diff
@@ -1118,11 +1122,13 @@ and an <a>event listener</a> <var>listener</var>, set <var>listener</var>'s
...
- <a for="service worker">script resource</a>'s
- <a for="script resource">has ever been evaluated flag</a> is set, then <a>throw</a> a
- <code>TypeError</code>. [[!SERVICE-WORKERS]]
+ <li><p>If <var>type</var> is one of the <a>functional events</a>'s <a for="event
+ listener">type</a>, the <a>context object</a> is a {{ServiceWorkerGlobalScope}} object, and its
+ associated <a>service worker</a>'s <a for="service worker">script resource</a>'s <a for="script
+ resource">has ever been evaluated flag</a> is set, then the user agent must <a>report a warning to
+ the console</a> with a description that explains that the user agent will still start the
+ <a>service worker</a> for the <a>functional event</a> that has been stored during the very first
+ evalution of the <a>service worker</a> script and is asynchronously removed. [[!SERVICE-WORKERS]]
```
jakearchibald: (any changes to addEventListener should also be made here)

dom.bs
jungkees: DOM uses "observe" as in:
> An event listener can be used to observe a specific event

If we use _listener's type_, we might need to compare it with the event's type attribute value, which makes it something like:
> If listener's type is the type attribute value of one of the events in this list, ...

I'm fine with either, but also think that we might be able to drop this first condition entirely and just start with:
> If eventTarget is a ServiceWorkerGlobalScope object, ...

This condition basically covers the listener's type is one in the list. WDYT?

jungkees: > and "functional event" doesn't cover the lifecycle events.

Good point!

dom.bs
jungkees: I updated it. Just kept the wording,"with a description", to denote an argument and tried to explain the service worker will not start with the asynchronously added event.

dom.bs
jungkees: I guess this note would help implementers understand *why* this behavior was added (while the warning description says what to display to the console as you pointed). I tried to simplify the words.

dom.bs
jungkees: I updated this accordingly too.

jungkees:
@jakearchibald, I uploaded a new snapshot. Please take another pass. This change references a change that I made in https://github.com/w3c/ServiceWorker/pull/1452.

dom.bs
```diff
@@ -1062,15 +1062,17 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
-  <a for="script resource">has ever been evaluated flag</a> is set, then <a>throw</a> a
-  <code>TypeError</code>.
-  [[!SERVICE-WORKERS]]
-
-  <p class="note no-backref">To optimize storing the event types allowed for the service worker and
-  to avoid non-deterministic changes to the event listeners, invocation of the method is allowed
-  only during the very first evaluation of the service worker script.
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object and its associated
+  <a>service worker</a>'s <a for="service worker">script resource</a>'s <a for="script resource">has
+  ever been evaluated flag</a> is set, then the user agent must <a>report a warning to the
```
annevk: `then <a>report ...` No need to say must here. (Also applies below.)

dom.bs
```diff
@@ -1062,15 +1062,17 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
+  console</a> with a description explaining that the <a>service worker event</a> should be added
+  synchronously, otherwise the user agent will not start</a> the <a>service worker</a> for the
```
annevk: This looks like a stray `</a>`. I'm surprised that builds.

dom.bs
```diff
@@ -1062,15 +1062,17 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
-  <code>TypeError</code>.
-  [[!SERVICE-WORKERS]]
-
-  <p class="note no-backref">To optimize storing the event types allowed for the service worker and
-  to avoid non-deterministic changes to the event listeners, invocation of the method is allowed
-  only during the very first evaluation of the service worker script.
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object and its associated
+  <a>service worker</a>'s <a for="service worker">script resource</a>'s <a for="script resource">has
+  ever been evaluated flag</a> is set, then the user agent must <a>report a warning to the
+  console</a> with a description explaining that the <a>service worker event</a> should be added
```
annevk: "is to be added" (let's not use "should" as it's meant to indicate requirements to the implementer within this document).

dom.bs
```diff
@@ -1062,15 +1062,17 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
+  synchronously, otherwise the user agent will not start</a> the <a>service worker</a> for the
+  <a>event</a>. [[!SERVICE-WORKERS]]
+
+  <p class="note no-backref">To optimize with the set of <a>service worker events</a> to handle, the
+  Service Workers specification does not store the asynchronously added event. The <a>event
```
annevk: It might be better to say what object does not store the event.

annevk: No newlines inside `<a>`.

dom.bs
jungkees: Done.

Out of my own curiosity, what'd be the actual meaning of it in terms of the conformance requirement after this change?

dom.bs
jungkees: Done.

dom.bs
jungkees: Removed.

dom.bs
jungkees: Done.

jungkees: Fixed.

jungkees:
Thanks for reviewing, @annevk. I addressed your comments and uploaded a new snapshot. Please take a look.

dom.bs
```diff
@@ -1118,11 +1120,12 @@ and an <a>event listener</a> <var>listener</var>, set <var>listener</var>'s
...
- <li><p>If the <a>context object</a>'s <a>relevant global object</a> is a
- {{ServiceWorkerGlobalScope}} object and its associated <a>service worker</a>'s
- <a for="service worker">script resource</a>'s
- <a for="script resource">has ever been evaluated flag</a> is set, then <a>throw</a> a
- <code>TypeError</code>. [[!SERVICE-WORKERS]]
+ <li><p>If the <a>context object</a> is a {{ServiceWorkerGlobalScope}} object and its associated
+ <a>service worker</a>'s <a for="service worker">script resource</a>'s
+ <a for="script resource">has ever been evaluated flag</a> is set, then
+ <a>report a warning to the console</a> with a description explaining that the user agent will still
+ start the <a>service worker</a> for the <a>service worker event</a> that was synchronously added.
```
annevk: We don't really know whether it was added at the correct time, right?

dom.bs
```diff
@@ -1062,15 +1062,17 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
-  to avoid non-deterministic changes to the event listeners, invocation of the method is allowed
-  only during the very first evaluation of the service worker script.
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object and its associated
+  <a>service worker</a>'s <a for="service worker">script resource</a>'s
+  <a for="script resource">has ever been evaluated flag</a> is set, then
+  <a>report a warning to the console</a> with a description explaining that the
+  <a>service worker event</a> is to be added synchronously, otherwise the user agent will not start
+  the <a>service worker</a> for the <a>event</a>. [[!SERVICE-WORKERS]]
+
+  <p class="note no-backref">To optimize with the set of <a>service worker events</a> to handle,
```
annevk: This doesn't read well. I also don't think we should use "synchronous" and "asynchronous" as it's not really clear relative to what they're being used. I'm not really sure what to suggest that doesn't get very wordy though.

dom.bs
```diff
@@ -1062,15 +1062,17 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
-
-  <p class="note no-backref">To optimize storing the event types allowed for the service worker and
-  to avoid non-deterministic changes to the event listeners, invocation of the method is allowed
-  only during the very first evaluation of the service worker script.
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object and its associated
+  <a>service worker</a>'s <a for="service worker">script resource</a>'s
+  <a for="script resource">has ever been evaluated flag</a> is set, then
+  <a>report a warning to the console</a> with a description explaining that the
+  <a>service worker event</a> is to be added synchronously, otherwise the user agent will not start
+  the <a>service worker</a> for the <a>event</a>. [[!SERVICE-WORKERS]]
```
mattto: The service worker spec makes skipping event dispatch optional, so "will not start" isn't necessarily true. See https://w3c.github.io/ServiceWorker/#should-skip-event

dom.bs
```diff
@@ -1062,15 +1062,17 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
+
+  <p class="note no-backref">To optimize with the set of <a>service worker events</a> to handle,
+  asynchronously added <a>service worker events</a> are not added to the <a>service worker</a>'s
+  <a for="service worker">set of event types to handle</a>. The <a>event listener</a> will still be
+  added under this condition, but the implmentations have to inform developers about this behavior.
```
erik-anderson: "implementations" is misspelled.

dom.bs
jungkees: I didn't notice the latest change in Service Workers made it optional. I changed the language in this step to *may*.

dom.bs
jungkees: We are able to know whether the event has been added to the service worker's *set of event types to handle*. But I thought that we'd want to show the warning message when the function is called outside of the first evaluation regardless of whether the event has been added synchronously.

dom.bs
jungkees: I tried to rephrase the note by using "deterministically added .. during the very first script evaluation".

dom.bs
jungkees: Fixed. Thanks.

jungkees:
I uploaded a new snapshot. Please take a look.

Also note that @grandfather1 added a test for this: https://github.com/web-platform-tests/wpt/pull/19496/commits/bc28ef58a83706de626d3d69d53cb2d0fd189141. Thanks!

dom.bs
annevk: Please use "might" instead as this is not a normative statement.

dom.bs
annevk: The event yes, but not this specific listener.

dom.bs
```diff
@@ -1062,15 +1062,19 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
-  [[!SERVICE-WORKERS]]
-
-  <p class="note no-backref">To optimize storing the event types allowed for the service worker and
-  to avoid non-deterministic changes to the event listeners, invocation of the method is allowed
-  only during the very first evaluation of the service worker script.
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object and its associated
+  <a>service worker</a>'s <a for="service worker">script resource</a>'s
+  <a for="script resource">has ever been evaluated flag</a> is set, then
+  <a>report a warning to the console</a> with a description explaining that the
+  <a>service worker event</a> is to be added synchronously, otherwise the user agent may not start
```
annevk: We don't know that it's a service worker event.

dom.bs
jungkees: Done.

dom.bs
jungkees: Right. I added a condition to do it only for service worker events in the follow-up commit.

dom.bs
jungkees: Thinking over it again, I think what we want to achieve for removeEventListener() is to warn, regardless whether it's called sync or async, removing the listener won't effect the service worker start-up if the service worker event was added in the script evaluation. Update the text in the follow-up commit.

jungkees:
I gave it another try. Please take a look. (I'd like to come up with simpler explanation but needs clarity about the behavior. Open to better language.)

The build is failing due to the reference to ServiceWorkerGlobalScope's service worker that I introduced. I made a PR to Service Workers to resolve it: https://github.com/w3c/ServiceWorker/pull/1476.

dom.bs
annevk: I think keeping it vaguer as I suggested might be nicer, since if you know what you're doing, adding event listeners late (for your own dispatch logic) is okay. Seems rather odd to do something like that, but I also don't really see reason to depend on so many details of service worker logic here for a warning.

annevk:
@jungkees I've made the changes I suggested above and I also moved the removal warning to "remove an event listener" as it should probably apply to event handlers equally. Please let me know what you think. Ideally we'd merge this early next week.

dom.bs
```diff
@@ -1151,21 +1151,16 @@ and an <a>event listener</a> <var>listener</var>, run these steps:
...
+  <p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object, its
+  <a for="ServiceWorkerGlobalScope">service worker</a>'s
+  <a for="service worker">script resource</a>'s
+  <a for="script resource">has ever been evaluated flag</a> is set, and <var>listener</var>'s
+  <a for="event listener">type</a> matches the {{Event/type}} attribute value of any of the
+  <a>service worker events</a>, then <a>report a warning to the console</a> that this might not give
+  the expected results. [[!SERVICE-WORKERS]]
+
+  <p class="note no-backref">Service workers are only started for event listeners added during the
+  first script evaluation.
```
jungkees: We have an occasion that is not covered by this note. Service Workers changed the skipping behavior to optional: https://w3c.github.io/ServiceWorker/#should-skip-event-algorithm step 1 uses "may". So, the listeners added after the first script evaluation still can see the service worker starts. Should we say: "Service workers may only be started for event listeners added during the first script evaluation?"

dom.bs
```diff
@@ -1195,31 +1190,35 @@ method, when invoked, must run these steps:
...
-<!-- Intentionally not exported. -->
+<p>To <dfn export>remove an event listener</dfn>, given an {{EventTarget}} object
+<var>eventTarget</var> and an <a>event listener</a> <var>listener</var>, run these steps:
+
+<ol>
+ <li><p>If the <a>context object</a> is a {{ServiceWorkerGlobalScope}} object and its
+ <a for="ServiceWorkerGlobalScope">service worker</a>'s
+ <a for="service worker">set of event types to handle</a> contains <var>type</var>, then
+ <a>report a warning to the console</a> that this might not give the expected results.
+ [[!SERVICE-WORKERS]]
```
jungkees: To give the context of what the expected results would be, does leaving a short note along the lines of "Service Workers will still be started after the removal." make sense?

jungkees:
@annevk, thanks for helping on this. It reads a lot simpler, indeed. I left two comments. I'm happy to see it merged early next week.

dom.bs
annevk: Why was that changed?

annevk: I ended up removing the note as the situation is complex and explaining it in detail here feels rather distracting for these general purpose methods.

jungkees: I think [the change](https://github.com/w3c/ServiceWorker/commit/2f7119069da24c095f6a6a8c47cd07f89d6bf143) was made to match the implementation. @mattto, can you give us more context on why you used *may* in that change?

mattto: Yes, I think it was mostly to match the implementations. It was the consensus at the F2Fs according to https://github.com/w3c/ServiceWorker/issues/1200#issuecomment-342655003 and https://github.com/w3c/ServiceWorker/issues/1200#issuecomment-432951697.

annevk: Limiting it to `fetch` seems okay, but making it random effectively requires user agents to look at each other's code of sites take a dependency. That's not great.

mattto: I've reopened the issue to track this.

PR #660, opened 2018-06-28
Define a "has listeners for" algorithm on EventTarget

### LINKED ISSUE #453: Define "has event listener" operation
For https://github.com/whatwg/html/issues/1900, service workers, and a couple of other cases in the platform, we need to determine whether an object has event listeners for a given event type. It would be good to tie these to a single operation that has a little warning alongside pointing out the issues with it.

gterzian:
Fix #453


gterzian:
```
curl: (22) The requested URL returned error: 400 Bad Request
make: *** [deploy] Error 22
```
Is this the result of a failed check, or is there a problem with the build?

gterzian:
@annevk ready for review(I think).

gterzian:
@annevk Ping on this one. From your Github timeline it looks like you were away for a couple of weeks, hope you had a nice holiday...

gterzian:
Ok thanks, I think an example of callers would be Step 9 of [unload a document](https://html.spec.whatwg.org/multipage/#unload-a-document), which I guess could read like:
*If document's Window object **has any event listeners** that were triggered by the earlier unload event step(...)*

gterzian:
@annevk thanks, comment addressed. Yes you are right about the unload steps, we could reword it and put the link to here in a step before the actual firing of the event. I'd be happy to follow up on that too...

annevk:
Thanks! @TimothyGu since you've been looking at events and similar algorithms, would you please double check?

TimothyGu:
It should be fine. I don't know the event dispatching process too well, but could it be possible that when "firing an event at the `Window` object with *legacy target override flag* set", event listeners registered on another object were also invoked? I'm guessing no, but it happened to be true then merely checking if the `Window` object has an `unload` event listener wouldn't been enough… (`body.onload` doesn't count, as the event _listener_ is registered on `Window` still.)

gterzian:
Reminds me of a discussion we had previously at https://github.com/whatwg/dom/issues/453#issuecomment-394246176, because I assumed one would have to check the entire "event-path"...

annevk:
Actually, we should probably also add a note here to discourage new usage as checking for listeners is bad practice (as we point out in https://dom.spec.whatwg.org/#observing-event-listeners).

gterzian:
Ok, I've added a note, let me know what you think.

annevk:
Thanks, I tweaked that a bit in a follow-up commit (will squash before landing). Let me know what you think.

As per above, I'm going to wait with landing until Service Workers / HTML are closer to adopting this.

gterzian:
Looks good to me, I previously got a warning when using `should not`, since it's apparently a non-normative section, and your 'can not' is a nice way around that :)

gterzian:
@TimothyGu thanks! By the way the change is from my use of 'can not'(which was an attempt at avoiding the use of 'should'), to a 'is not', so definitely progress if 'can not' is not a word :)

annevk:
The reason I wanted to avoid "can not" is that an API designer certainly can rely on it, but they are not supposed to.

Issue #657, opened 2018-06-21
Abstract common logic in toggleAttribute() and setAttribute()?


annevk:
In particular for creating a new attribute and appending it. If we don't create an abstraction, we should at least improve the language.

jonathanKingston:
I'm interested in moving that out.

annevk:
Trusted types made this harder to share. The current setup is probably okay.

PR #663, opened 2018-06-29
Refactor toggle attribute

jonathanKingston:
Solves:
https://github.com/whatwg/dom/issues/657
https://github.com/whatwg/html/issues/3779

As I understand it vars are not mutable and are instead copied, right? This mean I still have to copy from `"get an attribute by name"` the code.

Would there be a way to simplify this further perhaps? I could copy out the two common steps in the body of `"get an attribute by name"` and `"get valid an attribute by name"` into another concept?

Are there better names for the concept?

***

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FjonathanKingston%2Fdom%2F543336cbb36a6a2aa93a883f38048b604cff9f6c%2Fdom.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%20663)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/dom%23663.)._
</details>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6189,6 +6189,24 @@ in an <a for="/">element</a> <var>element</var>, run these steps:
 
 <hr>
 
+To <dfn export id=concept-element-attributes-get-valid-by-name>get a valid attribute by name</dfn>
+given a <var>qualifiedName</var>, and <a for="/">element</a> <var>element</var>, run these steps:
+
+<ol>
+ <li><p>If <var>qualifiedName</var> does not match the <code><a type>Name</a></code> production in
+ XML, then <a>throw</a> an "{{InvalidCharacterError!!exception}}" {{DOMException}}.
+
+ <li><p>If <var>element</var> is in the <a>HTML namespace</a> and its <a for=Node>node document</a>
+ is an <a>HTML document</a>, then set <var>qualifiedName</var> to <var>qualifiedName</var> in
+ <a>ASCII lowercase</a>.
+
+ <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
+ <li><p>Let <var>attribute</var> be the first <a>attribute</a> in <var>element</var>'s <a for=Element>attribute list</a>
+ whose <a for=Attr>qualified name</a> is <var>qualifiedName</var>, and null otherwise.
+
+ <li>Return <var>qualifiedName</var>, and <var>attribute</var>.
+</ol>
+
 To <dfn export id=concept-element-attributes-get-by-name>get an attribute by name</dfn> given a
 <var>qualifiedName</var> and <a for="/">element</a> <var>element</var>, run these steps:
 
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -6189,6 +6189,40 @@ in an <a for="/">element</a> <var>element</var>, run these steps:
 
 <hr>
 
+
+To <dfn export id=concept-element-attributes-set-by-name>create an attribute by value</dfn>
```
annevk: Maybe we should name this "create and append an attribute"? Or maybe I don't understand "by value".

The other thing I'm thinking is that it would be nice if https://dom.spec.whatwg.org/#concept-element-attributes-set-value could invoke this algorithm as well. That would require additional arguments for namespace/prefix, but seems doable.

dom.bs
```diff
@@ -6189,6 +6189,40 @@ in an <a for="/">element</a> <var>element</var>, run these steps:
...
+given a <var>qualifiedName</var>, <var>value</var> and <a for="/">element</a> <var>element</var>, run these steps:
+
+<ol>
+ <li><p>Create an <a>attribute</a> whose
+ <a for="Attr">local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
+ <var>value</var>, and <a for=Node>node document</a> is <a>context object</a>'s
+ <a for=Node>node document</a>.
+
+ <li><a lt="append an attribute">Append</a> this <a>attribute</a> to
+ <a>element</a>
```
annevk: You want `<var>element</var>` here and a dot.

dom.bs
```diff
@@ -6189,6 +6189,40 @@ in an <a for="/">element</a> <var>element</var>, run these steps:
...
 <hr>
 
+
+To <dfn export id=concept-element-attributes-set-by-name>create an attribute by value</dfn>
+given a <var>qualifiedName</var>, <var>value</var> and <a for="/">element</a> <var>element</var>, run these steps:
+
+<ol>
+ <li><p>Create an <a>attribute</a> whose
+ <a for="Attr">local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
+ <var>value</var>, and <a for=Node>node document</a> is <a>context object</a>'s
```
annevk: the context object* throughout

dom.bs
```diff
@@ -6189,6 +6189,40 @@ in an <a for="/">element</a> <var>element</var>, run these steps:
...
+ <a for=Node>node document</a>.
+
+ <li><a lt="append an attribute">Append</a> this <a>attribute</a> to
+ <a>element</a>
+
+ <li>Return.
+</ol>
+
+To <dfn export id=concept-element-attributes-get-valid-by-name>get a valid attribute by name</dfn>
```
annevk: "validate and get an attribute by name" perhaps?

dom.bs
```diff
@@ -6590,36 +6611,25 @@ when invoked, must run these steps:
...
- <li><p>Otherwise, if <var>force</var> is not given or is false,
- <a lt="remove an attribute by name">remove an attribute</a> given <var>qualifiedName</var> and the
- <a>context object</a>, and then return false.
+ <li><p>Otherwise, if <var>force</var> is given and is true, return true.
 
- <li><p>Return true.
+ <li><p><a lt="remove an attribute by name">Remove an attribute</a> given <var>qualifiedName</var> and the
+ <a>context object</a>.
+
+ <li>Return false.
```
annevk: `<li><p>` (also above)

annevk (commented):
I quite like the direction this is going in. We do end up with a few more lines, but it's much more clear what is going on.

annevk:
Closing per https://github.com/whatwg/dom/issues/657#issuecomment-4905616571.

PR #670, opened 2018-07-25
Make the DOMTokenList creation steps use "get an attribute value"

### LINKED ISSUE #666: "When a DOMTokenList object is created, then: ..."
https://dom.spec.whatwg.org/commit-snapshots/32efc48fef28c34ed914b60cfb3c29807e617365/#interface-domtokenlist

> When a DOMTokenList object is created, then:​
> 
>     1. Let element be associated element\.
> 
>     2. Let localName be associated attribute’s local name\.
> 
>     3. Let value be the result of getting an attribute given null, localName, and element\.
> 
>     4. Run the attribute change steps for element, localName, value, value, and null\. 

In step 3, "getting an attribute" links to the [get an attribute by namespace and local name](https://dom.spec.whatwg.org/#concept-element-attributes-get-by-namespace) algorithm, which returns an `Attr` node or `null`. This is a problem because in step 4, value is passed to the "attribute change steps" as the oldValue and value arguments, which only accept a `string` or `null`. So, step 3 should either link to the [get an attribute value](https://dom.spec.whatwg.org/#concept-element-attributes-get-value) algorithm or operate on the returned `Attr` node's value. Not really sure which is more appropriate here.

TRowbotham:
This makes it so that if the `DOMTokenList`'s associated element has an attribute that is the associated attribute's local name, at creation time, it will use the "get an attribute value" steps, which will always return a string. Previously, this used the "get an attribute by namespace and local name" steps, which would return an `Attr` node if the associated element had an attribute that is the associated attribute's local name. Fixes #666.

***
<a href="https://whatpr.org/dom/670.html" title="Last updated on Jul 25, 2018, 4:36 PM GMT (7f2c96c)">Preview</a> | <a href="https://whatpr.org/dom/670/32efc48...7f2c96c.html" title="Last updated on Jul 25, 2018, 4:36 PM GMT (7f2c96c)">Diff</a>

### DIFF
--- a/review-drafts/2018-06.bs
+++ b/review-drafts/2018-06.bs
@@ -1,0 +1,10053 @@
+<pre class='metadata'>
+Group: WHATWG
+Date: 2018-06-21
+H1: DOM
+Shortname: dom
+Text Macro: TWITTER thedomstandard
+Abstract: DOM defines a platform-neutral model for events, aborting activities, and node trees.
+Translation: ja https://triple-underscore.github.io/DOM4-ja.html
+Ignored Terms: EmptyString, Array, Document
+Indent: 1
+</pre>
+
+<pre class=anchors>
+urlPrefix: https://www.w3.org/TR/xml/#NT-
+ type: type
+  text: Name; url: Name
+  text: Char; url: Char
+  text: PubidChar; url: PubidChar
+urlPrefix: https://www.w3.org/TR/xml-names/#NT-
+ type: type
+  text: QName; url: QName
+url: https://w3c.github.io/DOM-Parsing/#dfn-createcontextualfragment-fragment
+ type: method; text: createContextualFragment(); for: Range
+type: interface
+ url: https://w3c.github.io/touch-events/#idl-def-touchevent
+  text: TouchEvent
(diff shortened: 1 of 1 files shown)


TRowbotham:
Just a few thoughts.

#### Current behavior
* Uses "get an attribute by namespace and local name", which returns an `Attr` node or `null`
* This probably works 99% of the time because the majority of elements are created without attributes and therefore, "get an attribute by namespace and local name" would return null.
* If the value is null, the "attribute change steps" would use "Step 1" and empty the *token set*.
* If the value is an `Attr` node, the "attribute change steps" will use "Step 2" and will try to run the "ordered set parsing" steps on the `Attr` node. (I guess the best case scenario here is that it gets serialized to "[object Attr]" and the "ordered set parsing" steps return a new *token set* containing "[object" and "Attr]")

#### New behavior
* Uses "get an attribute value", which always returns a string.
* The "attribute change steps" will now always use "Step 2", which runs the string through the "ordered set parsing" steps. (This now always replaces the *token set* with a new *token set* instead of just emptying the existing *token set* when an attribute with the associated local name is not present, since we always pass a string unlike the current behavior, which passes null when the attribute is not present.)

PR #686, opened 2018-09-06
Fix event path iteration

### LINKED ISSUE #685: Capturing event listeners are called during bubbling phase for shadow hosts
We have a [bug report](https://webkit.org/b/174288) saying that WebKit fires event listeners with [*capture* flag](https://dom.spec.whatwg.org/#event-listener-capture) set to true during bubbling phase.

Consider the following DOM:
```js
const hostParent = document.createElement('section');
hostParent.id = 'hostParent';
hostParent.innerHTML = '<div id="host"></div>';
const host = hostParent.firstChild;
const shadowRoot = host.attachShadow({mode: 'closed'});
shadowRoot.innerHTML = '<p id="parent"><span id="target"></span></p>';
...
target.dispatchEvent(new CustomEvent('test', {detail: {}, bubbles: true, composed: true}));
```

Then the event listeners on target, parent, host, and hostParent are invoked in the following order on WebKit & Chrome:
 - hostParent, capturing, eventPhase: CAPTURING_PHASE
 - parent, capturing, eventPhase: CAPTURING_PHASE
 - target, capturing, eventPhase: AT_TARGET
 - target, non-capturing, eventPhase: AT_TARGET
 - parent, non-capturing, eventPhase: BUBBLING_PHASE
 - host, capturing, eventPhase: AT_TARGET
 - host, non-capturing, eventPhase: AT_TARGET
 - hostParent, non-capturing, eventPhase: BUBBLING_PHASE

As far as I can tell, the current behavior of WebKit & Chrome is in accordance with the latest DOM specification. What happens is that step 5 (event path construction) in [dispatching an event](https://dom.spec.whatwg.org/#concept-event-dispatch) ends up appending an tuple/struct (we should probably stick with either term, btw) with *target* set to `null`. In step 14, we only invoke event listeners with capture flag set if *target* is `null`, meaning that shadow hosts' event listeners with capture flag set would NOT be called. Then in step 15, we set `eventPhase` to `AT_TARGET` and [invoke](https://dom.spec.whatwg.org/#concept-event-listener-invoke) event listeners. Because the concept [to inner invoke](https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke) an event listener doesn't skip event listeners with capture flag set to true when the event phase is AT_TARGET, we end up firing capturing event listeners during bubbling.

annevk:
Instead of shoehorning all target handling into the bubbling iteration, this separates "capturing" iteration from "bubbling" iteration and the Event object's phase is set to target as appropriate in both.

This also invokes the event listeners in a more natural order.

Tests: ...

Fixes #685.

***
<a href="https://whatpr.org/dom/686.html" title="Last updated on Mar 1, 2019, 10:49 AM UTC (7aaff50)">Preview</a> | <a href="https://whatpr.org/dom/686/42d2485...7aaff50.html" title="Last updated on Mar 1, 2019, 10:49 AM UTC (7aaff50)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1315,22 +1315,28 @@ for discussion).
    <var>clearTargetsStruct</var>'s <a for=Event/path>touch target list</a> is a <a for=/>node</a>
    and its <a for=tree>root</a> is a <a for=/>shadow root</a>, and false otherwise.
 
-   <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/CAPTURING_PHASE}}.
-
    <li><p>If <var>activationTarget</var> is non-null and <var>activationTarget</var> has
    <a for=EventTarget>legacy-pre-activation behavior</a>, then run <var>activationTarget</var>'s
    <a for=EventTarget>legacy-pre-activation behavior</a>.
 
    <li>
-    <p>For each <var>struct</var> in <var>event</var>'s <a for=Event>path</a>, in reverse order:
+    <p><a for=list>For each</a> <var>struct</var> in <var>event</var>'s <a for=Event>path</a>, in
+    reverse order:
 
     <ol>
-     <li><p>If <var>struct</var>'s <a for=Event/path>target</a> is null, then <a>invoke</a> with
-     <var>struct</var>, <var>event</var>, and <var>legacyOutputDidListenersThrowFlag</var> if given.
+     <li><p>If <var>struct</var>'s <a for=Event/path>target</a> is non-null, then set
+     <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/AT_TARGET}}.
+
+     <li><p>Otherwise, set <var>event</var>'s {{Event/eventPhase}} attribute to
+     {{Event/CAPTURING_PHASE}}.
+
+     <li><p><a>Invoke</a> with <var>struct</var>, <var>event</var>, "<code>capturing</code>", and
+     <var>legacyOutputDidListenersThrowFlag</var> if given.
     </ol>
 
    <li>
-    <p>For each <var>struct</var> in <var>event</var>'s <a for=Event>path</a>, in order:
+    <p>If <var>event</var>'s {{Event/bubbles}} attribute is true, then <a for=list>for each</a>
+    <var>struct</var> in <var>event</var>'s <a for=Event>path</a>:
 
     <ol>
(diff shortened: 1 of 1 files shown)


annevk:
@whatwg/components I think this is what we want here, but careful review appreciated. If I did it correctly this does seem like a much nicer model.

dom.bs
```diff
@@ -1306,43 +1306,47 @@ for discussion).
...
+   <li><p>Let <var>clearTargetsStruct</var> be the last struct in <var>event</var>'s
    <a for=Event>path</a> whose <a for=Event/path>target</a> is non-null.
 
-   <li><p>Let <var>clearTargets</var> be true if <var>clearTargetsTuple</var>'s
-   <a for=Event/path>target</a>, <var>clearTargetsTuple</var>'s <a for=Event/path>relatedTarget</a>,
-   or an {{EventTarget}} object in <var>clearTargetsTuple</var>'s
-   <a for=Event/path>touch target list</a> is a <a for=/>node</a> and its <a for=tree>root</a> is a
-   <a for=/>shadow root</a>, and false otherwise.
-
-   <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/CAPTURING_PHASE}}.
```
annevk: As far as I can tell "legacy-pre-activation behavior" cannot trigger side effects so setting this later is not observable.

rniwa:
The change looks sensible to me.

dom.bs
rniwa: Actually, it looks like we fire change event on radio buttons & checkboxes so the event phase change is observable. Gecko follows the current spec and WebKit & Blink uses the new behavior.

https://gist.github.com/rniwa/6c502dca3e16d5816db7958ce7bab4d7

annevk: Thanks, we should add that test to the test suite. None seems like a reasonable phase to me given that traversal hasn't started yet, WDYT @smaug----?

rniwa: Agreed that None seems more reasonable than capturing given the event dispatching hasn't even started yet.

smaug----: https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=6198

smaug----: From API user's point of view Gecko's behavior makes more sense, since for them the event dispatch clearly has started already once dispatchEvent is called.

rniwa: I disagree. Just because we're in `dispatchEvent`, doesn't mean `eventPhase` needs to be something other than "none".

Issue #688, opened 2018-09-09
Change `#time-origin` to `#sec-time-origin` for <anchor> tag.


payingattention:
Reading https://dom.spec.whatwg.org/commit-snapshots/28cdb47e557c814fcca60759f4cce43e9c0e9a47 an error: the <anchor> tag for "#time-origin" (https://w3c.github.io/hr-time/#time-origin) does not point to "#sec-time-origin" (https://w3c.github.io/hr-time/#sec-time-origin).

I do not know if to `blame` (https://git-scm.com/docs/git-blame) W3.org or WHATWG.org specifically. I am scared to do the code (just lines https://github.com/whatwg/dom/blob/28cdb47e557c814fcca60759f4cce43e9c0e9a47/dom.bs#L36, https://github.com/whatwg/dom/blob/28cdb47e557c814fcca60759f4cce43e9c0e9a47/dom.bs#L570 {basically add/remove of "`sec`" is the language/spelling problem?} I think?), honestly I need a second opinion publicly before I make a change like that.

Automatic bug generation text [From "File an issue about the selected text." tool]:

https://dom.spec.whatwg.org/#ref-for-time-origin

> time origin

annevk:
I think we want https://w3c.github.io/hr-time/#dfn-time-origin even. I suspect this changed in High Resolution Time at some point, but it's rather hard to determine as git doesn't hold the generated HTML.

A good way to fix this would be to make High Resolution Time export its definitions so we don't have to include a "hard coded" link.

cc @igrigorik

domenic:
I don't think ReSpec specs have a notion of "exporting a definition", or of stability over time in general.

igrigorik:
Indeed, you want the hr-time definition. Paging @marcoscaceres for input on respec. :)

annevk:
@payingattention if you are willing to sign https://participate.whatwg.org/agreement and want to fix that'd be great. `type:dfn; text: time origin` needs to become `type:dfn; text: time origin; url: dfn-time-origin`.

payingattention:
Okay, signed w/ https://github.com/whatwg/participant-data/commit/f36dee22d752491c60d1a0fe3dcf9b5360397885. Roger. Definitely thank you (tripped by `dfn` vs. `sec` apparently). Give me a sec.

marcoscaceres:
@domenic, you know I love you, but...

> I don't think ReSpec specs have a notion of "exporting a definition",

[...]

PR #693, opened 2018-09-10
Change `#time-origin` line, def <a> w/ `#dfn-time-origin`

payingattention:
Change `#time-origin` line to add URL link `#dfn-time-origin` (help https://w3c.github.io/hr-time/#{dfn-, sec-}time-origin broke) for <anchor> tag line just https://github.com/whatwg/dom/issues/688 reported.

***
<a href="https://whatpr.org/dom/693.html" title="Last updated on Sep 18, 2018, 7:33 AM GMT (0508cb4)">Preview</a> | <a href="https://whatpr.org/dom/693/42d2485...0508cb4.html" title="Last updated on Sep 18, 2018, 7:33 AM GMT (0508cb4)">Diff</a>


dom.bs
```diff
@@ -33,7 +33,9 @@ urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
  text: Realm; url: realm; type: dfn
 urlPrefix: https://w3c.github.io/hr-time/#; spec: HR-TIME
  type:typedef; urlPrefix: dom-; text: DOMHighResTimeStamp
- type:dfn; text: time origin
+ type:dfn;
+ text: time origin;
+ url: dfn-time-origin
```
annevk: These three have to be on the same line.

payingattention: Fair to ask if that is for style?

Just to answer for why I did that, I thought about how that is an exact (or should I say verbatim?) copy of the formatting standard for former/earlier lines, in fact the original <pre> section is the readable column with linebreak style, not the inlined/minified/abbreviated style, from the very beginning of the document code in question.

annevk: https://tabatkins.github.io/bikeshed/#infotree

payingattention: Thank you for the link.

So why are "`type`, `text`, `url`" otherwise herein given separate/individual lines?

I hear the documentation says otherwise as well (showing single and then nesting format), so is our spec in conflict, or is their specification tooling causing the problem, er,... still fair to ask if that's for style yo?

annevk: They are only on separate lines as well as appropriately indented when they are shared, as per the format. If you mean something else then please give an example, as it isn't entirely clear from your message where the confusion stems from.

payingattention: Hmm. Fair. I really think about how some coders might allow that for style.

I deferred to you however and made the correction to our specification, I did not mean my <style> to be mean, questioning about making a style statement.

annevk:
Do you want to add yourself to the Acknowledgments section?

payingattention:
id appreciate the acknowledgements section. thank you again.

annevk:
Great, please add another commit with a change to that section!

annevk:
Okay, so your name is "Prosody Vereable Context(", including the open parenthesis, and you want to be acknowledged as "prosody—Gab Vereable Context("?

payingattention:
Thank you. Exactly please. Ty for the help with coding commits.

PR #695, opened 2018-09-11
Manage slot manipulation centrally and special case replace operations

annevk:
This is an attempt at fixing https://github.com/w3c/webcomponents/issues/764.

I first wrote an algorithm that combined the operations done by remove and insert. I then used that for replace and replace all. I then abstracted it to avoid duplication.

I'm a little worried that this is not correct as this delays assigning slotables to a slot quite a bit, which I think might be observable in some cases.


dom.bs
```diff
@@ -2382,17 +2412,11 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
    <li>If <var>node</var> is a {{Text}} node, run the <a>child text content change steps</a> for
    <var>parent</var>.
 
-   <li><p>If <var>parent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
-   <var>parent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> is the empty list,
-   then run <a>signal a slot change</a> for <var>parent</var>.
-
-   <li><p>Run <a>assign slotables for a tree</a> with <var>node</var>'s <a for=tree>root</a>.
+   <li><p>If the <i>suppress observers flag</i> is unset, then <a>update slots</a> with « »,
+   <var>parent</var>, and « <var>node</var> ».
```
rniwa: Why do we need to run it on both `parent` and `node`? We just inserted `node` into the same tree as `parent`.

annevk: That's not what this does, right? It uses parent for different things then node.

annevk:
@EdgarChen maybe you could review this. The high level goal is to reduce the number of `slotchange` events for replace all operations such as setting `textContent`.

dom.bs
```diff
@@ -2203,6 +2203,36 @@ steps:
...
+<ol>
+ <li>
+  <p>For each <var>removedNode</var> in <var>removedNodes</var>:
+
+  <ol>
+   <li><p>If <var>removedNode</var> is <a for=slotable>assigned</a>, then run
+   <a>assign slotables</a> for <var>removedNode</var>'s <a>assigned slot</a>.
+
+   <li><p>If <var>removedNode</var> has an <a>inclusive descendant</a> that is a <a>slot</a>, then
+   run <a>assign slotables for a tree</a> with <var>removedNode</var>.
```
EdgarChen: I found that it seems the order of slotchange event is changed in a certain case, for example, https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=6882. If I read spec correctly, the event order of current spec is [slot1, slot3. slot2], and it changes to [slot2, slot1, slot3]. I am not sure if this will cause any problem? Current browsers have a different order than spec, and the order is not the same between browsers, either.

annevk: Thanks! @rniwa do you have any thoughts on order? I looked this through with Edgar and either way we are going to change the order here in order to centralize some of this work for "replace all". I don't care strongly for which slot we signal things first.

And also, given that browsers are already different, it probably does not matter much?

Not sure who to copy from Chrome. @domenic?

rniwa: Hm... looks like Blink & Gecko do: [slot1, slot2, slot3] and WebKit does [slot3, slot2, slot1]

annevk:
@EdgarChen I've made the wording consistent in #762.

dom.bs
annevk: I think the main point here is that if we want to centralize work and special "replace all" to avoid redundant events, the order will change somewhere, right? So we need to make some kind of call about whether that is okay and whether there are certain principles for deciding on an order. (I think it's okay and although I'd like tree order it doesn't seem worth the additional instructions that'd require.)

annevk:
@rniwa this still seems good, but it would be good if someone could implement this to identify the tests we'll need to adjust.

rniwa:
I'm not certain when I can get to it but I'd be interested in implementing this behavior assuming it's compatible with existing content.

dom.bs
```diff
@@ -2585,6 +2612,9 @@ within a <var>parent</var>, run these steps:
  <li><p>If <var>node</var> is non-null, then <a for=/>insert</a> <var>node</var> into
  <var>parent</var> before null with the <i>suppress observers flag</i> set.
 
+ <li><p>If the <i>suppress observers flag</i> is unset, then <a>update slots</a> with
+ « <var>removedNodes</var> », <var>parent</var>, and « <var>addedNodes</var> ».
```
annevk, line 2616: This is the "replace all" algorithm that doesn't have a suppress observers flag at the moment so that would have to be added.

PR #696, opened 2018-09-13
Fix composedPath() once more

### LINKED ISSUE #684: composedPath() is wrong
@pmdartus discovered in https://github.com/jsdom/jsdom/pull/2347#discussion_r212829365 that the algorithm as-is exposes closed shadow nodes. This also happens in Firefox.

See `shadow-dom/event-composed-path.html` in wpt.

It seems to solve this we'd need to record for each item in an event's path what nesting level it's at and whether it's inside a closed shadow tree?

(This was missed in #535.)

cc @whatwg/components

annevk:
The existing algorithm exposed nodes in shadow trees that should remain hidden. (This wasn't noticed in #535.)

This changes "get the parent" to return a tuple and therefore requires downstream changes in at least Indexed DB.

Tests: ...

Fixes #684.

***
<a href="https://whatpr.org/dom/696.html" title="Last updated on Sep 13, 2018, 12:40 PM GMT (fd6a070)">Preview</a> | <a href="https://whatpr.org/dom/696/42d2485...fd6a070.html" title="Last updated on Sep 13, 2018, 12:40 PM GMT (fd6a070)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1014,7 +1014,7 @@ object, an <a>event listener</a> is a broader concept as can be seen above.
 <p>Each {{EventTarget}} object also has an associated <dfn export>get the parent</dfn> algorithm,
 which takes an <a>event</a> <var>event</var>, and returns null or a
 <dfn export>parent event target</dfn>, which is a <a for=/>tuple</a> consisting of an
-<dfn export for="parent event target">node</dfn> (an {{EventTarget}} object) and a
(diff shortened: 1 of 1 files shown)


annevk:
(The WebKit patch suggests this is at least tested by https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/event-composed-path-after-dom-mutation.html.)

PR #698, opened 2018-09-17
Avoid setting target to the Window object

### LINKED ISSUE #697: Dispatch event always set the event.target to the global object after dispatch
https://dom.spec.whatwg.org/commit-snapshots/42d2485c4f84fb210452ca53e07503229716b147/#dispatching-events

> If parent is a node and target’s root is a shadow\-including inclusive ancestor of parent, then:​

When walking up the parent chain and *parent* is the global object, instead of going through the (5.9.6) branch of the algorithm it goes through the (5.9.8) branch. The last tuple of all the event path, would always have it's `target` set to the global object. Because of this, after being dispatched the `event.target` is always the global object.

The following test fails: [Event-defaultPrevented-after-dispatch.html](https://github.com/web-platform-tests/wpt/blob/master/dom/events/Event-defaultPrevented-after-dispatch.html)

```js
test(function() {
    var EVENT = "foo";
    var TARGET = document.getElementById("target");
    var evt = document.createEvent("Event");
    evt.initEvent(EVENT, true, true);

    TARGET.addEventListener(EVENT, this.step_func(function(e) {
		// <-- OK: event.target === TARGET
        e.preventDefault();
        assert_true(e.defaultPrevented, "during dispatch");
    }), true);
    TARGET.dispatchEvent(evt);

    assert_true(evt.defaultPrevented, "after dispatch");
    assert_equals(evt.target, TARGET); // <-- FAILS: event.target === window
    assert_equals(evt.srcElement, TARGET);
}, "Default prevention via preventDefault");
```

In order to get around this issue the 5.9.6 should also check if *parent* is the *global object*.

annevk:
fc165642b437c548a11222c85ddda0094ed40ea5 did not adequately consider the global object.

Fixes #697.

***
<a href="https://whatpr.org/dom/698.html" title="Last updated on Sep 17, 2018, 9:31 AM GMT (0e3e225)">Preview</a> | <a href="https://whatpr.org/dom/698/42d2485...0e3e225.html" title="Last updated on Sep 17, 2018, 9:31 AM GMT (0e3e225)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1271,8 +1271,9 @@ for discussion).
      <var>touchTarget</var> against <var>parent</var> to <var>touchTargets</var>.
 
      <li>
-      <p>If <var>parent</var> is a <a>node</a> and <var>target</var>'s <a for=tree>root</a> is a
-      <a>shadow-including inclusive ancestor</a> of <var>parent</var>, then:
+      <p>If <var>parent</var> is a {{Window}} object, or <var>parent</var> is a <a>node</a> and
+      <var>target</var>'s <a for=tree>root</a> is a <a>shadow-including inclusive ancestor</a> of
+      <var>parent</var>, then:
 
       <ol>
        <li><p>If <var>isActivationEvent</var> is true, <var>event</var>'s {{Event/bubbles}}


annevk:
@pmdartus thanks for finding that! I think this fixes it, but I should probably double check I didn't miss anything else.

PR #699, opened 2018-09-18
Fix composedPath() once more

### LINKED ISSUE #684: composedPath() is wrong

annevk:
The existing algorithm exposed nodes in shadow trees that should remain hidden. (This wasn't noticed in #535.)

This roughly matches the approach taken by jsdom to solve this issue and unlike #696 requires no downstream changes.

Tests: ...

Fixes #684.

***
<a href="https://whatpr.org/dom/699.html" title="Last updated on Sep 18, 2018, 10:34 AM GMT (60b4d4a)">Preview</a> | <a href="https://whatpr.org/dom/699/0b04ae8...60b4d4a.html" title="Last updated on Sep 18, 2018, 10:34 AM GMT (60b4d4a)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -587,38 +587,101 @@ was initialized to. When an <a>event</a> is created the attribute must be initia
 steps:
 
 <ol>
- <li><p>Let <var>reversedComposedPath</var> be an empty <a for=/>list</a>.
+ <li><p>Let <var>composedPath</var> be an empty <a for=/>list</a>.
 
- <li><p>Let <var>hiddenSubtreeLevel</var> be 0.
+ <li><p>Let <var>path</var> be the <a>context object</a>'s <a for=Event>path</a>.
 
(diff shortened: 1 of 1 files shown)


annevk:
@pmdartus fancy reviewing this?

annevk:
This is tested by `shadow-dom/event-composed-path.html` so no need for tests.

Bugs (only Firefox has an issue):
* https://bugzilla.mozilla.org/show_bug.cgi?id=1510204

PR #717, opened 2018-11-26
Ground primitives in agents

### LINKED ISSUE #88: _Which_ unit of related similar-origin browsing contexts?
Mutation observer algorithm steps say:

> Let notify list be a copy of unit of related similar-origin browsing contexts' list of MutationObserver objects. 

and

> append it to the unit of related similar-origin browsing contexts' list of MutationObserver objects, and then return it.

This seems pretty ambiguous to me. Shouldn't be trying to go from e.g. entry settings object to browsing context to unit of related similar-origin browsing contexts, or something?

Maybe HTML needs to define "current unit of related similar-origin browsing contexts"?

annevk:
Using "unit of related similar-origin browsing contexts" never made much sense as a browsing context could always hold several cross-site documents.

Helps with https://github.com/whatwg/html/issues/4198.

Fixes #88.

***
<a href="https://whatpr.org/dom/717.html" title="Last updated on Jan 16, 2019, 12:34 PM UTC (66d734c)">Preview</a> | <a href="https://whatpr.org/dom/717/15e2f41...66d734c.html" title="Last updated on Jan 16, 2019, 12:34 PM UTC (66d734c)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2208,8 +2208,8 @@ steps:
 
 <h5 id=signaling-slot-change>Signaling slot change</h5>
 
-<p>Each <a>similar-origin window agent</a> has <dfn noexport>signal slots</dfn> (a <a for=/>set</a>
-of <a>slots</a>). Unless stated otherwise it is empty. [[!HTML]]
+<p>Each <a>similar-origin window agent</a> has <dfn noexport id=signal-slot-list>signal slots</dfn>
+(a <a for=/>set</a> of <a>slots</a>), which is initially empty. [[!HTML]]
 
 <p>To <dfn noexport>signal a slot change</dfn>, for a <a>slot</a> <var>slot</var>, run these steps:
 
@@ -3193,8 +3193,9 @@ invoked, must run these steps:
 <dfn noexport>mutation observer compound microtask queued flag</dfn>, which is initially unset.
 [[!HTML]]
 
-<p>Each <a>similar-origin window agent</a> also has <dfn noexport>mutation observers</dfn> (a
-<a for=/>set</a> of zero or more {{MutationObserver}} objects), which is initially empty.
+<p>Each <a>similar-origin window agent</a> also has
+<dfn noexport id=mutation-observer-list>mutation observers</dfn> (a <a for=/>set</a> of zero or more
+{{MutationObserver}} objects), which is initially empty.
 
 <p>To <dfn export>queue a mutation observer compound microtask</dfn>, run these steps:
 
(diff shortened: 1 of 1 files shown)


annevk:
(FWIW, I realize that similar-origin window agent is currently defined in terms of browsing contexts, but that'll go away over time as it's inaccurate.)

domenic:
> Using "unit of related similar-origin browsing contexts" never made much sense as a browsing context could always hold several cross-site documents.

I don't understand this. If a browsing context held a cross-site document, it would no longer be in that UORSOBCs.

In general I disagree that agents are more primitive or useful than UORSOBCs, but we can discuss that over in the HTML issue I guess.

annevk:
How can a browsing context both be in it and not be in it? Remember that it holds a sequence of documents. Whether a document is active shouldn't be relevant.

domenic:
Why not? That's how it's defined.

> Each unit of related browsing contexts is then further divided into the smallest number of groups such that every member of each group **has an active document** with an origin that, through appropriate manipulation of the document.domain attribute, could be made to be same origin-domain with other members of the group, but could not be made the same as members of any other group. Each such group is a unit of related similar-origin browsing contexts.

That definition is quite nice IMO.

annevk:
I think it's rather terrible as it isn't stable over time and it's duplicative with agent clusters. And sometimes some of the documents are inactive, which then wouldn't be covered by this definition. Where are those documents located?

domenic:
It's not duplicative; they're layered.

Being stable over time is not a desirable property. Your setup is also not stable over time, you just need to manually add and remove things to it to make it "unstable", instead of automatically getting the right thing.

Stated another way, the spec's current setup is

```js
function getURSOBC(bc) {
}
```

whereas you're wanting to move toward a world of

```js
const bcsToBCGroup = new Map();
// manually call bcsToBCGroup.set(), bcsToBCGroup.delete(), etc.

function getBCGroup(bc) {
  return bcsToBCGroup.get(bc);
}
```

I think the former is nicer and simpler.

annevk:
The only reason we need "URSOBC" is for the definition of agents (i.e., event loop), not for anything related to browsing contexts. Browsing contexts use agents to host documents in. Conflating them doesn't make logical sense.

domenic:
I just disagree with that layering. The definition of agents needs to be grounded in something. URSOBC is a good grounding for them, and a much better one than manually adding and removing realms to them all the time.

I prefer the layering of agents, event loops, mutation observers, etc. depending on URSOBC, instead of manually maintaining agents and then layering event loops, mutation observers, etc. on top of them.

The choices are equivalent, it's just a matter of whether we churn from the current setup, or build on it. I don't see any benefit from the churn.

annevk:
I think we reached agreement in https://github.com/whatwg/html/issues/4198, can you look at the PR?

dom.bs
```diff
@@ -2206,16 +2208,17 @@ steps:
...
+<a>slots</a>). Unless stated otherwise it is empty. [[!HTML]]
 
 <p>To <dfn noexport>signal a slot change</dfn>, for a <a>slot</a> <var>slot</var>, run these steps:
 
 <ol>
- <li><p>If <var>slot</var> is not in <a>unit of related similar-origin browsing contexts</a>'
- <a>signal slot list</a>, append <var>slot</var> to
- <a>unit of related similar-origin browsing contexts</a>' <a>signal slot list</a>.
+ <li><p>Let <var>agent</var> be the <a>similar-origin window agent</a> to which <var>slot</var>'s
+ <a>relevant Realm</a> belongs.
```
domenic: We may want to consider introducing a shorthand in HTML so you can say "slot's relevant agent".

dom.bs
```diff
@@ -3213,15 +3215,17 @@ objects), which is initially empty.
 <ol>
  <li><p>Unset <a>mutation observer compound microtask queued flag</a>.
 
- <li><p>Let <var>notifyList</var> be a <a for=list>clone</a> of
- <a>unit of related similar-origin browsing contexts</a>' <a>mutation observer list</a>.
+ <li><p>Let <var>agent</var> be the <a>similar-origin window agent</a> <a>current Realm Record</a>
+ belongs to.
```
domenic: "the" before "current Realm Record".

Again we could define "the current agent" in HTML.

annevk:
I noticed two more things:

* These members on agents should not be exported.
* They are both sets.

Issue #664, opened 2018-07-04
Order of "contains" in removeEventListener


TimothyGu:
> The <a href="https://dom.spec.whatwg.org/#dom-eventtarget-removeeventlistener">**<code>removeEventListener(*type*, *callback*, *options*)</code>**</a> method, when invoked, must run these steps:
>
> …
>
> 3. If the context object’s [event listener list](https://dom.spec.whatwg.org/#eventtarget-event-listener-list) [contains](https://infra.spec.whatwg.org/#list-contain) an [event listener](https://dom.spec.whatwg.org/#concept-event-listener) whose type is *type*, callback is *callback*, and capture is *capture*, then <a href="https://dom.spec.whatwg.org/#remove-an-event-listener">remove an event listener</a> with the context object and that event listener.

> A [list](https://infra.spec.whatwg.org/#list) [**contains**](https://infra.spec.whatwg.org/#list-contain) an [item](https://infra.spec.whatwg.org/#list-item) if it appears in the list.

What happens if there are multiple event listeners in the `EventTarget` object's event listener list that fit those searching criteria (type, callback, capture)? Is the searching process in ascending or descending order? Or are *all* the listeners that fit those criteria removed? The order is observable through the use of the [`passive`](https://dom.spec.whatwg.org/#dom-addeventlisteneroptions-passive) and [`once`](https://dom.spec.whatwg.org/#dom-addeventlisteneroptions-once) options.

Inspired by nodejs/node#21635.

TimothyGu:
Never mind, [add an event listener](https://dom.spec.whatwg.org/#add-an-event-listener) looks through the list using those criteria before appending, and exits if one exists already.

annevk:
It seems reasonable to add a note.

PR #722, opened 2018-12-05
Add a note about duplicate event listeners for removeEventListener

LukeZielinski:
First crack at first spec change.

Addresses https://github.com/whatwg/dom/issues/664

***
<a href="https://whatpr.org/dom/722.html" title="Last updated on Dec 10, 2018, 9:44 AM GMT (6c15c49)">Preview</a> | <a href="https://whatpr.org/dom/722/23ec642...6c15c49.html" title="Last updated on Dec 10, 2018, 9:44 AM GMT (6c15c49)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1216,6 +1216,9 @@ method, when invoked, must run these steps:
  with the <a>context object</a> and that <a>event listener</a>.
 </ol>
 
+<p class="note">The event listener list will not contain duplicate event listeners matching
+<var>type</var>, <var>callback</var>, and <var>capture</var> because the <a>add an event listener</a> concept does not add duplicates.
+
 <p>The <dfn method for=EventTarget><code>dispatchEvent(<var>event</var>)</code></dfn> method, when
 invoked, must run these steps:
 
--- a/dom.bs
+++ b/dom.bs
@@ -1217,7 +1217,7 @@ method, when invoked, must run these steps:
 </ol>
 
 <p class="note">The event listener list will not contain duplicate event listeners matching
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -1216,6 +1216,9 @@ method, when invoked, must run these steps:
  with the <a>context object</a> and that <a>event listener</a>.
 </ol>
 
+<p class="note">The event listener list will not contain duplicate event listeners matching
+<var>type</var>, <var>callback</var>, and <var>capture</var> because the <a>add an event listener</a> concept does not add duplicates.
```
foolip:
```suggestion
<var>type</var>, <var>callback</var>, and <var>capture</var> because <a>add an event listener</a> does not add duplicates.
```

dom.bs
```diff
@@ -1216,6 +1216,9 @@ method, when invoked, must run these steps:
  with the <a>context object</a> and that <a>event listener</a>.
 </ol>
 
+<p class="note">The event listener list will not contain duplicate event listeners matching
```
foolip:
```suggestion
<p class="note">The event listener list will not contain multiple event listeners matching
```

annevk:
Thanks @LukeZielinski! I added your name to the Acknowledgments section and made some small tweaks. All good?

(Also, a further cleanup here might be to change "list" to "set" or perhaps some kind of "map" even, but that would be a little more involved.)

annevk:
@LukeZielinski (un)fortunately I don't get a notification for a thumbs up, so I missed this was all good to go. Landed now, thanks for your efforts!

PR #727, opened 2018-12-29
Fix a small but important error in composedPath()

emilio:
The comparison needs to be 'less than or equal', otherwise you don't prepend the items from the same level as the target.

Trivially, in a 'no shadow DOM' situation, all the items in the path before the current target wouldn't be in the path without this patch, which is wrong.

This is tested already by event-composed-path.html in WPT.

***
<a href="https://whatpr.org/dom/727.html" title="Last updated on Dec 29, 2018, 6:42 PM UTC (157b3af)">Preview</a> | <a href="https://whatpr.org/dom/727/ee24b02...157b3af.html" title="Last updated on Dec 29, 2018, 6:42 PM UTC (157b3af)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -633,7 +633,7 @@ steps:
    <li><p>If <var>path</var>[<var>index</var>]'s <a for=Event/path>root-of-closed-tree</a> is true,
    then increase <var>currentHiddenLevel</var> by 1.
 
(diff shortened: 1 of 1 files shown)


annevk:
Thanks and sorry for the mistake!

PR #732, opened 2019-02-08
Use a queue of deferred steps in the node insertion algorithm

nox:


dom.bs
```diff
@@ -2389,12 +2389,10 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
-<dfn export id=concept-node-insert-ext>insertion steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed <var ignore>insertedNode</var>, as indicated in the <a for=/>insert</a>
-algorithm below.
-<!-- See https://github.com/whatwg/dom/issues/34#issuecomment-125571750 for why we might need to
-     adjust this further based on the requirements of the script element. There might be other ways
-     to define that though as Olli suggests, so leaving that out for now. -->
+<dfn export id=concept-node-insert-ext>insertion steps</dfn> or
+<dfn export id=concept-node-children-added>children added steps</dfn> for all or some
+<a for=/>nodes</a>. The two algorithms are passed <var ignore>insertedNode</var> and
+<var ignore>deferredStepsQueue</var>, as indicated in the <a for=/>insert</a> algorithm below.
```
annevk: Should we mention child text content change steps here?

dom.bs
```diff
@@ -2440,8 +2438,12 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
  <li><p>Let <var>previousSibling</var> be <var>child</var>'s <a>previous sibling</a> or
  <var>parent</var>'s <a>last child</a> if <var>child</var> is null.
 
+ <li><p>Let <var>deferredStepsQueue</var> be the empty <a>queue</a>.
```
annevk: an empty queue*

dom.bs
```diff
@@ -2440,8 +2438,12 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+ <li><p>Let <var>textChanged</var> be false.
+
  <li>
-  <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
+  <p>For each <var>node</var> in <var>nodes</var>:
```
annevk: Optional nit: `<a for=list>For each</a>`

dom.bs
```diff
@@ -2453,8 +2455,7 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
    <li><p>If <var>parent</var> is a <a for=Element>shadow host</a> and <var>node</var> is a
    <a>slotable</a>, then <a>assign a slot</a> for <var>node</var>.
 
-   <li>If <var>node</var> is a {{Text}} node, run the <a>child text content change steps</a> for
-   <var>parent</var>.
+   <li>If <var>node</var> is a {{Text}} node, set <var>textChanged</var> to true.
```

dom.bs
```diff
@@ -2490,6 +2492,22 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
    </li>
   </ol>
 
+ <li><p>If <var>textChanged</var> is true, run the <a>child text content change steps</a> for
```
annevk: then run*

(so it's observable if we run these before the insertion steps? I guess there's a test?)

dom.bs
```diff
@@ -2490,6 +2492,22 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+ <var>parent</var>.
+
+ <li><p>If <var>nodes</var> is not empty, run the <a>children added steps</a> with
+ <var>parent</var> and <var>deferredStepsQueue</var>.
+
+ <li>
+  <p>While <var>deferredStepsQueue</var> is not empty:
```
annevk: `<a for=queue>is not empty</a>` iirc

dom.bs
nox: I don't know, they are defined elsewhere with `Text` and they don't take `deferredStepsQueue`.

dom.bs
nox: Changed that.

dom.bs
nox: Changed that.

dom.bs
nox: Changed that.

dom.bs
nox: It's not observable, just thought it would be better to specify them as running only once, otherwise for example if you follow the spec pedantically you would recreate a `style`'s stylesheet once for each added `Text` child.

dom.bs
nox: Changed that.

dom.bs
```diff
@@ -2490,6 +2492,22 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
   </ol>
 
+ <li><p>If <var>textChanged</var> is true, then run the <a>child text content change steps</a> for
+ <var>parent</var>.
+
+ <li><p>If <var>nodes</var> is not empty, then run the <a>children added steps</a> with
+ <var>parent</var> and <var>deferredStepsQueue</var>.
+
+ <li>
+  <p>While <var>deferredStepsQueue</var> <a for=list>is not empty</a>:
```
annevk: How about:

> While deferredStepsQueue is not empty, run the result of dequeueing deferredStepsQeue.

dom.bs
annevk: I see, I think it makes sense to group them, but that'd be a follow-up indeed.

dom.bs
nox: Amended that.

annevk:
We should note in the eventual commit that this helps fixing https://github.com/whatwg/html/issues/1127 and #575

dom.bs
```diff
@@ -2389,12 +2389,10 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
 
 <p><a lt="Other applicable specifications">Specifications</a> may define
-<dfn export id=concept-node-insert-ext>insertion steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed <var ignore>insertedNode</var>, as indicated in the <a for=/>insert</a>
-algorithm below.
-<!-- See https://github.com/whatwg/dom/issues/34#issuecomment-125571750 for why we might need to
-     adjust this further based on the requirements of the script element. There might be other ways
-     to define that though as Olli suggests, so leaving that out for now. -->
+<dfn export id=concept-node-insert-ext>insertion steps</dfn> or
+<dfn export id=concept-node-children-added>children added steps</dfn> for all or some
```
annevk: So I'm pretty sure implementations have a "children changed steps" callback. Removing children does not affect this? What if someone modifies a `Text` node directly?

I'm pretty sure we should merge "children added steps" and "child text content change steps".

The main place that might require some detailed tests around this is the `style` element. As far as I can tell it's mostly not observable for `textarea`. I could write some tests around this if that would help.

annevk: I've created #800 to add "children changed steps". Once that's accepted I'll rebase this accordingly and see what's left.

annevk:
The other piece of relevant feedback here is https://github.com/whatwg/html/issues/4611. We need these deferred steps for removal too. Hurray!

domenic:
I don't believe that's true? At least, not if we want to match 2/3 browsers; for those the current spec suffices.

annevk:
That would effectively mean keeping around mutation events though and I doubt the current specification accurately accounts for that.

domenic:
Hmm, maybe that argument could convince Chrome or WebKit to add the extra complexity; I'm not sure. unload events are kind of a compat minefield though so it'd be a hard sell.

dom.bs
```diff
@@ -2389,16 +2389,13 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
-     adjust this further based on the requirements of the script element. There might be other ways
-     to define that though as Olli suggests, so leaving that out for now. -->
+algorithm is passed <var ignore>insertedNode</var> and <var ignore>deferredStepsQueue</var>, as
+indicated in the <a for=/>insert</a> algorithm below.
 
 <p><a lt="other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-children-changed-ext>children changed steps</dfn> for all or some
-<a for=/>nodes</a>. The algorithm is passed no argument and is called from <a for=/>insert</a>,
-<a for=/>remove</a>, and <a for=/>replace data</a>.
+<a for=/>nodes</a>. The algorithm is passed <var ignore>deferredStepsQueue</var> and is called from
```
annevk: Addressing https://github.com/whatwg/html/pull/4354#issuecomment-476038918 by @tkent-google in a way that aligns with Chrome would also require passing "insert"/"change"/"remove" as an argument here and acting on it accordingly. (Script would ignore anything but "insert".)

Probably would have to double check whether setting `textContent` counts as "change" or "insert".

PR #747, opened 2019-04-05
Set current node to sibling when iterating the TreeWalker. Fixes #667.

### LINKED ISSUE #667: TreeWalker's nextNode() method can result in an infinite loop because the value of node never changes if it doesn't have a child
Given the following:
```html
<doctype html>
<html>
  <head></head>
  <body></body>
</html>
```
```javascript
const treeWalker = document.createTreeWalker(document, NodeFilter.SHOW_ELEMENT, null);
treeWalker.nextNode();
```

A walk through:

Step 1) *node* is set to `document`
Step 3.1) *result* is `FILTER_ACCEPT` and *node* has a child so we enter the loop
Step 3.1.1) *node* is set to `document`'s `doctype`
Step 3.1.2) *result* is set to `FILTER_SKIP` so we continue on
Step 4.2) *sibling* is set to *temporary's* next sibling, which is the `<html>` tag
Step 4.3) *sibling* is non-null, so we break out of the loop
Step 5) *result* is set to `FILTER_SKIP`
Step 6) since *result* is not `FILTER_ACCEPT`, we return to Step 3

Since a `doctype` doesn't have children, we never enter the while loop at Step 3.1 again, which is the only place where *node* changes, leaving *node* permanently set to `document`'s `doctype`, and thus an infinite loop is born.

oozcitak:
This PR fixes #667 where a sibling node is never returned from the nextNode() method of TreeWalker, resulting in an infinite loop.

***
<a href="https://whatpr.org/dom/747.html" title="Last updated on Apr 5, 2019, 9:33 AM UTC (124d06d)">Preview</a> | <a href="https://whatpr.org/dom/747/7a8406b...124d06d.html" title="Last updated on Apr 5, 2019, 9:33 AM UTC (124d06d)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9443,7 +9443,8 @@ steps:
 
      <li><p>Set <var>sibling</var> to <var>temporary</var>'s <a for=tree>next sibling</a>.
 
-     <li><p>If <var>sibling</var> is non-null, then <a for=iteration>break</a>.
+     <li><p>If <var>sibling</var> is non-null, then set <var>node</var> to <var>sibling</var> and
+     <a for=iteration>break</a>.
 
      <li><p>Set <var>temporary</var> to <var>temporary</var>'s <a for=tree>parent</a>.
     </ol>


annevk:
Thanks, do you know by any chance if there are tests for this already?

oozcitak:
Well there is the wpt, however `nextNode` and `previousNode` show as TODO: 

https://github.com/web-platform-tests/wpt/blob/master/dom/traversal/TreeWalker.html#L12

annevk:
Ah, thanks, per https://whatwg.org/working-mode#changes we'd need to fix coverage there. And also, as I noted in #667 earlier today, there might be some other changes required here.

oozcitak:
Would you like me to change the PR to follow mozilla's implementation?

annevk:
Yeah, that seems better, but we'll need some test coverage too before we can land the change.

oozcitak:
Can you please advise on how/where to add the tests? Should I send a second PR to WPT with `nextNode` and `previousNode` tests?

annevk:
@oozcitak sorry, yeah, that'd be the way to go if you want to take that on. Much appreciated!

oozcitak:
Oops, I just read https://whatwg.org/working-mode#tests. I'll send the tests to WPT. Sorry for taking your time.

annevk:
Resolved by 868f1734717c791261e75687ba3ce12bfedfae8e. https://github.com/whatwg/dom/issues/787#issuecomment-550485387 identified existing test coverage.

Issue #645, opened 2018-05-07
Editorial: improve clarity in event dispatch model


annevk:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1449560#c5 @bzbarsky wrote:

> In particular, the fact that the "target" in the dispatch algorithm becomes the "item" in the event path structs.

I'd welcome ideas on how to improve this. I agree it's a bit of a mess, but I also don't quite see how to make things substantially better. We could rename to "item" to "current target" perhaps, but I'm not sure if that helps or hurts things.

> I _really_ hope "the last tuple in event’s path whose target is non-null" is in fact the thing that we are thinking of as the target of the event... Is it?

It's the outermost target. The innermost target (the one passed to dispatch) will be the first tuple in event's path whose target is non-null.

bzbarsky:
> I'd welcome ideas on how to improve this.

The issue isn't really what `item` is named in the event path struct.

The issue is that the fact that `target` becomes `item` allows some other thing to become `target`, so that `target` can mean different things depending on where exactly you are looking...

bzbarsky:
One option is to rename `item` to `current target` in the dispatch structs and then rename `target` to `current target` in the algorithm that constructs them, so that the names match....

And maybe call the thing that becomes `target` in the dispatch struct `targetOverride` or `target` consistently?  Less sure about this one.

It would also help a lot if https://dom.spec.whatwg.org/#concept-event-path-append described the meaning of the members of the struct.  Right now nothing seems to do that, so you have to go dig to figure out what the difference is between the `target` and `targetOverride` arguments and when one would use one vs the other.  I _think_ the `target` argument is what will become the event's `currentTarget` when dispatching to this struct and `targetOverride`, if not null, is what retargets the event at shadow dom boundaries?  But I'm not 100% sure that's always the case even after reading through this stuff for a bit and double-checking with smaug.

annevk:
Your understanding is correct and those seem like reasonable suggestions. (We also use `targetOverride` to set `event.target` to document while dispatching on the `Window` object.)

PR #750, opened 2019-04-10
Non-bubbling event should run non-capture listeners

### LINKED ISSUE #742: Composed and non bubbling event semantic dispatched in shadow tree
What is the expected behavior for a *composed* and *non-bubbling* event when dispatched inside a shadow tree?

When an event listener is attached to the host element, and a composed and non-bubbling event is dispatched within a shadow tree, the listener attached on the host element is invoked. Here is an example: https://jsfiddle.net/pmdartus/6ft9gkc4/7/

This behavior is consistent on all the browsers implementing shadow DOM.

As far as I understand the recent changes made to the dispatch event algorithm (https://github.com/whatwg/dom/pull/686), the behavior changed and now the event will only invoke the host element event handler if the event *bubbles* (step 14). Is it the new expected behavior, or did I missed something?

----

Regardless of the outcome of this question, I think it worth adding more details to the non-normative section about `composed`.

annevk:
This fixes a regression from #686.

Tests: various tests were failing because of this. https://github.com/web-platform-tests/wpt/pull/16307 aligns the remaining failing test with the new model.

Fixes #742.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1405,15 +1405,21 @@ for discussion).
     </ol>
 
    <li>
-    <p>If <var>event</var>'s {{Event/bubbles}} attribute is true, then <a for=list>for each</a>
-    <var>struct</var> in <var>event</var>'s <a for=Event>path</a>:
+    <p><a for=list>For each</a> <var>struct</var> in <var>event</var>'s <a for=Event>path</a>:
 
     <ol>
      <li><p>If <var>struct</var>'s <a for=Event/path>target</a> is non-null, then set
      <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/AT_TARGET}}.
 
-     <li><p>Otherwise, set <var>event</var>'s {{Event/eventPhase}} attribute to
-     {{Event/BUBBLING_PHASE}}.
+     <li>
+      <p>Otherwise:
+
+      <ol>
+       <li><p>If <var>event</var>'s {{Event/bubbles}} attribute is false, then
+       <a for=iteration>continue</a>.
+
+       <li><p>Set <var>event</var>'s {{Event/eventPhase}} attribute to {{Event/BUBBLING_PHASE}}.
+      </ol>
 
      <li><p><a>Invoke</a> with <var>struct</var>, <var>event</var>, "<code>bubbling</code>", and
      <var>legacyOutputDidListenersThrowFlag</var> if given.


domenic:
@pmdartus and/or @Zirro if you have time to review this it would be appreciated.

dom.bs
```diff
@@ -1405,15 +1405,21 @@ for discussion).
     </ol>
 
    <li>
-    <p>If <var>event</var>'s {{Event/bubbles}} attribute is true, then <a for=list>for each</a>
-    <var>struct</var> in <var>event</var>'s <a for=Event>path</a>:
+    <p><a for=list>For each</a> <var>struct</var> in <var>event</var>'s <a for=Event>path</a>:
 
     <ol>
      <li><p>If <var>struct</var>'s <a for=Event/path>target</a> is non-null, then set
```
rniwa: I actually don't get this check. How can target ever be null?
Don't we need to check that target is equal to targetOverride instead?
In general, steps 5.9.7-9 seem to be in the need of a bit more clarification as to what they're trying to do.

annevk: An event's path consists of a list of structs. target is only non-null for the actual target and any shadow tree hosts that the actual target might be "nested" in.

rniwa: I understand that's what this sentence assumes but I don't see how such a condition is met given what's in https://dom.spec.whatwg.org/#concept-event-dispatch

annevk: Yeah, but that argument is what becomes "item" on the struct, not "target". See https://dom.spec.whatwg.org/#concept-event-path-append. I agree this is somewhat confusing and I'd appreciate if someone could come up with better variable names as I'm still somewhat stuck.

rniwa: Oh, I see. That is indeed very confusing. Maybe we can use different variable names or something? Having item, target, & targetOverride is confusing enough.

rniwa:
Okay, with that clarification, the change looks good. That's basically [what I implemented in WebKit](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/EventDispatcher.cpp?rev=236002#L76).

annevk:
I had forgotten that I already got some suggestions in #645. I'll see about fixing that after this.

rniwa:
We can probably merge this fix as is though. We make editorial improvements in #645.

rniwa:
FWIW, I added a [test](https://svn.webkit.org/repository/webkit/trunk/LayoutTests/fast/shadow-dom/capturing-and-bubbling-event-listeners-across-shadow-trees.html) in https://trac.webkit.org/browser/webkit/trunk/LayoutTests/fast/shadow-dom/capturing-and-bubbling-event-listeners-across-shadow-trees.html?rev=236002.

Feel free to create a PR to merge it into WPT. I can make one too if you'd like.

annevk:
@rniwa if you could turn that into a PR that'd be great, thanks!

rniwa:
@annevk : Ok, posted a PR on https://github.com/web-platform-tests/wpt/pull/16358

PR #751, opened 2019-04-12
Editorial: use IDL to simplify create an element

annevk:
Per https://github.com/whatwg/html/pull/4525 a custom element constructor is required to be an HTMLElement.

Also, rather than assume another specification will make foreign namespace objects implement HTMLElement, let's instead assert that does not happen as we cannot test it.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6035,20 +6035,10 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
      <li><p>Set <var>result</var> to the result of <a>constructing</a> <var>C</var>, with no
      arguments.
 
-     <li>
-      <p>If <var>result</var> does not implement the {{HTMLElement}} interface, then <a>throw</a> a
-      <code>TypeError</code>.
+     <li><p>Assert: <var>result</var>'s <a for=Element>custom element state</a> and
+     <a for=Element>custom element definition</a> are initialized.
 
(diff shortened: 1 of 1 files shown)


domenic:
How does this work given the following?

```js
customElements.define("x-y", class extends HTMLElement {
  constructor() {
    return {}; // not a HTMLElement
  }
});
```

dom.bs
```diff
@@ -6035,20 +6035,10 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
-
-       <p>If this check passes, then <var>result</var> will already have its
-       <a for=Element>custom element state</a> and <a for=Element>custom element definition</a>
-       initialized.
-      </div>
-     </li>
+     <li><p>Assert: <var>result</var>'s <a for=Element>custom element state</a> and
+     <a for=Element>custom element definition</a> are initialized.
+
+     <li><p>Assert: <var>result</var>'s <a for=Element>namespace</a> is the <a>HTML namespace</a>.
```
domenic: I think it's still worth keeping a note why this assert is true: it's because IDL enforced that _result_ is a HTMLElement instance, and at this point in time all HTMLElement instances have the HTML namespace.

annevk: That makes it sound like it might change to me, but I've added a similar note.

PR #753, opened 2019-04-15
Editorial: rename event path's item and target

### LINKED ISSUE #645: Editorial: improve clarity in event dispatch model

annevk:
Fixes #645.

***
<a href="https://whatpr.org/dom/753.html" title="Last updated on May 17, 2019, 10:06 AM UTC (d229069)">Preview</a> | <a href="https://whatpr.org/dom/753/98564fc...d229069.html" title="Last updated on May 17, 2019, 10:06 AM UTC (d229069)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -483,8 +483,8 @@ empty list.
 {{TouchEvent}} interface and related interfaces. [[TOUCH-EVENTS]]
 
 <p>An <a>event</a> has an associated <dfn export for=Event>path</dfn>. A <a for=Event>path</a> is a
-<a for=/>list</a> of <a for=/>structs</a>. Each <a for=/>struct</a> consists of a
-<dfn for=Event/path>invocation target</dfn> (an {{EventTarget}} object), a
+<a for=/>list</a> of <a for=/>structs</a>. Each <a for=/>struct</a> consists of an
+<dfn for=Event/path>invocation target</dfn> (an {{EventTarget}} object), an
 <dfn for=Event/path>invocation-target-in-shadow-tree</dfn> (a boolean), a
 <dfn for=Event/path>shadow-adjusted target</dfn> (a <a>potential event target</a>), a
 <dfn id=event-path-relatedtarget for=Event/path>relatedTarget</dfn> (a
--- a/dom.bs
+++ b/dom.bs
@@ -484,9 +484,9 @@ empty list.
 
 <p>An <a>event</a> has an associated <dfn export for=Event>path</dfn>. A <a for=Event>path</a> is a
 <a for=/>list</a> of <a for=/>structs</a>. Each <a for=/>struct</a> consists of a
-<dfn for=Event/path>current target</dfn> (an {{EventTarget}} object), a
+<dfn for=Event/path>invocation target</dfn> (an {{EventTarget}} object), a
 <dfn for=Event/path>current-target-in-shadow-tree</dfn> (a boolean), a
-<dfn for=Event/path>target</dfn> (a <a>potential event target</a>), a
+<dfn for=Event/path>shadow-adjusted target</dfn> (a <a>potential event target</a>), a
 <dfn id=event-path-relatedtarget for=Event/path>relatedTarget</dfn> (a
 <a>potential event target</a>), a <dfn for=Event/path>touch target list</dfn> (a <a for=/>list</a>
 of <a>potential event targets</a>), a <dfn for=Event/path>root-of-closed-tree</dfn> (a boolean), and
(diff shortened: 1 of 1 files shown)


annevk:
@bzbarsky @rniwa how does this look?

I could also rename _parent_ to _currentTarget_ or some such, as I think was suggested in #645, but it's not totally clear to me that adds clarity. I might be in a bit too deep though.

dom.bs
```diff
@@ -483,10 +483,10 @@ empty list.
...
 
 <p>An <a>event</a> has an associated <dfn export for=Event>path</dfn>. A <a for=Event>path</a> is a
-<a for=/>list</a> of <a for=/>structs</a>. Each <a for=/>struct</a> consists of an
-<dfn for=Event/path>item</dfn> (an {{EventTarget}} object), an
-<dfn for=Event/path>item-in-shadow-tree</dfn> (a boolean), a <dfn for=Event/path>target</dfn> (a
-<a>potential event target</a>), a
+<a for=/>list</a> of <a for=/>structs</a>. Each <a for=/>struct</a> consists of a
+<dfn for=Event/path>current target</dfn> (an {{EventTarget}} object), a
+<dfn for=Event/path>current-target-in-shadow-tree</dfn> (a boolean), a
+<dfn for=Event/path>target</dfn> (a <a>potential event target</a>), a
```
rniwa: Maybe we can call this one *dispatch target* to further distinguish itself from other kinds of targets?

dom.bs
```diff
@@ -613,7 +614,7 @@ steps:
    <li><p>If <var>path</var>[<var>index</var>]'s <a for=Event/path>root-of-closed-tree</a> is true,
    then increase <var>currentTargetHiddenSubtreeLevel</var> by 1.
 
-   <li><p>If <var>path</var>[<var>index</var>]'s <a for=Event/path>item</a> is
+   <li><p>If <var>path</var>[<var>index</var>]'s <a for=Event/path>current target</a> is
```
rniwa: *current target* being different form *currentTarget* is super confusing here...
Perhaps we should call *current target* in the struct *invocation target* instead??

dom.bs
annevk: Okay, so this is either the target passed into the dispatch algorithm or the target as adjusted for shadow tree encapsulation (some host). Does originating target reflect that well?

rniwa: Hm... maybe shadow-adjusted originating target? I think the key here is convey the semantics that the event looks as though it originated from that node in scripts.

Or we could call it *shadow adjusted target* or *visible originating target*. Or maybe even *embarkment target* or *source target*?

annevk:
Updated the PR to use invocation and shadow-adjusted target as terms.

annevk:
@smaug---- you might want to take a look as well.

dom.bs
```diff
@@ -483,10 +483,10 @@ empty list.
 {{TouchEvent}} interface and related interfaces. [[TOUCH-EVENTS]]
 
 <p>An <a>event</a> has an associated <dfn export for=Event>path</dfn>. A <a for=Event>path</a> is a
-<a for=/>list</a> of <a for=/>structs</a>. Each <a for=/>struct</a> consists of an
-<dfn for=Event/path>item</dfn> (an {{EventTarget}} object), an
-<dfn for=Event/path>item-in-shadow-tree</dfn> (a boolean), a <dfn for=Event/path>target</dfn> (a
-<a>potential event target</a>), a
+<a for=/>list</a> of <a for=/>structs</a>. Each <a for=/>struct</a> consists of a
+<dfn for=Event/path>invocation target</dfn> (an {{EventTarget}} object), a
+<dfn for=Event/path>current-target-in-shadow-tree</dfn> (a boolean), a
```
smaug----: current target has very specific meaning, so why do we use it here too?
Shouldn't it be invocation-target-in-shadow-tree?

annevk: You're right, thanks!

annevk:
@rniwa @smaug---- I think it's all good now. If this event terminology looks good hopefully it can stick around for a while.

rniwa:
The new terminology reads a lot better to me.

PR #754, opened 2019-04-15
Run adopt as part of insert

### LINKED ISSUE #744: Don't allow adopting DocumentFragment of a template element
Right now, we allow [adopting](https://dom.spec.whatwg.org/#concept-node-adopt) a [`DocumentFragment`](https://dom.spec.whatwg.org/#interface-documentfragment) of a [template element](https://html.spec.whatwg.org/multipage/scripting.html#the-template-element) because `DocumentFragment` never have a parent.

The subtree of such a `DocumentFragment` is still associated with the template element yet its owner document is of the adopted document. As far as I know, this is only case in which such an oddity arises because `ShadowRoot` is forbidden from being adopted into another tree, and every other node type would have a parent as long as it's part of a document and therefore will be removed before the adoption.

I don't think we should allow this. https://github.com/web-platform-tests/wpt/blob/master/html/semantics/scripting-1/the-template-element/template-element/template-content-hierarcy.html explicitly tests this.

annevk:
And do not allow non-null host DocumentFragment nodes to be adopted.

This means that a DocumentFragment is no longer implicitly adopted and a DocumentFragment with a non-null host cannot be put in a "weird" state.

Tests: https://github.com/web-platform-tests/wpt/pull/16348.

Fixes #744.

***
<a href="https://whatpr.org/dom/754.html" title="Last updated on Dec 6, 2019, 1:48 PM UTC (6aec9ae)">Preview</a> | <a href="https://whatpr.org/dom/754/37d8475...6aec9ae.html" title="Last updated on Dec 6, 2019, 1:48 PM UTC (6aec9ae)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2379,8 +2379,6 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
  <li><p>If <var>referenceChild</var> is <var>node</var>, then set <var>referenceChild</var> to
  <var>node</var>'s <a for=tree>next sibling</a>.
 
- <li><p><a>Adopt</a> <var>node</var> into <var>parent</var>'s <a for=Node>node document</a>.
-
  <li><p><a for=/>Insert</a> <var>node</var> into <var>parent</var> before <var>referenceChild</var>.
 
  <li><p>Return <var>node</var>.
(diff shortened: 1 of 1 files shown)


annevk:
@rniwa I think this addresses your concerns.

I discussed this with @smaug---- and I think he agrees with this (note that Firefox did do implicit adoption of DocumentFragment nodes).

One thing we were unsure of is whether it's worth trying to align with `ShadowRoot` for `adoptNode()` by either trying out throwing behavior or making both not throw.

@tkent-google opinions on behalf of Chrome?

tkent-google:
> @tkent-google opinions on behalf of Chrome?

Looks a reasonable change.

annevk:
@rniwa heya, could you please take a look at this?

rniwa:
Looks reasonable to me.

annevk:
Bugs:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1602200
* https://bugs.webkit.org/show_bug.cgi?id=204980
* https://bugs.chromium.org/p/chromium/issues/detail?id=1031811

PR #755, opened 2019-04-16
Add ParentNode.prototype.replaceContents(nodes)

### LINKED ISSUE #478: Proposal: ParentNode.replaceAll() / ParentNode.replaceChildren()
The **ParentNode.replaceAll()** or **ParentNode.replaceChildren()** method would remove all child nodes from the ParentNode while allowing the insertion of a new set of nodes. The best name is yet settled upon.

A “replace all” concept [already exists](https://dom.spec.whatwg.org/#concept-node-replace-all) in the spec.

### Syntax

```
ParentNode.replaceAll();
ParentNode.replaceChildren();
```

### Example:

1. In a form, changing a **Country** `<select>` causes options to change in a corresponding **Region** `<select>`:

```js
countrySelectElement.addEventListener('change', () => {
  regionSelectElement.replaceAll(...optionElementsById[countrySelectElement.value]);
});
```

2. Any time an element is to be emptied.

```js
elementToBeEmptied.replaceChildren();
```

### Rationale:

Like `append()`, `prepend()`, `before()`, `after()`, and `replaceWith()`, we’re allowing developers to do something useful without a loop, e.g:

```js
// remove all child nodes from the ParentNode
while (parentNode.lastChild) {
  parentNode.removeChild(parentNode.lastChild);
}

// insert a new set of Node objects or DOMString objects
parentNode.append(...addedNodes);
```

### Fallback Strategies

There are 2 common strategies to empty a ParentNode before inserting a new set of nodes (using [`ParentNode.append`](https://dom.spec.whatwg.org/#dom-parentnode-append)).

The 1st strategy is to erase the innerHTML of a ParentNode. This is less meaningful and leads one to question the underlying implementation.

```js
ParentNode.innerHTML = ''
```

The 2nd strategy is to loop over and remove each child of the ParentNode. This is verbosE and leads one to wish for the same loop-free “sugar” provided to `append()`, `prepend()`, `before()`, `after()`, and `replaceWith()`.

```
while (parentNode.lastChild) {
  parentNode.removeChild(parentNode.lastChild);
}
```

---

## Previous proposal (replaceAll)

> A previous proposal named this method `replaceAll`, a name that mirrors a section in the spec. Recent feedback has been that `replaceChildren` would be a more accurate and less confusing name. Therefore, both are represented in this proposal.

---

## Original Proposal (empty)

> The original proposal’s functionality is fully captured in the updated proposal. The updated proposal also reflects an existing concept in the spec.

The **ParentNode.empty()** method would remove all child nodes from the parent node.

### Syntax

```
ParentNode.empty();
```

### Example:

In a filtering component, a change to one `<select>` swaps out the `<option>` elements available to a second `<select>`:

```js
selectEl1.addEventListener('change', () => {
  selectEl2.empty();

  selectEl2.append(...optionElsByValue[selectEl1.value]);
});
```

### Rationale:

Like `append()`, `prepend()`, `before()`, `after()`, and `replaceWith()`, we’re allowing developers to do something useful without a loop, e.g:

```js
while (parentNode.lastChild) {
  parentNode.removeChild(parentNode.lastChild);
}
```

jonathantneal:
The **ParentNode.replaceContents()** method would remove all child nodes from the ParentNode while allowing the insertion of a new set of nodes.

Resolves #478

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2839,6 +2839,7 @@ interface mixin ParentNode {
 
   [CEReactions, Unscopable] void prepend((Node or DOMString)... nodes);
   [CEReactions, Unscopable] void append((Node or DOMString)... nodes);
+  [CEReactions, Unscopable] void replaceContents((Node or DOMString)... nodes);
 
   Element? querySelector(DOMString selectors);
   [NewObject] NodeList querySelectorAll(DOMString selectors);
@@ -2870,6 +2871,15 @@ Element includes ParentNode;
   <!-- "NotFoundError" is impossible -->
 
  <dt><code><var>node</var> . <a method for=ParentNode lt="append()">append</a>(<var>nodes</var>)</code>
+ <dd>
+  <p>Inserts <var>nodes</var> after the <a>last child</a> of <var>node</var>, while replacing
+  strings in <var>nodes</var> with equivalent {{Text}} <a for=/>nodes</a>.
+
+  <p><a>Throws</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}} if the constraints of
+  the <a>node tree</a> are violated.
+  <!-- "NotFoundError" is impossible -->
+
+ <dt><code><var>node</var> . <a method for=ParentNode lt="replaceContents()">replaceContents</a>(<var>nodes</var>)</code>
  <dd>
   <p>Inserts <var>nodes</var> after the <a>last child</a> of <var>node</var>, while replacing
   strings in <var>nodes</var> with equivalent {{Text}} <a for=/>nodes</a>.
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -2839,6 +2839,7 @@ interface mixin ParentNode {
 
   [CEReactions, Unscopable] void prepend((Node or DOMString)... nodes);
   [CEReactions, Unscopable] void append((Node or DOMString)... nodes);
+  [CEReactions, Unscopable] void replaceContents((Node or DOMString)... nodes);
```
annevk: I think per the discussion we ended up with `replaceChildren()` as a name, as "contents" doesn't always refer to nodes. Did I miss something?

dom.bs
```diff
@@ -2878,6 +2879,15 @@ Element includes ParentNode;
   the <a>node tree</a> are violated.
   <!-- "NotFoundError" is impossible -->
 
+ <dt><code><var>node</var> . <a method for=ParentNode lt="replaceContents()">replaceContents</a>(<var>nodes</var>)</code>
+ <dd>
+  <p>Inserts <var>nodes</var> after the <a>last child</a> of <var>node</var>, while replacing
```
annevk: This needs rewording.

dom.bs
```diff
@@ -2925,6 +2935,45 @@ must run these steps:
...
 </ol>
 
+<p>The <dfn method for=ParentNode><code>replaceContents(<var>nodes</var>)</code></dfn> method, when invoked,
+must run these steps:
+
+<ol>
+  <li><a for=/>Remove</a> all
+  <var>parent</var>'s <a>children</a>, in
+  <a>tree order</a>, with the
+  <i>suppress observers flag</i> set.
```
annevk: I think we should strive to use the https://dom.spec.whatwg.org/#concept-node-replace-all primitive here, which would take care of most of the work, preceded by a call to https://dom.spec.whatwg.org/#converting-nodes-into-a-node.

And I guess we need to run https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity as well.

jonathantneal:
I intend to make corrections next week. Thanks for the great feedback, and also for getting it back to me so quickly.

PR #760, opened 2019-05-13
Add "disable shadow" flag check to Element.attachShadow()

tkent-google:
This fixes a part of https://github.com/w3c/webcomponents/issues/758

This and https://github.com/whatwg/html/pull/4324 should be merged at the same time.

Test: https://github.com/web-platform-tests/wpt/pull/16795

***
<a href="https://whatpr.org/dom/760.html" title="Last updated on May 16, 2019, 4:14 PM UTC (2982935)">Preview</a> | <a href="https://whatpr.org/dom/760/bcdab96...2982935.html" title="Last updated on May 16, 2019, 4:14 PM UTC (2982935)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6767,6 +6767,23 @@ invoked, must run these steps:
  "<code>section</code>", or
  "<code>span</code>", then <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
 
+ <li>
+  <p>If <a>context object</a>'s <a for=Element>local name</a> is a
+  <a>valid custom element name</a>, or <a>context object</a>'s
+  <a for=Element><code>is</code> value</a> is not null, then:
+
+  <ol>
+   <li><p>Let <var>definition</var> be the result of
+   <a lt="look up a custom element definition">looking up a custom element definition</a> given
+   <a>context object</a>'s <a for=Node>node document</a>, its <a for=Element>namespace</a>, its
+   <a for=Element>local name</a>, and its <a for=Element><code>is</code> value</a>.
+
+   <li><p>If <var>definition</var> is not null and <var>definition</var>'s
+   <a for="custom element definition">disable shadow</a> is true, then <a>throw</a> a
+   "{{NotSupportedError!!exception}}" {{DOMException}}.
+  </ol>
+ </li>
+
  <li><p>If <a>context object</a> is a <a for=Element>shadow host</a>, then <a>throw</a> an
  "{{NotSupportedError!!exception}}" {{DOMException}}.
 
@@ -10034,6 +10051,7 @@ J. S. Choi,
 Jungkee Song,
 Justin Summerlin,
 呂康豪 (Kang-Hao Lu),
+田村健人 (Kent TAMURA),
 Kevin Sweeney,
 Kirill Topolyan,
 Koji Ishii,


dom.bs
```diff
@@ -6770,6 +6770,22 @@ invoked, must run these steps:
...
+  <a>valid custom element name</a>, then:</p>
+
+  <ol>
+   <li><p>Let <var>definition</var> be the result of
+   <a lt="look up a custom element definition">looking up a custom element definition</a> given
+   <a>context object</a>'s <a for=Node>node document</a>, its <a for=Element>namespace</a>, its
+   <a for=Element>local name</a>, and null as <a><code>is</code> value</a>.</p></li>
+
+   <li><p>If <var>definition</var> is not null and <var>definition</var>'s
+   <a lt="concept-custom-element-definition-disable-shadow">disable shadow</a> is true,
```
tkent-google: I'm not sure what's the correct way to make a link to "disable shadow" flag.

annevk: Once HTML has exported this term it should work. I wonder if we should throw InvalidStateError instead as this would allow distinguishing between it being a shadow host and it being disabled, no?

tkent-google: I thought that ```InvalidStateError``` was not suitable because the context object can't be 'valid' state again.  But I found the context object was valid before calling custom element constructor.  So ```InvalidStateError``` seems to make sense.

```attachInternals()``` has the same issue.

Comment (annevk) [reply dom.bs]: Ah okay, let's change both then.

Comment (tkent-google) [reply dom.bs]: BTW, the previous step checks element types, and this step is also a kind of element type check.  NotSupportedError is consistent with the previous step.

Comment (annevk) [reply dom.bs]: But how would we address this issue then? Or you don't think it's an issue? I suppose we could also change the exception of it already having a shadow tree further down.

Comment (tkent-google) [reply dom.bs]: I think either of NotSupportedError or InvalidStateError is ok.  I prefer NotSupportedError slightly.

Comment (annevk) [reply dom.bs]: So we use NotSupportedError when there's already a shadow tree? I'd be okay with such a change.

Comment (tkent-google) [reply dom.bs]: > So we use NotSupportedError when there's already a shadow tree? 

I don't think so.  It's definitely an issue of context object's state.  I think InvalidStateError is appropriate.

Comment (annevk) [reply dom.bs]: You appear to be arguing for exposing the distinction between a shadow root being forbidden and one being used. I care about removing that distinction. I don't really care about types of exceptions as it's not clear to me we've been consistent enough for the differences between them to be meaningful. What do you propose?

Comment (tkent-google) [reply dom.bs]: ok, I understand your intention.
Then I think NotSupportedError is ok for all of four DOMExceptions thrown by attachShadow(). IMO InvalidStateError is a specific subset of NotSupportedError.

Anyway, this PR doesn't change the DOMException type of the already-attached case.  It's not in the scope of this PR.

Comment (annevk) [reply dom.bs]: If you want to do them separately I think we'll need to first change the existing exceptions then and then land this PR. Otherwise landing this would result in a broken state.

Comment (domenic) [reply dom.bs]: I posted https://github.com/whatwg/dom/pull/761 and will now work on tests and browser bugs.

**Comment by domenic** [comment]
Doesn't this fail to respect the disable-shadow flag for customized built-in elements on the safelist? E.g. `<h4 is="special-h4">` with

```js
customElements.define("special-h4", class extends HTMLHeadingElement {
  disabledFeatures = ['shadow'];
}, { extends: 'h4' });
```

I think we need to pass the node's `is value`. I will add a commit to do that, but we should also update the tests.

dom.bs
```diff
@@ -6767,6 +6767,22 @@ invoked, must run these steps:
  "<code>section</code>", or
  "<code>span</code>", then <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
 
+ <li>
+  <p>If <a>context object</a>'s <a for=Element>local name</a> is a
```
tkent-google: Need to change this to something like:

    If context object's local name is a valid custom element name, or
    context object's is value is not null, then:

tkent-google: Fixed in a new commit.

annevk:
Assuming this is rebased on top of / lands after #761 this seems fine to me.

emilio:
This is a bit racy isn't it? As in, if you attach the shadow before the custom element definition then it works, otherwise it doesn't...

annevk:
Yeah, not sure what can be done about that though. I suppose we could let "upgrade an element" fail, because the "preconditions" were broken.

annevk:
@tkent-google @domenic thoughts on @emilio's comment? If we want to make "upgrade an element" fail we should probably decide sooner rather than later.

domenic:
I don't have strong feelings on disable-shadow in general.

tkent-google:
```attachInternals()``` has the same issue.

> If we want to make "upgrade an element" fail

It sounds a reasonable fix.

annevk:
Follow-up at https://github.com/whatwg/html/pull/4673.

PR #761, opened 2019-05-15
Use "NotSupportedError" when trying to attachShadow() twice

domenic:
Discussed in #760. This makes the various cases less distinguishable,
which is desirable. With this change, all three cases behave the same
between:

- A custom element with disabled-shadow
- A custom element with a shadow tree
- A built-in element on the blocklist

Tests: https://github.com/web-platform-tests/wpt/pull/16853

***
<a href="https://whatpr.org/dom/761.html" title="Last updated on May 15, 2019, 7:09 PM UTC (7bcdca6)">Preview</a> | <a href="https://whatpr.org/dom/761/e4e0ea8...7bcdca6.html" title="Last updated on May 15, 2019, 7:09 PM UTC (7bcdca6)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6768,7 +6768,7 @@ invoked, must run these steps:
  "<code>span</code>", then <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
 
  <li><p>If <a>context object</a> is a <a for=Element>shadow host</a>, then <a>throw</a> an
- "{{InvalidStateError!!exception}}" {{DOMException}}.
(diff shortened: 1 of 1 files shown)


annevk:
I think I can say that Mozilla is fine with this change.

domenic:
Bugs:

- https://bugs.chromium.org/p/chromium/issues/detail?id=963920
- https://bugs.webkit.org/show_bug.cgi?id=197948
- https://bugzilla.mozilla.org/show_bug.cgi?id=1552198

PR #762, opened 2019-05-17
Editorial: use Infra list syntax in mutations

annevk:
As pointed out at https://github.com/whatwg/dom/pull/695#pullrequestreview-233547961 it would be nice if this were more consistent.

***
<a href="https://whatpr.org/dom/762.html" title="Last updated on May 20, 2019, 2:38 PM UTC (09c786e)">Preview</a> | <a href="https://whatpr.org/dom/762/daa525d...09c786e.html" title="Last updated on May 20, 2019, 2:38 PM UTC (09c786e)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2427,10 +2427,8 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
    its <a for=range>end offset</a> by <var>count</var>.
   </ol>
 
- <li>Let <var>nodes</var> be <var>node</var>'s
- <a>children</a> if <var>node</var> is
- a {{DocumentFragment}} <a>node</a>, and a
- list containing solely <var>node</var> otherwise.
+ <li><p>Let <var>nodes</var> be <var>node</var>'s <a>children</a>, if <var>node</var> is a
+ {{DocumentFragment}} <a>node</a>, « <var>node</var> » otherwise.
 
  <li>If <var>node</var> is a {{DocumentFragment}}
  <a>node</a>,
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -2637,11 +2635,12 @@ To <dfn export for=Node id=concept-node-replace-all>replace all</dfn> with a
...
- null, <var>node</var>'s <a>children</a> if
- <var>node</var> is a {{DocumentFragment}}
- <a>node</a>, and a list containing <var>node</var>
- otherwise.
+ <li><p>Let <var>addedNodes</var> be the empty list.
+
+ <li><p>If <var>node</var> is {{DocumentFragment}} <a>node</a>, then set <var>addedNodes</var> to
+ <var>node</var>'s <a>children</a>.
+
+ <li><p>Otherwise, if <var>node</var> is non-null, set <var>addedNodes</var> to « <var>node</var> ».
```
annevk: @domenic this is the only change that felt a little weird here due to it getting so much longer, but it's probably better?

dom.bs
```diff
@@ -2427,10 +2427,8 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
    its <a for=range>end offset</a> by <var>count</var>.
   </ol>
 
- <li>Let <var>nodes</var> be <var>node</var>'s
- <a>children</a> if <var>node</var> is
- a {{DocumentFragment}} <a>node</a>, and a
- list containing solely <var>node</var> otherwise.
+ <li><p>Let <var>nodes</var> be <var>node</var>'s <a>children</a>, if <var>node</var> is a
+ {{DocumentFragment}} <a>node</a>, « <var>node</var> » otherwise.
```
domenic: This reads really wrong; I'm not happy with where we've landed if this is what happens. Commas just don't work that way. Semicolon could work, or using a conjunction (either "and" or "or").

dom.bs
domenic: Yeah, this matches the code better. Nested ternaries are not great in either English or code, IMO. You could also use dl class="switch".

dom.bs
annevk: Same if otherwise comes first? I see now I put the wrong way around. I guess I can do some follow-up on Monday in that original thread on this.

domenic: Ah I see. If otherwise comes first it isn't as viscerally wrong, but I still have a slight preference for a conjunction or semicolon.

PR #767, opened 2019-06-17
Review Draft Publication: June 2019

annevk:
A Review Draft for this Workstream will be published shortly, by merging this pull request.

Under the [WHATWG IPR Policy](https://whatwg.org/ipr-policy), Participants may, within 45 days after publication of a Review Draft, exclude certain Essential Patent Claims from the Review Draft Licensing Obligations. See the [IPR Policy](https://whatwg.org/ipr-policy) for details.

### DIFF
--- a/review-drafts/2019-06.bs
+++ b/review-drafts/2019-06.bs
@@ -1,0 +1,10147 @@
+<pre class='metadata'>
+Group: WHATWG
+Date: 2019-06-17
+H1: DOM
+Shortname: dom
+Text Macro: TWITTER thedomstandard
+Abstract: DOM defines a platform-neutral model for events, aborting activities, and node trees.
+Translation: ja https://triple-underscore.github.io/DOM4-ja.html
+Ignored Terms: EmptyString, Array, Document
+Indent: 1
+</pre>
+
+<pre class=anchors>
+urlPrefix: https://www.w3.org/TR/xml/#NT-
+ type: type
+  text: Name; url: Name
+  text: Char; url: Char
+  text: PubidChar; url: PubidChar
+urlPrefix: https://www.w3.org/TR/xml-names/#NT-
+ type: type
+  text: QName; url: QName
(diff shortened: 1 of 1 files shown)


domenic:
https://dom.spec.whatwg.org/review-drafts/2019-06/

foolip:
This URL will be updated in-place, so since we don't have the build output in version control I've made sure the Wayback Machine has a copy:
https://web.archive.org/web/20200302174446/https://dom.spec.whatwg.org/review-drafts/2019-06/

It ([add `id_` to URL](https://web.archive.org/web/20200302174446id_/https://dom.spec.whatwg.org/review-drafts/2019-06/)) currently matches the live URL byte-for-byte.

foolip:
https://dom.spec.whatwg.org/review-drafts/2019-06/ has been updated in-place now, thanks @sideshowbarker!

domenic:
Hmm. The logo ordering and fonts don't match the agreement.

annevk:
The agreement doesn't spell out an ordering as far as I can tell. The sample has them (WHATWG DOM, W3C) and this publication has them (W3C, WHATWG DOM), but I think either is okay per the agreement.

domenic:
True, I meant the sample, which is linked to as part of the agreement.

sideshowbarker:
The logo ordering was simply due to the way I appended the W3C logo in from the Python script I wrote. I think that’s just how it ended up when I did the first pass, and it seemed to look fine — and it had the effect of not moving the DOM logo but instead just leaving in the same position relative to the margins — so I just kept it like that in the https://domspec.herokuapp.com/ output that I gave to the W3C as the basis for the CR transition request, and there were no request from the W3C side about moving it. So it ended up just staying there.

But anyway in hindsight I should’ve somewhere noted that as a difference from the MoU sample.

Going forward, for the HTML CR draft, I can ensure that the logos follow the order in the MoU sample. Or else if we actually prefer the logos in the same order was what we ended up with for DOM, we can make them follow that order for HTML, too.

PR #771, opened 2019-07-01
Align with IDL dictionary changes

annevk:
See https://github.com/heycam/webidl/pull/750.

***
<a href="https://whatpr.org/dom/771.html" title="Last updated on Aug 8, 2019, 3:50 PM UTC (810f321)">Preview</a> | <a href="https://whatpr.org/dom/771/b5eac4d...810f321.html" title="Last updated on Aug 8, 2019, 3:50 PM UTC (810f321)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5879,7 +5879,7 @@ interface Element : Node {
   [CEReactions] Attr? setAttributeNodeNS(Attr attr);
   [CEReactions] Attr removeAttributeNode(Attr attr);
 
-  ShadowRoot attachShadow(ShadowRootInit init = {});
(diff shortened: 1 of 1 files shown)


bzbarsky:
@annevk The IDL here was correct before.  The dictionary has a required member, so it's not required to be optional per the IDL spec (and in fact not passing it would always throw, so there's no reason to pretend like that's a thing someone would ever try to do).

bzbarsky:
Er, that was specifically about attachShadow.

annevk:
Yeah, my bad. I'll revert that part.

PR #774, opened 2019-07-14
Expand title of spec

SamB:
Honestly, the reason for this patch is that "DOM" looks *really* short in MDN's spec tables, especially on e.g. <https://developer.mozilla.org/en-US/docs/Web/API/DOMError>.

(Yes, I realize that changing the title tag in the spec won't directly alter what's shown on MDN, but I assume they prefer that the names in their JSON file to be *official* names, and this seems like the best way to get one of those.)

***
<a href="https://whatpr.org/dom/774.html" title="Last updated on Jul 14, 2019, 8:35 PM UTC (7c6696a)">Preview</a> | <a href="https://whatpr.org/dom/774/b5eac4d...7c6696a.html" title="Last updated on Jul 14, 2019, 8:35 PM UTC (7c6696a)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1,6 +1,6 @@
 <pre class='metadata'>
 Group: WHATWG
-H1: DOM
+H1: Document Object Model (DOM)
 Shortname: dom
 Text Macro: TWITTER thedomstandard
 Abstract: DOM defines a platform-neutral model for events, aborting activities, and node trees.


annevk:
HTML isn't that much longer. Perhaps MDN should have special styling for shorter titles. (They could also expand the hit region a bit to be the entire cell.)

Issue #590, opened 2018-03-12
Introduce a StaticRange constructor


annevk:
Somehow `StaticRange` got added to browsers without the TAG noting the lack of a constructor.  #589 adds `StaticRange` to the DOM Standard.

The proposal by @garykac in https://w3c.github.io/staticrange/index.html#interface-staticrange seems reasonable. You require a dictionary to be passed specifying all the bits.

One question is whether those bits should be validated as https://github.com/w3c/staticrange/issues/13 suggests or whether we should allow any inputs and just pass them through. After all, you need to validate a `StaticRange` object whenever you use it, so why do it here as well? (Do we have any APIs yet that take `StaticRange` objects?)

It's not entirely clear to me what the tradeoffs are here.

cc @rniwa @johanneswilm @bzbarsky @dstorey

annevk:
(As part of adding a constructor we should also expand the introduction of these objects a bit and mention the relative high cost of `Range` objects.)

bzbarsky:
I dunno that there's any point to doing validation here.  Anything working with StaticRange needs to handle invalid ones somehow anyway.

Apart from that, the constructor generally makes sense to me, assuming we're doing StaticRange at all.  Are we?  Has anyone other than WebKit added it?  Where is WebKit using it?

johanneswilm:
Chromium has added it as well. It's used with Input Events.

rniwa:
WebKit uses it in input events.

annevk:
Right, https://w3c.github.io/input-events/index.html has a method that can return them (not defined in detail though).

It seems like we should restrict the boundary node from being a doctype and possibly from being an attribute as well, depending on #607.

sanketj:
Hi all! I work on the Microsoft Edge team and we were interesting in using static ranges as part of our Highlight API proposal. In case you’re unfamiliar with the highlight API, it provides a way for web developers to style the text inside arbitrary ranges within the DOM, which can be useful for editing applications that are trying to implement their own selection, find-on-page, spellchecking, etc. If you would like more details, please refer to this explainer: https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/highlight/explainer.md. We believe the highlight API is a great use case for StaticRange; however, we would need StaticRanges to be constructible, so that authors can create these range objects and add them to one or more HighlightRangeGroups (see examples in the explainer).

[...]

PR #778, opened 2019-08-22
Add StaticRange constructor

sanketj:
Adding a constructor to the StaticRange interface. The shape of the constructor is based on how it is defined here: https://w3c.github.io/staticrange/index.html#interface-staticrange, and the construction algorithm is written based on the discussions here: #590 .

Also adding a section mentioning the high cost of maintaining ranges, and that static ranges are not subject to this cost because they do not update when the DOM tree mutates.

Corresponding WPT tests PR: https://github.com/web-platform-tests/wpt/pull/18619

***
<a href="https://whatpr.org/dom/778.html" title="Last updated on Aug 26, 2019, 12:27 PM UTC (17f27b2)">Preview</a> | <a href="https://whatpr.org/dom/778/a257369...17f27b2.html" title="Last updated on Aug 26, 2019, 12:27 PM UTC (17f27b2)">Diff</a>


rniwa:
I've posted a WebKit patch on https://bugs.webkit.org/show_bug.cgi?id=201055.

annevk:
Thanks! I wonder if we should make it a bit more friendly. Default `startOffset` to 0 (this can be done in IDL), default `endContainer` to `startContainer`, and default `endOffset` to `endContainer`'s child nodes length (perhaps only if `endContainer` is `startContainer`).

annevk:
Also, please add your name to the Acknowledgments section.

rniwa:
>  I wonder if we should make it a bit more friendly. Default startOffset to 0 (this can be done in IDL), default endContainer to startContainer, and default endOffset to endContainer's child nodes length (perhaps only if endContainer is startContainer).

We can make those changes too but I think this is a good first cut. I'll land the WebKit change anyway since that seems like more of an enhancement.

sanketj:
> Also, please add your name to the Acknowledgments section.

Thanks, added! :)

sanketj:
> > I wonder if we should make it a bit more friendly. Default startOffset to 0 (this can be done in IDL), default endContainer to startContainer, and default endOffset to endContainer's child nodes length (perhaps only if endContainer is startContainer).
> 
> We can make those changes too but I think this is a good first cut. I'll land the WebKit change anyway since that seems like more of an enhancement.

I tend to agree. Having some default values seems useful, but it might be better to wait and observe some patterns in StaticRange constructor usage before deciding on what those defaults should be.

annevk:
Okay, that approach is fine with me. I pushed a number of editorial nits I'd appreciate your review on.

annevk:
Thanks again @sanketj, please do return with more PRs 😊

sanketj:
> Thanks again @sanketj, please do return with more PRs 😊

Thanks @annevk for the guidance and merging the PRs! Thanks @rniwa for the review! And yes, hopefully there will be more in the future. :)

foolip:
https://dom.spec.whatwg.org/ wasn't updated after this was merged, see https://travis-ci.org/whatwg/dom/builds/578197397. That's because "a new promise" is ambiguous, which I saw another issue for but can't find now.

Should we work around this so we can get the live spec updated?

foolip:
Hmm, fixed by https://github.com/w3ctag/promises-guide/pull/61, now that just has to propagate through... Shepherd?

foolip:
Looks like it was https://github.com/tabatkins/bikeshed-data that needed an update, see https://github.com/tabatkins/bikeshed-data/issues/3. I pushed an update, then `bikeshed update` + `make local` works for DOM. I'll restart the Travis build to see if it'll deploy now.

annevk:
Thanks, all good now.

Issue #780, opened 2019-08-29
Update to constructor operations


Ms2ger:
See https://github.com/heycam/webidl/issues/778.

PR #782, opened 2019-09-07
Use Web IDL constructor operations

foolip:
Fixes https://github.com/whatwg/dom/issues/780.


foolip:
Hrm, the `constructor` keyword in the IDL snippets are no longer linked to the constructor definitions as the `Constructor` extended attributes were.

foolip:
Alright, this works now.

dom.bs
```diff
@@ -993,9 +995,9 @@ for historical reasons.
 <h3 id=interface-eventtarget>Interface {{EventTarget}}</h3>
 
 <pre class=idl>
-[Constructor,
- Exposed=(Window,Worker,AudioWorklet)]
+[Exposed=(Window,Worker,AudioWorklet)]
 interface EventTarget {
+  constructor();
```
annevk: Newline?

dom.bs
```diff
@@ -3348,9 +3351,9 @@ they do not get lost when {{MutationObserverInit/subtree}} is set to true on <a
 <h4 id=interface-mutationobserver>Interface {{MutationObserver}}</h4>
 
 <pre class="idl">
-[Constructor(MutationCallback callback),
- Exposed=Window]
+[Exposed=Window]
 interface MutationObserver {
+  constructor(MutationCallback callback);
```
annevk: Newline?

dom.bs
```diff
@@ -7318,9 +7322,9 @@ method, when invoked, must <a>replace data</a> with node <a>context object</a>,
 <h3 id=interface-text>Interface {{Text}}</h3>
 
 <pre class=idl>
-[Constructor(optional DOMString data = ""),
- Exposed=Window]
+[Exposed=Window]
 interface Text : CharacterData {
+  constructor(optional DOMString data = "");
```
annevk: Newline?

annevk:
"Align with IDL constructor changes" would be consistent with the prior commit message so I'll use that when squash merging.

dom.bs
foolip: I tried to be consistent with what was previously done within each interface. Where methods and attributes were already separated by a blank line I added a newline, and where there were no blank lines I didn't add one.

Is there a rule I can apply to all of the affected interfaces?

annevk: I was thinking we'd add a newline after each constructor (unless there's nothing following it). The other newlines don't have a good rule I think. They're mostly added for some subjective measure of readability.

PR #785, opened 2019-09-20
Add onabort definition and add onslotchange

domenic:
***
<a href="https://whatpr.org/dom/785.html" title="Last updated on Sep 23, 2019, 10:09 AM UTC (07a9172)">Preview</a> | <a href="https://whatpr.org/dom/785/dea5d92...07a9172.html" title="Last updated on Sep 23, 2019, 10:09 AM UTC (07a9172)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1835,6 +1835,11 @@ service worker.
 <p>The <dfn attribute for=AbortSignal>aborted</dfn> attribute's getter, when invoked, must return
 true if the <a>context object</a>'s [=AbortSignal/aborted flag=] is set, and false otherwise.
 
+<p>The <dfn attribute for=AbortSignal><code>onabort</code></dfn> attribute is an
+<a>event handler IDL attribute</a> for the <dfn export for=AbortSignal><code>onabort</code></dfn>
+<a>event handler</a>, whose <a>event handler event type</a> is
+<dfn event for=AbortSignal><code>abort</code></dfn>.
+
 <p class=note>Changes to an {{AbortSignal}} object represent the wishes of the corresponding
 {{AbortController}} object, but an API observing the {{AbortSignal}} object can chose to ignore
 them. For instance, if the operation has already completed.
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -1835,6 +1835,11 @@ service worker.
 <p>The <dfn attribute for=AbortSignal>aborted</dfn> attribute's getter, when invoked, must return
 true if the <a>context object</a>'s [=AbortSignal/aborted flag=] is set, and false otherwise.
 
+<p>The <dfn attribute for=AbortSignal><code>onabort</code></dfn> attribute is an
+<a>event handler IDL attribute</a> for the <dfn for=AbortSignal><code>onabort</code></dfn>
```
annevk: Bikeshed fails on this because we don't export onabort here. Maybe we should not dfn it?

annevk:
* https://bugs.webkit.org/show_bug.cgi?id=191310
* https://bugzilla.mozilla.org/show_bug.cgi?id=1501983
* https://bugs.chromium.org/p/chromium/issues/detail?id=1006096

dom.bs
domenic: Hmm I swore this built locally.

I'm trying following HTML, which seems to have three definitions in these cases: the IDL attribute, the event handler, and the event.

I think probably the right thing to do is export it, so I'll update it to do that. (And I'll make sure it builds locally.)

domenic:
Force-pushed a version that exports them both.

foolip:
Sorry to revive this old PR, but is there anything that ever causes a "slotchange" event to be fired on `ShadowRoot`? In Chromium source I can't see it ever happening, and a quick search of wpt also doesn't find any tests around this.

Context: https://github.com/mdn/browser-compat-data/pull/16733

domenic:
Hmm. Does it bubble to shadow roots? I don't really remember how bubbling works with shadow DOM...

foolip:
Oh right, it would bubble to the `ShadowRoot`. The composed flag is what causes the event to also bubble from `ShadowRoot` to its host and onward up.

But why does `ShadowRoot` have an `onslotchange` event handler attribute specifically, and not one for every event that might bubble up to it?

domenic:
Yeah, that's a fair question. For my part I am pretty sure that I just followed @annevk's lead in https://github.com/whatwg/html/issues/3487.

Maybe it's because this is a common event to want to listen to on ShadowRoot? Or maybe we made a consistency error.

annevk:
I think it might be a net improvement to add most that can bubble there? Not sure why we'd treat it differently to other objects in that respect. Probably worth a new issue though.

foolip:
I've filed https://github.com/whatwg/dom/issues/1097.

PR #794, opened 2019-11-08
Address infinite loop in TreeWalker's nextNode()

### LINKED ISSUE #787: TreeWalker::nextNode() method, loops infinite
https://dom.spec.whatwg.org/commit-snapshots/57512fac17cf2f1c4c85be4aec178c8086ee5ee4/#interface-treewalker

> The nextNode\(\) method, when invoked, must run these steps:​
> 
> 1. Let node be the context object’s current\.
> 2. Let result be FILTER\_ACCEPT\.
> 3. While true:​
>     1. While result is not FILTER\_REJECT and node has a child:​
>         1. Set node to its first child\.
>         2.Set result to the result of filtering node within the context object\.
>         3. If result is FILTER\_ACCEPT, then set the context object’s current to node and return node\.
>     2. Let sibling be null\.
>     3. Let temporary be node\.
>     4. While temporary is non\-null:​
>         1. If temporary is the context object’s root, then return null\.
>         2. Set sibling to temporary’s next sibling\.
>         3. If sibling is non\-null, then break\.
>         4. Set temporary to temporary’s parent\.
>     5. If sibling is null, break
>     6. Set node to sibling
>     7. Set result to the result of filtering node within the context object\.
>     8. If result is FILTER\_ACCEPT, then set the context object’s current to node and return node\.

In the algorithm are missing the steps "V" and "VI", without them, it loops infinite

annevk:
See https://github.com/whatwg/dom/issues/787#issuecomment-550485387 onward for analysis and test references.

Fixes #787.

@pwombwell care to review? I'll land it next week I suppose if all is in order.

***
<a href="https://whatpr.org/dom/794.html" title="Last updated on Nov 8, 2019, 2:21 PM UTC (e2b172b)">Preview</a> | <a href="https://whatpr.org/dom/794/b9383a3...e2b172b.html" title="Last updated on Nov 8, 2019, 2:21 PM UTC (e2b172b)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9519,7 +9519,8 @@ steps:
 
      <li><p>Set <var>sibling</var> to <var>temporary</var>'s <a for=tree>next sibling</a>.
 
-     <li><p>If <var>sibling</var> is non-null, then <a for=iteration>break</a>.
+     <li><p>If <var>sibling</var> is non-null, then set <var>node</var> to <var>sibling</var> and
+     <a for=iteration>break</a>.
 
      <li><p>Set <var>temporary</var> to <var>temporary</var>'s <a for=tree>parent</a>.
     </ol>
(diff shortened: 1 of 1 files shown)


annevk:
(Got review via DM.)

PR #805, opened 2019-12-06
Editorial: embrace that attributes are objects

### LINKED ISSUE #801: Editorial: remove parent argument to remove
A number of callers don't give it and callers do check for parent being non-null before calling. It seems we should assert it's non-null and then use that. I.e., turn the internal concept into being closer to `child.remove()` rather than `parent.removeChild(child)`.

Bonus: clean up attribute removal in a similar fashion.

annevk:
Closes #801.

***
<a href="https://whatpr.org/dom/805.html" title="Last updated on Dec 11, 2019, 5:34 PM UTC (6ad5327)">Preview</a> | <a href="https://whatpr.org/dom/805/37d8475...6ad5327.html" title="Last updated on Dec 11, 2019, 5:34 PM UTC (6ad5327)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6123,8 +6123,6 @@ its <a for=Element>attribute list</a> <a for=list>is empty</a>.
 <a>attribute</a> <var>attribute</var> to <var>value</var>, run these steps:
 
 <ol>
- <li><p>Assert: <var>attribute</var>'s <a for=Attr>element</a> is non-null.
-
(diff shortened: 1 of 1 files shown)


annevk:
@nox @pmdartus this might be of interest. It mainly reduces redundancy, which the attribute section had plenty of due to the desire to get rid of `Attr` and that then failing.

annevk:
@domenic this also refactors some of the custom element attribute callback steps.

Issue #789, opened 2019-10-11
Trusted Types integration


koto:
We've outlined what we think would be the HTML integration points needed for [Trusted Types](https://w3c.github.io/webappsec-trusted-types/dist/spec/). This accompanies [HTML#3052](https://github.com/whatwg/html/issues/3052) describing the integration with HTML.

## Approaches
The current draft spec implements the TT checks at the DOM sinks (JS functions) layer, and then discards the type information, such that TT become invisible to most other Web APIs and algorithms. The upside is that TT are simpler to spec & implement, the downside is that the future sinks might be introduced that skip the TT logic, and re-introduce DOM XSS-proneness. Or that we might have missed some existing algorithms that would bypass TT already. There's also a few of bypasses that require some additional custom protections (e.g. [this](https://w3c.github.io/webappsec-trusted-types/dist/spec/#enforcement-in-script-text)). Let's call that **tt-at-sinks approach**.

@annevk [proposed](https://github.com/w3c/webappsec-trusted-types/issues/176) an alternative approach - to keep the type information intact such that it can be verified by the algorithms running later on (e.g. when script is to be prepared, check that its URL was a [`TrustedScriptURL`](https://w3c.github.io/webappsec-trusted-types/dist/spec/#trused-script-url). Let's call that **tt-at-primitives approach**.

## TT-at-sinks 
* [Add text node validation steps](https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-text-node-validation) to the [insert](https://dom.spec.whatwg.org/#concept-node-insert) algorithm.
Note: Perhaps this needs to be changed - the checks might be problematic in step 7 of [node insert](https://dom.spec.whatwg.org/#concept-node-insert) as they can partially fail, and then step 8 doesn't run. Perhaps hook after insertion but pre execution.

* Changes to `Element.setAttribute*`. That's to add TT checks for IDL attributes that reflect content attributes with those checks. Currently only explained as a [note](https://w3c.github.io/webappsec-trusted-types/dist/spec/#!trustedtypes-extended-attribute). This needs some work, but in general the approach is to perform the checks (via relevant global object) that ascertain that TT type checks are performed <del>based</del>, determining the right type for (context object, qualified name) pair. These callouts may be added to [set an attribute value](https://dom.spec.whatwg.org/#concept-element-attributes-set-value) algorithm.

[...]

PR #809, opened 2019-12-10
Add attribute validate steps.

koto:
When [handling attribute change](https://dom.spec.whatwg.org/#handle-attribute-changes), validate the value before setting it. 

Attribute [append](https://dom.spec.whatwg.org/#concept-element-attributes-append) now also takes a value, and sets it only after handling the changes (this should be side-effects free, as no callbacks in handle attribute change get passed the attribute node - whose value is not yet set).

See #789 for the background. This PR doesn't allow the validate steps to change the value, only to possibly reject setting it.

***

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fkoto%2Fdom%2Fcc379857acc3a5b05eff9f64ca5ec9f989913332%2Fdom.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%20809)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/dom%23809.)._
</details>


koto:
Rebased now that #805 is merged.

koto:
From my pov this is ready for a review, see discussion in #789.

koto:
Friendly ping, this seems to be the most efficient way we can prevent [attribute node bypasses](https://github.com/w3c/webappsec-trusted-types/issues/47) in Trusted Types by rejecting the value without potentially calling the default policy on all attr mutations.

annevk:
I think this makes sense, but we should add a note explaining this setup somewhere close to where it's relevant.

lukewarlow:
This should have implementor interest now given Mozilla's positive standards position on trusted types as a whole.

annevk:
I think the main reason it wasn't merged was because it was abandoned.

I do wonder if we want an arbitrary extension point for this or just call into Trusted Types directly. With an arbitrary extension point we have the risk of running into ordering issues so we tend to avoid those these days, even though the DOM has some precedent for it (mainly historical).

annevk:
That's not really an issue, especially for integration points. And in fact, we tend to prefer referencing the latest editor's draft.

domenic:
Definite +1 for a direct call out.

koto:
https://github.com/whatwg/dom/pull/1247 and https://github.com/w3c/trusted-types/pull/418 supercede this one. Unfortunately handling the attr mutation in DOM happens now after setting the new value, so I had to add an explicit new step to the algorithms.

PR #815, opened 2020-01-02
Remove document.origin

### LINKED ISSUE #410: Consider removing `document.origin`

annevk:
Use self.origin instead. (The document's origin internal concept is kept for now. Maybe at some point we can only use that of the relevant global object, but not now.)

Tests: ...

Closes #410.

***
<a href="https://whatpr.org/dom/815.html" title="Last updated on Jan 2, 2020, 8:09 AM UTC (7e2469c)">Preview</a> | <a href="https://whatpr.org/dom/815/0913dac...7e2469c.html" title="Last updated on Jan 2, 2020, 8:09 AM UTC (7e2469c)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4747,7 +4747,6 @@ interface Document : Node {
   [SameObject] readonly attribute DOMImplementation implementation;
   readonly attribute USVString URL;
   readonly attribute USVString documentURI;
-  readonly attribute USVString origin;
   readonly attribute DOMString compatMode;
   readonly attribute DOMString characterSet;
   readonly attribute DOMString charset; // historical alias of .characterSet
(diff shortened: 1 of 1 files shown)


annevk:
Filed https://bugs.webkit.org/show_bug.cgi?id=205681.

annevk:
Note to self: maybe remove the parenthetical as it isn't entirely clear to me what the best design is here. Mostly using document as it has less mutable state for all things might well be preferable.

sideshowbarker:
@whatwg/documentation after this spec change is merged, https://developer.mozilla.org/en-US/docs/Web/API/Document/origin will need to be updated (and https://github.com/mdn/browser-compat-data/blob/master/api/Document.json#L7441 flipped to `standard_track: false, deprecated: true`)

sideshowbarker:
Docs:
* updated https://wiki.developer.mozilla.org/en-US/docs/Web/API/Document/origin
* raised https://github.com/mdn/browser-compat-data/pull/5448

Issue #814, opened 2019-12-31
replaceNode() has odd mutation observer behavior after #754


TimothyGu:
After implementing #754 in jsdom, we are no longer passing [wpt/dom/nodes/MutatioObserver-childList.html](https://github.com/web-platform-tests/wpt/blob/878422d33b13cf46ec0d4f4201536426137ecc62/dom/nodes/MutationObserver-childList.html#L277-L297) due to the following test:

```html
<p id='n52'><span>NO </span><span>CHANGED</span></p>
<p id='n53'><span>text content</span></p>
```
```diff
   var n52 = document.getElementById('n52');
   runMutationTest(n52,
                   {"childList":true},
                   [{type: "childList",
-                    removedNodes: [n52.lastChild],
-                    previousSibling: n52.firstChild},
+                    removedNodes: [n52.lastChild]},
                    {type: "childList",
                     removedNodes: [n52.firstChild],
                     addedNodes: [n52.lastChild]}],
                   function() { n52.replaceChild(n52.lastChild, n52.firstChild); },
                   "childList Node.replaceChild: internal replacement mutation");
 
   var n53 = document.getElementById('n53');
   runMutationTest(n53,
                   {"childList":true},
                   [{type: "childList",
-                    removedNodes: [n53.firstChild]},
-                   {type: "childList",
-                    addedNodes: [n53.firstChild]}],
+                    addedNodes: [n53.firstChild],
+                    removedNodes: [n53.firstChild]}],
                   function() { n53.replaceChild(n53.firstChild, n53.firstChild); },
                   "childList Node.replaceChild: self internal replacement mutation");
```
(`-` is the previous expected result; `+` is what #754 gives.)

While the n53 change looks fine if in need of an update, n52 is perhaps more problematic.

Prior to #754, `n52.lastChild` first gets removed as part of adoption process in [replace](https://dom.spec.whatwg.org/#concept-node-replace), between steps 9 and 10, leading to a mutation record. At the time of its removal, `n52.firstChild` is still in place, so it's recorded as `previousSibling` in the record.

After #754, however, the explicit adoption process was removed. So `n52.firstChild` first gets removed silently at step 11, and then `n52.lastChild` is removed as part of the adoption process in insert in step 13, leading to a mutation record – when `n52.firstChild` is no longer present in the document.

[...]

Issue #813, opened 2019-12-31
What does adoptNode() return for DocumentFragment?


TimothyGu:
#754 adds to [`adoptNode()`](https://dom.spec.whatwg.org/#dom-document-adoptnode):

> 3. If _node_ is a `DocumentFragment` node whose host is non-null, then return.

But `adoptNode()` must return a `Node`. What does this "return" mean?

@annevk

annevk:
Sigh, it seems that change introduced a bunch of errors. If we can fix the `replaceChild()` issue somehow I'd say it should return the node passed in, even though it was not adopted.

Perhaps I should first revert that change and test changes though while we figure out a new design. Thoughts welcome.

TimothyGu:
I think that would be good, especially since none of the major browsers have implemented it. I'd be glad to use jsdom as a testbed for any such change.

PR #819, opened 2020-01-10
Adoption and DocumentFragment, part two

annevk:
This reverts commit c825ceaf3cb77943f487f85533d8be79bccd2fe3 and adds a new solution for `DocumentFragment` adoption on top. See #813 and #814 for context. 

HTML: https://github.com/whatwg/html/pull/5413

Tests: https://github.com/web-platform-tests/wpt/pull/22504

jsdom: https://github.com/jsdom/jsdom/pull/2925

***
<a href="https://whatpr.org/dom/819.html" title="Last updated on Oct 18, 2022, 9:21 AM UTC (7afc950)">Preview</a> | <a href="https://whatpr.org/dom/819/bf5f6c2...7afc950.html" title="Last updated on Oct 18, 2022, 9:21 AM UTC (7afc950)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5329,8 +5329,8 @@ method steps are:
 algorithm is passed <var>node</var> and <var>oldDocument</var>, as indicated in the <a>adopt</a>
 algorithm.
 
-<p>To <dfn export id=concept-node-adopt>adopt</dfn> a <var>node</var> into a <var>document</var>, run
-these steps:
+<p>To <dfn export id=concept-node-adopt>adopt</dfn> a <var>node</var> into a <var>document</var>,
+with an optional <var>forceDocumentFragmentAdoption</var> (default false):
 
 <ol>
  <li><p>Let <var>oldDocument</var> be <var>node</var>'s <a for=Node>node document</a>.
@@ -5347,6 +5347,16 @@ these steps:
     <a>shadow-including inclusive descendants</a>:
 
     <ol>
+     <li>
+      <p>If <var>forceDocumentFragmentAdoption</var> is false, <var>inclusiveDescendant</var> is a
+      {{DocumentFragment}} <a for=/>node</a>, <var>inclusiveDescendant</var> is <var>node</var>, and
+      <var>node</var>'s <a for=DocumentFragment>host</a> is non-null, then
+      <a for=iteration>continue</a>.
+
+      <p class=note>This is only reasonable as long as all <a>adopt</a> callers remove the children
+      of <var>node</var>. HTML's <{template}> element passes true for
+      <var>forceDocumentFragmentAdoption</var>.
+
      <li><p>Set <var>inclusiveDescendant</var>'s <a for=Node>node document</a> to <var>document</var>.
 
      <li><p>If <var>inclusiveDescendant</var> is an <a for=/>element</a>, then set the
@@ -5374,6 +5384,12 @@ these steps:
 
  <li><p>If <var>node</var> is a <a for=/>shadow root</a>, then <a>throw</a> a
  "{{HierarchyRequestError!!exception}}" {{DOMException}}.
+
+ <li>
+  <p>If <var>node</var> is a {{DocumentFragment}} <a for=/>node</a> and its
+  <a for=DocumentFragment>host</a> is non-null, then return <var>node</var>.
+
+  <p class=note>Unfortunately this does not throw for web compatibility.
 
  <li><p><a>Adopt</a> <var>node</var> into <a>this</a>.
 
--- a/dom.bs
+++ b/dom.bs
@@ -2509,6 +2509,8 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
  <li><p>If <var>referenceChild</var> is <var>node</var>, then set <var>referenceChild</var> to
  <var>node</var>'s <a for=tree>next sibling</a>.
 
+ <li><p><a>Adopt</a> <var>node</var> into <var>parent</var>'s <a for=Node>node document</a>.
+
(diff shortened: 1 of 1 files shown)


annevk:
Bugs (still the same):
* https://bugzilla.mozilla.org/show_bug.cgi?id=1602200
* https://bugs.webkit.org/show_bug.cgi?id=204980
* https://bugs.chromium.org/p/chromium/issues/detail?id=1031811

domenic:
Anything I can help with on the jsdom side, or on reviews, to land this and the associated HTML/WPT PRs?

annevk:
Verifying that this is correct would help. Perhaps @mfreed7 is interested in prototyping in Chromium?

mfreed7:
> Verifying that this is correct would help as well as a review. Perhaps @mfreed7 is interested in prototyping in Chromium?

The change looks reasonable, but from #814 it might be tricky to get exactly right. I'm happy to prototype in Chromium, but I'd prefer to do that after jsdom has had a chance to confirm the change passes existing WPTs... @TimothyGu?

dom.bs
```diff
@@ -5260,10 +5265,15 @@ when invoked, must run these steps:
 algorithm is passed <var>node</var> and <var>oldDocument</var>, as indicated in the <a>adopt</a>
 algorithm.
 
-<p>To <dfn export id=concept-node-adopt>adopt</dfn> a <var>node</var> into a <var>document</var>, run
-these steps:
+<p>To <dfn export id=concept-node-adopt>adopt</dfn> a <var>node</var> into a <var>document</var>,
+with an optional <var>forceDocumentFragmentAdoption</var>, run these steps:
```
rniwa, line 5269: When is this argument ever set to true??

meshl555 (commented):
Noneo8

meshl555:
Noneo8

rniwa:
FWIW, I've implemented this new behavior in https://commits.webkit.org/252098@main

dom.bs
```diff
@@ -5370,8 +5385,11 @@ these steps:
  <li><p>If <var>node</var> is a <a for=/>shadow root</a>, then <a>throw</a> a
  "{{HierarchyRequestError!!exception}}" {{DOMException}}.
 
- <li><p>If <var>node</var> is a {{DocumentFragment}} <a for=/>node</a> whose
- <a for=DocumentFragment>host</a> is non-null, then return.
+ <li>
+  <p>If <var>node</var> is a {{DocumentFragment}} <a for=/>node</a> and its
+  <a for=DocumentFragment>host</a> is non-null, then return <var>node</var>.
+
+  <p class=note>Unfortunately this does not throw for web compatibility.
```
bigopon, line 5392: What does "does not throw for web compatibility" mean? Any one who employs `template.content` node adoption before will now face with breakage with the sudden change of "whose host is non-null, then return".

If we no longer guarantee old API evolves without breaking, how do I find some rule of thumbs what is dangerous and what is not? An example issue is here https://github.com/aurelia/aurelia/issues/1264

jded76: > From reading the issue you linked it seems that implementations never matched the specification and some applications (such as yours) relied on that. 
That's not true. The implementations were according to the spec until #754, that inserted a rule about template's fragment for the first time. 
That rule returned undefined, there were some issues with that and the browsers didn't implemented (according to [this ](https://github.com/whatwg/dom/issues/813#issuecomment-570191640).
Then with #819 the spec changed again. 
On July webkit implemented the new rule. Before one month Safari went public with the new version of webkit and many applications broke on **production**.
And since then we are trying to figure out what's going on.

mfreed7: So it sounds from the linked issues and comment like this change isn't web compatible as-is, right? Sounds like another approach is needed.

jded76: Any updates on this?
Right now our application is not working on Safari on production. 
Webkit [says ](https://bugs.webkit.org/show_bug.cgi?id=246899) that this is intentional behavior change and we don't know how to proceed to fix our application...

jded76:
Sorry for coming back on this again and again, but we have a [problem ](https://github.com/whatwg/dom/pull/819#discussion_r1006583041) in production with Safari and it seems that we got stuck...
Just ignoring it doesn't solve anything. I think a [decision ](https://github.com/whatwg/dom/pull/819#discussion_r1007199723) has to be made here soon...

bigopon:
Is there a process that I need to follow or do I just create a PR to fix this? The easiest option is to revert the change and find a more reasonable one later.

bigopon:
It's been awhile without answer, where can I shoot my concerns over this issues? It's breaking both apps and hearts.

mfreed7:
> It's been awhile without answer, where can I shoot my concerns over this issues? It's breaking both apps and hearts.

I believe you're referring to the WebKit/Safari behavior described here, correct? https://bugs.webkit.org/show_bug.cgi?id=246899

If so, that bug is the better place to raise the issue, since this issue is about the spec. That bug is informative about the direction the spec needs to take, though, since it seems the proposed direction isn't web compatible.

karlcow:
so adoptNode behavior has been reverted in WebKit 
https://github.com/WebKit/WebKit/pull/11347
because of https://github.com/aurelia/framework/issues/1003
WebKit being the only to have adopted the new behavior, it was having the burden of the breakage. 

There is a need for a better path forward.

bigopon:
> There is a need for a better path forward.

What we want is the ability to adopt the document fragment. Currently adopting the template element doesn't adopt the document fragment itself. What if we do that?

I dont know all the reasons why the fragment of the template element isn't associated with the document that created it, but if calling adoptNode explicitly on template does it then maybe this behavior is safe to go in?

annevk:
Unfortunately this doesn't work. `DocumentFragment` nodes with a host need different handling. To be continued in #813.

PR #820, opened 2020-01-13
Editorial: add missing comma in "add an event listener" definition

### LINKED ISSUE #788: Editorial: move contains check into "remove an event listener"
That way cssom-view can invoke it directly. We might want to rename the existing "remove an event listener" to "remove an event listener inner" or some such so that HTML can invoke that directly, or let HTML use the one that does a check that will always noop for HTML.

cc @zcorpan

shvaikalesh:
Fixes #788.

***
<a href="https://whatpr.org/dom/820.html" title="Last updated on Jan 16, 2020, 2:04 PM UTC (2b7b081)">Preview</a> | <a href="https://whatpr.org/dom/820/55379a7...2b7b081.html" title="Last updated on Jan 16, 2020, 2:04 PM UTC (2b7b081)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1146,7 +1146,7 @@ must return a new {{EventTarget}}.
 if this would be useful for your programs. For now, all author-created {{EventTarget}}s do not
 participate in a tree structure.</p>
 
-<p>To <dfn export>add an event listener</dfn> given an {{EventTarget}} object <var>eventTarget</var>
+<p>To <dfn export>add an event listener</dfn>, given an {{EventTarget}} object <var>eventTarget</var>
 and an <a>event listener</a> <var>listener</var>, run these steps:
 
 <ol>
--- a/dom.bs
+++ b/dom.bs
@@ -1146,8 +1146,8 @@ must return a new {{EventTarget}}.
 if this would be useful for your programs. For now, all author-created {{EventTarget}}s do not
 participate in a tree structure.</p>
 
-<p>To <dfn export>add an event listener</dfn>, given an {{EventTarget}} object <var>eventTarget</var>
-and an <a>event listener</a> <var>listener</var>, run these steps:
+<p>To <dfn export>add an event listener</dfn>, given an {{EventTarget}} object
+<var>eventTarget</var> and an <a>event listener</a> <var>listener</var>, run these steps:
 
 <ol>
  <li><p>If <var>eventTarget</var> is a {{ServiceWorkerGlobalScope}} object, its
--- a/dom.bs
+++ b/dom.bs
@@ -10007,9 +10007,9 @@ what it is today.
 <p>With that, many thanks to
 Adam Klein,
 Adrian Bateman,
-Aleksey Shvayka,
 Alex Komoroske,
 Alex Russell,
+Alexey Shvayka,
 Anthony Ramine,
 Arkadiusz Michalski,
 Arnaud Le Hors,


dom.bs
```diff
@@ -1146,7 +1146,7 @@ must return a new {{EventTarget}}.
 if this would be useful for your programs. For now, all author-created {{EventTarget}}s do not
 participate in a tree structure.</p>
 
-<p>To <dfn export>add an event listener</dfn> given an {{EventTarget}} object <var>eventTarget</var>
+<p>To <dfn export>add an event listener</dfn>, given an {{EventTarget}} object <var>eventTarget</var>
```
annevk: The last thing needs to go on the next line due to exceeding 100 columns.

dom.bs
```diff
@@ -1196,9 +1196,13 @@ method, when invoked, must run these steps:
...
+ <li><p>If <var>eventTarget</var>'s <a for=EventTarget>event listener list</a> does not
+ <a for=list>contain</a> an <a>event listener</a> whose <a for="event listener">type</a> is
+ <var>listener</var>'s <a for="event listener">type</a>, <a for="event listener">callback</a> is
+ <var>listener</var>'s <a for="event listener">callback</a>, and <a for="event listener">capture</a>
+ is <var>listener</var>'s <a for="event listener">capture</a>, then set <var>listener</var>'s
+ <a for="event listener">removed</a> to true and <a for=list>remove</a> <var>listener</var> from
+ <var>eventTarget</var>'s <a for=EventTarget>event listener list</a>.
 </ol>
 
 <p class=note>HTML needs this to define event handlers. [[HTML]]
```
annevk: Maybe update this note to indicate that the contains check above always no-ops when HTML calls this? Or maybe call that out in an HTML comment. Just keep that knowledge around somewhere.

annevk:
Also, thanks for taking this up!

dom.bs
shvaikalesh: Fixed in 3142d90.

dom.bs
shvaikalesh: Appended this note in a203af5 instead of HTML spec, so we can refer `contains` check more easily (as "above").

I was a bit hesitant to use "check above always no-ops" wording, as it may be understood as "listener is never found, so step 3 does nothing" and went with "always returns true".

dom.bs
```diff
@@ -1196,12 +1196,17 @@ method, when invoked, must run these steps:
...
+ <a for=list>contains</a> an <a>event listener</a> whose <a for="event listener">type</a> is
+ <var>listener</var>'s <a for="event listener">type</a>, <a for="event listener">callback</a> is
+ <var>listener</var>'s <a for="event listener">callback</a>, and <a for="event listener">capture</a>
+ is <var>listener</var>'s <a for="event listener">capture</a>, then set that <a>event listener</a>'s
+ <a for="event listener">removed</a> to true and <a for=list>remove</a> it from
+ <var>eventTarget</var>'s <a for=EventTarget>event listener list</a>.
 </ol>
 
-<p class=note>HTML needs this to define event handlers. [[HTML]]
+<p class=note>HTML needs this to define event handlers. When called from HTML,
```
```suggestion
<p class=note>HTML needs this to define event handlers. When called from HTML, the
```

annevk:
I'm so sorry, I realize now that this doesn't make sense.

`removeEventListener()` first needs to find the event listener to remove. It creating an event listener that is then compared to existing event listeners makes little sense.

And this also goes for CSSOM View, so I think the layering as-is is correct.

We should still fix your name though.

annevk:
I'll rebase it to drop the other commits, but please let me know if you have any concerns.

shvaikalesh:
Yeah, it felt a little weird with all the "set that event listener's ..." wording I have not found anywhere else in the spec and intermediate event listener. Current prose "contains event listener whose" is very concise.

I've considered extracting "inner remove an event listener", yet it has the same cons.

annevk:
Yeah, sorry, I clearly didn't think it through when I filed that issue and since then have been assuming I was correct until I did one last read through just now.

shvaikalesh:
No problem at all, thank you for taking time to thoroughly review this change.

I've rebased and renamed the PR.

PR #822, opened 2020-01-16
Avoid enqueuing tree mutation record when there is no mutation

### LINKED ISSUE #796: Avoid enqueuing tree mutation record when there is no mutation
Currently, the [replace all](https://dom.spec.whatwg.org/#concept-node-replace-all) steps queues a tree mutation record even when `addedNodes` and `removedNodes` are empty. When following the spec strictly, the [childList Node.textContent: textContent no mutation](https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/MutationObserver-childList.html#L103-L107) fails.

IMHO, the replace all step should only enqueue a tree mutation record when either `addedNodes` or `removedNodes` is not empty. In fact, servo implementation already add [this extra branch check](https://github.com/servo/servo/blob/5ab2d4ea835374c2ca88f5d9b50100669f142273/components/script/dom/node.rs#L2008-L2016).

shvaikalesh:
Fixes #796.

***
<a href="https://whatpr.org/dom/822.html" title="Last updated on Jan 22, 2020, 3:23 PM UTC (de10bc3)">Preview</a> | <a href="https://whatpr.org/dom/822/55379a7...de10bc3.html" title="Last updated on Jan 22, 2020, 3:23 PM UTC (de10bc3)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2416,7 +2416,7 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
    <li><p><a for=/>Remove</a> its <a>children</a> with the <i>suppress observers flag</i> set.
 
    <li>
-    <p><a>queue a tree mutation record</a> for <var>node</var> with « », <var>nodes</var>, null, and
+    <p><a>Queue a tree mutation record</a> for <var>node</var> with « », <var>nodes</var>, null, and
     null.
 
     <p class="note no-backref">This step intentionally does not pay attention to the
--- a/dom.bs
+++ b/dom.bs
@@ -2604,7 +2604,7 @@ within a <var>parent</var>, run these steps:
  <li><p>If <var>node</var> is non-null, then <a for=/>insert</a> <var>node</var> into
  <var>parent</var> before null with the <i>suppress observers flag</i> set.
 
- <li><p>If either <var>addedNodes</var> or <var>removedNodes</var> <a for=list>is not empty</a>,
+ <li><p>If either <var>addedNodes</var> or <var>removedNodes</var> <a for=set>is not empty</a>,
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -2604,7 +2604,8 @@ within a <var>parent</var>, run these steps:
  <li><p>If <var>node</var> is non-null, then <a for=/>insert</a> <var>node</var> into
  <var>parent</var> before null with the <i>suppress observers flag</i> set.
 
- <li><p><a>Queue a tree mutation record</a> for <var>parent</var> with <var>addedNodes</var>,
+ <li><p>If either <var>addedNodes</var> or <var>removedNodes</var> <a for=list>is not empty</a>,
```
```suggestion
 <li><p>If either <var>addedNodes</var> or <var>removedNodes</var> <a for=set>is not empty</a>,
```

dom.bs
```diff
@@ -3591,6 +3592,9 @@ run these steps:
 <var>target</var> with null, null, null, <var>addedNodes</var>, <var>removedNodes</var>,
 <var>previousSibling</var>, and <var>nextSibling</var>.
 
+<p class="note no-backref">Either <var>addedNodes</var> or <var>removedNodes</var>
+<a for=list>is not empty</a>.
```
annevk: This seems nice to call out, but then we should turn the thing above this into an algorithm with two steps and the first step is this, as an assert.

dom.bs
shvaikalesh: Please also see #823.

dom.bs
shvaikalesh: Resolved in de10bc3.

annevk:
Thanks @shvaikalesh, really appreciate it!

PR #858, opened 2020-04-17
Add declarative Shadow DOM features

mfreed7:
The explainer for this feature is here: https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md
The issue discussion is here: https://github.com/whatwg/dom/issues/831
There is a corresponding Pull Request for the HTML spec that goes along with this PR.

   * I'm working to get this implemented in Chromium, tracking bug [here](https://crbug.com/1042130).
   * Other implementers have at least not been negative on the [discussion thread](https://github.com/whatwg/dom/issues/831). Not explicitly positive (yet).

   * In WPT: https://wpt.fyi/results/shadow-dom/declarative?label=master&label=experimental&aligned&q=declarative
   * More tests are being implemented, but the basics are there.

   * Chrome: https://crbug.com/1042130
   * Firefox: None yet
   * Safari: None yet

***
<a href="https://whatpr.org/dom/858.html" title="Last updated on Jun 11, 2020, 9:43 PM UTC (191a88c)">Preview</a> | <a href="https://whatpr.org/dom/858/94679a2...191a88c.html" title="Last updated on Jun 11, 2020, 9:43 PM UTC (191a88c)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5727,6 +5727,9 @@ or "<code>closed</code>").</p>
 <p><a for=/>Shadow roots</a> have an associated <dfn export for=ShadowRoot>delegates focus</dfn>.
 It is initially set to false.</p>
 
+<p><a for=/>Shadow roots</a> have an associated <dfn export for=ShadowRoot>is declarative shadow root</dfn>.
+It is initially set to false.</p>
+
 <p><a for=/>Shadow roots</a>'s associated <a for=DocumentFragment>host</a> is never null.</p>
 <!-- If we ever change this, e.g., add a ShadowRoot object constructor, that would have serious
      consequences for innerHTML. -->
@@ -6750,6 +6753,8 @@ invoked, must run these steps:
 
  <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>delegates focus</a> to <var>init</var>'s
  {{ShadowRootInit/delegatesFocus}}.
+
+ <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>is declarative shadow root</a> to false.
 
  <li><p>Set <a>this</a>'s <a for=Element>shadow root</a> to <var>shadow</var>.
 
@@ -10133,6 +10138,7 @@ Manish Tripathi,
 Marcos Caceres,
 Mark Miller,
 Martijn van der Ven,
+Mason Freed,  <!-- mfreed7 on GitHub -->
 Mats Palmgren,
 Mounir Lamouri,
 Michael Stramel,
--- a/dom.bs
+++ b/dom.bs
@@ -4232,8 +4232,8 @@ elements. SVG ought to do the same for its <{script}> elements, but does not cal
 at the moment.
 
 <p>To <dfn export id=concept-node-clone lt="clone a node" local-lt="clone">clone</dfn> a
-<var>node</var>, with an optional <var>document</var> and <i>clone children flag</i>, run these
-steps:
+<var>node</var>, with an optional <var>document</var>, <i>clone children flag</i>, and
+<i>clone shadows flag</i>, run these steps:
 <!-- This algorithm is used by dom-Node-cloneNode, dom-Document-importNode,
 dom-Range-extractContents, dom-Range-cloneContents -->
 
@@ -4307,6 +4307,22 @@ dom-Range-extractContents, dom-Range-cloneContents -->
  <a>children</a> of <var>node</var> and append them to <var>copy</var>, with
  <var>document</var> as specified and the <i>clone children flag</i> being set.
 
+ <li>If the <i>clone shadows flag</i> is set, run these steps:
+
+ <ol>
+   <li><p>Run <a>attach a shadow root</a> with <var>shadow host</var> equal to <var>copy</var>,
+ <var>mode</var> equal to <var>node</var>'s <a for=/>shadow root</a>'s <a for=ShadowRoot>mode</a>, and <var>delegates focus</var>
+ equal to <var>node</var>'s <a for=/>shadow root</a>'s <a>delegates focus</a>.</p></li>
+
+   <li><p>If <var>node</var>'s <a for=/>shadow root</a>'s <a for=ShadowRoot>is declarative shadow root</a> is true, then
+   set <var>copy</var>'s <a for=/>shadow root</a>'s <a for=ShadowRoot>is declarative shadow root</a> property to true.</p></li>
+
+   <li>If the <i>clone children flag</i> is set, <a lt="clone a node">clone</a> all the
+   <a>children</a> of <var>node</var>'s <a for=/>shadow root</a> and append them to <var>copy</var>'s <a for=/>shadow root</a>, with
+   <var>document</var> as specified and the <i>clone children flag</i> being set. 
+
+ </ol>
+
  <li>Return <var>copy</var>.
 </ol>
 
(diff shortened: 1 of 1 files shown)


mfreed7:
Ok, I believe this PR is ready for review. This is my first time with a PR of this size against DOM so please let me know what I need to change. All of this PR has now been implemented in Chromium behind the DeclarativeShadowDOM flag, and WPT tests have been written.

annevk:
Could you rebase it and force push so that the build output will (hopefully) be more useful?

mfreed7:
Ok, I just force-pushed. Do I also need to fix the (pre-existing) issue with en-GB spelling of "serialisation", to get Travis to build?

```
en-GB spelling (use lang="en-GB", or <!-- en-GB -->, on the same line to override):
6922:<a href="https://www.w3.org/TR/html5/single-page.html#html-fragment-serialisation-algorithm">fragment
Makefile:22: recipe for target 'deploy' failed
make: *** [deploy] Error 1
The command "make deploy" exited with 2.
```

As for implementer interest and @rniwa's last comment:
> > b. In the "backwards" case, the `attachShadow()` call from the component "wins" the race. Then the declarative content finishes parsing, and the parser tries to attach the declarative shadow root. In this case, that attachment will fail, and the declarative contents will be discarded. Here again, not performance-ideal, but everything keeps working correctly.
> 
> I don't think this behavior is acceptable. There needs to be a way for a component to let the declarative content arrive into its shadow root. Maybe we need an option to attachShadow which indicates whether the declarative content could be inserted or not.

I believe this is exactly [issue 871](https://github.com/w3c/webcomponents/issues/871), to expose shadowRoot on element internals. I would like to take up that issue also, but separately, and it sounds like we're close to agreement on that thread.

> Also, this poses another problem. Since the component would never know when the children had finished parsing, there is no point in which the component is safe to attach a shadow root until your sibling or your ancestor's sibling starts appearing. Having to write that kind of code manually everywhere seems very fragile & developer hostile.

This second point is exactly [issue 809](https://github.com/w3c/webcomponents/issues/809), to add a callback when parsing is "done". I also think we need to work on that issue, but again, separately. I'm in favor of a solution to that problem, and that is @rniwa's own issue, so presumably he agrees.

Maybe it is my optimistic reading of the above, but it seemed like we might be close. I suppose we will need to wait for @rniwa to return to find out?

In the meantime, what are Mozilla's thoughts on this API? One of your comments requested more clarity about the interaction between declarative and imperative shadow roots. Hopefully that's clear with this PR and the corresponding [HTML PR](https://github.com/whatwg/html/pull/5465)?

annevk:
If you want to reference the HTML Standard's fragment serialization algorithm you want something like `<a>HTML fragment serialization algorithm</a>`. Out of curiosity, where did you copy that URL from?

It might take some time for Mozilla to determine a position. I recommend filing an issue at https://github.com/mozilla/standards-positions.

mfreed7:
> If you want to reference the HTML Standard's fragment serialization algorithm you want something like `<a>HTML fragment serialization algorithm</a>`. Out of curiosity, where did you copy that URL from?

Thanks - done. I still don't quite understand the magic happening in the Makefile.

I got that link from the HTML spec itself - go to https://html.spec.whatwg.org/#serialising-html-fragments and check out the fragment link there for "HTML fragment serialization algorithm".

> It might take some time for Mozilla to determine a position. I recommend filing an issue at https://github.com/mozilla/standards-positions.

Ok, thanks. I've [filed an issue](https://github.com/mozilla/standards-positions/issues/335) there.

annevk:
Well, the link was  to a completely different domain, too. 😊

mfreed7:
> Well, the link was to a completely different domain, too. 😊

Oh, ha! I didn't notice it was to w3. Yeah now I'm not sure where I got that link. 😬

mfreed7:
Are there any comments on this PR, technically speaking? It is waiting for other implementer support, but it would be good to take care of any obvious spec-writing problems if possible.

dom.bs
```diff
@@ -6734,31 +6787,26 @@ invoked, must run these steps:
...
 
- <li><p>Let <var>shadow</var> be a new <a for=/>shadow root</a> whose <a for=Node>node document</a>
- is <a>this</a>'s <a for=Node>node document</a>, <a for=DocumentFragment>host</a> is <a>this</a>,
- and <a for=ShadowRoot>mode</a> is <var>init</var>'s {{ShadowRootInit/mode}}.
-
- <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>delegates focus</a> to <var>init</var>'s
- {{ShadowRootInit/delegatesFocus}}.
+ <li><p>If <var>shadow host</var> has a non-null <a for=/>shadow root</a> whose
+ <a for=ShadowRoot>is declarative shadow root</a> property is true, then <a for=/>remove</a> all of
+ <a for=/>shadow root</a>'s <a>children</a>, in <a>tree order</a>. Return <var>shadow host</var>'s <a for=/>shadow root</a>.
```
annevk, line 6796: I think it would be clearer to have a single non-null step and then have these algorithm exists (throw and return) as substeps.

dom.bs
```diff
@@ -5872,6 +5896,10 @@ dictionary ShadowRootInit {
   required ShadowRootMode mode;
   boolean delegatesFocus = false;
 };
+
+dictionary GetInnerHTMLOptions {
+  boolean includeShadowRoots = true;
```
annevk, line 5901: At the very least this should say OpenShadowRoots, but it seems better to me if we did not have a separate API here for open and closed.

dom.bs
```diff
@@ -6694,10 +6722,35 @@ when invoked, must run these steps:
...
+
+ <li><p>Return <a>this</a>'s <a for=Element>shadow root</a>.</p></li>
+
+</ol>
+
+<p>The <dfn attribute for=Element><code>shadowRoot</code></dfn> attribute's getter must run these
+steps:
+
+<ol>
+ <li><p>Let <var>shadow</var> be <var>shadow host</var>'s <a for=Element>shadow root</a>.
```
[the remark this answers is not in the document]
mfreed7: Yep, thanks. Fixed.

dom.bs
mfreed7: Done. Let me know if this is what you were expecting.

dom.bs
mfreed7: I've added the "rest" of this interface, as explained [here](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md#serialization) in the explainer. In particular, the section just below that, [Closed shadow roots](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md#closed-shadow-roots) goes into the details of the interface. You essentially get two knobs. `includeShadowRoots` is an opt-in for *any* shadow roots to be serialized. The second input, `closedRoots` is a list of closed shadow roots that you would like to be included. If a closed shadow root is encountered that is not in the list, it will be skipped to preserve encapsulation.

I am going to add the corresponding change to my HTML spec PR.

LMK if the above clarifies the situation, or if you have further questions.

mfreed7:
Thanks for the review! I think i I addressed all of your comments, but let me know what else needs attention.

> FWIW, I think that apart from formatting (and it needs rebasing), this largely looks fine. Cloning shadow trees looks rather odd to me though. The explainer also doesn't go into this from a quick scan.

This is a good point, which wasn't previously in the explainer. I came across this behavior as I was implementing. Based on this comment, I've added a section to the explainer that details why this change was necessary - you can see that [here](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md#ordinary-templates-containing-declarative-shadow-dom). The TL/DR is that without this change (or something similar), you can't use declarative Shadow DOM inside an "ordinary" template.

mfreed7:
Just checking in on this PR. Is there anything missing, other than multi-implementer interest?

mfreed7:
Closing this, and replacing with [PR 892](https://github.com/whatwg/dom/pull/892).

PR #860, opened 2020-04-24
Add Imperative Slot API

yuzhe-han:
The explainer for this feature is here: https://github.com/w3c/webcomponents/blob/gh-pages/proposals/Imperative-Shadow-DOM-Distribution-API.md  
The issue discussion is here: [3534](https://github.com/whatwg/html/issues/3534)  
There is a corresponding [Pull Request](https://github.com/whatwg/html/pull/5483) for the HTML spec that goes along with this PR.  

   * I'm working to get this implemented in Chromium, tracking bug [here](http://crbug.com/869308).
   * The implementation is based on what was agreed in the last [F2F Meeting](https://github.com/whatwg/html/issues/3534#issuecomment-537802687).

   * In WPT: https://wpt.fyi/results/shadow-dom/slots-imperative-slot-api.tentative.html?label=master&label=experimental&aligned&q=slots-imperative-slot-api.tentative.html
   * In WPT: https://wpt.fyi/results/shadow-dom/slots-imperative-api-slotchange.tentative.html?label=master&label=experimental&aligned
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://crbug.com/869308
   * Firefox: None yet
   * Safari: None yet

***

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fyuzhe-han%2Fdom%2F49ce9861f326c9e4337338316bc1e8d29a2495a9%2Fdom.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%20860)

```
<html><head>
</head><body>
your request.</p>
in the server error log.</p>
<hr>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/dom%23860.)._
</details>


dom.bs
```diff
@@ -2195,6 +2195,11 @@ steps:</p>
  <li><p>If the <i>open flag</i> is set and <var>shadow</var>'s <a for=ShadowRoot>mode</a> is
  <em>not</em> "<code>open</code>", then return null.</p></li>
 
+ <li><p>If the shadow's <a for=ShadowRoot>slot assignment</a> is set to "<code>manual</code>",
+ return the associated <a>slot</a> in shadow where the <var>slottable</var> is
+ <a href="https://html.spec.whatwg.org/multipage/scripting.html#dom-slot-assign"><code>assigned</code></a>,
```
annevk, line 2200: This is too much action-at-a-distance. A `slot` probably needs to keep track of what was manually assigned to it in some "associated" thingie.

dom.bs
```diff
@@ -5731,6 +5738,9 @@ It is initially set to false.</p>
 <!-- If we ever change this, e.g., add a ShadowRoot object constructor, that would have serious
      consequences for innerHTML. -->
 
+<p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>slot assignment</dfn>
+("<code>manual</code>" or "<code>auto</code>"), whose default value is "<code>auto</code>"</p>
```
annevk, line 5742: Misses a dot.

dom.bs
```diff
@@ -5713,9 +5718,11 @@ interface ShadowRoot : DocumentFragment {
   readonly attribute ShadowRootMode mode;
   readonly attribute Element host;
   attribute EventHandler onslotchange;
+  readonly attribute SlotAssignmentMode slotAssignment;
```
annevk, line 5721: I think this should be listed under `mode`.

dom.bs
yuzhe-han: Thanks Anne. I've update this portion to be more of an algorithm with internal variables, manuallyAssignedNodes and nodeSlotMap, to keep track of the association between a manually assigned node to its slot.

dom.bs
yuzhe-han: done.

dom.bs
yuzhe-han: moved.

dom.bs
```diff
@@ -2195,6 +2195,25 @@ steps:</p>
  <li><p>If the <i>open flag</i> is set and <var>shadow</var>'s <a for=ShadowRoot>mode</a> is
  <em>not</em> "<code>open</code>", then return null.</p></li>
 
+ <li><p>If the shadow's <a for=ShadowRoot>slot assignment</a> is set to "<code>manual</code>", then:</p>
```
annevk, line 2198: This should be `<var>shadow</var>`.

dom.bs
```diff
@@ -2195,6 +2195,25 @@ steps:</p>
...
+  <ol>
+   <li><p>Let <var>manuallyAssignedNodes</var> be nodes that are
+   <a href="https://html.spec.whatwg.org/multipage/scripting.html#dom-slot-assign"><code>assigned</code></a>
```
annevk, line 2201: This is a method, it should reference an internal/associated concept, no?

dom.bs
```diff
@@ -2195,6 +2195,25 @@ steps:</p>
...
+   to a <a>slot</a> in <var>shadow</var>'s <a for=tree>descendants</a>.</p></li>
+
+   <li><p>If <var>nodeSlotMap</var> doesn't exists, let <var>nodeSlotMap</var> be an empty <a for=/>map</a>.</p></li>
```
annevk, line 2204: How could it exist here?

dom.bs
```diff
@@ -2195,6 +2195,25 @@ steps:</p>
...
+
+   <li><p>For each <var>node</var> in <var>manuallyAssignedNodes</var>, set <var>nodeSlotMap[node]</var>
```
annevk, line 2206: You want `<var>nodeSlotMap</var>[<var>node</var>]` here.

dom.bs
```diff
@@ -2195,6 +2195,25 @@ steps:</p>
...
+   to <a>slot</a>.</p></li>
```
annevk, line 2207: Which slot?

dom.bs
```diff
@@ -2195,6 +2195,25 @@ steps:</p>
...
+
+   <li><p>Set <var>manuallyAssignedNodes</var> to empty list.</p></li>
```
annevk, line 2209: an empty list*

yuzhe-han:
@annevk,

Thanks for your review. I updated the corresponding HTML spec to include the concept of manuallyAssignedNodes and used that concept within [find a slot].  In addition, I removed a number of steps under [find a slot][step 5] where slotAssignment == 'manual'. I found those steps complicates things and unnecessary.  Hopefully, the new revision is  more clear.

I didn't want my local branch to be too far out of date, so I fetched from whatwg:master and rebased my changes on top. It changed history so now your previous reviews aren't matched. Maybe I should have done a merge instead?  

Sorry for any inconvenience.  Thanks again for your review.

dom.bs
```diff
@@ -2195,6 +2195,10 @@ steps:</p>
  <li><p>If the <i>open flag</i> is set and <var>shadow</var>'s <a for=ShadowRoot>mode</a> is
  <em>not</em> "<code>open</code>", then return null.</p></li>
 
+ <li><p>If the <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> is set to "<code>manual</code>",
+ return the associated slot in <var>shadow</var>'s tree whose <a>manuallyAssignedNodes</a> includes
```
annevk, line 2199: You want to change this to be more similar to the next step in terms of wording. Although I suppose you can leave out "first" and "tree order" since there's guaranteed to be a single slot if I understand it correctly.

dom.bs
```diff
@@ -5731,6 +5737,9 @@ It is initially set to false.</p>
 <!-- If we ever change this, e.g., add a ShadowRoot object constructor, that would have serious
      consequences for innerHTML. -->
 
+<p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>slot assignment</dfn>
+("<code>manual</code>" or "<code>auto</code>"), whose default value is "<code>auto</code>".</p>
```
annevk, line 5741: I don't think we have to state a default/initial value as it's always given. This is also a problem with delegates focus.

dom.bs
yuzhe-han: Updated wording.

dom.bs
yuzhe-han: Removed default value.

yuzhe-han:
Thanks for reviewing. I've uploaded a new revision for review.

annevk:
I don't really understand the setup anymore. I thought manually assigned nodes was mutually exclusive with assigned nodes, but here the assign slottables algorithm can end up setting both.

yuzhe-han:
annevk@

I removed the additional step in assign slottables. I also sent you an email for some clarification based on your comments in the corresponding HTML spec. 

Thanks.

rniwa:
FWIW, what's being proposed here seems fine.

dom.bs
```diff
@@ -2195,6 +2195,10 @@ steps:</p>
  <li><p>If the <i>open flag</i> is set and <var>shadow</var>'s <a for=ShadowRoot>mode</a> is
  <em>not</em> "<code>open</code>", then return null.</p></li>
 
+ <li><p>If <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> is set to "<code>manual</code>",
+ return the <a>slot</a> in <var>shadow</var>'s <a for=tree>descendants</a> whose <a>manually assigned nodes</a>
```
annevk, line 2199: then return*

dom.bs
```diff
@@ -2195,6 +2195,10 @@ steps:</p>
...
+ includes <var>slottable</var>, if any, and null otherwise.</p></li>
```
annevk, line 2200: you want `<a for=set>contains</a>` instead of includes here I think. It's a set right?

And I guess we're not bothering with tree order and such since there can only be one such slot.

dom.bs
```diff
@@ -2450,7 +2454,8 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
    <li><p>Otherwise, <a for=set>insert</a> <var>node</var> into <var>parent</var>'s
    <a for=tree>children</a> before <var>child</var>'s <a for=tree>index</a>.
 
-   <li><p>If <var>parent</var> is a <a for=Element>shadow host</a> and <var>node</var> is a
+   <li><p>If <var>parent</var> is a <a for=Element>shadow host</a> with its <a for=tree>root</a>'s
+   <a for=ShadowRoot>slot assignment</a> set to "<code>auto</code>" and <var>node</var> is a
    <a>slottable</a>, then <a>assign a slot</a> for <var>node</var>.
 
    <li><p>If <var>parent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
```
annevk, line 2461: Why should this step not be skipped?

dom.bs
```diff
@@ -2667,8 +2672,16 @@ indicated in the <a for=/>remove</a> algorithm below.
...
- <var>node</var>'s <a>assigned slot</a>.
+ <li>
+  <p>If <var>node</var> is <a for=slottable>assigned</a>, then:
+
+  <ol>
+   <li><p><p>If <var>parent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> with its
+   <a for=tree>root</a>'s <a for=ShadowRoot>slot assignment</a> set to "<code>auto</code>",
+   then <a for=set>remove</a> <var>node</var> from its <a>assigned slot</a>'s <a>manually assigned nodes</a>.
+
+   <li><p>Run <a>assign slottables</a> for <var>node</var>'s <a>assigned slot</a>.
```
annevk, line 2683: Here too I have the feeling we're not skipping enough for the manual case. Same below.

dom.bs
```diff
@@ -2667,8 +2672,16 @@ indicated in the <a for=/>remove</a> algorithm below.
 
  <li><p><a for=set>Remove</a> <var>node</var> from its <var>parent</var>'s <a for=tree>children</a>.
 
- <li><p>If <var>node</var> is <a for=slottable>assigned</a>, then run <a>assign slottables</a> for
- <var>node</var>'s <a>assigned slot</a>.
+ <li>
+  <p>If <var>node</var> is <a for=slottable>assigned</a>, then:
+
+  <ol>
+   <li><p><p>If <var>parent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> with its
```
annevk, line 2679: Drop one `<p>`.

domenic:
@yuzhe-han would you mind responding to @annevk's comments? It'd be good to get this merged; I understand Chrome has shipped this.

yuzhe-han:
@domenic, 
Yes, sorry for the delay. I'll response to @annevk comment Tomorrow.

dom.bs
yuzhe-han: This step, 
   ** If parent’s root is a shadow root, and parent is a slot whose assigned nodes is the empty list, then run signal a slot change for parent. **
   is to signal slot change event because the slot's fallback content has changed. 

For slot assignment "manual", the slot still supports fallback content.

dom.bs
yuzhe-han: I'm not sure how much optimization can be done here. Once the assigned node has been removed from slot's manually assigned node, we need to recalculate the slot's assigned nodes and signal slot change event. Both occurs inside <a>assign slottables</a>.

dom.bs
yuzhe-han: Thanks.  Yes, contains is correct. I see how the verbiage is being used. 
Right, there's can only be one slot because the corresponding HTML spec for HTMLSlotElement::assign() ensures that.

yuzhe-han:
@annevk,

I've update the spec. Please review.  
The build error is due to the definition of "manually assigned nodes" not available in HTML spec yet. 
Would that get resolved once the corresponding HTML spec is merged? Or do I need to do something?

Thanks.

dom.bs
```diff
@@ -2440,7 +2444,8 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
    <li><p>Otherwise, <a for=set>insert</a> <var>node</var> into <var>parent</var>'s
    <a for=tree>children</a> before <var>child</var>'s <a for=tree>index</a>.
 
-   <li><p>If <var>parent</var> is a <a for=Element>shadow host</a> and <var>node</var> is a
+   <li><p>If <var>parent</var> is a <a for=Element>shadow host</a> with its <a for=tree>root</a>'s
```
annevk, line 2447: Don't you mean parent's shadow root here? Parent's root might well be a document or something else.

dom.bs
```diff
@@ -2657,8 +2662,16 @@ indicated in the <a for=/>remove</a> algorithm below.
...
  <li><p><a for=set>Remove</a> <var>node</var> from its <var>parent</var>'s <a for=tree>children</a>.
 
- <li><p>If <var>node</var> is <a for=slottable>assigned</a>, then run <a>assign slottables</a> for
- <var>node</var>'s <a>assigned slot</a>.
+ <li>
+  <p>If <var>node</var> is <a for=slottable>assigned</a>, then:
+
+  <ol>
+   <li><p>If <var>parent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> with its
+   <a for=tree>root</a>'s <a for=ShadowRoot>slot assignment</a> set to "<code>manual</code>",
```
annevk, line 2670: instead of "with its root" you want "whose" I think (and "is" rather than "set to").

dom.bs
annevk: I guess that's fair.

dom.bs
```diff
@@ -2668,6 +2681,11 @@ indicated in the <a for=/>remove</a> algorithm below.
   <p>If <var>node</var> has an <a>inclusive descendant</a> that is a <a>slot</a>, then:
 
   <ol>
+   <li><p>for each <a>slot</a> <var>slot</var> in <var>node</var>'s <a for=tree>inclusive descendants</a>,
+   if <var>slot</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> with its <a for=tree>root</a>'s
+   <a for=ShadowRoot>slot assignment</a> set to "<code>manual</code>", then set <var>slot</var>'s
```
annevk, line 2686: Same comments as above, you want "whose" and "is".

dom.bs
```diff
@@ -2668,6 +2681,11 @@ indicated in the <a for=/>remove</a> algorithm below.
...
+   <a>manually assigned nodes</a> to an empty list.
```
annevk, line 2687: set, right?

annevk:
(And yeah, if manually assigned slots is the only error that will be resolved once the HTML PR merges.)

rniwa:
https://bugs.webkit.org/show_bug.cgi?id=218692

dom.bs
yuzhe-han: Right! Parent's shadow root.  Good to know I had to make this super clear.

dom.bs
yuzhe-han: right.

dom.bs
yuzhe-han: done.

dom.bs
yuzhe-han: done.

dom.bs
```diff
@@ -5701,11 +5719,13 @@ invoked, must return a new {{DocumentFragment}} <a>node</a> whose <a for=Node>no
 [Exposed=Window]
 interface ShadowRoot : DocumentFragment {
   readonly attribute ShadowRootMode mode;
+  readonly attribute SlotAssignmentMode slotAssignment;
   readonly attribute Element host;
   attribute EventHandler onslotchange;
 };
 
 enum ShadowRootMode { "open", "closed" };
+enum SlotAssignmentMode { "manual", "auto" };
```
rniwa, line 5728: Perhaps we want to future-proof this and call "auto" as "name" instead.
That way, we can add more *automatic" mode like based on tag name, etc...

dom.bs
```diff
@@ -2668,6 +2681,11 @@ indicated in the <a for=/>remove</a> algorithm below.
   <p>If <var>node</var> has an <a>inclusive descendant</a> that is a <a>slot</a>, then:
 
   <ol>
+   <li><p>for each <a>slot</a> <var>slot</var> in <var>node</var>'s <a for=tree>inclusive descendants</a>,
+   if <var>slot</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a> whose <a for=tree>root</a>'s
+   <a for=ShadowRoot>slot assignment</a> is "<code>manual</code>", then set <var>slot</var>'s
+   <a>manually assigned nodes</a> to an empty set.
```
rniwa, line 2687: Hm... this would be different from what we've been discussing in AOM:
https://github.com/whatwg/html/issues/4925

dom.bs
rniwa: How can this be right? *assign slottables* would set the list of assigned nodes to slottable matching the slot in tree order. It would mean that we're always re-order slottable in the tree order regardless of in what order they're inserted.

dom.bs
annevk: How does this relate to that?

dom.bs
yuzhe-han: This is a good point. However, I'm not sure "name" is meaningful enough. 
How about "default" -- indicating default slotting?

dom.bs
yuzhe-han: Thanks for catching that. I've updated the find-slottables algorithm to preserve the order of manually assigned nodes.

dom.bs
yuzhe-han: Update spec to use "name"

dom.bs
```diff
@@ -2177,6 +2177,10 @@ steps:</p>
  <li><p>If the <i>open flag</i> is set and <var>shadow</var>'s <a for=ShadowRoot>mode</a> is
  <em>not</em> "<code>open</code>", then return null.</p></li>
 
+ <li><p>If <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> is "<code>manual</code>",
+ then return the <a>slot</a> in <var>shadow</var>'s <a for=tree>descendants</a> whose <a>manually assigned nodes</a>
+ <a for=set>contains</a> <var>slottable</var>, if any, and null otherwise.</p></li>
```
domenic, line 2182: The for=set here is questionable per https://github.com/whatwg/html/pull/5483#issuecomment-784464588

dom.bs
```diff
@@ -2660,6 +2682,11 @@ indicated in the <a for=/>remove</a> algorithm below.
   <p>If <var>node</var> has an <a>inclusive descendant</a> that is a <a>slot</a>, then:
 
   <ol>
+   <li><p>for each <a>slot</a> <var>slot</var> in <var>node</var>'s <a for=tree>inclusive descendants</a>,
```
domenic, line 2685:
```suggestion
   <li><p>For each <a>slot</a> <var>slot</var> in <var>node</var>'s <a for=tree>inclusive descendants</a>,
```

domenic (commented):
This LGTM except for the set vs. list question.

dom.bs
rniwa: Well, it would mean that that this slot.assingedNodes would behave differently from element ID reflection, which seems bad.

rniwa:
Can we make the behavior of manually assigned nodes getting removed from a tree aligned with ID-ref as currently proposed in https://github.com/whatwg/html/pull/3917/files ? Namely, removal doesn't delete the assigned'ness permanently but merely makes it invisible. Once the node is inserted back as a child of the same shadow host, it would reappear.

dom.bs
```diff
@@ -2195,7 +2199,16 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
  <a for=DocumentFragment>host</a>.</p></li>
 
  <li>
-  <p>For each <a>slottable</a> <a for=tree>child</a> of <var>host</var>, <var>slottable</var>, in
+  <p>If <var>slot</var>'s <a for=/>shadow root</a>'s <a for=ShadowRoot>slot assignment</a> is "<code>manual</code>",
+  then:</p>
+
+  <ol>
+   <li><p>Set <var>result</var> to <var>slot</var>'s <a>manually assigned nodes</a>.</p></li>
```
mfreed7, line 2206: Ok, I have the changes made locally, but:
 a) I don't think I have permission to push to this PR, so I'll need to create a fresh one. Is that correct?
 b) It seems that `master` became `main`, and while I support that change, my Git-fu precludes me from being able to get changes. I'm going to pull a fresh fork and see if that helps. I wish git and GitHub were less hostile.

Suggestions for either of the above appreciated.

dom.bs
```diff
@@ -5693,11 +5720,13 @@ invoked, must return a new {{DocumentFragment}} <a>node</a> whose <a for=Node>no
 [Exposed=Window]
 interface ShadowRoot : DocumentFragment {
   readonly attribute ShadowRootMode mode;
+  readonly attribute SlotAssignmentMode slotAssignment;
   readonly attribute Element host;
   attribute EventHandler onslotchange;
 };
 
 enum ShadowRootMode { "open", "closed" };
+enum SlotAssignmentMode { "manual", "name" };
```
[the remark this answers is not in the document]
rniwa: It was intentional because we may want to add a new assignment mode in the future. e.g. you could imagine we'd introduce a new mode where slots are assigned based on each custom element's local name, or maybe we'd add some kind of "brand" to custom element so that custom elements can be slotted based on "brand" so that all subclasses of a custom custom element will be assigned to a given slot, etc...

dom.bs
mfreed7: Alright, I ended up just making a fresh PR for this set of changes. That is located here: https://github.com/whatwg/dom/pull/966/files

I suppose we should close this PR.

dom.bs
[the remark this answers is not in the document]
mfreed7: Ahh ok, that makes sense. Thanks. I'll get the Chromium implementation changed ASAP.

PR #877, opened 2020-06-29
Meta: update repository files

foolip:
See https://github.com/whatwg/spec-factory for details.


foolip:
I've filed https://github.com/whatwg/spec-factory/issues/20 and fixed up this branch manually.

PR #883, opened 2020-07-31
Define TreeNode mixin and add .closest() to more nodes

### LINKED ISSUE #161: Text nodes .closest
Text nodes have no .closest method, when working with Range, I find useful to do `Text.prototype.closest = function(s) {return this.parentNode.closest(s); } //with this.parentNode&& if we want to be strict` 

For me it makes sense to be able to search from a textnode, .matches wouldn't make sense, but .closest would

shvaikalesh:
Fixes #161.

I will submit a WebKit patch as soon as WPT PR is merged and synced.
Given this change adds a nice feature without growing API surface area, I have high hopes of it being landed.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3075,6 +3075,53 @@ steps:
 </ol>
 
 
+<h4 id=interface-treenode>Mixin {{TreeNode}}</h4>
+
+<pre class=idl>
+interface mixin TreeNode {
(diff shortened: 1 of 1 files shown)


domenic:
Can you restore the pull request template, instead of deleting it?

/cc @mfreed7 to speak for Chrome's interest.

Issue #398, opened 2017-01-15
Consider adding localNameFilter to MutationObserver


smaug----:
I was looking at a slow (cross) browser addon code, and one thing they need to do is to figure out when new input elements have been added to the tree.
This means that all the added (and removed) elements are touched, which forces browsers to have
JS wrappers for those and create MutationRecords too.

localNamefilter, similar to attributeFilter would optimize some of that out.
Using css for filtering might be also one option, but it is way more heavy weight in general, so if possible, I'd stick with localNames. (and localNameFilter could be used for text node filtering too, unlike css)

annevk:
Using selectors is #77. How would localNameFilter be used for text nodes?

@dtapuska @rniwa thoughts?

rniwa:
Filtering based on element names sense to me but why not just `elementFilter`?

Also, should this follow the usual case-insensitive in HTML but case-sensitive in XML semantics?

smaug----:
> How would localNameFilter be used for text nodes?

Er, silly me, that would need to be nodeName, not localName. And since attributeFilter uses localName, elementFilter should too.

I don't care too much about the name. elementFilter would be ok to me. But it should be case sensitive, similar to attributeFilter.

annevk:
Case-sensitive `elementFilter` matching on local name sounds fairly straightforward. I could work on specifying that and writing tests, but also happy to hand it to someone else if there's interest since this seems like a nice starter bug.

hemant1612:
I am ready to check for this chance

annevk:
@hemant1612 what you want to do is fork this repository and then [create a pull request from the fork](https://help.github.com/articles/creating-a-pull-request-from-a-fork/). The other thing that will need to be done is writing new web-platform-tests.

zbraniecki:
While `elementFilter` may sound straightforward it also is limiting. Could we consider a `querySelector` instead?

zbraniecki:
> Filtering based on element names sense to me but why not just elementFilter?

My main issue with `elementFilter` is that it solves a subclass of the problem and motivates future additions like `attributeFilter` or combinations for cases where elements and attributes of particular type are expected.

querySelector handle all of that and can be internally optimized by engines for common cases (like elementFilter, attributeFilter etc.) while keeping the "slow case" still faster than doing that in the JS code

rniwa:
We definitely don't want to execute CSS selectors on every DOM mutation. That's simply non-starter.

zbraniecki:
[...]

PR #885, opened 2020-08-12
Add new mutation observer init options elementHasAttributeNameFilter and elementLocalNameFilter

AndrewRot:
Adds additional options when constructing a `MutationObserver`.

MutationObserverInit now contains two additional options: `elementHasAttributeNameFilter` and `elementLocalNameFilter` which are used to enhance the precision of filtering DOM nodes to observe.

See #398 for the background.

   * @rniwa
   * @smaug----
   * … TODO
   * Chrome: …
   * Firefox: …
   * Safari: …


dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
 <ol>
  <li><p>Assert: either <var>addedNodes</var> or <var>removedNodes</var> <a for=set>is not empty</a>.
 
+ <li><p>If either <var>options</var>'s' {{MutationObserverInit/elementLocalNameFilter}} or
+ {{MutationObserverInit/elementHasAttributeNameFilter}} is present, then:
+
+ <ol>
+  <li><p>Let <var>nodeAttributesList</var> be <var>nodeAttributesList</var> of the node's attributes, "<code>node.getAttributeNames()</code>".
```
rniwa, line 3632: This isn't right. node.getAttributeNames returns a list of qualified names.
In general, we don't want to be replying on other DOM APIs to define these algorithmic steps.
Also, node's attribute has a bunch of things: https://dom.spec.whatwg.org/#concept-attribute
What we want here is the local name so either we pick that here, or reference it when we filter it.

dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
...
+   <li>
+     <p>If none of the following are true
+
+     <ul class=brief>
+      <li>if <var>options</var>'s
+      {{MutationObserverInit/elementLocalNameFilter}} is present, and
+      "<code>node.localName</code>" is in {{MutationObserverInit/elementLocalNameFilter}}
+
+      <li><var>options</var>'s {{MutationObserverInit/elementHasAttributeNameFilter}} is
+      present,and <var>options</var>'s  {{MutationObserverInit/elementHasAttributeNameFilter}}
```
rniwa, line 3650: Nit: Need a space after ",".

dom.bs
```diff
@@ -3320,6 +3320,8 @@ dictionary MutationObserverInit {
   boolean attributeOldValue;
   boolean characterDataOldValue;
   sequence&lt;DOMString> attributeFilter;
+  sequence&lt;DOMString> elementHasAttributeNameFilter;
```
rniwa, line 3323: I don't think this option name is great. How about elementFilterByAttribute?

annevk:
@AndrewRot can Shape Security sign https://participate.whatwg.org/agreement? Previously that ended up not happening, but not sure what the reason was.

dom.bs
AndrewRot: Is this correct? Later it is used to check if node's attributes contains any attributes from the `elementFilterByAttribute` list. If `nodeAttributesList` is only the local name of the node, it cannot filter properly. Maybe I'm mistaken?

https://github.com/whatwg/dom/pull/885/files/7eb45a14fff8de0268e64bd2787b8d7f08357c69#diff-a467e681501e456ce8c1ef31425e1b41R3651

AndrewRot: @rniwa - would the following suffice putting it inside of a code block? Or how is it preferred to be written here?
```
<code>
var nodeAttributesList = [];
for (var i = 0; i < element.attributes.length; i ++) { 
    nodeAttributesList.push(element.attributes[i].localName);
}
</code>
```

rniwa: No, these DOM normative text shouldn't be written as JavaScript or invoking other DOM API. We should probably just say Let nodeAttributesList be the set of local names of node's attributes.

AndrewRot: Works for me 👍

AndrewRot:
@rniwa I don't see anything that could be causing this error:

```
LINK ERROR: Multiple possible 'dfn' local refs for 'element'.
```

dom.bs
```diff
@@ -3398,6 +3400,19 @@ dictionary MutationObserverInit {
...
    observed and {{MutationObserverInit/attributes}} is true
    or omitted.
+
+   <dt>{{MutationObserverInit/elementFilterByAttribute}}
+   <dd>Set to a list of <a>attribute</a>
+   <a for=Attr>local names</a> (without <a for=Attr>namespace</a>) to observe changes
+   to the presence of an elements <a>attributes</a>.
+
+   <dt>{{MutationObserverInit/elementLocalNameFilter}}
+   <dd>Set to a list of <a>element</a>
```
annevk, line 3410: You want `<a for=/>element</a>` since there are multiple element definitions.

AndrewRot: > @AndrewRot can Shape Security sign https://participate.whatwg.org/agreement? Previously that ended up not happening, but not sure what the reason was.

I have signed the participation agreement on behalf of [shapesecurity](https://github.com/shapesecurity) however, the Participation below still states: Pending — @AndrewRot participates on behalf of an unverified entity

Is there something that needs to be verified on behalf of whatwg?

dom.bs
```diff
@@ -3320,6 +3320,8 @@ dictionary MutationObserverInit {
   boolean attributeOldValue;
   boolean characterDataOldValue;
   sequence&lt;DOMString> attributeFilter;
+  sequence&lt;DOMString> elementFilterByAttribute;
+  sequence&lt;DOMString> elementLocalNameFilter;
```
annevk, line 3324: Shall we name this `elementFilter` to match `attributeFilter`?

dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
 <ol>
  <li><p>Assert: either <var>addedNodes</var> or <var>removedNodes</var> <a for=set>is not empty</a>.
 
+ <li><p>If either <var>options</var>'s' {{MutationObserverInit/elementLocalNameFilter}} or
+ {{MutationObserverInit/elementFilterByAttribute}} is present, then:
+
+ <ol>
+  <li><p>Let <var>nodeAttributesList</var> be the set of local names of node's attributes.
```
annevk, line 3632: What is node referring to here? I feel like this is lacking some references to other variables.

dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
 <ol>
  <li><p>Assert: either <var>addedNodes</var> or <var>removedNodes</var> <a for=set>is not empty</a>.
 
+ <li><p>If either <var>options</var>'s' {{MutationObserverInit/elementLocalNameFilter}} or
+ {{MutationObserverInit/elementFilterByAttribute}} is present, then:
```
annevk, line 3629: Since this `<p>` isn't the only child of the `<li>`, it needs to be on its own line, indented one. This also goes for the following `<ol>`, which is not indented accurately.

dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
...
+
+ <ol>
+  <li><p>Let <var>nodeAttributesList</var> be the set of local names of node's attributes.
+
+  <li><p>Let <var>filteredAddedNodes</var> be an empty <a for=/>list</a>.
+
+  <li><p>Let <var>filterRemovedNodes</var> be an empty <a for=/>list</a>.
+
+  <li>
+   <p>For each <var>node</var> in <var>addedNodes</var></a>.
```
annevk, line 3639: A for each should end in a `:`.

dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
...
+
+   <li>
```
annevk, line 3641: I would expect a for each to be followed by a nested list so there's an `<ol>` missing here I think.

dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
...
+     <p>If none of the following are true
+
+     <ul class=brief>
+      <li>if <var>options</var>'s
```
annevk, line 3645: s/if// I think? Same below. If not I'd try to reword these to be more clearly true/false statements.

dom.bs
```diff
@@ -3594,6 +3625,60 @@ run these steps:
...
+      {{MutationObserverInit/elementLocalNameFilter}} is present, and
+      "<code>node.localName</code>" is in {{MutationObserverInit/elementLocalNameFilter}}
```
annevk, line 3647: What is node.localName and why is it formatted this way? Is that a special value?

annevk (commented):
Generally this looks great, thanks for working on this! I have left a couple of formatting nits and I'm happy to help with those once we're further along.

We'll need a lot of tests to cover the various conditions and new behavior of `observe()`.

I'm a little concerned about the agreement. Typically a person in a management or a very senior engineer would sign it (or at least be an alternate), but neither appears to be the case here, assuming I have interpreted your title correctly. As it seems this is a very big company, have legal et al signed off on this?

annevk:
@mfreed7 any thoughts from Google on this?

dom.bs
annevk: Also, @smaug---- suggested that maybe all filters should end with `Filter`, which seems reasonable. That would suggest `elementByAttributeFilter` and `elementFilter` to me.

rniwa: That sounds like a reasonable suggestion.

rniwa:
FWIW, we're supportive of this proposal.

AndrewRot:
@annevk yes - I've gotten written approval to sign off on the agreement from our legal department as well as my team managers.

dom.bs
```diff
@@ -3594,6 +3625,67 @@ run these steps:
...
+   <li><p>Let <var>filteredAddedNodes</var> be an empty <a for=/>list</a>.
+
+   <li><p>Let <var>filterRemovedNodes</var> be an empty <a for=/>list</a>.
+
+   <li>
+    <p>For each <var>node</var> in <var>addedNodes</var></a>:
+
+    <ol>
+     <li><p>Let <var>nodeAttributesList</var> be the set of local names of <var>node</var>'s
+     attributes.
```
annevk, line 3642: "local names" and "attributes" should xref the relevant definitions. (We should also have a test with two attributes that have the same local name and see if this is all working as designed.)

dom.bs
```diff
@@ -3594,6 +3625,67 @@ run these steps:
...
+
+     <li>
+      <p>If none of the following are true
+
+      <ul class=brief>
+       <li>if <var>options</var>'s {{MutationObserverInit/elementFilter}} is present, and the
```
annevk, line 3648: I don't think the nested "if" works here. This needs rephrasing.

dom.bs
```diff
@@ -3594,6 +3625,67 @@ run these steps:
...
+       <a>node</a>'s <var>localName</var> is in {{MutationObserverInit/elementFilter}}
```
annevk, line 3649: This is quite confusing. Isn't node the variable and local name the field?

dom.bs
```diff
@@ -3594,6 +3625,67 @@ run these steps:
...
+
+       <li><var>options</var>'s {{MutationObserverInit/elementByAttributeFilter}} is present, and
+       <var>options</var>'s {{MutationObserverInit/elementByAttributeFilter}} contains any of the
+       attribute name in <var>nodeAttributesList</var>
+      </ul>
+
+      <p>then <a for=queue>enqueue</a> <var>record</var> node in <var>filteredAddedNodes</var>.
```
annevk, line 3656: I think you might mean `<a for=list>append</a> <var>node</var> to <var>filteredAddedNodes</var>` here?

annevk (commented):
Thank you for confirming the IPR stuff, that's great! There's a couple things remaining with the PR (note that several comments apply to multiple places) and the other thing we'll need here are tests. Do you think you can work on those?

AndrewRot:
@annevk looks like agreement was filled out. Though I just added my manager as a [secondary contact](https://github.com/whatwg/participant-data/pull/40).

All comments appear to have been addressed. At this point, the only thing missing seems to be the tests? I can start on those.

annevk:
@AndrewRot that sounds great, thank you! Once the tests are there I'm happy to give this another pass.

bershanskiy:
Has there been any progress on the tests? I'm working on a product which would benefit from this Mutation Observer filter feature, and might be able to assist with tests if there is interest.

annevk:
@mfreed7 ping on Chromium interest. @smaug---- Gecko?

@bershanskiy I'd be happy to pick up the remaining specification work assuming @AndrewRot no longer has the bandwidth if you can sort the tests. Thanks for volunteering!

AndrewRot:
Hi @bershanskiy & @annevk - I'm glad this feature is useful for others! Yes, feel free to take the reins from here. I do not currently have the bandwidth to allocate for this feature.

dom.bs
```diff
@@ -3635,6 +3666,67 @@ method steps are:
...
+       <li><var>options</var>'s {{MutationObserverInit/elementByAttributeFilter}} is present, and
+       <var>options</var>'s {{MutationObserverInit/elementByAttributeFilter}} contains any of the
+       attribute names in <var>nodeAttributesList</var>
+      </ul>
+
+      <p>then <a for=list>append</a> <var>node</var> to <var>filterRemovedNodes</var>.
+    </ol>
+
+   <li><p>Assert: either <var>filteredAddedNodes</var> or <var>filteredRemovedNodes</var>
+   <a for=set>is not empty</a>.
```
smaug----, line 3723: How can we assert this?

dom.bs
```diff
@@ -3635,6 +3666,67 @@ method steps are:
...
+      <ul class=brief>
+       <li><var>options</var>'s {{MutationObserverInit/elementFilter}} is present, and the
+       <a>node</a>'s <var>localName</var> is in {{MutationObserverInit/elementFilter}}
+
+       <li><var>options</var>'s {{MutationObserverInit/elementByAttributeFilter}} is present, and
+       <var>options</var>'s {{MutationObserverInit/elementByAttributeFilter}} contains any of the
+       attribute names in <var>nodeAttributesList</var>
+      </ul>
+
+      <p>then <a for=list>append</a> <var>node</var> to <var>filterRemovedNodes</var>.
```
smaug----, line 3719: I must be confused here.
Don't we want something like
if elementFilter is present, and elementByAttributeFilter is not present and localName is in elementfilter, or if 
elementByAttributeFilter is present and elementFilter is not present and elementByAttributeFilter contains any of the attribute names in nodeAttribute list ...  or combination of those two

smaug----:
Gecko would implement this or some variant of it.
(which isn't super surprising since looks like I filed https://github.com/whatwg/dom/issues/398 )

mfreed7:
> @mfreed7 ping on Chromium interest. @smaug---- Gecko?
> 
> @bershanskiy I'd be happy to pick up the remaining specification work assuming @AndrewRot no longer has the bandwidth if you can sort the tests. Thanks for volunteering!

Looks like a good addition to `MutationObserver` which would avoid boilerplate JS and speed things up a lot. I'm supportive!

dom.bs
AndrewRot: Both of these are now defined in scope. Check if filteredAddedNodes length > 0 or filteredRemovedNodes length > 0.

dom.bs
```diff
@@ -3442,6 +3444,19 @@ dictionary MutationObserverInit {
    <a>attribute</a> mutations need to be
    observed and {{MutationObserverInit/attributes}} is true
    or omitted.
+
+   <dt>{{MutationObserverInit/elementByAttributeFilter}}
+   <dd>Set to a list of <a>attribute</a>
+   <a for=Attr>local names</a> (without <a for=Attr>namespace</a>) to observe changes
+   to the presence of an elements <a>attributes</a>.
```
LeaVerou, line 3451:
```suggestion
   to the presence of an element's <a>attributes</a>.
```

PR #890, opened 2020-09-09
DOM Events introduction: clarify how listeners are invoked

### LINKED ISSUE #888: "When an event is dispatched to an object that p..."
https://dom.spec.whatwg.org/commit-snapshots/8f20138c9510c7f4a7d01dbb5e24175e72fc9734/#synthetic-events

> When an event is dispatched to an object that participates in a tree \(e\.g\. an element\), it can reach event listeners on that object’s ancestors too\. First all object’s ancestor event listeners whose capture variable is set to true are invoked, in tree order\. Second, object’s own event listeners are invoked\. And finally, and only if event’s bubbles attribute value is true, object’s ancestor event listeners are invoked again, but now in reverse tree order\.
> 
> Let’s look at an example of how events work in a tree:​
> 
> &lt;\!doctype html&gt;
> &lt;html&gt;
>  &lt;head&gt;
>   &lt;title&gt;Boring example&lt;/title&gt;
>  &lt;/head&gt;
>  &lt;body&gt;
>   &lt;p&gt;Hello &lt;span id=​x&gt;world&lt;/span&gt;\!&lt;/p&gt;
>   &lt;script&gt;
>    function test\(e\) \{
>      debug\(e\.target, e\.currentTarget, e\.eventPhase\)
>    \}
>    document\.addEventListener\("hey", test, \{capture:​ true\}\)
>    document\.body\.addEventListener\("hey", test\)
>    var ev =​ new Event\("hey", \{bubbles:​true\}\)
>    document\.getElementById\("x"\)\.dispatchEvent\(ev\)
>   &lt;/script&gt;
>  &lt;/body&gt;
> &lt;/html&gt;
> 
> The debug function will be invoked twice. Each time the event’s target attribute value will be the span element. The first time currentTarget attribute’s value will be the document, the second time the body element. eventPhase attribute’s value switches from CAPTURING_PHASE to BUBBLING_PHASE. If an event listener was registered for the span element, eventPhase attribute’s value would have been AT_TARGET.

I find this hard to follow, so I think it could use clarification (or perhaps I am just missing something). Following the first paragraph, this is what I would expect to happen in the example:
1. > First all object’s ancestor event listeners whose capture variable is set to true are invoked, in tree order.

The object (the span element) has exactly one ancestor event listener with capture variable set to true (the document). So that listener is invoked.

2. > Second, object’s own event listeners are invoked\.

The object (the span element) has no event listeners, so nothing is invoked.

3. > And finally, and only if event’s bubbles attribute value is true, object’s ancestor event listeners are invoked again, but now in reverse tree order\.

The wording here is unclear as to whether the invoked ancestor event listeners must have capture variable set to true. The use of the word "again" suggests that this should be so. In either case, since the event listeners are to be invoked "again," I expect the event listener of the document to be invoked again. Instead, only the body element has its event listener invoked, even though it was not invoked during the capturing phase.

annevk:
Fixes #888.

***
<a href="https://whatpr.org/dom/890.html" title="Last updated on Sep 9, 2020, 10:35 AM UTC (cd802c8)">Preview</a> | <a href="https://whatpr.org/dom/890/8f20138...cd802c8.html" title="Last updated on Sep 9, 2020, 10:35 AM UTC (cd802c8)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1,4 +1,4 @@
-<pre class='metadata'>
+<pre class=metadata>
 Group: WHATWG
 H1: DOM
 Shortname: dom
@@ -358,34 +358,24 @@ use of this functionality through <a>events</a>
 (synthetic events) can make use of the return value of the
 {{EventTarget/dispatchEvent()}} method:
 
-<pre class='lang-javascript'>
+<pre class=lang-javascript>
 if(obj.dispatchEvent(event)) {
   // event was not canceled, time for some magic
   &hellip;
 }
 </pre>
 
-When an <a>event</a> is
-<a>dispatched</a> to an object that
-<a>participates</a> in a
-<a>tree</a> (e.g. an
-<a for="/">element</a>), it can reach
-<a>event listeners</a> on that object's
-<a>ancestors</a> too. First all object's
-<a>ancestor</a>
-<a>event listeners</a> whose
-<a for="event listener">capture</a> variable is set to true are invoked, in
-<a>tree order</a>. Second, object's own
-<a>event listeners</a> are invoked. And
-finally, and only if <a>event</a>'s
-{{Event/bubbles}} attribute value is true,
-object's  <a>ancestor</a>
-<a>event listeners</a> are invoked again,
-but now in reverse <a>tree order</a>.
+<p>When an <a>event</a> is <a>dispatched</a> to an object that <a>participates</a> in a <a>tree</a>
+(e.g., an <a for=/>element</a>), it can reach <a>event listeners</a> on that object's
+<a>ancestors</a> too. Effectively, all the object's <a>inclusive ancestor</a> <a>event listeners</a>
+whose <a for="event listener">capture</a> is true are invoked, in <a>tree order</a>. And then, if
+<a>event</a>'s {{Event/bubbles}} is true, all the object's <a>inclusive ancestor</a>
+<a>event listeners</a> whose <a for="event listener">capture</a> is false are invoked, now in
+reverse <a>tree order</a>.
 
-Let's look at an example of how <a>events</a> work in a <a>tree</a>:
+<p>Let's look at an example of how <a>events</a> work in a <a>tree</a>:
 
-<pre class='lang-markup'>
+<pre class=lang-markup>
 &lt;!doctype html>
 &lt;html>
  &lt;head>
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -358,34 +358,24 @@ use of this functionality through <a>events</a>
...
-object's  <a>ancestor</a>
-<a>event listeners</a> are invoked again,
-but now in reverse <a>tree order</a>.
-
-Let's look at an example of how <a>events</a> work in a <a>tree</a>:
-
-<pre class='lang-markup'>
+<p>When an <a>event</a> is <a>dispatched</a> to an object that <a>participates</a> in a <a>tree</a>
+(e.g., an <a for=/>element</a>), it can reach <a>event listeners</a> on that object's
+<a>ancestors</a> too. Effectively, all the object's <a>inclusive ancestor</a> <a>event listeners</a>
```
kevinsung, line 370: The use of the word "Effectively" leaves ambiguity as to what the true behavior is. If there is in fact no ambiguity, then I would not use that word; I would replace it with "Firstly" or "In particular" or something like that. If there is in fact ambiguity (I do not know) then it's fine to leave it, though a note of clarification may be helpful.

annevk: It's an approximation as shadow trees and the `Window` object make it more involved than this. Not sure how to capture that succinctly though.

kevinsung: Ok, that's fine then.

kevinsung (approved):
I can vouch for the clarity of the wording and its consistency with the example, but not its correctness.

sideshowbarker (approved):
General comment (relevant to @kevinsung’s specific comment about “effectively”): To avoid ambiguity about what statements in the DOM spec are normative and which aren’t, it seems like it would be useful for non-normative sections of the published DOM spec to be rendered with explicit *This section is non-normative* text after the section headings — in the way the HTML spec is. But I guess more generally, that’s an issue for Bikeshed (in that, Bikeshed currently doesn’t have a feature for automatically emitting explicit *This section is non-normative* text in the way that the HTML spec has.

annevk:
I'll leave that for a follow-up as that might require a more thorough review of the remainder of the specification.

annevk:
Thanks again for your help in this @kevinsung!

PR #892, opened 2020-09-15
Add declarative Shadow DOM features

mfreed7:
The explainer for this feature is here: https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md
The issue discussion is here: https://github.com/whatwg/dom/issues/831
There is a corresponding [Pull Request](https://github.com/whatwg/html/pull/5465) for the HTML spec that goes along with this PR.

Note, due to some local branch issues, this is a duplicate of [PR 858](https://github.com/whatwg/dom/pull/858). Please see that PR for more history and review comments here. In particular, the last comment there [is from me](https://github.com/whatwg/dom/pull/858#issuecomment-668875565) - it is my belief that I've addressed all of the outstanding comments.

   * I'm working to get this implemented in Chromium, tracking bug [here](https://crbug.com/1042130).
   * Other implementers have at least not been negative on the [discussion thread](https://github.com/whatwg/dom/issues/831). Not explicitly positive (yet).

   * https://wpt.fyi/results/shadow-dom/declarative?label=master&label=experimental&aligned&q=declarative

   * Chrome: https://crbug.com/1042130
   * Firefox: None yet
   * Safari: None yet

Closes https://github.com/whatwg/dom/issues/1137.

***
<a href="https://whatpr.org/dom/892.html" title="Last updated on Oct 4, 2023, 9:17 AM UTC (17ede93)">Preview</a> | <a href="https://whatpr.org/dom/892/3bddf91...17ede93.html" title="Last updated on Oct 4, 2023, 9:17 AM UTC (17ede93)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5005,8 +5005,8 @@ known as <dfn export id=concept-document lt="document">documents</dfn>.
 <a for=/>encoding</a>, <a for=Document>content type</a> is
 "<code>application/xml</code>", <a for=Document>URL</a> is "<code>about:blank</code>",
 <a for=Document>origin</a> is an <a>opaque origin</a>,
-<a for=Document>type</a> is "<code>xml</code>", <a for=Document>mode</a> is "<code>no-quirks</code>",
-and <a for=Document>allow declarative shadow roots</a> is false.
+<a for=Document>type</a> is "<code>xml</code>", <a for=Document>mode</a> is
+"<code>no-quirks</code>", and its <a for=Document>allow declarative shadow roots</a> is false.
 
 <p>A <a>document</a> is said to be an <dfn export>XML document</dfn> if its <a for=Document>type</a>
 is "<code>xml</code>"; otherwise an <dfn export>HTML document</dfn>. Whether a <a>document</a> is an
@@ -6890,9 +6890,10 @@ are:
 <p>The <dfn method for=Element><code>attachShadow(<var>init</var>)</code></dfn> method steps are:
 
 <ol>
- <li><p>Run <a>attach a shadow root</a> with <a>this</a>, <var>init</var>["{{ShadowRootInit/mode}}"],
- <var>init</var>["{{ShadowRootInit/clonable}}"], <var>init</var>["{{ShadowRootInit/delegatesFocus}}"],
- and <var>init</var>["{{ShadowRootInit/slotAssignment}}"].
+ <li><p>Run <a>attach a shadow root</a> with <a>this</a>,
+ <var>init</var>["{{ShadowRootInit/mode}}"], <var>init</var>["{{ShadowRootInit/clonable}}"],
+ <var>init</var>["{{ShadowRootInit/delegatesFocus}}"], and
+ <var>init</var>["{{ShadowRootInit/slotAssignment}}"].
 
  <li><p>Return <a>this</a>'s <a for=Element>shadow root</a>.
 </ol>
@@ -6907,8 +6908,8 @@ a boolean <var>delegatesFocus</var>, and a boolean <var>slotAssignment</var>:
  <li><p>If <var>element</var>'s <a for=Element>namespace</a> is not the <a>HTML namespace</a>,
  then <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
 
- <li><p>If <a>this</a>'s <a for=Element>local name</a> is not a <a>valid shadow host name</a>, then
- <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
+ <li><p>If <var>element</var>'s is not a <a>valid shadow host name</a>, then <a>throw</a> a
+ "{{NotSupportedError!!exception}}" {{DOMException}}.
 
  <li>
   <p>If <var>element</var>'s <a for=Element>local name</a> is a <a>valid custom element name</a>, or
@@ -6933,11 +6934,12 @@ a boolean <var>delegatesFocus</var>, and a boolean <var>slotAssignment</var>:
    <li><p>If <var>element</var>'s <a for=Element>shadow root</a>'s <a for=ShadowRoot>declarative</a>
    is false, then <a>throw</a> an "{{NotSupportedError!!exception}}" {{DOMException}}.
 
-   <li><p>Otherwise, <a for=/>remove</a> all of <var>element</var>'s
-   <a for=Element>shadow root</a>'s <a>children</a>, in <a>tree order</a>, and return.
+   <li>
+    <p>Otherwise, <a for=/>remove</a> all of <var>element</var>'s <a for=Element>shadow root</a>'s
+    <a>children</a>, in <a>tree order</a>, and return.
 
-   <p class=note>This means that if multiple declarative shadow roots are contained within a single
-   shadow host, only the last one will remain.
+    <p class=note>This means that if multiple declarative shadow roots are contained within a single
+    shadow host, only the last one will remain.
   </ol>
 
  <li><p>Let <var>shadow</var> be a new <a for=/>shadow root</a> whose <a for=Node>node document</a>
@@ -6946,8 +6948,8 @@ a boolean <var>delegatesFocus</var>, and a boolean <var>slotAssignment</var>:
 
  <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>delegates focus</a> to <var>delegatesFocus</var>.
 
- <li><p>If <a>this</a>'s <a for=Element>custom element state</a> is "<code>precustomized</code>" or
- "<code>custom</code>", then set <var>shadow</var>'s
+ <li><p>If <var>element</var>'s <a for=Element>custom element state</a> is
+ "<code>precustomized</code>" or "<code>custom</code>", then set <var>shadow</var>'s
  <a for=ShadowRoot>available to element internals</a> to true.
 
  <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> to <var>slotAssignment</var>.
--- a/dom.bs
+++ b/dom.bs
@@ -6936,7 +6936,7 @@ a boolean <var>delegatesFocus</var>, and a boolean <var>slotAssignment</var>:
 
    <li>
     <p>Otherwise, <a for=/>remove</a> all of <var>element</var>'s <a for=Element>shadow root</a>'s
-    <a>children</a>, in <a>tree order</a>, and return.
+    <a for=tree>children</a>, in <a>tree order</a>, and return.
 
     <p class=note>This means that if multiple declarative shadow roots are contained within a single
     shadow host, only the last one will remain.
--- a/dom.bs
+++ b/dom.bs
@@ -4465,15 +4465,36 @@ dom-Range-extractContents, dom-Range-cloneContents -->
  <var>copy</var>, if <var>copy</var> is a <a>document</a>, and set <var>copy</var>'s
  <a for=Node>node document</a> to <var>document</var> otherwise.
 
- <li>Run any <a>cloning steps</a> defined for <var>node</var> in
+ <li><p>Run any <a>cloning steps</a> defined for <var>node</var> in
  <a>other applicable specifications</a> and pass <var>copy</var>, <var>node</var>,
- <var>document</var> and the <i>clone children flag</i> if set, as parameters.
+ <var>document</var>, and the <i>clone children flag</i> if set, as parameters.
 
- <li>If the <i>clone children flag</i> is set, <a lt="clone a node">clone</a> all the
- <a for=tree>children</a> of <var>node</var> and append them to <var>copy</var>, with
- <var>document</var> as specified and the <i>clone children flag</i> being set.
+ <li><p>If the <i>clone children flag</i> is set, then for each <a for=tree>child</a>
+ <var>child</var> of <var>node</var>, in <a>tree order</a>: <a>append</a> the result of
+ <a lt="clone a node">cloning</a> <var>child</var> with <var>document</var> and the
+ <i>clone children flag</i> set, to <var>copy</var>.
 
- <li>Return <var>copy</var>.
+ <li>
+  <p>If <var>node</var> is a <a for=Element>shadow host</a> whose <a for=/>shadow root</a>'s
+  <a for=ShadowRoot>clonable</a> is true:
+
+  <ol>
+   <li><p>Run <a>attach a shadow root</a> with <var>copy</var>, <var>node</var>'s
+   <a for=Element>shadow root</a>'s <a for=ShadowRoot>mode</a>, true, <var>node</var>'s
+   <a for=Element>shadow root</a>'s <a for=ShadowRoot>delegates focus</a>, and <var>node</var>'s
+   <a for=Element>shadow root</a>'s <a for=ShadowRoot>slot assignment</a>.
+
+   <li><p>Set <var>copy</var>'s <a for=Element>shadow root</a>'s <a for=ShadowRoot>declarative</a>
+   to <var>node</var>'s <a for=Element>shadow root</a>'s <a for=ShadowRoot>declarative</a>.
+
+   <li><p>If the <i>clone children flag</i> is set, then for each <a for=tree>child</a>
+   <var>child</var> of <var>node</var>'s <a for=Element>shadow root</a>, in <a>tree order</a>:
+   <a>append</a> the result of <a lt="clone a node">cloning</a> <var>child</var> with
+   <var>document</var> and the <i>clone children flag</i> set, to <var>copy</var>'s
+   <a for=Element>shadow root</a>.
+  </ol>
+
+ <li><p>Return <var>copy</var>.
 </ol>
 
 <p>The <dfn method for=Node><code>cloneNode(<var>deep</var>)</code></dfn> method steps are:
@@ -4973,8 +4994,9 @@ known as <dfn export id=concept-document lt="document">documents</dfn>.
 <dfn export for=Document id=concept-document-content-type>content type</dfn> (a string),
 <dfn export for=Document id=concept-document-url>URL</dfn> (a <a for=/>URL</a>),
 <dfn export for=Document id=concept-document-origin>origin</dfn> (an <a for=/>origin</a>),
-<dfn export for=Document id=concept-document-type>type</dfn> ("<code>xml</code>" or "<code>html</code>"), and
-<dfn export for=Document id=concept-document-mode>mode</dfn> ("<code>no-quirks</code>", "<code>quirks</code>", or "<code>limited-quirks</code>").
+<dfn export for=Document id=concept-document-type>type</dfn> ("<code>xml</code>" or "<code>html</code>"),
+<dfn export for=Document id=concept-document-mode>mode</dfn> ("<code>no-quirks</code>", "<code>quirks</code>", or "<code>limited-quirks</code>"), and
+<dfn export for=Document>allow declarative shadow roots</dfn> (a boolean).
 [[!ENCODING]]
 [[!URL]]
 [[!HTML]]
@@ -4983,8 +5005,8 @@ known as <dfn export id=concept-document lt="document">documents</dfn>.
 <a for=/>encoding</a>, <a for=Document>content type</a> is
 "<code>application/xml</code>", <a for=Document>URL</a> is "<code>about:blank</code>",
 <a for=Document>origin</a> is an <a>opaque origin</a>,
-<a for=Document>type</a> is "<code>xml</code>", and its
-<a for=Document>mode</a> is "<code>no-quirks</code>".
+<a for=Document>type</a> is "<code>xml</code>", <a for=Document>mode</a> is "<code>no-quirks</code>",
+and <a for=Document>allow declarative shadow roots</a> is false.
 
 <p>A <a>document</a> is said to be an <dfn export>XML document</dfn> if its <a for=Document>type</a>
 is "<code>xml</code>"; otherwise an <dfn export>HTML document</dfn>. Whether a <a>document</a> is an
@@ -5857,12 +5879,18 @@ It is initially set to false.</p>
 <p><a for=/>Shadow roots</a> have an associated
 <dfn export for=ShadowRoot>available to element internals</dfn>. It is initially set to false.</p>
 
+<p><a for=/>Shadow roots</a> have an associated <dfn export for=ShadowRoot>declarative</dfn>
+(a boolean). It is initially set to false.</p>
+
 <p><a for=/>Shadow roots</a>'s associated <a for=DocumentFragment>host</a> is never null.</p>
 <!-- If we ever change this, e.g., add a ShadowRoot object constructor, that would have serious
      consequences for innerHTML. -->
 
 <p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>slot assignment</dfn>
 ("<code>manual</code>" or "<code>named</code>").
+
+<p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>clonable</dfn> (a boolean).
+It is initially set to false.</p>
 
 <p>A <a for=/>shadow root</a>'s <a>get the parent</a> algorithm, given an <var>event</var>, returns
 null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow root</a> is the
@@ -6010,6 +6038,7 @@ dictionary ShadowRootInit {
   required ShadowRootMode mode;
   boolean delegatesFocus = false;
   SlotAssignmentMode slotAssignment = "named";
+  boolean clonable = false;
 };
 </pre>
 
@@ -6857,23 +6886,38 @@ are:
  "<code>span</code>"
 </ul>
 
+<div algorithm>
 <p>The <dfn method for=Element><code>attachShadow(<var>init</var>)</code></dfn> method steps are:
 
 <ol>
- <li><p>If <a>this</a>'s <a for=Element>namespace</a> is not the <a>HTML namespace</a>,
+ <li><p>Run <a>attach a shadow root</a> with <a>this</a>, <var>init</var>["{{ShadowRootInit/mode}}"],
+ <var>init</var>["{{ShadowRootInit/clonable}}"], <var>init</var>["{{ShadowRootInit/delegatesFocus}}"],
+ and <var>init</var>["{{ShadowRootInit/slotAssignment}}"].
+
+ <li><p>Return <a>this</a>'s <a for=Element>shadow root</a>.
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn id=concept-attach-a-shadow-root>attach a shadow root</dfn>, given an
+<a for=/>element</a> <var>element</var>, a string <var>mode</var>, a boolean <var>clonable</var>,
+a boolean <var>delegatesFocus</var>, and a boolean <var>slotAssignment</var>:
+
+<ol>
+ <li><p>If <var>element</var>'s <a for=Element>namespace</a> is not the <a>HTML namespace</a>,
  then <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
 
  <li><p>If <a>this</a>'s <a for=Element>local name</a> is not a <a>valid shadow host name</a>, then
  <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
 
  <li>
-  <p>If <a>this</a>'s <a for=Element>local name</a> is a <a>valid custom element name</a>, or
-  <a>this</a>'s <a for=Element><code>is</code> value</a> is not null, then:
+  <p>If <var>element</var>'s <a for=Element>local name</a> is a <a>valid custom element name</a>, or
+  <var>element</var>'s <a for=Element><code>is</code> value</a> is non-null, then:
 
   <ol>
    <li><p>Let <var>definition</var> be the result of
    <a lt="look up a custom element definition">looking up a custom element definition</a> given
-   <a>this</a>'s <a for=Node>node document</a>, its <a for=Element>namespace</a>, its
+   <var>element</var>'s <a for=Node>node document</a>, its <a for=Element>namespace</a>, its
    <a for=Element>local name</a>, and its <a for=Element><code>is</code> value</a>.
 
    <li><p>If <var>definition</var> is not null and <var>definition</var>'s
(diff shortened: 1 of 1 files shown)


mfreed7:
This PR has been updated to include the "opt-in" mechanics described in [#912](https://github.com/whatwg/dom/issues/912). The changes I made here roughly match the current implementation of Chromium, so I'm hoping they're functional. But comments appreciated.

Here, I am unable to figure out how to get the "magic" `{{include shadow roots state}}` to link over to the equivalent section of HTML. I get a build error there. I cannot find where in dom.bs these strings are defined.

dom.bs
```diff
@@ -6835,9 +6887,15 @@ for <a>this</a>.
  <dd><p><a>Throw</a> a "{{SyntaxError!!exception}}" {{DOMException}}.
 </dl>
 
+<p>The
+<dfn method for=Element><code>getInnerHTML(<var>options</var>)</code></dfn>
+method, when invoked, must return the result of running <a>HTML fragment serialization algorithm</a>,
+given <a>this</a> as <var>node</var>, <i>options.includeShadowRoots</i> as <i>include shadow roots</i>,
+and <i>options.closedRoots</i> as <i>closed shadow roots</i>.
```
annevk, line 6894: If we are to add this feature, I suspect we want to define this in HTML instead, @domenic?

domenic: Yeah, the plan is to move https://w3c.github.io/DOM-Parsing/#the-innerhtml-mixin into HTML, not DOM. One day...

annevk:
As for the build error, once HTML is indexed by Shepherd you should be able to cross-reference, but that does mean broken builds until that happens unfortunately. We don't have a good workaround. (One of the nice things about Bikeshed is that it doesn't need a full local index of terms defined elsewhere.)

dom.bs
mfreed7: So just to be clear here - you'd prefer I move this definition (lines 6890-6894) to the HTML spec, but leave the IDL that defines `getInnerHTML` here in DOM. Is that right? Would you mind pointing me to another example that auto-magically links over to HTML from the IDL like this, so I can copy it? I looked through the [IDL Index](https://dom.spec.whatwg.org/#idl-index) but didn't find any such examples. I'm sure I missed them.

mfreed7:
> As for the build error, once HTML is indexed by Shepherd you should be able to cross-reference, but that does mean broken builds until that happens unfortunately. We don't have a good workaround. (One of the nice things about Bikeshed is that it doesn't need a full local index of terms defined elsewhere.)

Ahh ok, that makes sense. Thanks.

dom.bs
annevk: No, we should also move the IDL. HTML will define all of `innerHTML`, including that.

mfreed7: So as part of this PR (adding declarative Shadow DOM), you also want me to move all of the Element IDL out of here and into HTML? I suppose I can try doing that, but it seems like perhaps that should be a separate PR? This one is already fairly large, and is moving along pretty slowly. I would hate to slow it down further. Having said that, I'll do it if you think it's necessary here.

annevk: My thought was that HTML would have `partial interface Element {}` to extend this interface, but having said that, that's also a thing we're considering to stop doing. Hmm. (We definitely don't want to move all of `Element`.)

domenic: `partial interface Element {}` is what I anticipated too. I don't think we're going to stop using partial interfaces any time soon (for `Document` alone, that would basically require merging the HTML and DOM specs). That also matches how https://w3c.github.io/DOM-Parsing/#extensions-to-the-element-interface works.

dom.bs
```diff
@@ -5828,6 +5855,11 @@ dictionary ShadowRootInit {
   required ShadowRootMode mode;
   boolean delegatesFocus = false;
 };
+
+dictionary GetInnerHTMLOptions {
+  boolean includeShadowRoots = true;
+  sequence&lt;ShadowRoot> closedRoots;
```
rniwa, line 5861: I continue to find this API shape problematic. We really want to have a consistent API surface. Given the [current consensus](https://github.com/WICG/webcomponents/issues/79#issuecomment-432974389) for `getComposedRange` for `Selection` is that it takes a list of shadow roots regardless of their closed-ness, we should match that convention and always take a list of shadow roots here as well.

mfreed7: Hmm, I only have the context from the issue you linked, and not directly, but I assume you're referring to this:

> Rough consensus at TPAC F2F:
> 
> * Add `getComposedRange`, which takes a list of closed trees, to [`Selection`](https://w3c.github.io/selection-api/#dom-selection).

I can't help but notice the "which takes a list of **closed trees**" there...? Also, that thread seems dead for 2.5 years. Perhaps **this** could be the consistent API surface, which selection tries to match, once the discussion restarts?

rniwa: To match the selection API, this should be instead instead called getComposedInnerHTML and simply take an array of closed shadow roots.

I really don't think we should add this one off API before we can reach a better consensus on what the general approach on how to define all these APIs for the composed tree & flat tree. We need to stop introducing these one-off APIs with slightly different semantics. We've done enough damages to the web platform already. We need to stop.

Beyond that, I have some issues with the semantics of how serialization is proposed. I'd add a new comment on https://github.com/whatwg/html/pull/5465/files.

mfreed7: Sorry - can you link me to the spec for `getComposedInnerHTML`? I can't seem to find that one. Perhaps you're just referring to the [unimplemented 2018 proposal](https://github.com/WICG/webcomponents/issues/79#issuecomment-432974389) for `getComposedRange`?

This seems different from that. The word "composed" I'd even object to here. We're just trying to return the full HTML for a node, including shadow content. Since we can't break the web by starting to do that with `innerHTML`, we need a new API. The closest thing to the existing `innerHTML` seems to be `getInnerHTML()` which allows the introduction of parameters, and is easily understandable by developers. Renaming this to `getComposedInnerHTML` will make everyone go look up "composed" to try to see if this'll just return them the HTML, which it will. So we make it longer to type, and more confusing to developers. Can you help me understand what we gain in the bargain?

Thanks for your comments on https://github.com/whatwg/html/pull/5465/files. I've put some replies there.

rniwa: So that people don't accidentally use this variant instead of innerHTML. We really shouldn't be encouraging people to serialize contents of shadow roots. Just like `Element.prototype.shadowRoot` should only be used when it's absolutely necessary, this variant of innerHTML should only be used when it's absolutely necessary.

annevk: I had also raised this in https://github.com/mfreed7/declarative-shadow-dom/issues/9.

mfreed7: Please see my comment in https://github.com/mfreed7/declarative-shadow-dom/issues/9#issuecomment-809793894. I've checked with the early implementers (e.g. Origin Trial participants) and they would be ok with these changes. But I don't want to allow too much time for people to start making use of the `getInnerHTML()` API such that we cause compat issues by changing the API shape.

vekunz:
Is there still progress here or is the topic dead?

mfreed7:
> Is there still progress here or is the topic dead?

This feature is **fully shipped** on chromium browsers. The other implementers are waiting for production performance data, to help convince them that SSR (and DSD) offers actual performance benefits on real sites. That’s despite [recent blog posts](https://webvision.mozilla.org/full/) that seem to have a different view:

> We also continue to see room for improving or adding primitives that provide declarative alternatives for situations that currently require JavaScript, such as…

This feature is not dead.

annevk:
@mfreed7 I think that's a mischaracterization of Mozilla's views and seems somewhat unwarranted? We've been plenty supportive of exploring declarative web components. We're just not quite convinced of declarative shadow trees in isolation.

mfreed7:
> @mfreed7 I think that's a mischaracterization of Mozilla's views and seems somewhat unwarranted? We've been plenty supportive of exploring declarative web components. We're just not quite convinced of declarative shadow trees in isolation.

I’m sorry you feel that way! I honestly didn’t mean to mischaracterize anything. (The link to your blog post might not have been fair, I suppose, though it **is** a very public blog post.) But I’ve been working for ~a year trying to gather production performance data on this feature precisely because I thought this was Mozilla’s view. E.g. see the latest comment from @smaug---- on [the issue thread](https://github.com/whatwg/dom/issues/831#issuecomment-988394185), which is talking about performance concerns related to (declarative/all) shadow roots. Since declarative custom elements (which are vaporware at this point) would almost surely build on top of declarative shadow dom, I would think those concerns are shared, no?

dom.bs
```diff
@@ -4271,7 +4287,10 @@ invoked, must run these steps:
  "{{NotSupportedError!!exception}}" {{DOMException}}.
 
  <li><p>Return a <a lt="clone a node">clone</a> of <a>this</a>, with the
- <i>clone children flag</i> set if <var>deep</var> is true.
+ <i>clone children flag</i> set if <var>deep</var> is true, and the
+ <i>clone shadows flag</i> set if <a>this</a> is a {{DocumentFragment}} whose
```
rniwa, line 4291: It's really bizarre to special case when calling clone on template element's content fragment.

dom.bs
```diff
@@ -4185,8 +4185,8 @@ elements. SVG ought to do the same for its <{script}> elements, but does not cal
 at the moment.
 
 <p>To <dfn export id=concept-node-clone lt="clone a node" local-lt="clone">clone</dfn> a
-<var>node</var>, with an optional <var>document</var> and <i>clone children flag</i>, run these
-steps:
+<var>node</var>, with an optional <var>document</var>, <i>clone children flag</i>, and
+<i>clone shadows flag</i>, run these steps:
```
rniwa, line 4189: I don't think we should allow cloning of shadow trees like this.
Instead, we should keep nested declarative shadow DOM as templates until the algorithm runs to turn it into a shadow tree. This algorithm can then recursively process inner declarative shadow DOMs.

dom.bs
mfreed7: Yeah, I don't disagree that this is a bit odd. But it's necessary (or something similar) to allow a "normal" `<template>` element to contain a declarative shadow root and get cloned properly into DOM. If you have suggestions for how to better implement that capability, I'm open to them!

dom.bs
mfreed7: Ahh I see. I suppose that would work, but then you'd have a special case in the **parser** rather than just here in the cloning code, right? I'd always prefer to avoid adding more special cases to the parser if possible.

annevk: Presumably this will have to change anyway as part of enabling streaming, right? I would expect all logic to end up in the parser, but I might be missing something.

annevk (commented):
I got the impression this isn't ready for a more detailed review as of yet, especially as there appears to be agreement to make somewhat significant changes. Let me know if I'm mistaken.

dom.bs
rniwa: So I'd much prefer adding a new method like `createInstance` on `template` element to process declarative shadow DOM over declarative shadow DOM being a thing inside `template` content, and adding a special case to cloning steps.

`importNode` has a bad default of not deep cloning anyway and many people end up sub-optimally calling `template.content.cloneNode(true)` instead of `importNode`. Adding an explicit instantiating method on `template` would solve those issues.

mfreed7: I think there's already confusion about the difference between `cloneNode` and `importNode`. E.g. [this SO post](https://stackoverflow.com/questions/39372886/document-importnode-vs-node-clonenode-real-example), or even [this post](https://github.com/GoogleChromeLabs/howto-components/issues/133) where strong experts aren't sure of the difference. In my own testing, I don't find a performance difference.

Having said all of that, the solution doesn't seem to be adding a third way to clone templates...

mfreed7:
> I got the impression this isn't ready for a more detailed review as of yet, especially as there appears to be agreement to make somewhat significant changes. Let me know if I'm mistaken.

Let's definitely get the HTML PR cleaned up and agreed before tackling this one.

mfreed7:
I updated this PR to go along with the latest html spec PR. See https://github.com/whatwg/html/pull/5465 for context/discussion.

dom.bs
```diff
@@ -6774,15 +6832,23 @@ are:
   </ol>
  </li>
 
- <li><p>If <a>this</a> is a <a for=Element>shadow host</a>, then <a>throw</a> an
- "{{NotSupportedError!!exception}}" {{DOMException}}.
+ <li><p>If <var>shadow host</var> has a non-null <a for=/>shadow root</a>, then:
```
annevk, line 6835: Why can this not use the original language of "If _shadow host_ is a shadow host"?

dom.bs
```diff
@@ -6727,11 +6759,37 @@ are:
...
+ <li><p>Let <var>shadow</var> be <a>this</a>'s <a for=Element>shadow root</a>.
+
+ <li><p>If <var>shadow</var> is null or its <a for=ShadowRoot>mode</a> is "<code>closed</code>",
+ then return null.</p></li>
+
+ <li><p>Return <var>shadow</var>.
+</ol>
+
+<p>To <dfn id=concept-attach-a-shadow-root>attach a shadow root</dfn>, given a
+<var>shadow host</var>, <var>mode</var>, <var>clonable</var>, and <var>delegates focus</var>,
```
annevk, line 6785: Also, _shadowHost_ and _delegatesFocus_.

dom.bs
```diff
@@ -6774,15 +6832,23 @@ are:
   </ol>
  </li>
 
- <li><p>If <a>this</a> is a <a for=Element>shadow host</a>, then <a>throw</a> an
- "{{NotSupportedError!!exception}}" {{DOMException}}.
+ <li><p>If <var>shadow host</var> has a non-null <a for=/>shadow root</a>, then:
+   <ol>
+     <li><p>If <var>shadow host</var>'s <a for=/>shadow root</a>'s <a for=ShadowRoot>is declarative
```
annevk, line 6837: Nit: indentation.

dom.bs
```diff
@@ -6791,20 +6857,9 @@ are:
  <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> to
  <var>init</var>["{{ShadowRootInit/slotAssignment}}"].
 
- <li><p>Set <a>this</a>'s <a for=Element>shadow root</a> to <var>shadow</var>.
+ <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>is declarative shadow root</a> property to false.
```
annevk, line 6860:
```suggestion
 <li><p>Set <var>shadow</var>'s <a for=ShadowRoot>is declarative shadow root</a> to false.
```

dom.bs
```diff
@@ -6727,11 +6759,37 @@ are:
...
 <ol>
- <li><p>If <a>this</a>'s <a for=Element>namespace</a> is not the <a>HTML namespace</a>,
+
+ <li><p>Run <a>attach a shadow root</a> with <var>shadow host</var> equal to <a>this</a>,
+   <var>mode</var> equal to <var>init</var>'s {{ShadowRootInit/mode}}, <var>clonable</var>
+   equal to <var>init</var>'s {{ShadowRootInit/clonable}}, and <var>delegates focus</var>
+   equal to <var>init</var>'s {{ShadowRootInit/delegatesFocus}}.</p></li>
+
+ <li><p>Return <a>this</a>'s <a for=Element>shadow root</a>.</p></li>
+
```
annevk, line 6769: Nit: spurious newline.

dom.bs
```diff
@@ -6774,15 +6832,23 @@ are:
   </ol>
  </li>
 
- <li><p>If <a>this</a> is a <a for=Element>shadow host</a>, then <a>throw</a> an
- "{{NotSupportedError!!exception}}" {{DOMException}}.
+ <li><p>If <var>shadow host</var> has a non-null <a for=/>shadow root</a>, then:
+   <ol>
+     <li><p>If <var>shadow host</var>'s <a for=/>shadow root</a>'s <a for=ShadowRoot>is declarative
+     shadow root</a> property is false, then <a>throw</a> an "{{NotSupportedError!!exception}}" {{DOMException}}.
```
annevk, line 6838: Not a property.

dom.bs
```diff
@@ -6774,15 +6832,23 @@ are:
...
+
+     <li><p>Otherwise, <a for=/>remove</a> all of <a for=/>shadow root</a>'s <a>children</a>, in
+     <a>tree order</a>. Return <var>shadow host</var>'s <a for=/>shadow root</a>.
```
annevk, line 6841: It seems weird for this algorithm to sometimes return something. That doesn't seem right.

dom.bs
```diff
@@ -4363,6 +4363,28 @@ dom-Range-extractContents, dom-Range-cloneContents -->
  <a for=tree>children</a> of <var>node</var> and append them to <var>copy</var>, with
  <var>document</var> as specified and the <i>clone children flag</i> being set.
 
+ <li>If <var>node</var> is a <a for=Element>shadow host</a> whose <a for=/>shadow root</a>'s
+ <a for=ShadowRoot>clonable</a> flag is set, run these steps:
```
annevk, line 4367:
```suggestion
 <a for=ShadowRoot>clonable</a> is true:
```
Right?

Also, needs to use `<p>` on a newline as the `<li>` has multiple children.

annevk (commented):
It largely looks okay, but still a lot of nits that make it hard to spot the real issues.

dom.bs
mfreed7: Done, here and above.

dom.bs
```diff
@@ -6727,11 +6759,37 @@ are:
...
+<var>shadow host</var>, <var>mode</var>, <var>clonable</var>, and <var>delegates focus</var>,
+run these steps:</p>
+
+<ol>
+ <li><p>If <var>shadow host</var>'s <a for=Element>namespace</a> is <em>not</em> the <a>HTML namespace</a>,
  then <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
 
- <li>
-  <p>If <a>this</a>'s <a for=Element>local name</a> is not one of the following:
+ <li><p>If <var>shadow host</var>'s <a for=Element>local name</a> is <em>not</em> a
```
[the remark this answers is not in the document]

dom.bs

dom.bs

dom.bs

dom.bs
mfreed7: Agreed - removed.

dom.bs

dom.bs
```diff
@@ -4363,6 +4363,26 @@ dom-Range-extractContents, dom-Range-cloneContents -->
  <a for=tree>children</a> of <var>node</var> and append them to <var>copy</var>, with
  <var>document</var> as specified and the <i>clone children flag</i> being set.
 
+ <li>
+  <p>If <var>node</var> is a <a for=Element>shadow host</a> whose <a for=/>shadow root</a>'s
+  <a for=ShadowRoot>clonable</a> is true:
+
+  <ol>
+   <li><p>Run <a>attach a shadow root</a> with <var>copy</var>, <var>node</var>'s <a for=/>shadow
```
annevk, line 4371: No wrapping in phrasing-level elements such as `<a>`. (Unfortunately the requirements for DOM and HTML are still different.)

dom.bs
```diff
@@ -4363,6 +4363,26 @@ dom-Range-extractContents, dom-Range-cloneContents -->
...
+   root</a>'s <a for=ShadowRoot>mode</a>, true, <var>node</var>'s <a for=/>shadow root</a>'s
+   <a>delegates focus</a>.</p></li>
+
+   <li><p>If <var>node</var>'s <a for=/>shadow root</a>'s
+   <a for=ShadowRoot>is declarative shadow root</a> is true, then set <var>copy</var>'s
+   <a for=/>shadow root</a>'s <a for=ShadowRoot>is declarative shadow root</a> property to
```
annevk, line 4377: It's not a property.

dom.bs
```diff
@@ -4363,6 +4363,26 @@ dom-Range-extractContents, dom-Range-cloneContents -->
...
  <var>document</var> as specified and the <i>clone children flag</i> being set.
 
+ <li>
+  <p>If <var>node</var> is a <a for=Element>shadow host</a> whose <a for=/>shadow root</a>'s
+  <a for=ShadowRoot>clonable</a> is true:
+
+  <ol>
+   <li><p>Run <a>attach a shadow root</a> with <var>copy</var>, <var>node</var>'s <a for=/>shadow
+   root</a>'s <a for=ShadowRoot>mode</a>, true, <var>node</var>'s <a for=/>shadow root</a>'s
+   <a>delegates focus</a>.</p></li>
```
annevk, line 4373: No need for closing tags in DOM.

dom.bs
```diff
@@ -4363,6 +4363,26 @@ dom-Range-extractContents, dom-Range-cloneContents -->
...
+
+   <li><p>If <var>node</var>'s <a for=/>shadow root</a>'s
+   <a for=ShadowRoot>is declarative shadow root</a> is true, then set <var>copy</var>'s
+   <a for=/>shadow root</a>'s <a for=ShadowRoot>is declarative shadow root</a> property to
+   true.</p></li>
+
+   <li>If the <i>clone children flag</i> is set, <a lt="clone a node">clone</a> all the
+   <a>children</a> of <var>node</var>'s <a for=/>shadow root</a> and append them to
```
annevk, line 4381: append needs an xref and I suspect we don't want to append them, but rather append the results?

dom.bs
```diff
@@ -4862,8 +4882,9 @@ known as <dfn export id=concept-document lt="document">documents</dfn>.
 <dfn export for=Document id=concept-document-content-type>content type</dfn> (a string),
 <dfn export for=Document id=concept-document-url>URL</dfn> (a <a for=/>URL</a>),
 <dfn export for=Document id=concept-document-origin>origin</dfn> (an <a for=/>origin</a>),
-<dfn export for=Document id=concept-document-type>type</dfn> ("<code>xml</code>" or "<code>html</code>"), and
-<dfn export for=Document id=concept-document-mode>mode</dfn> ("<code>no-quirks</code>", "<code>quirks</code>", or "<code>limited-quirks</code>").
+<dfn export for=Document id=concept-document-type>type</dfn> ("<code>xml</code>" or "<code>html</code>"),
+<dfn export for=Document id=concept-document-mode>mode</dfn> ("<code>no-quirks</code>", "<code>quirks</code>", or "<code>limited-quirks</code>"), and
+<dfn export for=Document>allow declarative shadow roots</dfn> ("<code>allow</code>", "<code>deny</code>", or "<code>unset</code>").</p>
```
annevk, line 4887: No need for an end tag.

annevk: Can you explain the need for a tri-state?

dom.bs
```diff
@@ -5762,13 +5785,19 @@ It is initially set to false.</p>
 <p><a for=/>Shadow roots</a> have an associated
 <dfn export for=ShadowRoot>available to element internals</dfn>. It is initially set to false.</p>
 
+<p><a for=/>Shadow roots</a> have an associated <dfn export for=ShadowRoot>is declarative shadow
```
annevk, line 5788: No wrapping inside a phrasing-level element. So wrap before `<dfn` I suspect.

dom.bs
```diff
@@ -6727,11 +6757,34 @@ are:
 <p>The <dfn method for=Element><code>attachShadow(<var>init</var>)</code></dfn> method steps are:
 
 <ol>
- <li><p>If <a>this</a>'s <a for=Element>namespace</a> is not the <a>HTML namespace</a>,
+ <li><p>Run <a>attach a shadow root</a> with <a>this</a>, <var>init</var>'s
+ {{ShadowRootInit/mode}}, <var>init</var>'s {{ShadowRootInit/clonable}}, and <var>init</var>'s
```
annevk, line 6761: It needs to be `<var>init<var>["{{ShadowRootInit/mode}}"]` when you reference members from a dictionary (map syntax).

dom.bs
```diff
@@ -6727,11 +6757,34 @@ are:
...
+ {{ShadowRootInit/delegatesFocus}}.</p></li>
+
+ <li><p>Return <a>this</a>'s <a for=Element>shadow root</a>.</p></li>
+</ol>
+
+<p>The <dfn attribute for=Element><code>shadowRoot</code></dfn> attribute's getter steps are:
```
annevk, line 6767:
```suggestion
<p>The <dfn attribute for=Element><code>shadowRoot</code></dfn> getter steps are:
```

dom.bs
```diff
@@ -6727,11 +6757,34 @@ are:
...
+<ol>
+ <li><p>Let <var>shadow</var> be <a>this</a>'s <a for=Element>shadow root</a>.
+
+ <li><p>If <var>shadow</var> is null or its <a for=ShadowRoot>mode</a> is "<code>closed</code>",
+ then return null.</p></li>
+
+ <li><p>Return <var>shadow</var>.
+</ol>
+
+<p>To <dfn id=concept-attach-a-shadow-root>attach a shadow root</dfn>, given a <a for=Element>shadow
```
annevk, line 6778: No newline in phrasing-level elements.

annevk: But also, the element isn't necessarily a shadow host at this point, right?

dom.bs
annevk: Also, I think we can just call this "declarative" or some such as it's directly tied to a shadow root so no need to go for such a long name.

dom.bs

dom.bs

dom.bs
mfreed7: Done, and good catch, done.

mfreed7 (commented):
Ok, I think I addressed all of the comments.

I had to also fix up a bunch of lines that broke the build, for missing `for=` on `<a>length</a>` and `<a>activation behavior</a>`. I wasn't sure of the proper reference for all of the `length`s and some may really refer to `string` or something else.

dom.bs
annevk: It makes inline search harder.

dom.bs
annevk: I would hope that the file has internal consistency, but perhaps there are some errors here and there? In general they are disallowed, yes.

dom.bs
annevk: I think this would be better as a boolean that document loading overrides. There are only two possible behaviors after all.

annevk:
I put your build fixes in #1143 and rebased this. I also pushed a fixup commit to correct a number of references to shadow root. Improving "allow declarative shadow roots" is left, but otherwise this is in order now I think.

mfreed7:
> I put your build fixes in #1143 and rebased this. I also pushed a fixup commit to correct a number of references to shadow root, adopt some modern practices and address a nit. Improving "allow declarative shadow roots" is left still.

Thanks. As for making "allow declarative shadow roots" non-tri-state, the problem I wanted to avoid was finding all of the places where a `create-the-document-object` is used, and instead defer the decision to the parser. That's how I implemented it in code, and it felt quite a bit simpler than always initializing the document correctly.

Perhaps @rniwa can weigh in on how he implemented this in WebKit? 

> 
> Adding `<div algorithm>` also uncovered an error. "attach a shadow root" didn't account for the addition of "slot assignment". It's probably best if you address that? To make sure it's good for all callers.

Thanks - fixed.

dom.bs
mfreed7: I meant consistency between WHATWG/html and WHATWG/dom.

dom.bs
mfreed7: But if you're working on WHATWG/html you're already used to searching for "my\s+multi\s+word\s+concept", right?

annevk:
But that is essentially a blocklist approach, which is not sound. It's fine for implementations to cut corners at times, but I'm not entirely convinced we should do it in the specification, especially since HTML does stipulate where it creates documents and all that has gotten improved as well.

annevk:
Looking at the Preview version the "domintro" for `attachShadow()` also needs updating.

domenic:
I don't really understand the objection to making it bi-state. Just modifying https://html.spec.whatwg.org/#initialise-the-document-object should be fine? That is only used iff the document is created by navigations, so it includes exactly this list:

- HTML parser documents
- XML parser documents
- text documents
- media documents

Note that text and media documents don't run scripts from the inside, so unless you add a JS parser entrypoint that makes use of the "allow declarative shadow roots" flag, it doesn't matter what the value is set to for those. If you do add such a JS entrypoint, it seems to me it should be allowed to do `iframeContainingATextFile.contentDocument.body.setHTMLWithDeclarativeShadowRootSupport("...")` just like `iframeContainingATextFile.contentDocument.body.innerHTML = "..."` is allowed today.

(Also user agent pages / PDF viewer pages, technically, but not really since those are opaque-origin and don't run author-controlled script, so nobody can ever insert declarative shadow roots into them even if such a parser entrypoint exists.)

It does *not* include initial about:blank `Document`s, i.e. the initial page of `<iframe src="about:blank"></iframe>`. So if you want to allow `iframeContainingATextFile.contentDocument.body.setHTMLWithDeclarativeShadowRootSupport("...")` on those you'd also need to modify https://html.spec.whatwg.org/#creating-a-new-browsing-context .

mfreed7:
> I don't really understand the objection to making it bi-state. Just modifying [html.spec.whatwg.org/#initialise-the-document-object](https://html.spec.whatwg.org/#initialise-the-document-object) should be fine? That is only used iff the document is created by navigations, so it includes exactly this list:

Ok, done. I converted the tri-state to a boolean, and (because that allowed it) removed the `parser-allow-declarative-shadow-roots` concept completely. I just read back from the intended parent's document's "allow declarative shadow dom".

PTAL.

dom.bs
```diff
@@ -5633,6 +5654,8 @@ method steps are:
 
  <li><p>Set <var>doc</var>'s <a for=Document>content type</a> to "<code>text/html</code>".
 
+ <li><p>Set <var>doc</var>'s <a for=Document>allow declarative shadow roots</a> to false.
+
```
annevk, line 5658: This is no longer needed, right?

annevk (commented):
Thank you! I think this looks good now, modulo a tiny nit. On reflection it seems unfortunate that we have to store parser state on the document, but it's not too bad and given that invoking the HTML parser isn't formalized all that well it's probably for the best. It's also something we can refactor down the line if it ever does get formalized better.

And "HTML fragment parsing algorithm" creates its own document so there's no potential for issues there.

dom.bs

annevk (approved):
This can merge once the HTML PR is ready.

annevk:
This needed rebasing due to #1200.

mfreed7:
> This needed rebasing due to #1200.

Done.

annevk:
Wait what? Shit. I did the work to rebase. Your subsequent force push overwrote a number of fixes: https://github.com/whatwg/dom/compare/17ede93b019cb319c9b786dfce950c2ba17ec3c3..f01b44dd4d6e88d39216a36c4cb47d58125301d2.

annevk:
Fortunately I still had my local work so I just forced pushed it again. Review appreciated.

mfreed7:
> This needed rebasing due to #1200.

Gah, sorry! I read the above comment as "This **needs** rebasing..." so I just went for it. I didn't notice that you also pushed some things to my branch. Glad you got it back - thanks. And sorry for the trouble.

mfreed7:
> Wait what? Shit. I did the work to rebase. Your subsequent force push overwrote a number of fixes: https://github.com/whatwg/dom/compare/17ede93b019cb319c9b786dfce950c2ba17ec3c3..f01b44dd4d6e88d39216a36c4cb47d58125301d2.

All of these changes look good, much appreciated.

annevk (approved):

PR #893, opened 2020-09-15
Add is-available-to-element-internals

mfreed7:
Add `"is available to element internals"` flag, for use with the accompanying [HTML PR](https://github.com/whatwg/html/pull/5912) which restricts access to `ElementInternals.shadowRoot` for shadow roots which were pre-existing before a custom element upgrade/construction. See the discussion on [Issue 871](https://github.com/w3c/webcomponents/issues/871#issuecomment-692383222) for more context.

- [X] At least two implementers are interested (and none opposed):
   * Chromium - me.
   * WebKit: see [this comment](https://github.com/w3c/webcomponents/issues/871#issuecomment-692383222).
- [X] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/pull/25554
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: https://crbug.com/1042130
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/dom/893.html" title="Last updated on Oct 5, 2020, 4:20 PM UTC (8625fe7)">Preview</a> | <a href="https://whatpr.org/dom/893/12beda2...8625fe7.html" title="Last updated on Oct 5, 2020, 4:20 PM UTC (8625fe7)">Diff</a>

PR #894, opened 2020-09-16
Add "precustomized" custom element state

mfreed7:
Add the "precustomized" state needed by the [HTML PR](https://github.com/whatwg/html/pull/5909) that prevents usage of `attachInternals()` prior to the constructor being run.

   * Chromium: me
   * Safari: based on [this comment](https://github.com/w3c/webcomponents/issues/871#issuecomment-690879744)
   * https://github.com/web-platform-tests/wpt/pull/25402
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: Tracked as part of the declarative Shadow DOM [tracking bug](https://crbug.com/1042130).
   * Firefox: none yet
   * Safari: none yet

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 28, 2020, 11:11 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fmfreed7%2Fdom%2F7431705a29fd08d44e13520f5fcac04ad6bd1bea%2Fdom.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%20894)

```
<html><body><p><strong>WARNING: </strong>mysqli_connect(): (HY000/2002): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (111)</p></body></html>
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/dom%23894.)._
</details>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5875,12 +5875,12 @@ dictionary ShadowRootInit {
 initialized.
 
 <p>An <a for=/>element</a>'s <a for=Element>custom element state</a> is one of
-"<code>undefined</code>", "<code>failed</code>", "<code>uncustomized</code>",
-"<code>precustomized</code>", or "<code>custom</code>". An <a for=/>element</a> whose
-<a for=Element>custom element state</a> is "<code>uncustomized</code>" or "<code>custom</code>"
-is said to be <dfn export id=concept-element-defined for=Element>defined</dfn>. An
-<a for=/>element</a> whose <a for=Element>custom element state</a> is "<code>precustomized</code>"
-or "<code>custom</code>" is said to be <dfn export id=concept-element-custom for=Element>custom</dfn>.
+"<code>undefined</code>", "<code>failed</code>", "<code>uncustomized</code>", "<code>precustomized</code>", or
+"<code>custom</code>". An <a for=/>element</a> whose <a for=Element>custom element state</a> is
+"<code>uncustomized</code>" or "<code>custom</code>" is said to be
+<dfn export id=concept-element-defined for=Element>defined</dfn>. An <a for=/>element</a> whose
+<a for=Element>custom element state</a> is "<code>custom</code>" is said to be
+<dfn export id=concept-element-custom for=Element>custom</dfn>.
 
 <p class=note>Whether or not an element is <a for=Element>defined</a> is used to determine the
 behavior of the '':defined'' pseudo-class. Whether or not an element is <a for=Element>custom</a> is
--- a/dom.bs
+++ b/dom.bs
@@ -5875,19 +5875,19 @@ dictionary ShadowRootInit {
 initialized.
 
 <p>An <a for=/>element</a>'s <a for=Element>custom element state</a> is one of
-"<code>undefined</code>", "<code>failed</code>", "<code>uncustomized</code>", or
-"<code>custom</code>". An <a for=/>element</a> whose <a for=Element>custom element state</a> is
-"<code>uncustomized</code>" or "<code>custom</code>" is said to be
-<dfn export id=concept-element-defined for=Element>defined</dfn>. An <a for=/>element</a> whose
-<a for=Element>custom element state</a> is "<code>custom</code>" is said to be
-<dfn export id=concept-element-custom for=Element>custom</dfn>.
+"<code>undefined</code>", "<code>failed</code>", "<code>uncustomized</code>",
+"<code>precustomized</code>", or "<code>custom</code>". An <a for=/>element</a> whose <a
+for=Element>custom element state</a> is "<code>uncustomized</code>", "<code>precustomized</code>",
+or "<code>custom</code>" is said to be <dfn export id=concept-element-defined
+for=Element>defined</dfn>. An <a for=/>element</a> whose <a for=Element>custom element state</a> is
+"<code>custom</code>" is said to be <dfn export id=concept-element-custom for=Element>custom</dfn>.
 
 <p class=note>Whether or not an element is <a for=Element>defined</a> is used to determine the
 behavior of the '':defined'' pseudo-class. Whether or not an element is <a for=Element>custom</a> is
 used to determine the behavior of the <a href=#mutation-algorithms>mutation algorithms</a>. The
-"<code>failed</code>" state is used to ensure that if a <a>custom element constructor</a> fails to
-execute correctly the first time, it is not executed again by an
-<a lt="upgrade an element">upgrade</a>.</p>
+"<code>failed</code>" and "<code>precustomized</code>" states are used to ensure that if a <a>custom
+element constructor</a> fails to execute correctly the first time, it is not executed again by an <a
+lt="upgrade an element">upgrade</a>.</p>
 
 <div class=example id=example-c5b21302>
  <p>The following code illustrates elements in each of these four states:</p>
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -5875,19 +5875,19 @@ dictionary ShadowRootInit {
...
+for=Element>defined</dfn>. An <a for=/>element</a> whose <a for=Element>custom element state</a> is
+"<code>custom</code>" is said to be <dfn export id=concept-element-custom for=Element>custom</dfn>.
 
 <p class=note>Whether or not an element is <a for=Element>defined</a> is used to determine the
 behavior of the '':defined'' pseudo-class. Whether or not an element is <a for=Element>custom</a> is
 used to determine the behavior of the <a href=#mutation-algorithms>mutation algorithms</a>. The
-"<code>failed</code>" state is used to ensure that if a <a>custom element constructor</a> fails to
-execute correctly the first time, it is not executed again by an
-<a lt="upgrade an element">upgrade</a>.</p>
+"<code>failed</code>" and "<code>precustomized</code>" states are used to ensure that if a <a>custom
```
annevk, line 5888: No newlines in inline elements. Sorry that this is different from HTML.

dom.bs
```diff
@@ -5875,19 +5875,19 @@ dictionary ShadowRootInit {
...
+element constructor</a> fails to execute correctly the first time, it is not executed again by an <a
```
annevk, line 5889: Same.

annevk (commented):
Thanks, this looks good. And I guess also suggests we need a test for defined.

dom.bs
```diff
@@ -5875,19 +5875,19 @@ dictionary ShadowRootInit {
...
 <p>An <a for=/>element</a>'s <a for=Element>custom element state</a> is one of
-"<code>undefined</code>", "<code>failed</code>", "<code>uncustomized</code>", or
-"<code>custom</code>". An <a for=/>element</a> whose <a for=Element>custom element state</a> is
-"<code>uncustomized</code>" or "<code>custom</code>" is said to be
-<dfn export id=concept-element-defined for=Element>defined</dfn>. An <a for=/>element</a> whose
-<a for=Element>custom element state</a> is "<code>custom</code>" is said to be
-<dfn export id=concept-element-custom for=Element>custom</dfn>.
+"<code>undefined</code>", "<code>failed</code>", "<code>uncustomized</code>",
+"<code>precustomized</code>", or "<code>custom</code>". An <a for=/>element</a> whose <a
+for=Element>custom element state</a> is "<code>uncustomized</code>", "<code>precustomized</code>",
```
rniwa, line 5880: This seems wrong. Why would a custom element be considered as *defined* before `super()` call is made in the constructor? This will be incompatible with the shipping behavior of custom elements.

dom.bs
mfreed7: Ahh ok. Fixed.

dom.bs

dom.bs
mfreed7: I think I agree. I've removed "precustomized" from the definition of "defined".

@annevk do you still think there is a need for more testing of `:defined`?

mfreed7:
Ok, I think I addressed everything here. Note that I did add "precustomized" to the definition of "custom element". I'm using that in [my other PR](https://github.com/whatwg/dom/pull/893).

dom.bs
```diff
@@ -5875,19 +5875,19 @@ dictionary ShadowRootInit {
...
-"<code>uncustomized</code>" or "<code>custom</code>" is said to be
-<dfn export id=concept-element-defined for=Element>defined</dfn>. An <a for=/>element</a> whose
-<a for=Element>custom element state</a> is "<code>custom</code>" is said to be
-<dfn export id=concept-element-custom for=Element>custom</dfn>.
+"<code>undefined</code>", "<code>failed</code>", "<code>uncustomized</code>",
+"<code>precustomized</code>", or "<code>custom</code>". An <a for=/>element</a> whose
+<a for=Element>custom element state</a> is "<code>uncustomized</code>" or "<code>custom</code>"
+is said to be <dfn export id=concept-element-defined for=Element>defined</dfn>. An
+<a for=/>element</a> whose <a for=Element>custom element state</a> is "<code>precustomized</code>"
+or "<code>custom</code>" is said to be <dfn export id=concept-element-custom for=Element>custom</dfn>.
```
rniwa, line 5883: I don't think it's right to consider "precustomized" state as **custom** either.
This would mean that in the upgrade case, you can insert the node into a document tree before `super()` is the called, and the spec text will try to look for its **custom element definition** and fail, which would be self contradictory.

mfreed7: Hmm, I added this in response to a comment from @annevk on [DOM-893](https://github.com/whatwg/dom/pull/893). How would you recommend writing [this check](https://github.com/whatwg/dom/pull/893/files#diff-a467e681501e456ce8c1ef31425e1b41R6744), which is meant to check for "precustomized" or "custom". As @annevk pointed out, the definition of "custom element" is "custom element state is custom". I would assume you would just prefer that check (in 893) to say this?

> If this's custom element state is "precustomized" or "custom", set shadow's **available to element internals** property to true. Otherwise, set it to false.

?

rniwa: It seems like we should just check that custom element state is either "custom" or "precusotmized"?

dom.bs
annevk: I think we should have a test yes, to ensure that it does not match that.

mfreed7: Ok, I'll add a test that `:defined` doesn't match, pre-`super()`.

dom.bs
mfreed7: Ok, done, both here and in [893](https://github.com/whatwg/dom/pull/893/files).

dom.bs
mfreed7: So we already have [a test for :defined](https://github.com/web-platform-tests/wpt/blob/master/custom-elements/pseudo-class-defined.html#L103) during an upgrade, and I think it's good enough to catch problems here. Note a couple things:
 * `:defined` already doesn't match during an upgrade, even **after** `super()`, because the custom element state is only set to "custom" [after the constructor completes](https://html.spec.whatwg.org/#upgrades).
 * Testing `:defined` prior to `super()` shouldn't be typical anyway, since you can't access `this` prior to `super()`. You could use an external reference to the element to test, of course, but that doesn't seem like it'd be common.

I'm sure I'm missing something here - please let me know what it is, and what you'd like me to test for. I can add a test for the second bullet above, but it doesn't seem terribly useful. It would catch the case where `:defined` matches **prior** to `super()` but stops matching it **after** `super()`.

mfreed7: I added [this additional testing](https://github.com/web-platform-tests/wpt/pull/25794/files), in case that's what you're looking for here.

annevk: Thanks, I think those asserts are helpful.

annevk (approved):

PR #898, opened 2020-09-29
Meta: put custom element acknowledgment in the IPR section

annevk:
***
<a href="https://whatpr.org/dom/898.html" title="Last updated on Sep 30, 2020, 1:30 PM UTC (0660235)">Preview</a> | <a href="https://whatpr.org/dom/898/03d45d1...0660235.html" title="Last updated on Sep 30, 2020, 1:30 PM UTC (0660235)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -10194,10 +10194,8 @@ for being awesome!
 with substantial contributions from Aryeh Gregor (<a href=mailto:ayg@aryeh.name>ayg@aryeh.name</a>)
 and Ms2ger (<a href="mailto:ms2ger@gmail.com">ms2ger@gmail.com</a>).
 
-<p>Part of the revision history of the integration points related to <a for=Element>custom</a>
-elements can be found in
+<p boilerplate=ipr>Part of the revision history of the integration points related to
+<a for=Element>custom</a> elements can be found in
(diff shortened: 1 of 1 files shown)


annevk:
Hmm, HTML has some of this still at the end of https://html.spec.whatwg.org/#acknowledgments it seems.

What do you think we should do @domenic? @foolip?

domenic:
Two options (which I think should apply to both HTML and DOM, whichever way we decide):

- Keep "Part of the revision history ... repository" in acknowledgments, and add a separate sentence such as "Some content is derived from the w3c/webcomponents repository, which is available under..." in the IPR section.

- Move the entire sentence into the IPR section.

(Relatedly, it seems HTML could probably due to cite some more licenses, e.g. ResponsiveImagesCG/picture-element has a license that wants folks to cite it... and the custom elements [license](https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document) seems to have changed its name and has rather specific suggestions for how to cite which we aren't fully following... hmm.)

foolip:
Without having thought too hard about what the duplication would look like, doing "thank you" acknowledgement and attribution required by license separately makes sense to me.

annevk:
I put up an equivalent fix for HTML at https://github.com/whatwg/html/pull/5963 which I consider to be good enough for now. Let me know if you feel otherwise.

PR #918, opened 2020-11-06
Remove Goals and Historical sections

annevk:
Now that the relevant W3C documents redirect here, there is no need to explain the difference.

(I'm also getting reference errors, e.g., in #819.)

***
<a href="https://whatpr.org/dom/918.html" title="Last updated on Nov 9, 2020, 4:24 PM UTC (643e4d8)">Preview</a> | <a href="https://whatpr.org/dom/918/0de5ae3...643e4d8.html" title="Last updated on Nov 9, 2020, 4:24 PM UTC (643e4d8)">Diff</a>


domenic (commented):
Hmm, I find the list of things that were explicitly removed kind of helpful... some of it could be cleaned up, e.g. things now defined elsewhere, and in general some of the section tries to act like a changelog, and there's no way that changelog is up to date.

But, explaining that we successfully removed getUserData()/xmlEncoding/setIdAttribute()/etc. seems good for people who might be coming from Java or similar? And it's kind of similar to HTML's https://html.spec.whatwg.org/#non-conforming-features section.

So my preference would be to remove the Goals section, and morph the Historical section to just list all the removed interfaces/attributes/operations, broken down by interface. You could do so without referencing what specs they were originally in, just an intro sentence saying "previous DOM specifications" or something like that.

WDYT?

annevk:
Yeah, that's fair. I still think we should clean up the words used in HTML as obsolete is really confusing for things not actually gone. I think I'm personally settling on "legacy" for "don't like/redundant, but have to keep around" and "historical" for "gone".

dom.bs
```diff
@@ -9877,47 +9840,11 @@ methods on {{Document}}.
...
 <h2 id=historical>Historical</h2>
 
-<p>As explained in <a href="#goals">goals</a> this standard is a significant revision of various
-obsolete DOM specifications. This section enumerates the incompatible changes.
-
-
-<h3 id=dom-events-changes>DOM Events</h3>
+<p id=domstringlist>Contrary to earlier revisions of this standard, {{DOMString}}, {{DOMException}},
+and {{DOMTimeStamp}} are now defined in Web IDL. {{DOMStringList}} is now defined in HTML.
+[[WEBIDL]] [[HTML]]
```
domenic, line 9845: I'd lean toward omitting these too, since they have a home that people can find without too much effort. It's the ones that totally disappeared that it seems nice to keep some record of. (But, no big deal either way.)

domenic (approved):
It like the rename to legacy!

PR #925, opened 2020-11-17
Fix the comparing boundary points of `concept-range-bp-set`

### LINKED ISSUE #924: concept-range-bp-set should compare root nodes before bp relative position
In https://dom.spec.whatwg.org/#concept-range-bp-set:

> If bp is after the range’s end, or if range’s root is not equal to node’s root, set range’s end to bp.

(and similar for "set the end")

This implies that implementations should check if "bp is after the range’s end", however the first step of comparing boundary points is to "Assert: nodeA and nodeB have the same root.".

I would suggest this be re-ordered to:

> If range’s root is not equal to node’s root, or if bp is after the range’s end, set range’s end to bp.

(and similar for "set the end")

chansuke:
Fixes #924

***
<a href="https://whatpr.org/dom/925.html" title="Last updated on Nov 17, 2020, 11:36 PM UTC (44eecc3)">Preview</a> | <a href="https://whatpr.org/dom/925/52f7a52...44eecc3.html" title="Last updated on Nov 17, 2020, 11:36 PM UTC (44eecc3)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -7803,14 +7803,14 @@ steps:
    <dt>If these steps were invoked as "set the start"
    <dd>
     <ol>
-     <li>If <var>bp</var> is
-     <a for="boundary point">after</a> the
-     <var>range</var>'s <a for=range>end</a>, or
-     if <var>range</var>'s
+     <li>If <var>range</var>'s
      <a for="live range">root</a> is not equal to
-     <var>node</var>'s <a for=tree>root</a>, set
-     <var>range</var>'s <a for=range>end</a> to
-     <var>bp</var>.
+     <var>node</var>'s <a for=tree>root</a>,
+     or if <var>bp</var> is
+     <a for="boundary point">after</a> the
+     <var>range</var>'s <a for=range>end</a>, set
+     <var>range</var>'s <a for=range>end</a>
+     to <var>bp</var>.
 
      <li>Set <var>range</var>'s
      <a for=range>start</a> to <var>bp</var>.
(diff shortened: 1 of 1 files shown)


annevk (approved):
Thanks, this looks good. @edg2s FYI.

annevk:
@chansuke if you want you can add your name to the Acknowledgments section.

edg2s:
+1

dom.bs
```diff
@@ -7803,28 +7803,28 @@ steps:
...
-     <var>range</var>'s <a for=range>start</a>,
-     or if <var>range</var>'s
+     <li>If <var>range</var>'s
      <a for="live range">root</a> is not equal to
-     <var>node</var>'s <a for=tree>root</a>, set
-     <var>range</var>'s <a for=range>start</a>
+     <var>node</var>'s <a for=tree>root</a>,
+     or if <var>bp</var> is
+     <a for="boundary point">before</a> the
+     <var>range</var>'s <a for=range>end</a>, set
```
edg2s, line 7826: this should be "start"

dom.bs
```diff
@@ -7803,28 +7803,28 @@ steps:
...
+     <var>range</var>'s <a for=range>end</a>
```
edg2s, line 7827: this should be "start"

edg2s (changes requested):

edg2s:
"start" was inadvertently changed to "end", presumably and copy-paste error, as can be seen in the HTML diff:
image

chansuke:
@edg2s (CC: @annevk )I appreciate your review. I fixed and rebased it(since it's a small change).

annevk (approved):

annevk:
Thanks all!

PR #966, opened 2021-04-02
Add Imperative Slot API

mfreed7:
### This is a fresh PR, with updates from https://github.com/whatwg/dom/pull/860. I don't have permissions to update that PR, and the master-to-main transition made it difficult anyway.

The explainer for this feature is here: https://github.com/w3c/webcomponents/blob/gh-pages/proposals/Imperative-Shadow-DOM-Distribution-API.md
The issue discussion is here: [3534](https://github.com/whatwg/html/issues/3534)
There is a corresponding [Pull Request](https://github.com/whatwg/html/pull/6561) for the HTML spec that goes along with this PR.

* [ ]  At least two implementers are interested (and none opposed):
  
  * Chromium tracking bug is [here](http://crbug.com/1196842), with the original implementation tracker [here](http://crbug.com/869308).
  * The implementation is based on what was agreed in the last [F2F Meeting](https://github.com/whatwg/html/issues/3534#issuecomment-537802687).
* [x]  [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
  
  * In WPT: https://wpt.fyi/results/shadow-dom/slots-imperative-slot-api.tentative.html?label=master&label=experimental&aligned&q=slots-imperative-slot-api.tentative.html
  * In WPT: https://wpt.fyi/results/shadow-dom/slots-imperative-api-slotchange.tentative.html?label=master&label=experimental&aligned
  * More WPTs to be implemented as part of the [Chromium implementation](http://crbug.com/1196842).
* [x]  [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
  
  * Chrome: https://crbug.com/869308
  * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1705141
  * Safari: https://bugs.webkit.org/show_bug.cgi?id=218692

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 15, 2021, 7:24 AM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwhatpr.org%2Fdom%2F966%2Fa73380f.html&doc2=https%3A%2F%2Fwhatpr.org%2Fdom%2F966.html)

```
<html><head>
</head><body>
your request.</p>
 sysreq@w3.org to inform them of the time this error occurred,
in the server error log.</p>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/dom%23966.)._
</details>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2222,7 +2222,8 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
    <li><p>Let <var>result</var> be an empty list.</p></li>
 
    <li><p>For each <a>slottable</a> in <var>slot</var>'s <a>manually assigned nodes</a>,
-   if <a>slottable</a>'s <a for=tree>parent</a> is <var>host</var>, add <a>slottable</a> to array.</p></li>
+   if <a>slottable</a>'s <a for=tree>parent</a> is <var>host</var>, add <a>slottable</a>
+   to <var>result</var>.</p></li>
   </ol>
  </li>
 
--- a/dom.bs
+++ b/dom.bs
@@ -2228,9 +2228,8 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
 
    <li><p><a for=set>For each</a> <a>slottable</a> <var>slottable</var> of <var>slot</var>'s
    <a>manually assigned nodes</a>, if <var>slottable</var>'s <a for=tree>parent</a> is
-   <var>host</var>, <a for=list>append</a> <var>slottable</var> to <var>result</var>.</p></li>
+   <var>host</var>, <a for=list>append</a> <var>slottable</var> to <var>result</var>.
   </ol>
- </li>
 
  <li>
   <p>Otherwise, for each <a>slottable</a> <a for=tree>child</a> <var>slottable</var> of
@@ -5756,7 +5755,7 @@ It is initially set to false.</p>
      consequences for innerHTML. -->
 
 <p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>slot assignment</dfn>
-("<code>manual</code>" or "<code>named</code>").</p>
+("<code>manual</code>" or "<code>named</code>").
 
 <p>A <a for=/>shadow root</a>'s <a>get the parent</a> algorithm, given an <var>event</var>, returns
 null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow root</a> is the
--- a/dom.bs
+++ b/dom.bs
@@ -2683,6 +2683,9 @@ indicated in the <a for=/>remove</a> algorithm below.
 
  <li><p><a for=set>Remove</a> <var>node</var> from its <var>parent</var>'s <a for=tree>children</a>.
 
+ <li><p>If <var>node</var> is an {{HTMLSlotElement}}, set <var>node</var>'s <a>manually assigned nodes</a>
+ to an empty set.
+
  <li><p>If <var>node</var> is <a for=slottable>assigned</a>, then run <a>assign slottables</a> for
  <var>node</var>'s <a>assigned slot</a>.
 
--- a/dom.bs
+++ b/dom.bs
@@ -2177,8 +2177,10 @@ non-null.</p>
 
 <p>A <a>slottable</a> has an associated
 <dfn export for=slottable id=slottable-manual-slot-assignment>manual slot assignment</dfn> (null
-or a <a>slot</a>). Unless stated otherwise, it is null. The <a>manual slot assignment</a> is
-a weak reference to the slot, such that it can be garbage collected if not referenced elsewhere.</p>
+or a <a>slot</a>). Unless stated otherwise, it is null.</p>
+
+<p class=note>The <a>manual slot assignment</a> can be implemented using a weak reference
+to the <a>slot</a>, because this variable is not directly accessible from script.</p>
 
 <h5 id=finding-slots-and-slotables>Finding slots and slottables</h5>
 
--- a/dom.bs
+++ b/dom.bs
@@ -2473,7 +2473,7 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
    <a for=tree>children</a> before <var>child</var>'s <a for=tree>index</a>.
 
    <li><p>If <var>parent</var> is a <a for=Element>shadow host</a> whose <a for=/>shadow root</a>'s
-   <a for=ShadowRoot>slot assignment</a> is "<code>name</code>" and <var>node</var> is a
+   <a for=ShadowRoot>slot assignment</a> is "<code>named</code>" and <var>node</var> is a
    <a>slottable</a>, then <a>assign a slot</a> for <var>node</var>.
 
    <li><p>If <var>parent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
@@ -5738,7 +5738,7 @@ interface ShadowRoot : DocumentFragment {
 };
 
 enum ShadowRootMode { "open", "closed" };
-enum SlotAssignmentMode { "manual", "name" };
+enum SlotAssignmentMode { "manual", "named" };
 </pre>
 
 <p>{{ShadowRoot}} <a for=/>nodes</a> are simply known as
@@ -5758,7 +5758,7 @@ It is initially set to false.</p>
      consequences for innerHTML. -->
 
 <p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>slot assignment</dfn>
-("<code>manual</code>" or "<code>name</code>").</p>
+("<code>manual</code>" or "<code>named</code>").</p>
 
 <p>A <a for=/>shadow root</a>'s <a>get the parent</a> algorithm, given an <var>event</var>, returns
 null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow root</a> is the
@@ -5903,7 +5903,7 @@ interface Element : Node {
 dictionary ShadowRootInit {
   required ShadowRootMode mode;
   boolean delegatesFocus = false;
-  SlotAssignmentMode slotAssignment = "name";
+  SlotAssignmentMode slotAssignment = "named";
 };
 </pre>
 
--- a/dom.bs
+++ b/dom.bs
@@ -2175,12 +2175,11 @@ Unless stated otherwise it is null. A <a>slottable</a> is
 <dfn export for=slottable id=slotable-assigned>assigned</dfn> if its <a>assigned slot</a> is
 non-null.</p>
 
-<p>A <a>slottable</a> has an associated
-<dfn export for=slottable id=slottable-manual-slot-assignment>manual slot assignment</dfn> (null
-or a <a>slot</a>). Unless stated otherwise, it is null.</p>
+<p>A <a>slottable</a> has an associated <dfn export for=slottable>manual slot assignment</dfn> (null
+or a <a>slot</a>). Unless stated otherwise, it is null.
 
-<p class=note>The <a>manual slot assignment</a> can be implemented using a weak reference
-to the <a>slot</a>, because this variable is not directly accessible from script.</p>
+<p class=note>A <a>slottable</a>'s <a>manual slot assignment</a> can be implemented using a weak
+reference to the <a>slot</a>, because this variable is not directly accessible from script.
 
 <h5 id=finding-slots-and-slotables>Finding slots and slottables</h5>
 
@@ -2199,9 +2198,10 @@ steps:</p>
  <li><p>If the <i>open flag</i> is set and <var>shadow</var>'s <a for=ShadowRoot>mode</a> is
  <em>not</em> "<code>open</code>", then return null.</p></li>
 
- <li><p>If <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> is "<code>manual</code>",
- then return the <a>slot</a> in <var>shadow</var>'s <a for=tree>descendants</a> whose <a>manually assigned nodes</a>
- <a for=set>contains</a> <var>slottable</var>, if any, and null otherwise.</p></li>
+ <li><p>If <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> is "<code>manual</code>", then
+ return the <a>slot</a> in <var>shadow</var>'s <a for=tree>descendants</a> whose
+ <a>manually assigned nodes</a> <a for=set>contains</a> <var>slottable</var>, if any, and null
+ otherwise.
 
  <li><p>Return the first <a>slot</a> in <a>tree order</a> in <var>shadow</var>'s
  <a for=tree>descendants</a> whose <a for=slot>name</a> is <var>slottable</var>'s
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -2192,6 +2192,10 @@ steps:</p>
  <li><p>If the <i>open flag</i> is set and <var>shadow</var>'s <a for=ShadowRoot>mode</a> is
  <em>not</em> "<code>open</code>", then return null.</p></li>
 
+ <li><p>If <var>shadow</var>'s <a for=ShadowRoot>slot assignment</a> is "<code>manual</code>",
+ then return the <a>slot</a> in <var>shadow</var>'s <a for=tree>descendants</a> whose <a>manually assigned nodes</a>
```
rniwa, line 2196: This probably needs to find the first slot because there could be multiple slots referring to the same node (using element reflection style weak reference would mean that we also want to allow re-ordering of slots without losing their assigned nodes.

dom.bs
```diff
@@ -2203,14 +2207,28 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
...
- <a for=DocumentFragment>host</a>.</p></li>
+ <li><p>Let <var>host</var> be <var>root</var>'s <a for=DocumentFragment>host</a>.</p></li>
 
  <li>
-  <p>For each <a>slottable</a> <a for=tree>child</a> of <var>host</var>, <var>slottable</var>, in
+  <p>If <var>root</var>'s <a for=ShadowRoot>slot assignment</a> is "<code>manual</code>",
+  then:</p>
+
+  <ol>
+   <li><p>Let <var>result</var> be an empty list.</p></li>
```
rniwa, line 2222: This probably also needs to check that this slot is the first slot of a given name / default slot.

dom.bs
mfreed7: Hmm, if I'm thinking about this correctly, there should only be one slot referring to a given node, because the [assign](https://github.com/whatwg/html/pull/5483/files#diff-41cf6794ba4200b839c53531555f0f3998df4cbb01a4d5cb0b94e3ca5e23947dR59369) method loops through all slots within the shadow root and removes the given nodes from all other slots. It also de-duplicates assigned nodes within a single slot. So I think this algorithm should always return a unique slot. Correct me if I'm wrong.

Separately, I do think we need to add a step to [remove a node](https://dom.spec.whatwg.org/#concept-node-remove) that clears out all manually assigned nodes from `<slot>` elements when they get removed from the tree. Without that, there is the definite possibility of the multiple slot problem you bring up, because the `assign()` steps won't be able to find all slots that might point to a node.

dom.bs
mfreed7: I think this is ok also, for the same reasons as above.

dom.bs
rniwa: I mean... that would defeat the whole point of using weak reference defined for element reflection. The whole point of using that technology is so that slot can be removed & inserted into another place without losing its assigned nodes.

mfreed7: Hmm, I thought the purpose was allowing the movement of **light dom** children of the host, not the `<slot>` elements themselves. Is that an important use case?

The (potentially huge?) downside of allowing movement of both light dom children **and** slot elements is that the `manually assigned nodes` vector can start to contain duplicates all over the place. Since that vector is not developer-visible, it might get very confusing, if the order of the `<slot>` elements starts to dictate which manually-assigned node gets assigned to which slot. Consider:

```html
<div id=host>
  <template shadowroot=open shadowrootmanualslotting>
    <slot id=slot1></slot>
    <slot id=slot2></slot>
  </template>
  <div id=node1></div>
  <div id=node2></div>
</div>

<script>
slot1.assign([node1, node2]);
slot1.remove();
slot2.assign([node1, node2]);
// Here, both node1/node2 will be assigned to `slot2`.
host.shadowRoot.appendChild(slot1);
// Still, both node1/node2 will be assigned to `slot2`, because `slot1` comes last.
host.shadowRoot.appendChild(slot2);
// Now, both node1/node2 will be assigned to `slot1`.
</script>
```

It seems more tractable (from a developer's point of view) to have all slots within a shadow root contain unique sets of manually assigned nodes, no? Otherwise, they'll always have to call `.assign()` on any slots they append to the shadow root, to make sure the (invisible) manually assigned nodes are still what they think.

rniwa: The simplest solution to that problem is that whenever a node is assigned to a new slot, then it gets removed from all other slots.

rniwa: It is problematic to require the list of nodes that can be assigned to a given slot to be in the shadow tree of their parent node as that would force the slot elements to be in the shadow tree before any nodes can be assigned. That is the kind of construction relationship problem we discussed during AOM / element reflection and concluded that it is a show stopper issue for that use case. I don't see why it would be any different here.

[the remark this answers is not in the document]
domenic: Ah, OK, that's not really avoidable then. No worries.

mfreed7: > The simplest solution to that problem is that whenever a node is assigned to a new slot, then it gets removed from all other slots.

Hmm ok, this seems like a reasonable solution. It does require another `weak_ptr` from `Node` back to its assigned slot. I suppose this isn't too bad. If you're ok with that solution, I'll see if I can add it to the spec. It definitely simplifies things to have only one slot per node.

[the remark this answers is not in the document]
mfreed7: Ok, good to hear you confirm this. I'm never quite sure if broken builds here are my fault or not.

rniwa: Right. Only caveat is that a given node can be simultaneously assigned to a slot of a shadow tree which uses automatic "name" based assignment and to a slot which is not in that shadow tree. In that case, the manual assignment should work but the automatic name based assignment should prevail since the manual assignment is not in effect as the assigned node is not a slottable of the host of a shadow root of the slot to which it was manually assigned. But it is true that at any given point, each node can only be assigned to at most one slot manually.

There is a bit of oddity here in that each shadow tree determines which mode of assignment we're using and slot doesn't despite of the fact manual assignment API appears on slot element. So in theory, it's possible for someone to assign a node to a slot and then insert that into a shadow tree which uses automatic name based assignment. I guess we'd ignore the manual assignment in that case.

mfreed7: Ok great. And that's a great point about a slot with manually assigned nodes that is living in a shadow host in "named" mode. I agree with the behavior you proposed, and I'll try to get it written up.

There are a number of potentially confusing things that the developer can do here, and I'm wondering which of them (if any) should get helpful console warnings. I feel like there will be a tough balance to strike between log spam and a lack of understanding of what's going on.

dom.bs
```diff
@@ -2175,6 +2175,11 @@ Unless stated otherwise it is null. A <a>slottable</a> is
 <dfn export for=slottable id=slotable-assigned>assigned</dfn> if its <a>assigned slot</a> is
 non-null.</p>
 
+<p>A <a>slottable</a> has an associated
+<dfn export for=slottable id=slottable-manual-slot-assignment>manual slot assignment</dfn> (null
+or a <a>slot</a>). Unless stated otherwise, it is null. The <a>manual slot assignment</a> is
+a weak reference to the slot, such that it can be garbage collected if not referenced elsewhere.</p>
```
mfreed7, line 2181: Here also, comments appreciated on the "weak pointer" stuff.

dom.bs
```diff
@@ -5729,6 +5755,9 @@ It is initially set to false.</p>
 <!-- If we ever change this, e.g., add a ShadowRoot object constructor, that would have serious
      consequences for innerHTML. -->
 
+<p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>slot assignment</dfn>
+("<code>manual</code>" or "<code>name</code>").</p>
```
mfreed7, line 5759: Did we bikeshed "name" vs. "named" at all here? Using "named" feels better to me for some reason, but no strong opinions.

dom.bs
rniwa: That is a fair point. You might want to sync up with @alice about this.

mfreed7 (commented):
Ok, this should be updated with the new "weak pointer in both directions" logic. The intention was:

 - you can call `slot.assign()` at any time with any nodes, regardless of where the nodes or the slot lives at the time.
 - Each node can be manually assigned to only a single `<slot>` at any time.
 - If a `<slot>` currently lives inside a shadow tree that is in "manual" mode, then any nodes in its manually assigned nodes list that are currently light-dom direct children of the shadow host will be slotted in. Any other manually assigned nodes that don't meet these criteria will not be slotted, but will also not be removed from the list.
 - If a `<slot>` with manually assigned nodes currently lives inside a shadow tree that is in "name" mode, the manually assigned nodes will be ignored, and nodes will be assigned based on normal "name" assignment rules. Again, nodes will not be removed from the list here.
 - Given the above, both slots and nodes should be able to be moved around the tree and among documents without destroying the slot-to-node relationships established by `slot.assign()`. Nodes will be (re-)assigned to slots if/when they meet the criteria above.

rniwa:
> Ok, this should be updated with the new "weak pointer in both directions" logic. The intention was:
> 
> * you can call `slot.assign()` at any time with any nodes, regardless of where the nodes or the slot lives at the time.
> * Each node can be manually assigned to only a single `<slot>` at any time.
> * If a `<slot>` currently lives inside a shadow tree that is in "manual" mode, then any nodes in its manually assigned nodes list that are currently light-dom direct children of the shadow host will be slotted in. Any other manually assigned nodes that don't meet these criteria will not be slotted, but will also not be removed from the list.
> * If a `<slot>` with manually assigned nodes currently lives inside a shadow tree that is in "name" mode, the manually assigned nodes will be ignored, and nodes will be assigned based on normal "name" assignment rules. Again, nodes will not be removed from the list here.
> * Given the above, both slots and nodes should be able to be moved around the tree and among documents without destroying the slot-to-node relationships established by `slot.assign()`. Nodes will be (re-)assigned to slots if/when they meet the criteria above.

@alice does this match your current understanding / expectation of how element reflection works as well?

dom.bs
mfreed7: Will do. I reached out to her privately to see what she thinks.

dom.bs
```diff
@@ -2203,14 +2214,28 @@ for a given <a>slot</a> <var>slot</var>, run these steps:</p>
...
+
+   <li><p>For each <a>slottable</a> in <var>slot</var>'s <a>manually assigned nodes</a>,
+   if <a>slottable</a>'s <a for=tree>parent</a> is <var>host</var>, add <a>slottable</a>
+   to <var>result</var>.</p></li>
+  </ol>
+ </li>
 
  <li>
-  <p>For each <a>slottable</a> <a for=tree>child</a> of <var>host</var>, <var>slottable</var>, in
+  <p>Otherwise, for each <a>slottable</a> <a for=tree>child</a> of <var>host</var>, <var>slottable</var>, in
```
rniwa, line 2238: I guess there is a question about what we're gonna do about slots which has non-empty manually assigned nodes inserted inside a shadow tree whose assignment mode is "name". Maybe clearing the manual assignment is the simplest thing to do in that case? It would be very mysterious if manually assigned nodes are completely ignored and it was treated as the default slot and/or it was silently ignored.

dom.bs
rniwa: Yeah, since "manual" is an adjective (not a noun in the sense of a book of instructions), "named" would make it more consistent with it. I guess an alternative is to make "manual" a noun like "script" or "none".

I've also started to think that maybe we want to say "slotname" instead of just "name" since if we wanted to add, for example, the [ability to assign nodes based on element's local names](https://github.com/whatwg/html/issues/6051) then it would be regrettable to have "name" and then "elementname" / "localname".

domenic: +1 for "named" (or "auto", but I guess we decided against that at some point in the past).

domenic (approved):
LGTM modulo naming of SlotAssignmentMode.

dom.bs
mfreed7: Sounds like we're roughly agreed that "named" might be better. I'll modify...

dom.bs
mfreed7: Hmm, I thought we were leaving these as-is. I.e. while that slot is located in a "named" mode shadow root, then "named" mode takes precedence. But nothing, other than re-assigning a node to another slot, removes anything from manually assigned nodes. I.e. `.assign()` is the only thing with the power to add/remove from manually assigned nodes. And no amount of node movement or slot movement can cause assignments to get cleared. I kind of feel like that's the easiest mental model for developers to have, rather than having to remember never to put a manually assigned slot into the wrong shadow root. It seems like a common pattern might be:

```javascript
shadow.appendChild(slot);
slot.assign([a,b,c]);
shadow.slotAssignment = 'named';
```

WDYT?

dom.bs

dom.bs
annevk: I don't understand the last line of your example. `slotAssignment` is readonly. I do tend to agree that `assign()` being the only way to clear seems more straightforward than having insertion in certain trees have side effects.

annevk:
Is there a WPT PR for dropping `.tentative.` from the tests? Do the tests cover the newer never-remove assigned nodes model?

Notes for the final commit message:

* Needs to close the older PR.
* Needs to give both authors committer credit.

Also note that all implementation bugs need to be filed before that criteria is met (unless there's a reason not to file them, but none seems to be given).

dom.bs
```diff
@@ -2175,6 +2175,13 @@ Unless stated otherwise it is null. A <a>slottable</a> is
 <dfn export for=slottable id=slotable-assigned>assigned</dfn> if its <a>assigned slot</a> is
 non-null.</p>
 
+<p>A <a>slottable</a> has an associated
+<dfn export for=slottable id=slottable-manual-slot-assignment>manual slot assignment</dfn> (null
+or a <a>slot</a>). Unless stated otherwise, it is null.</p>
```
annevk, line 2180: Nit I can fix before landing: generally we don't use optional closing tags in this document.

dom.bs
annevk, line 2238: for each slottable child _slottable_ of host,*

dom.bs
```diff
@@ -5746,6 +5777,9 @@ null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow roo
 <dfn for=ShadowRoot export><code>onslotchange</code></dfn> <a>event handler</a>, whose
 <a>event handler event type</a> is {{HTMLSlotElement/slotchange}}.
 
+<p>The <dfn attribute for=ShadowRoot><code>slotAssignment</code></dfn> attribute's getter must
```
annevk, line 5780: Nit I can fix: The X getter steps are to return*

(Once this lands I can take care of the other getters in this section. It seems slightly preferable to have this correct from the start for blame.)

dom.bs
mfreed7: Yeah sorry, good point. That's not a great example. So you'd prefer this?

```javascript
const shadow = host.attachShadow({mode: 'open'}); // "named" slot assignment
shadow.appendChild(slot);
slot.assign([a,b,c]); // Has no effect
```

I suppose I could live with this. It just feels weird that the entire point of all of this conversation was to make the node-to-slot assignments "sticky" across all kinds of tree mutations. But then this would be the one exception - you can move a node or a slot anywhere in any tree and the references will be maintained, **except** if any stop is ever within a "named"-assignment shadow root.

Given the fact that the node references are unobservable, there will definitely be cases that are a bit hard to understand for developers. E.g. assigning a node to a slot, and then making the node a grandchild (instead of direct child) of the host. We'll need to develop developer tooling to make this situation better. But I don't think adding even more magic is the answer. Again, I'm ok either way, just trying to get this right.

mfreed7: Sorry, re-reading your comment @annevk, it does sound like you might agree that `.assign()` should be the only way to clear/change node assignments.

I think this is the only remaining material issue to resolve.

mfreed7:
> Is there a WPT PR for dropping `.tentative.` from the tests? Do the tests cover the newer never-remove assigned nodes model?

I don't have that ready yet, but I'll own doing it. I have the WPT updates included as part of the [Chromium tracking bug](http://crbug.com/1196842). I'm just waiting until we nail down the exact behavior before I go implement it.

dom.bs
annevk: Yeah, we're on the same page. @rniwa can you take another look at this?

rniwa: So `assign(~)` has no side effect and its results get simply ignored? That's a bit strange but all other options are equally strange so maybe it's okay. I guess this might be something we want to solicit some developer feedback.

annevk: Oh no, it got confusing. That's not what Mason or I mean. That particular comment was based on a misunderstanding.

`assign()` always succeeds (and doesn't no-op or throw), as per the current HTML PR. And moving a slot around doesn't affect its manually assigned nodes, as per this PR.

And the answer to your question in this subthread OP is that manually assigned nodes would be ignored in that case as it would be quite esoteric if moving a `slot` element around is a no-op for manually assigned nodes except if you move it to a shadow tree with named slot assignment.

rniwa: Yeah, I mean that `assign(~)` will succeed but it won't have any observable side effect until the slot is moved to another shadow root with manual slot assignment along with the assigned nodes, right?

annevk: Ah yes, indeed. 😊

If we can get developer feedback in a timely fashion it seems reasonable to block on that, but otherwise I'd be inclined to land these PRs and count on the fact that making a change for that particular scenario is without risk and can be made once developers have some more experience with the API.

rniwa: Yeah, sounds good to me.

annevk:
Okay, so I think once we have bugs for Firefox and Safari all is in order here (and ditto for the HTML PR), assuming my editorial changes look good to you all. It's fine to handle the test renaming as part of a Chromium changeset.

mfreed7:
> Okay, so I think once we have bugs for Firefox and Safari all is in order here (and ditto for the HTML PR), assuming my editorial changes look good to you all. It's fine to handle the test renaming as part of a Chromium changeset.

Thanks! LGTM. I just updated both PRs with tracking bugs.

mfreed7:
Thanks everyone!

PR #989, opened 2021-06-09
Add "mode locked" flag to Document, plus an algorithm to use it

mfreed7:
This goes along with https://github.com/whatwg/html/pull/6745. This PR adds a "mode locked" flag to Document, which can be used to lock the [mode](https://dom.spec.whatwg.org/#concept-document-mode), so that it cannot be later changed.

***
<a href="https://whatpr.org/dom/989.html" title="Last updated on Jun 9, 2021, 10:30 PM UTC (748d3c0)">Preview</a> | <a href="https://whatpr.org/dom/989/f346858...748d3c0.html" title="Last updated on Jun 9, 2021, 10:30 PM UTC (748d3c0)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4268,8 +4268,8 @@ dom-Range-extractContents, dom-Range-cloneContents -->
   <dl class=switch>
    <dt>{{Document}}
    <dd><p>Set <var>copy</var>'s <a for=Document>encoding</a>, <a for=Document>content type</a>,
-   <a for=Document>URL</a>, <a for=Document>origin</a>, <a for=Document>type</a>, and
-   <a for=Document>mode</a>, to those of <var>node</var>.
+   <a for=Document>URL</a>, <a for=Document>origin</a>, <a for=Document>type</a>,
+   <a for=Document>mode</a>, and <a for=Document>mode locked</a> to those of <var>node</var>.
 
    <dt>{{DocumentType}}
    <dd><p>Set <var>copy</var>'s <a for=DocumentType>name</a>, <a>public ID</a>, and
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -4860,6 +4862,14 @@ null if <var>event</var>'s {{Event/type}} attribute value is "<code>load</code>"
 <a>document</a> does not have a <a for=Document>browsing context</a>; otherwise the
 <a>document</a>'s <a>relevant global object</a>.
 
+<p>A <a>document</a>'s <dfn export id=concept-document-set-the-mode>set the mode</dfn> algorithm,
+given a <var>new mode</var>, performs the following steps:
+
+<ol>
+ <li><p>If <a>this</a>'s <a for=Document>mode locked</a> is true, return.
```
annevk, line 4869: then return*

Add a newline here as well to separate from the following `<li>`.

dom.bs
```diff
@@ -4860,6 +4862,14 @@ null if <var>event</var>'s {{Event/type}} attribute value is "<code>load</code>"
 <a>document</a> does not have a <a for=Document>browsing context</a>; otherwise the
 <a>document</a>'s <a>relevant global object</a>.
 
+<p>A <a>document</a>'s <dfn export id=concept-document-set-the-mode>set the mode</dfn> algorithm,
```
annevk, line 4865: It might be slightly clearer to redo this as

> To set the document mode, given a document _document_ and a string _newMode_, ...

No need to add the ID for new `<dfn>`s. One will be generated.

annevk (commented):
How often can this state change? Can it only change once from no-quirks to either quirks or limited-quirks? An alternative might be having an "initial-no-quirks" state, but perhaps that requires too many changes elsewhere.

annevk:
Closing this as https://github.com/whatwg/html/pull/6745 went in a different direction.

PR #1000, opened 2021-07-08
Editorial: Update Event timeStamp initialization

npm1:
HR-Time now exposes a method that can be used to obtain the current high resolution time.

***
<a href="https://whatpr.org/dom/1000.html" title="Last updated on Jul 19, 2021, 2:39 PM UTC (3c569d3)">Preview</a> | <a href="https://whatpr.org/dom/1000/76a0b04...3c569d3.html" title="Last updated on Jul 19, 2021, 2:39 PM UTC (3c569d3)">Diff</a>


npm1:
@annevk can you take a look?

npm1:
Or @domenic perhaps

dom.bs
```diff
@@ -906,8 +906,9 @@ correct defaults.</p>
  <li><p>Set <var>event</var>'s <a>initialized flag</a>.
 
  <li>
-  <p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
-  representing the high resolution time from the <a>time origin</a> to <var>time</var>.
+  <p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to the
+  <a>relative high resolution coarse time</a> given <var>time</var> and <var>realm</var>'s
+  <a for=Realm>global object</a>.
```
domenic, line 911: _realm_ might be null. Probably the easiest way to handle this is to use _event_'s relevant global object, deferring the null-handling to that location.

npm1: Thanks, done

npm1:
Can I have another review please?

domenic:
It looks like from https://whatpr.org/dom/1000.html the link for "current high resolution time" goes to https://www.w3.org/TR/hr-time-2/#dfn-current-high-resolution-time which is not correct. It should probably go to https://w3c.github.io/hr-time/#dfn-current-high-resolution-time .

Since the spec appears to be a ReSpec one the only way to fix this is to add something to the anchors block, similar to what is already done in https://github.com/whatwg/dom/blob/76a0b048e8d30cdd79529eadfe8271d5eb6e00f0/dom.bs#L36-L39

npm1:
> It looks like from https://whatpr.org/dom/1000.html the link for "current high resolution time" goes to https://www.w3.org/TR/hr-time-2/#dfn-current-high-resolution-time which is not correct. It should probably go to https://w3c.github.io/hr-time/#dfn-current-high-resolution-time .
> 
> Since the spec appears to be a ReSpec one the only way to fix this is to add something to the anchors block, similar to what is already done in
> 
> https://github.com/whatwg/dom/blob/76a0b048e8d30cdd79529eadfe8271d5eb6e00f0/dom.bs#L36-L39

Done

dom.bs
```diff
@@ -906,8 +908,9 @@ correct defaults.</p>
  <li><p>Set <var>event</var>'s <a>initialized flag</a>.
 
  <li>
-  <p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to a {{DOMHighResTimeStamp}}
-  representing the high resolution time from the <a>time origin</a> to <var>time</var>.
+  <p>Initialize <var>event</var>'s {{Event/timeStamp}} attribute to the
+  <a>relative high resolution coarse time</a> given <var>time</var> and <var>event</var>'s
+  <a>relevant global object</a>.
 
   <p class=warning>User agents should set a minimum resolution of <var>event</var>'s
```
annevk, line 915: It seems we  can remove this at which point the remaining p needs some rewrapping as it's the sole child of the li.

npm1: Done

domenic (approved):

annevk:
Thanks @npm1! (I ended up removing "Editorial:" as I think it represents a significant change, now it aligns with cross-origin isolation.)

Issue #770, opened 2019-06-27
Add note that "A ShadowRoot node has no parent."


mbrodesser:
The spec:

https://dom.spec.whatwg.org/commit-snapshots/b5eac4d624b229232e8c2e347d6b532172f73439/#interface-node

already notes:
> An Attr node has no parent\. 

It'd be helpful if the note additionally contained "A ShadowRoot node has no parent.".

annevk:
I'm going to remove these Attr node notes, but I'll fix https://dom.spec.whatwg.org/#node-trees instead. That seems like a better place to put all the information about parent/child relationships nodes can have. It already seems adequate for DocumentFragment/ShadowRoot, but #1004 will make it better still.

Issue #719, opened 2018-12-02
"If node is a Text node"


domenic:
https://dom.spec.whatwg.org/commit-snapshots/23ec6421f47ab629bd14dfc6d4910414dc443d0b/#mutation-algorithms

> If _node_ is a `Text` node, ...

Apparently, per tests and browser behavior, this includes both `Text` and `CDATASection` nodes. While I might say that both are "text nodes", the current spec's typography of "`Text` node" is a bit misleading, implying it would not include CDATA section nodes.

I'd like to suggest that the spec try to be as clear as possible and use either "text node" or (better) "inclusive text node" when referring to both `Text` and `CDATASection` nodes.

Via https://github.com/jsdom/jsdom/pull/2441, where @Zirro remarks that we might not have test coverage for very many other places in the spec that have this problem.

annevk:
See #636 and in particular #597.

I really want [[PlatformBrand]] in IDL before trying to fix this.

domenic:
I would really rather not block on that, especially since I consider it a dubious and over-formal solution. We don't need all that machinery just to change "Text node" to "Text or CDATASection node" or similar.

Issue #636, opened 2018-04-16
Editorial: note that CDATASections are Texts inside switches


domenic:
I made a mistake and thought CDATASection was "Any other node", but actually it's a Text node. So in switches like textContent and nodeValue, I'd suggest changing "`Text`" to "`Text` (including `CDATASection`)"

Thoughts?

annevk:
See #597. Note that if we did this, how should we handle `Element`, which has many subclasses? I guess they're not different node types, but it does seem somewhat inconsistent in a way.

domenic:
Right, the mistake I made was switching on node type, which is probably easier than switching on inherited-from-ness in most implementations.

And yeah, no web platform tests seemed to catch this either.

annevk:
Maybe that's a good solution in the absence of [[PlatformBrand]]. Make "type" an internal slot of nodes and branch on that everywhere.

Issue #597, opened 2018-03-13
Lack of [[PlatformBrand]] is biting us


annevk:
Consider https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity. Step 4 bans `DocumentFragment`. Does it ban `ShadowRoot`? If it does, that's fine, but also a problem, since it shouldn't ban `CDATASection`.

I thought we had an issue tracking this, but it seems like we don't.

This is somewhat related to #457, which is about updating various algorithms to account for `Attr` nodes, which can now be passed to them again (and not all are anticipating that).

annevk:
([[PlatformBrand]] is https://github.com/heycam/webidl/issues/97 by the way, which would give us a way more accurate way of describing what "_instance_ is an _interface_" actually means.)

dwimmer90:
Hi,

Do you know how I can be removed from this group?

Danielle Wimmer

annevk:
Use "Unwatch" in the GitHub UI (cannot be done via email).

annevk:
I guess just like "exclusive Text node" we should have "exclusive DocumentFragment node" for now and use that. And add some tests.

annevk:
Maybe this is not a problem. It seems Chrome and Safari happily treat a ShadowRoot as a DocumentFragment and don't really branch on it when appending. Probably still worth going through all the places, but this might be all mostly fine. (Would be good to check test coverage though.)
```
<!DOCTYPE html>
...<script>
x = document.createElement("div");
s = x.attachShadow({mode: "closed"});
is = s.appendChild(document.createElement("blah"));
document.body.appendChild(s);
</script>
```

annevk:
I started looking into this to see if I could solve https://github.com/heycam/webidl/issues/659 (and also https://github.com/whatwg/dom/issues/636 and https://github.com/whatwg/dom/issues/719) and unfortunately it's not super straightforward.

1. I noticed that many instances of the term "node" reference the wrong definition. I have a patch for this (lacking `for=/`).
2. I noticed that we talk about Attr nodes, despite Attr not being defined as a node. 
3. If we go to the language of implements X, it doesn't really make sense anymore to distinguish between "abstract" interfaces and "concrete" ones. Or does it?

I'll upload a PR that takes a stab at this for further discussion.

PR #1004, opened 2021-08-03
Revamp the way DOM talks about nodes

annevk:
1. Address numerous xref issues around the term node and prevent some from occurring for the term children.
2. Embrace Attr as a node even more.
3. Rely on Web IDL's implements and interface primitives to reduce the opportunities for confusion.

Fixes https://github.com/heycam/webidl/issues/659, #597, #636, #719, and #770.

***
<a href="https://whatpr.org/dom/1004.html" title="Last updated on Aug 31, 2021, 11:08 AM UTC (1b388b0)">Preview</a> | <a href="https://whatpr.org/dom/1004/f2a2ded...1b388b0.html" title="Last updated on Aug 31, 2021, 11:08 AM UTC (1b388b0)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2023,9 +2023,10 @@ can be used to explore this matter in more detail.
 </div>
 
 <p>For brevity, this specification refers to an object that <a>implements</a> {{Node}} and an
-inherited interface <var>NodeInterface</var>, as a <var>NodeInterface</var> <a for=/>node</a>.
+inherited interface <code><var>NodeInterface</var></code>, as a
+<code><var>NodeInterface</var></code> <a for=/>node</a>.
 
-<p>A <a>node tree</a> is constrained as follows, expressed as a relationship between an interface of
+<p>A <a>node tree</a> is constrained as follows, expressed as a relationship between a
 <a for=/>node</a> and its allowed <a for=tree>children</a>:
 
 <dl>
@@ -2353,8 +2354,8 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
  <li><p>If <var>child</var> is non-null and its <a for=tree>parent</a> is not <var>parent</var>,
  then <a>throw</a> a "{{NotFoundError!!exception}}" {{DOMException}}.
 
- <li><p>If <var>node</var> is not a {{DocumentFragment}}, {{DocumentType}}, {{Element}}, {{Text}},
- {{ProcessingInstruction}}, or {{Comment}} <a for=/>node</a>, then <a>throw</a> a
+ <li><p>If <var>node</var> is not a {{DocumentFragment}}, {{DocumentType}}, {{Element}}, or
+ {{CharacterData}} <a for=/>node</a>, then <a>throw</a> a
  "{{HierarchyRequestError!!exception}}" {{DOMException}}.
 
  <li><p>If either <var>node</var> is a {{Text}} <a for=/>node</a> and <var>parent</var> is a
@@ -2542,8 +2543,8 @@ within a <var>parent</var>, run these steps:
  <li><p>If <var>child</var>'s <a for=tree>parent</a> is not <var>parent</var>, then <a>throw</a> a
  "{{NotFoundError!!exception}}" {{DOMException}}.
 
- <li><p>If <var>node</var> is not a {{DocumentFragment}}, {{DocumentType}}, {{Element}}, {{Text}},
- {{ProcessingInstruction}}, or {{Comment}} <a for=/>node</a>, then <a>throw</a> a
+ <li><p>If <var>node</var> is not a {{DocumentFragment}}, {{DocumentType}}, {{Element}}, or
+ {{CharacterData}} <a for=/>node</a>, then <a>throw</a> a
  "{{HierarchyRequestError!!exception}}" {{DOMException}}.
 
  <li><p>If either <var>node</var> is a {{Text}} <a for=/>node</a> and <var>parent</var> is a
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):
Reviewed source and tested/reviewed Bikeshed output

dom.bs
```diff
@@ -2001,54 +2001,69 @@ can be used to explore this matter in more detail.
...
+ as {{HTMLAnchorElement}}.
+
+ <p>Objects that implement {{CharacterData}} also implement an inherited interface: {{Text}},
+ {{ProcessingInstruction}}, or {{Comment}}.
+
+ <p>Objects that implement {{Text}} sometimes implement {{CDATASection}}.
+</div>
+
+<p>For brevity, this specification refers to an object that <a>implements</a> {{Node}} and an
+inherited interface <var>NodeInterface</var>, as a <var>NodeInterface</var> <a for=/>node</a>.
```
domenic, line 2026: Hmm maybe `<code><var>NodeInterface</var></code>`?

dom.bs
```diff
@@ -2001,54 +2001,69 @@ can be used to explore this matter in more detail.
...
+
+<p>A <a>node tree</a> is constrained as follows, expressed as a relationship between an interface of
```
domenic, line 2028:
```suggestion
<p>A <a>node tree</a> is constrained as follows, expressed as a relationship between a
```

?

domenic (commented):
Looks good!

dom.bs
```diff
@@ -4071,35 +4041,30 @@ return <a>this</a>'s <a>shadow-including root</a> if
 
 <!-- TODO: domintro -->
 
-The <dfn attribute for=Node>nodeValue</dfn> attribute
-must return the following, depending on <a>this</a>:
+<p>The <dfn attribute for=Node>nodeValue</dfn> getter steps are to return the following, switching
+on <a>this</a>:
```
domenic, line 4045: The switching is still a bit weird. It might be best to do this as numbered steps so that you can say "this implements Attr / this implements CharacterData"

dom.bs
```diff
@@ -2353,8 +2354,8 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
  <li><p>If <var>child</var> is non-null and its <a for=tree>parent</a> is not <var>parent</var>,
  then <a>throw</a> a "{{NotFoundError!!exception}}" {{DOMException}}.
 
- <li><p>If <var>node</var> is not a {{DocumentFragment}}, {{DocumentType}}, {{Element}}, {{Text}},
- {{ProcessingInstruction}}, or {{Comment}} <a for=/>node</a>, then <a>throw</a> a
+ <li><p>If <var>node</var> is not a {{DocumentFragment}}, {{DocumentType}}, {{Element}}, or
+ {{CharacterData}} <a for=/>node</a>, then <a>throw</a> a
```
domenic, line 2358: "does not implement"?

dom.bs
annevk: Changing all instances of X node to that would be a rather massive undertaking and require changes in a lot more places. I was hoping that the new definition of node would avoid the need for doing that, but I wouldn't object to someone making that change.

domenic: It's specifically the switches that are strange. E.g. in most programming languages if you do `switch (this) { case X: ... }` that compares `this === X`. But here you seem to be instead doing a `this instanceof X` comparison, which is not how switches usually work.

annevk: Okay, your comment below made me think it was both. I'll look at reworking the switches tomorrow.

domenic: Well, I think implements would be an improvement, but I can see not wanting to tackle that.

dom.bs
```diff
@@ -2001,54 +2001,74 @@ can be used to explore this matter in more detail.
...
-<a>node</a> and its allowed <a>children</a>:
+<div class=note>
+ <p>In practice you deal with more specific objects.
+
+ <p>Objects that <a>implement</a> {{Node}} also implement an inherited interface: {{Document}},
+ {{DocumentType}}, {{DocumentFragment}}, {{Element}}, {{CharacterData}}, or {{Attr}}.
+
+ <p>Objects that implement {{DocumentFragment}} sometimes implement {{ShadowRoot}}.
+
+ <p>Objects that <a>implement</a> {{Element}} also typically implement an inherited interface, such
```
domenic, line 2016: The inconsistent linking of "implement" here is a bit confusing. I'd suggest either in every sentence in this note, or in just the first instance.

dom.bs
```diff
@@ -2001,54 +2001,74 @@ can be used to explore this matter in more detail.
...
- {{Comment}}.
+ <dd><p>Zero or more {{Element}} or {{CharacterData}} <a for=/>nodes</a>.
+
  <dt>{{DocumentType}}
- <dt>{{Text}}
- <dt>{{ProcessingInstruction}}
- <dt>{{Comment}}
+ <dt>{{CharacterData}}
+ <dt>{{Attr}}
  <dd><p>None.
```
domenic, line 2051: "No allowed children"? I had to go back and re-read the intro paragraph to see what you meant.

dom.bs
```diff
@@ -2001,54 +2001,74 @@ can be used to explore this matter in more detail.
...
 </dl>
 
-<p>To determine the <dfn export id=concept-node-length for=Node>length</dfn> of a <a>node</a>
-<var>node</var>, switch on <var>node</var>:
+<p class=note>{{Attr}} <a for=/>nodes</a> <a>participate</a> in a <a>tree</a> for historical
+reasons; they never have a <a for=tree>parent</a> or <a for=tree>children</a> and are therefore
```
domenic, line 2055: I think "their parent is always null" is a bit more technically accurate. Because https://whatpr.org/dom/1004.html#concept-tree-participate says "An object that participates in a tree has a parent", which seems to contradict this.

I guess technically "never have children" should be "always have an empty set for their children", but that one bothers me much less for some reason.

domenic (approved):
LGTM with nits

PR #1009, opened 2021-08-27
Add validity criteria for StaticRange

### LINKED ISSUE #947: Define criteria for StaticRange validation
Because static ranges are not kept up-to-date during DOM mutations, UAs/authors need to validate the static ranges before actually using them. The criteria for checking if a StaticRange is valid seems to be:
- The start and end containers of the static range should be in the same (non-flat) DOM tree.
- The start offset is between 0 and the start container’s node length (inclusive).
- The end offset is between 0 and the end container’s node length (inclusive).
- The position of the start boundary point should be before or equal to the position of the end boundary point.

My questions are:
(1) Should we include these criteria as part of the StaticRange spec so that all UAs and authors can follow the same?
(2) If the answer to (1) is yes, I wonder if it is also worth having a public DOM API that does this, something like: `StaticRange.IsValid`?

Note that I'm only talking about specifying _what_ the validity criteria should be, not _when_ these validity checks should be run. The _when_, in my opinion, should be left up to the individual use case. (Ideally, validity is checked on-demand, only when necessary.)

dandclark:
Add criteria defining whether a StaticRange is valid.

This will be used by the CSS [Highlight API](https://drafts.csswg.org/css-highlight-api-1/) to determine whether or not to paint highlights specified by StaticRanges. See https://drafts.csswg.org/css-highlight-api-1/#range-invalidation (which will be updated to point to the DOM spec definition) and the CSSWG [resolution](https://github.com/w3c/csswg-drafts/issues/4597#issuecomment-905688814) that Highlight API should use StaticRanges internally rather than backing them with live Ranges.

Resolves #947.

   * Chromium has implemented Highlight API based on StaticRange using this validity criteria. 
   * Webkit has implemented an experimental version of Highlight API, and agreed to the use of StaticRange per the [resolution notes](https://github.com/w3c/csswg-drafts/issues/4597#issuecomment-905688814).
   * https://github.com/web-platform-tests/wpt/blob/master/css/css-highlight-api/painting/
     * Note, the validity criteria aren't exposed directly, but these Highlight API painting tests depend on the validity criteria to 
       determine when to paint or paint a given StaticRange. See in particular custom-highlight-painting-invalidation-*.html
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed (for Highlight API):
   * Chrome: [crbug.com/1164461](https://crbug.com/1164461)
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1703961
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=204903

***
<a href="https://whatpr.org/dom/1009.html" title="Last updated on Aug 31, 2021, 10:58 AM UTC (2aeabe9)">Preview</a> | <a href="https://whatpr.org/dom/1009/f2a2ded...2aeabe9.html" title="Last updated on Aug 31, 2021, 10:58 AM UTC (2aeabe9)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -7665,6 +7665,7 @@ constructor steps are:
  <a for="boundary point">equal</a> to its <a for=range>end</a>.
 </ul>
 
+
 <h3 id=interface-range>Interface {{Range}}</h3>
 
 <pre class=idl>
--- a/dom.bs
+++ b/dom.bs
@@ -7650,8 +7650,7 @@ constructor steps are:
  <var>init</var>["{{StaticRangeInit/endOffset}}"]).
 </ol>
 
-<p>A {{StaticRange}} is <dfn for=StaticRange export>valid</dfn> if and only if all of the following
-conditions are true:
+<p>A {{StaticRange}} is <dfn for=StaticRange export>valid</dfn> if all of the following are true:
 
 <ul>
  <li><p>Its <a for=range>start</a> and <a for=range>end</a> are in the same <a>node tree</a>.
--- a/dom.bs
+++ b/dom.bs
@@ -7650,6 +7650,21 @@ constructor steps are:
  <var>init</var>["{{StaticRangeInit/endOffset}}"]).
 </ol>
 
+<p>A {{StaticRange}} is <dfn for=StaticRange export>valid</dfn> if and only if all of the following
+conditions are true:
+
+<ul>
+ <li><p>Its <a for=range>start</a> and <a for=range>end</a> are in the same <a>node tree</a>.
+
+ <li><p>Its <a for=range>start offset</a> is between 0 and its <a for=range>start node</a>'s
+ <a>length</a>, inclusive.
+
+ <li><p>Its <a for=range>end offset</a> is between 0 and its <a for=range>end node</a>'s
+ <a>length</a>, inclusive.
+
+ <li><p>Its <a for=range>start</a> is <a for="boundary point">before</a> or
+ <a for="boundary point">equal</a> to its <a for=range>end</a>.
+</ul>
 
 <h3 id=interface-range>Interface {{Range}}</h3>
 
@@ -10013,6 +10028,7 @@ Chris Paris,
 Chris Rebert,
 Cyrille Tuzi,
 Dan Burzo,
+Daniel Clark,
 Daniel Glazman,
 Darin Fisher,
 David Bruant,


dom.bs
```diff
@@ -7650,6 +7650,21 @@ constructor steps are:
  <var>init</var>["{{StaticRangeInit/endOffset}}"]).
 </ol>
 
+<p>A {{StaticRange}} is <dfn for=StaticRange export>valid</dfn> if and only if all of the following
+conditions are true:
```
annevk, line 7654: I'd suggest to shorten this to

> if all of the following are true:

annevk (commented):
It would be good to clarify that https://github.com/w3c/csswg-drafts/issues/4597#issuecomment-908223745 is intentional before we move ahead with this, but overall this seems reasonable to me.

Have any measurements been done to show that this is indeed faster?

cc @smaug----

dandclark:
> It would be good to clarify that [w3c/csswg-drafts#4597 (comment)](https://github.com/w3c/csswg-drafts/issues/4597#issuecomment-908223745) is intentional before we move ahead with this, but overall this seems reasonable to me.

Commented [here](https://github.com/w3c/csswg-drafts/issues/4597#issuecomment-908512815).

> Have any measurements been done to show that this is indeed faster?
> 
> cc @smaug----

Yes, @ffiori built some benchmarks for this and ran them on Chromium's Highlight API prototype, with the findings  [documented here](https://docs.google.com/document/d/1byMUPIcx2cldNlppyvRP3HX9w1NIryO_F7mgQNytMzs/edit#heading=h.7nki9mck5t64). As expected, adding a large number of live Ranges to a document increases the cost of DOM node moves due to the cost of keeping the Ranges up-to-date for every DOM operation. For StaticRange, these costs were avoided, and since it's possible to cache the results of the validity checks StaticRanges were in some cases significantly fater, and never significantly slower.

smaug----:
Use of StaticRanges forces one to use MutationObserver or other mechanisms to detect mutations in order to keep StaticRanges up-to-date all the time, so the tests were very much unrealistic.
(Also, live Ranges need to observe changes only in the subtree they live in
"The Document stores a collection of pointers to all the Ranges that are alive inside it.
Whenever there’s a DOM mutation that involves deletion/addition of nodes or text for example, all these Ranges are notified about the mutation." smells like some implementations have rather heavyweight Ranges, but that is an implementation detail.)

 But anyhow, this change as such is fine.

dom.bs
```diff
@@ -7650,6 +7650,20 @@ constructor steps are:
...
+
+ <li><p>Its <a for=range>start offset</a> is between 0 and its <a for=range>start node</a>'s
+ <a>length</a>, inclusive.
+
+ <li><p>Its <a for=range>end offset</a> is between 0 and its <a for=range>end node</a>'s
+ <a>length</a>, inclusive.
+
+ <li><p>Its <a for=range>start</a> is <a for="boundary point">before</a> or
+ <a for="boundary point">equal</a> to its <a for=range>end</a>.
+</ul>
```
annevk, line 7666:
```suggestion
</ul>

```

annevk (approved):

PR #1027, opened 2021-10-18
Add abort reason to AbortSignal

nidhijaju:
Based on the discussion in https://github.com/whatwg/streams/issues/1165, it was suggested that we add an abort reason property to [AbortSignal](https://dom.spec.whatwg.org/#abortsignal). This PR makes the initial changes to the base primitives in the DOM spec, before we make changes to all other related specs that use the AbortSignal.

@domenic Is it okay if you please help take a look?
/cc @yutakahirano

   * Google
   * Mozilla
   * https://github.com/web-platform-tests/wpt/pull/31291
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1263410
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1737771
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=232299

***
<a href="https://whatpr.org/dom/1027.html" title="Last updated on Nov 4, 2021, 5:24 AM UTC (723f5e8)">Preview</a> | <a href="https://whatpr.org/dom/1027/0715c2f...723f5e8.html" title="Last updated on Nov 4, 2021, 5:24 AM UTC (723f5e8)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1114,8 +1114,8 @@ participate in a tree structure.</p>
  <a>service worker events</a>, then <a>report a warning to the console</a> that this might not give
  the expected results. [[!SERVICE-WORKERS]]
 
- <li><p>If <a for="event listener">signal</a> is not null and is [=AbortSignal/aborted=], then
- return.
+ <li><p>If <var>listener</var>'s <a for="event listener">signal</a> is not null and is
+ [=AbortSignal/aborted=], then return.
 
  <li><p>If <var>listener</var>'s <a for="event listener">callback</a> is null, then return.
 
@@ -1741,11 +1741,11 @@ interface AbortController {
  <dt><code><var>controller</var> . <a attribute for=AbortController>signal</a></code>
  <dd>Returns the {{AbortSignal}} object associated with this object.
 
- <dt><code><var>controller</var> . <a method for=AbortController lt=abort()>abort</a>
- (<var>reason</var>)</code>
+ <dt><code><var>controller</var> . <a method for=AbortController lt=abort()>abort</a>(<var>reason</var>)</code>
  <dd>Invoking this method will store <var>reason</var> in this object's {{AbortSignal}}'s
  [=AbortSignal/abort reason=], and signal to any observers that the associated activity is to be
- aborted.
+ aborted. If <var>reason</var> is undefined, then an "{{AbortError!!exception}}" {{DOMException}}
+ will be stored.
 </dl>
 
 <p>An {{AbortController}} object has an associated <dfn for=AbortController>signal</dfn> (an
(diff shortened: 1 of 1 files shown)


domenic:
Hmm, I wonder if the property should be `signal.reason` or `signal.abortReason`... Given that `AbortSignal` is being used as a base class for things like `TaskSignal`, and that the variable is usually named `signal` instead of `abortSignal`, spelling it out as `abortReason` might be a good idea.

nidhijaju:
Thank you @jasnell! Do you have any opinions on the naming of the abort reason by the way (i.e. `signal.reason` vs. `signal.abortReason`)?

jasnell:
I think just 'reason' for the name works

annevk:
@nidhijaju can you update OP to indicate support from Google (I'm assuming that's the case) and Mozilla? To be clear, I'm still happy to land this ahead of the Fetch PR as we have a good idea of what to do there, but the tests do need updating still.

@cdumez does this seem reasonable to WebKit?

cdumez:
What concerns me is that reason has type "any" and not DOMString. Why is that?

annevk:
@cdumez the idea is that it's a generic mechanism, similar to `throw`, `Promise`, the various stream objects, the `postMessage()` family, etc.

cdumez:
> @cdumez the idea is that it's a generic mechanism, similar to `throw`, `Promise`, the various stream objects, the `postMessage()` family, etc.

Ok, seems fine to me then.

nidhijaju:
> @nidhijaju can you update OP to indicate support from Google (I'm assuming that's the case) and Mozilla? To be clear, I'm still happy to land this ahead of the Fetch PR as we have a good idea of what to do there, but the tests do need updating still.

Updated 😊 Thank you!
I'll update the tests very soon as well.

nidhijaju:
Hi @annevk, the tests have been updated. Just wondering if you think this PR is ready to merge now?

annevk:
Hey @nidhijaju, thanks for the reminder and congratulations on adding this feature to the DOM Standard! This looks like it will be a very useful primitive for aborting.

nidhijaju:
Thank you @annevk and everyone else who helped patiently review the PR as well! :)

dom.bs
```diff
@@ -1762,43 +1764,51 @@ constructor steps are:
...
 
  <dt><code><var>signal</var> . <a attribute for=AbortSignal>aborted</a></code>
  <dd>Returns true if this {{AbortSignal}}'s {{AbortController}} has signaled to abort; otherwise
  false.
+
+ <dt><code><var>signal</var> . <a attribute for=AbortSignal>reason</a></code>
+ <dd>Returns this {{AbortSignal}}'s <a for=AbortSignal>abort reason</a>.
 </dl>
 
-<p>An {{AbortSignal}} object has an associated <dfn export for=AbortSignal>aborted flag</dfn>. It is
```
jyasskin, line 1790: FYI, this broke the builds of specs that use the aborted flag. No objections to the change; we just need to keep looking for better ways to do this sort of migration.

annevk: Let's continue to discuss in #1030.

Issue #920, opened 2020-11-10
AbortController, AbortSignal, and garbage collection


annevk:
@smaug---- made me realize we could probably be a little bit more specific about what keeps what alive.

Thoughts:

* AbortController weakly references AbortSignal. The controller getting GC'd is fine, since it doesn't learn about results.
* AbortSignal cannot be collected when it has a controller and event listeners/abort algorithms. I think that means it needs a weak reference to its controller.

cc @jakearchibald @domenic

jakearchibald:
The controller already has a strong reference to the signal, no? Via `controller.signal`. I guess that value could be overridden, but do we generally try to cater for cases like that?

annevk:
Sorry, yeah, the controller keeps the signal alive, but not the other way around. I'll fix OP.

jakearchibald:
Abort algorithms need to stick around while the controller is in reference and the signal is not aborted.

Abort events listeners need to stick around while the signal is in reference, or the controller is in reference and the signal is not aborted. But some of that seems redundant given the controller has a strong reference to the signal.

annevk:
Oh yeah, I think the basic scenarios are fine already, indeed. My bad.

I guess the one thing that complicates things somewhat might be signals that follow other signals.

MattiasBuelens:
> I guess the one thing that complicates things somewhat might be signals that follow other signals. Edit: would that be as simple as keeping it alive from the signal that it is following? At least as long as the aborted flag isn't set.

Again, isn't that already the case by virtue of the *followingSignal* attaching abort steps to its *parentSignal*, and those steps necessarily keeping the *followingSignal* alive?

MattiasBuelens:
On a slightly unrelated note: what **is** missing here is if the *followingSignal* is aborted separately (i.e. through its own abort controller, not as a result of the *parentSignal* becoming aborted). In that case, the *followingSignal* should remove the abort steps it had attached to its *parentSignal*, so that the *parentSignal* does not keep the *followingSignal* alive unnecessarily.

I don't think this shouldn't be too difficult to fix in the specification. I'll give it a try.

annevk:
Hmm yeah. I'm not sure you can ever be a _followingSignal_ with your own controller. Fetch uses this mechanism and always creates an "unowned" signal for this (as part of creating a new `Request` object).

MattiasBuelens:
[...]

PR #1032, opened 2021-11-19
Introduce AbortSignal.timeout()

### LINKED ISSUE #951: Add AbortSignal.timeout(ms)
This has come up in the context of fetch in the past: see https://github.com/whatwg/fetch/issues/951#issuecomment-541369940

And it has recently come up for streams as well: https://github.com/WICG/serial/issues/122

The proposal is essentially

```js
AbortSignal.timeout = ms => {
  setTimeout(() => controller.abort(), ms);
  return controller.signal;
};
```

at a high level. Some low-level details:

- I think we should share the timer task source with `setTimeout()`.
- I think we should not include the clamping/nesting behavior of `setTimeout()`, although I'm open to being persuaded otherwise.
- We should use `[EnforceRange] unsigned long long` so that you get a quick exception on negative or too-large numbers, instead of clamping negative numbers to zero and taking too-large numbers modulo 2**32.

I'd be happy to write spec text for this if there's implementer interest. Maybe @josepharhar would be interested in implementing this too, similar in spirit to the other recent `AbortSignal` integration?

domenic:
Closes #951.

### Mini-explainer for TAG review/blink-dev purposes:

This introduces a new static method, `AbortSignal.timeout(milliseconds)`, which returns a newly-created `AbortSignal` that, after `milliseconds` milliseconds, becomes automatically aborted. The `AbortSignal`'s `reason` property will be a `"TimeoutError"` `DOMException`.

The main motivating use case for this is helping web developers easily time out async operations, such as `fetch()`. For example, now you can write:

```js
fetch(url, { signal: AbortSignal.timeout(10_000) });
```

to ensure your fetch operation has a 10-second timeout. A more full example would be:

```js
try {
  const res = await fetch(url, { signal: AbortSignal.timeout(10_000) });
  const result = await res.text();
  // ...
} catch (e) {
  if (e.name === "TimeoutError") {
    // It took more than 10 seconds to get the result!
  } else if (e.name === "AbortError") {
    // The fetch was explicitly aborted, e.g. by the user pressing the stop button!
  } else {
    // Something horrible went wrong, like a network error!
  }
}
```

This has come up in the context of several specs; see discussions on fetch at https://github.com/whatwg/fetch/issues/951 and streams at https://github.com/WICG/serial/issues/122#issuecomment-785427831. It is generally helpful for any operation that takes an `AbortSignal`, including web-developer-defined operations.

Interesting points of discussion are:

- It is specifically important that the reason be a `"TimeoutError"` `DOMException`, instead of the default-for-`AbortSignal` `"AbortError"` `DOMException`. This is because developers almost always want to discriminate between those two: a timeout is a proper failure which they might show to users, whereas an abort usually more of a "third state" (alongside success and failure) which usually they don't want to tell users about. (E.g., because the user is the one that caused the abort in the first place.)

- We treat the `milliseconds` argument the same as [`scheduler.postTask()`](https://wicg.github.io/scheduling-apis/#dom-scheduler-posttask), and almost the same as `setTimeout()`. (The difference is, we don't have the nested-clamping-to-4-ms behavior that `setTimeout()` does.) In particular this means that the timer does *not* advance while the document is in bfcache, or the worker is suspended. See https://github.com/whatwg/dom/pull/1032#issuecomment-976900712 for the reasoning there.

Future work: this feature would benefit greatly from a solution for combining `AbortSignal`s, so that people could have an operation that aborts on _either_ a timeout _or_ an explicit abort. For example, something like this:

```js
// NOT REAL CODE, YET:
fetch(url, { signal: AbortSignal.all([AbortSignal.timeout(10_000), controller.signal]) });

abortButton.onclick = () => controller.abort();
```

Unfortunately there are a lot of possible shapes for such an API and so we still need to do a bit more work and research to find the best one. That's tracked in https://github.com/whatwg/dom/issues/920. Until then, people can combine such signals manually, as discussed in that issue thread.

   * [Chromium](https://github.com/whatwg/dom/issues/951#issuecomment-785432044)
   * [Gecko](https://github.com/whatwg/dom/pull/1032#pullrequestreview-811812958)
   * https://github.com/web-platform-tests/wpt/pull/32622
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1181925
   * Firefox: TODO
   * Safari: TODO

----

<details>
<summary>Now-obsolete discussion of "Subtle things to deal with"</summary>

* Need to export "timer task source" from HTML
* What portions of the delay logic from [setTimeout](https://html.spec.whatwg.org/#timer-initialisation-steps) and [postTask](https://wicg.github.io/scheduling-apis/#schedule-a-posttask-task) should we port over here?
  * I so far ported: wait milliseconds; ensure ordering within a global; and allow implementation-defined extra waiting (e.g. to align on CPU wakeups or allow throttling while in the background).
  * I'm on the fence as to whether we should port suspending the timer while the document is in bfcache or the worker is suspended. For e.g. fetch timeouts, it doesn't seem to matter much whether you were in bfcache/suspended; if the server took 30 wall-clock seconds to respond, you probably want to give up. On the other hand, it might be strange (or hard to implement) for this timer mechanism to behave differently from others on the platform.
  * Should these timers impact [idle callback deadline computation](https://html.spec.whatwg.org/#event-loop-processing-model:map-of-active-timers)? postTask timers do not, but that may be an oversight. @shaseley @noamr
  * I think that, like postTask, we should not port the nesting and clamping messiness from setTimeout. The implementation-defined extra time is enough.
* Probably it's time to centralize this delay logic so that this spec, HTML, and postTask can reuse it. But we'd need to decide on the above questions first.
</details>

***
<a href="https://whatpr.org/dom/1032.html" title="Last updated on Dec 6, 2021, 10:09 AM UTC (7df51fe)">Preview</a> | <a href="https://whatpr.org/dom/1032/982c7e2...7df51fe.html" title="Last updated on Dec 6, 2021, 10:09 AM UTC (7df51fe)">Diff</a>

***
<a href="https://whatpr.org/dom/1032.html" title="Last updated on Feb 2, 2022, 10:07 PM UTC (7df51fe)">Preview</a> | <a href="https://whatpr.org/dom/1032/982c7e2...7df51fe.html" title="Last updated on Feb 2, 2022, 10:07 PM UTC (7df51fe)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1773,6 +1773,7 @@ to <a for=AbortSignal>signal abort</a> on <a>this</a>'s <a for=AbortController>s
 [Exposed=(Window,Worker)]
 interface AbortSignal : EventTarget {
   [NewObject] static AbortSignal abort(optional any reason);
+  [NewObject] static AbortSignal timeout([EnforceRange] unsigned long long milliseconds);
 
   readonly attribute boolean aborted;
   readonly attribute any reason;
@@ -1784,6 +1785,11 @@ interface AbortSignal : EventTarget {
  <dt><code>AbortSignal . <a method for=AbortSignal>abort</a>(<var>reason</var>)</code>
  <dd>Returns an {{AbortSignal}} instance whose <a for=AbortSignal>abort reason</a> is set to
  <var>reason</var> if not undefined; otherwise to an "{{AbortError!!exception}}" {{DOMException}}.
+
+ <dt><code>AbortSignal . <a method for=AbortSignal lt=timeout(milliseconds)>timeout</a>(<var>milliseconds</var>)</code>
+ <dd>Returns an {{AbortSignal}} instance which will be aborted in <var>milliseconds</var>
+ milliseconds. Its <a for=AbortSignal>abort reason</a> will be set to a
+ "{{TimeoutError!!exception}}" {{DOMException}}.
 
  <dt><code><var>signal</var> . <a attribute for=AbortSignal>aborted</a></code>
  <dd>Returns true if this {{AbortSignal}}'s {{AbortController}} has signaled to abort; otherwise
@@ -1832,6 +1838,31 @@ are:
  otherwise to a new "{{AbortError!!exception}}" {{DOMException}}.
 
  <li>Return <var>signal</var>.
+</ol>
+
+<p>The static <dfn method for=AbortSignal><code>timeout(<var>milliseconds</var>)</code></dfn> method
+steps are:
+
+<ol>
+ <li><p>Let <var>signal</var> be a new {{AbortSignal}} object.
+
+ <li><p>Let <var>global</var> be <var>signal</var>'s <a>relevant global object</a>.
+
+ <li>
+  <p><a>Run steps after a timeout</a> given <var>global</var>, "<code>AbortSignal-timeout</code>",
+  <var>milliseconds</var>, and the following step:</p>
+
+  <ol>
+   <li><p><a>Queue a global task</a> on the <a>timer task source</a> given <var>global</var> to
+   <a for=AbortSignal>signal abort</a> given <var>signal</var> and a new
+   "{{TimeoutError!!exception}}" {{DOMException}}.
+  </ol>
+
+  <p>For the duration of this timeout, if <var>signal</var> has any event listeners registered for
+  its {{AbortSignal/abort}} event, there must be a strong reference from <var>global</var> to
+  <var>signal</var>.
+
+ <li><p>Return <var>signal</var>.
 </ol>
 
 <p>The <dfn attribute for=AbortSignal>aborted</dfn> getter steps are to return true if <a>this</a>


shaseley:
> * Should these timers impact [idle callback deadline computation](https://html.spec.whatwg.org/#event-loop-processing-model:map-of-active-timers)? postTask timers do not, but that may be an oversight. @shaseley @noamr

Ah, yeah it's an oversight if `postTask` timers don't affect the idle deadline computation, and they do in our implementation. Looks like we'll need to insert an entry into that map, but it looks like that might also require moving `handle` outside of `setTimeout` spec?

domenic:
Yeah, in that case I will work on a HTML spec PR pulling out the common infrastructure. For now it will behave exactly like the shared parts of postTask/setTimeout. If we decide AbortSignal.timeout() should behave the same as those with regard to bfcache/suspended workers then this PR can just use that; otherwise we can add a flag.

domenic:
I did the centralization work at https://github.com/whatwg/html/pull/7349, and sent https://github.com/WICG/scheduling-apis/pull/54 plus updated this PR to match.

Now this PR *does* pause the timeout while in bfcache/while the worker is suspended. We should discuss whether that's a good idea or not; it's certainly the path of least resistance.

dom.bs
```diff
@@ -1831,6 +1837,27 @@ are:
...
+  <p><a>Wait for a timeout</a> given <var>global</var>, "<code>AbortSignal-timeout</code>",
+  <var>milliseconds</var>, and the following step:</p>
+
+  <ol>
+   <li><p><a>Queue a global task</a> on the <a>timer task source</a> given <var>global</var> to
+   <a for=AbortSignal>signal abort</a> given <var>signal</var> and a new
+   "{{TimeoutError!!exception}}" {{DOMException}}.
+  </ol>
+
+ <li>Return <var>signal</var>.
```
annevk, line 1858:
```suggestion
 <li><p>Return <var>signal</var>.
```

annevk (commented):
I like it! And yeah, Mozilla can be considered supportive.

domenic:
I was going to write a big comment with the pros/cons of suspending while in bfcache/in a worker. But in the end it mostly ended up with pros for suspended. So I am currently advocating we keep the `setTimeout()` behavior here, i.e. if you do `AbortSignal.timeout(10_000)`, you need to spend 10 seconds _active_ before the signal aborts, not 10 seconds _wall time_.

Here's was my thought process on the pros/cons of suspending:

- (+) Matches setTimeout() so might be more expected for developers
- (+) Easier to spec and probably easier to implement
- (+/-) Avoids "timer storms" upon reactivation
- (-? but not really) Seemingly doesn't match the primary use case that well, of erroring an operation if it took too long. E.g. if you use `fetch(..., { signal: AbortSignal.timeout(30_000) })`, and the server still hasn't responded after 30 seconds, you don't care whether some of those 30 seconds were spent in bfcache: the server is probably not going to respond.
  - This is why the "timer storms" argument is both a pro and a con: having a bunch of stuff get aborted upon bfcache restore, because it took too long, actually seems OK! It's very different than _running a bunch of code_ upon bfcache restore, which is what you would expect with `setTimeout()`.
  - On the other hand, I believe this scenario isn't realistic: if you go into bfcache, I'm pretty sure all ongoing fetches will be aborted immediately anyway. (Or maybe we just won't let you go into bfcache with ongoing fetches; not sure.)
  - Finally, what really swung this for me was remembering that we use `fetch()`'s `signal` for more than just the network operation: we use it for, e.g., JSON decoding or other consume-body operations. So consider the scenario where the server responds, and perhaps even all of the body is read into memory. But then we go into bfcache, before the promise returned from `json()` can fulfill. If we spend a minute in bfcache and then go back, we don't really want the promise to reject; we want it to do that last tiny bit of work, and fulfill! I think this is probably a specific case of a more general scenario, possibly involving non-browser APIs that use `AbortSignal` and care about timeouts.

@shaseley, does this reasoning seem reasonable to you? I'd like to have your blessing on our decision here, as the person who has thought most about these sorts of timing and scheduling things in Chromium :)

shaseley:
Yep, this sounds reasonable; suspending in bfcache LGTM. These roughly match what I've been thinking, and great point about how `fetch`'s signal is used for more than just the network operation. A few more thoughts inline below:

> Here's was my thought process on the pros/cons of suspending:
> 
> * (+) Matches setTimeout() so might be more expected for developers

Note: this is unfortunately [broken in  in Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1202687), but I do like the consistency.

> * (+) Easier to spec and probably easier to implement
> * (+/-) Avoids "timer storms" upon reactivation
> * (-? but not really) Seemingly doesn't match the primary use case that well, of erroring an operation if it took too long. E.g. if you use `fetch(..., { signal: AbortSignal.timeout(30_000) })`, and the server still hasn't responded after 30 seconds, you don't care whether some of those 30 seconds were spent in bfcache: the server is probably not going to respond.
>   
>   * This is why the "timer storms" argument is both a pro and a con: having a bunch of stuff get aborted upon bfcache restore, because it took too long, actually seems OK! It's very different than _running a bunch of code_ upon bfcache restore, which is what you would expect with `setTimeout()`.

Yeah aborting stuff on restore would be good in this scenario, although timer storms could still result if these signals have `abort` event listeners that do a bunch of work in response to the error.

>   * On the other hand, I believe this scenario isn't realistic: if you go into bfcache, I'm pretty sure all ongoing fetches will be aborted immediately anyway. (Or maybe we just won't let you go into bfcache with ongoing fetches; not sure.)

We're [recommending](https://web.dev/bfcache/#always-close-open-connections-before-the-user-navigates-away) that devs close connections in `pagehide`, and that article is leading me to believe pages with ongoing fetches aren't eligible for bfcache in Chromium. I'm not sure about other browsers.

Combining a timeout signal with one that gets aborted on `pagehide` seems like maybe a good option? If we can make it more ergonomic to combine signals that would be even better. (Although, you might get into scenarios where you have the data but end up aborting the processing like you're describing below.)

>   * Finally, what really swung this for me was remembering that we use `fetch()`'s `signal` for more than just the network operation: we use it for, e.g., JSON decoding or other consume-body operations. So consider the scenario where the server responds, and perhaps even all of the body is read into memory. But then we go into bfcache, before the promise returned from `json()` can fulfill. If we spend a minute in bfcache and then go back, we don't really want the promise to reject; we want it to do that last tiny bit of work, and fulfill! I think this is probably a specific case of a more general scenario, possibly involving non-browser APIs that use `AbortSignal` and care about timeouts.

dom.bs
```diff
@@ -1831,6 +1837,27 @@ are:
...
+<p>The static <dfn method for=AbortSignal><code>timeout(<var>milliseconds</var>)</code></dfn> method
+steps are:
+
+<ol>
+ <li><p>Let <var>signal</var> be a new {{AbortSignal}} object.
+
+ <li><p>Let <var>global</var> be <var>signal</var>'s <a>relevant global object</a>.
+
+ <li>
+  <p><a>Wait for a timeout</a> given <var>global</var>, "<code>AbortSignal-timeout</code>",
```
annevk, line 1849: Needs to reference the new algorithm.

annevk (commented):
The main thing I'm worried about for Firefox is that I think @farre or @smaug---- at some point mentioned to me that the clamping logic sits pretty deep in our timeout code. And beyond that, I worry that if we don't enforce clamping for these new timer mechanisms we'll eventually end up back there due to misbehaving websites. But perhaps if no browser budges it's feasible to hold the door.

jasnell:
You can add Node.js and Cloudflare Workers to the list of runtime implementations in support of this. I have PRs pending to land in both environments implementing the `setTimeout()` compatible behavior.

domenic:
> The main thing I'm worried about for Firefox is that I think @farre or @smaug---- at some point mentioned to me that the clamping logic sits pretty deep in our timeout code. And beyond that, I worry that if we don't enforce clamping for these new timer mechanisms we'll eventually end up back there due to misbehaving websites. But perhaps if no browser budges it's feasible to hold the door.

I feel like we discussed this in the context of postTask() and our conclusion was that the allowance for implementation-defined waiting was enough, at least to start with.

annevk (approved):
Thanks, that makes sense.

domenic:
Alright, I've updated this with a mini-explainer. All that remains are someone to write some web platform tests, probably as part of an implementation :)

smaug----:
So the signal on the fetch (in the example) would be aborted even if the fetch() succeeds, abort would just happen when the timeout happens to run? Isn't that error prone, getting abort event, even though the fetch() itself has succeeded?

domenic:
I'm not sure I understand the concern. Nobody else is listening for the abort event in the example... and if they were, then the contract is that the AbortSignal should signal abort to both the fetch and that other thing, so it seems like it's working as intended.

smaug----:
The concern, which may not be that major, is just that one gets now always an aborted signal eventually, so the event listener on the signal is called, even though fetch() or other APIs using signal  succeed.
One could think of some way to cancel the timeout.

jasnell:
I think the issue is more that the timer can continue well after the set of operations completed successfully. Take a pathological case where I create an `AbortSignal` with a timeout of 20 minutes and I attach a handler to it with a closure.

```js
const signal = AbortSignal.timeout(1_200_000);
signal.addEventListener('abort', () => {
  throw signal.reason;
});
```

The complete set of operations that I pass the `signal` to complete successfully, but the internal timer for the `signal` keeps right on going, holding the `AbortSignal` and the closure (along with whatever memory it's closing over) for the whole 20 minutes. Then, it unconditionally throws a synchronous error. 

Ideally there would be a way of canceling the timeout. Obviously that doesn't help with the closure holding on to things but it ensures that the underlying timer it at least no holding the reference for way longer than it needs to.

```js
const signal = AbortSignal.timeout(1_200_000);
await fetch('http://example.org', { signal });
signal.clearTimeout();
```

This case is similar to why we allowed adding an `AbortSignal` to `addEventListener()` itself. We can even follow a similar pattern here if necessary:

```js
const cancelAbort = AbortController();
const signal = AbortSignal.timeout(1_200_000, { signal: cancelAbort.signal });
signal.addEventListener('abort', () => {
  throw signal.reason;
}, { signal: cancelAbort.signal });
await fetch('http://example.org', { signal });
cancelAbort.abort();
```

domenic:
I'm still having trouble understanding the concern. Could someone produce a semi-realistic example of what someone might do with the current proposal, and how they think that the code would give unexpected results?

jasnell:
Is this example not realistic?

```js
const signal = AbortSignal.timeout(1_200_000);
signal.addEventListener('abort', () => { throw signal.reason; });
await fetch('http://example.org', { signal });
```

domenic:
Well, I'm not sure what the consumer on the second line is trying to do. What about that example is not working as expected? I.e. what are the expected results, vs. the actual results in this proposal?

jasnell:
It's not so much that it's not working the way it's written. It's just a bit of a footgun that we can fairly easily do something about.

annevk:
@smaug---- resetting the timer was discussed in https://github.com/whatwg/dom/issues/951#issuecomment-970515484. That would require a change to the controller. (Either a new class or we change the existing class to also support timer functionality.) It's already possible for a signal to become aborted after an operation to which it is passed succeeds. That's inherent in its design. After all, the operation to which it is passed does not get to control the signal, only the controller does.

annevk:
Chatting with @smaug---- I think the concern is usage of the signal `abort` event outside the realm of abortable operations. E.g., using the signal `abort` event as an indication you need to show some UI.

However, the design is such that the `abort` event is for the purposes of implementing abortable operations consuming signals only. And not as an independent "trigger" point to go do something else.

The "trigger" to go do something else (such as displaying UI) should always be (an) (abortable) operation(s) and its/their completion value(s) (which could be abort or timeout).

Perhaps this is something we need to further clarify in the specification.

jasnell:
Well, FWIW, I've updated the Node.js implementation so that the underlying timer holds only a weak reference to the `AbortSignal` that is created so that the timer itself does not prevent the signal from being garbage collected. When the timer fires, the signal is only tripped if it still exists.

annevk:
@jasnell why not clear the timer when the signal can be collected?

smaug----:
If the timer doesn't have a strong reference to the signal, then it is possible to observe GC behavior, no?  One may have an abort listener on the signal after all.

jasnell:
@annevk:
> why not clear the timer when the signal can be collected?

That's a good idea also.

smaug----:
If the signal has an event listener, it shouldn't be collected if there is a timer.

annevk:
Yeah, I guess that is something we'd need to spell out. At least, we typically do that for other objects with event listeners.

jasnell (approved):
LGTM. I've got implementations ready for both Node.js and Cloudflare Workers. With the gc note included, I think this is good to go.

domenic:
We're just waiting on web platform tests at this point :)

domenic:
Tests have landed! Merging, and will file browser bugs and edit the OP with them momentarily.

PR #1079, opened 2022-05-05
Allow more characters when creating various nodes

### LINKED ISSUE #769: A way for createAttribute() to bypass XML Name validation in HTML documents
`createAttribute(localName)` and other attribute creating methods (`setAttribute` and such) require attribute name to be a valid XML Name, and throw otherwise.

However, the HTML parser is less restrictive in attribute names, allowing HTML documents to have some attribute names that are illegal in XML.

Whenever it is needed to programmatically create an attribute with a name that is allowed in HTML but illegal in XML, this is possible through the parser, but not with DOM attribute creating methods, it seems.

```javascript
document.createAttribute('impo$$ible'); // throws

const createHTMLAttribute = name =>
  document.importNode(
    new DOMParser().parseFromString(`<html ${name}>`, 'text/html')
      .documentElement.getAttributeNode(name)
  );

createHTMLAttribute('impo$$ible'); // 'impo$$ible=""'
```

Could `createAttribute` (and maybe `setAttribute` and friends) be somehow allowed to create such attributes in HTML documents?

### LINKED ISSUE #849: Valid/Invalid characters in document.createElement()
A difference between the spec and browser implementations was raised here: https://bugs.chromium.org/p/chromium/issues/detail?id=1061436

[In the spec](https://dom.spec.whatwg.org/#dom-document-createelement), document.createElement should allow names that "match the Name production," which links to [this definition](https://www.w3.org/TR/xml/#NT-Name), which allows characters in many ranges, including `[#x10000-#xEFFFF]`, which I am guessing includes emojis and would therefore allow `
document.createElement('test-\u{1f602}')`

However, Chrome, Firefox, and Safari all throw exceptions when this code snippet is run.
Should the browsers update to match the spec, or should the spec be updated?
Or am I misinterpreting the spec? If the spec should be updated, what should it say instead?

### LINKED ISSUE #1373: Consider more name validation loosening

After #1079, the following parts of the spec will still reference XML name stuff:

- `createProcessingInstruction()` checks against the XML `Name` production
- `createDocumentType()` checks against the XML Namespaces `QName` production

(Additionally, there are historical references to the `Char` and `Name` production, and how we _don't_ validate against them.)

I suspect in both cases you can create PIs/doctypes that violate these rules using the HTML parser, so we could consider doing something similar to #1079.

However, the priority here is significantly lower than for attributes and elements.

domenic:
Stop using XML grammar productions for validating element, attribute, and doctype names. These were overly-restrictive, as it is possible to create nodes with names that don't match those productions using the HTML parser. After this change, each construct has its own custom name validation algorithm.

The only remaining dependency on XML naming rules is for processing instructions, which are uncommon and cannot be created with the HTML parser anyway.

Closes #769. Closes #849. Closes #1373.

   * Gecko
   * Chromium
   * https://github.com/web-platform-tests/wpt/pull/38503
   * https://chromium-review.googlesource.com/c/chromium/src/+/6570951
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1334640
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1773312
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=241419

---

<details>
<summary>Original points for discussion, discussed and concluded on in following comments. These do not reflect the currently proposed change.</summary>

- I did not disallow `=` inside attribute local names. Both the parser and DOM APIs currently disallow them, except the parser allows it for the first character. I'm happy to change this if people prefer; I started with the simpler version.
- This does not disallow lone surrogates, the Unicode replacement character U+FFFD, single quotes, or < in any position, because the HTML parser allows introducing those already and it seems nicer to align.
- I did not change validation for `createProcessingInstruction()` or `createDocumentType()`. We could try to simplify those too, perhaps after investigating parser behavior. But they didn't seem to be causing any real web developer pain, unlike elements and local names, so I thought it'd be better to just leave them as-is.
</details>

***
<a href="https://whatpr.org/dom/1079.html" title="Last updated on Jun 3, 2025, 1:04 AM UTC (406f683)">Preview</a> | <a href="https://whatpr.org/dom/1079/388779b...406f683.html" title="Last updated on Jun 3, 2025, 1:04 AM UTC (406f683)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -221,7 +221,7 @@ reasons, but beyond ASCII anything is allowed.
  [=valid element local name=]:
 
  <pre><code class=lang-javascript>
-  /^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*)|(?:[:_\u0080-\u{10FFFF}][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*)$/u
+  /^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*|[:_\u0080-\u{10FFFF}][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*)$/u
  </code></pre>
 </div>
 
--- a/dom.bs
+++ b/dom.bs
@@ -15,11 +15,6 @@ Indent: 1
 urlPrefix: https://www.w3.org/TR/xml/#NT-; spec: XML
  type: dfn
   text: Name; url: Name; for: XML
-  text: Char; url: Char
-  text: PubidChar; url: PubidChar
-urlPrefix: https://www.w3.org/TR/xml-names/#NT-; spec: XML-NAMES
- type: dfn
-  text: QName; url: QName
 url: https://w3c.github.io/DOM-Parsing/#dom-range-createcontextualfragment
  type: method; text: createContextualFragment(); for: Range
 type: interface
@@ -225,8 +220,10 @@ reasons, but beyond ASCII anything is allowed.
  </code></pre>
 </div>
 
-<p>A string is a <dfn>valid doctype name</dfn> if it does not contain [=ASCII whitespace=], U+0000
-NULL, or U+003E (>).
+<p>A [=string=] is a <dfn>valid doctype name</dfn> if it does not contain [=ASCII whitespace=],
+U+0000 NULL, or U+003E (>).
+
+<p class="note">The empty string is a [=valid doctype name=].
 
 <p>To <dfn>validate and extract</dfn> a <var>namespace</var> and <var>qualifiedName</var>, given a
 <var>context</var>:
@@ -275,6 +272,17 @@ NULL, or U+003E (>).
 
  <li><p>Return (<var>namespace</var>, <var>prefix</var>, <var>localName</var>).
 </ol>
+
+<div class="note">
+ <p>Various APIs in this specification used to validate namespace prefixes, attribute local names,
+ element local names, and doctype names more strictly. This was done in a way that aligned with
+ various XML-related specifications. (Although not all rules from the those specifications were
+ enforced.)
+
+ <p>This was found to be annoying for web developers, especially since it meant there were some
+ names that could be created by the HTML parser, but not by DOM APIs. So, the validations have been
+ loosened to just those described above.
+</div>
 
 
 
@@ -5848,9 +5856,6 @@ return a new {{DocumentFragment}} <a for=/>node</a> whose <a for=Node>node docum
 to return a new {{Text}} <a for=/>node</a> whose <a for=CharacterData>data</a> is <var>data</var>
 and <a for=Node>node document</a> is <a>this</a>.
 
-<p class=note>No check is performed that <var>data</var> consists of
-characters that match the <code><a>Char</a></code> production.
-
 <p>The <dfn method for=Document><code>createCDATASection(<var>data</var>)</code></dfn> method steps
 are:
 
@@ -5868,10 +5873,6 @@ are:
 <p>The <dfn method for=Document><code>createComment(<var>data</var>)</code></dfn> method steps are
 to return a new {{Comment}} <a for=/>node</a> whose <a for=CharacterData>data</a> is <var>data</var>
 and <a for=Node>node document</a> is <a>this</a>.
-
-<p class=note>No check is performed that <var>data</var> consists of
-characters that match the <code><a>Char</a></code> production
-or that it contains two adjacent hyphens or ends with a hyphen.
 
 <p>The
 <dfn method for=Document><code>createProcessingInstruction(<var>target</var>, <var>data</var>)</code></dfn>
@@ -5892,11 +5893,6 @@ method steps are:
  <a for=CharacterData>data</a> set to <var>data</var>, and
  <a for=Node>node document</a> set to <a>this</a>.
 </ol>
-
-<p class=note>No check is performed that <var>target</var> contains
-"<code>xml</code>" or "<code>:</code>", or that
-<var>data</var> contains characters that match the
-<code><a>Char</a></code> production.
 
 <hr>
 
@@ -6225,10 +6221,6 @@ method steps are:
  as its <a>system ID</a>, and with its <a for=Node>node document</a> set to the associated
  <a for=/>document</a> of <a>this</a>.
 </ol>
-
-<p class=note>No check is performed that <var>name</var> matches the <code><a>QName</a></code>
-production, that <var>publicId</var>'s code points match the <code><a>PubidChar</a></code>
-production, or that <var>systemId</var> does not contain both a U+0022 (") and a U+0027 (').
 
 <p>The
 <dfn method for=DOMImplementation><code>createDocument(<var>namespace</var>, <var>qualifiedName</var>, <var>doctype</var>)</code></dfn>
--- a/dom.bs
+++ b/dom.bs
@@ -13,12 +13,12 @@ Indent: 1
 
 <pre class=anchors>
 urlPrefix: https://www.w3.org/TR/xml/#NT-
- type: type
-  text: Name; url: Name
+ type: dfn
+  text: Name; url: Name; for: XML
   text: Char; url: Char
   text: PubidChar; url: PubidChar
 urlPrefix: https://www.w3.org/TR/xml-names/#NT-
- type: type
+ type: dfn
   text: QName; url: QName
 url: https://w3c.github.io/DOM-Parsing/#dom-range-createcontextualfragment
  type: method; text: createContextualFragment(); for: Range
@@ -177,19 +177,59 @@ first <a>following</a> <a for=tree>sibling</a> or null if it has no <a for=tree>
 added.
 
 
-<h3 id=namespaces>Namespaces</h3>
+<h3 id=name-validation oldids=namespaces>Name validation</h3>
 
-<p>To <dfn export>validate</dfn> a <var>qualifiedName</var>, <a>throw</a> an
-"{{InvalidCharacterError!!exception}}" {{DOMException}} if <var>qualifiedName</var> does not match
-the <code><a type>QName</a></code> production.
+<p>A [=string=] is a <dfn>valid namespace prefix</dfn> if its [=string/length=] is at least 1 and it
+does not contain [=ASCII whitespace=], U+0000 NULL, U+002F (/), or U+003E (>).
 
-<p>To <dfn export>validate and extract</dfn> a <var>namespace</var> and <var>qualifiedName</var>,
-run these steps:
+<p>A [=string=] is a <dfn>valid attribute local name</dfn> if its [=string/length=] is at least 1
+and it does not contain [=ASCII whitespace=], U+0000 NULL, U+002F (/), U+003D (=), or U+003E (>).
+
+<p>A [=string=] |name| is a <dfn>valid element local name</dfn> if the following steps return true:
+
+<ol>
+ <li><p>If |name|'s [=string/length=] is 0, then return false.
+
+ <li>
+  <p>If |name|'s 0th [=code point=] is an [=ASCII alpha=], then:
+
+  <ol>
+   <li><p>If |name| contains [=ASCII whitespace=], U+0000 NULL, U+002F (/), or U+003E (>), then
+   return false.
+
+   <li><p>Return true.
+  </ol>
+
+ <li><p>If |name|'s 0th [=code point=] is not U+003A (:), U+005F (_), or in the range U+0080
+ to U+10FFFF, inclusive, then return false.
+
+ <li><p>If |name|'s subsequent [=code points=], if any, are not [=ASCII alphas=], [=ASCII digits=],
+ U+002D (-), U+002E (.), U+003A (:), U+005F (_), or in the range U+0080 to U+10FFFF, inclusive, then
+ return false.
+
+ <li><p>Return true.
+</ol>
+
+<p class=note>This concept is used to validate [=/element=] [=Element/local names=], when
+constructed by DOM APIs. The intention is to allow any name that is possible to construct using the
+HTML parser (the branch where the first [=code point=] is an [=ASCII alpha=]), plus some additional
+possibilities. For those additional possibilities, the ASCII range is restricted for historical
+reasons, but beyond ASCII anything is allowed.
+
+<div class=note>
+ <p>The following JavaScript-compatible regular expression is an implementation of the above
+ definition:
+
+ <pre class=lang-javascript>
+  /^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*)|(?:[:_\u0080-\u{10FFFF}][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*)$/u
+ </pre>
+</div>
+
+<p>To <dfn>validate and extract</dfn> a <var>namespace</var> and <var>qualifiedName</var>, given a
+<var>context</var>:
 
 <ol>
  <li><p>If <var>namespace</var> is the empty string, then set it to null.
-
- <li><p><a>Validate</a> <var>qualifiedName</var>.
 
  <li><p>Let <var>prefix</var> be null.
 
@@ -206,6 +246,15 @@ run these steps:
 
    <li><p>Set <var>localName</var> to <var>splitResult</var>[1].
   </ol>
+
+ <li><p>If <var>prefix</var> is not a [=valid namespace prefix=], then [=throw=] an
+ "{{InvalidCharacterError}}" {{DOMException}}.
+
+ <li><p>If <var>context</var> is "<code>attribute</code>" and <var>localName</var> is not a
+ [=valid attribute local name=], then [=throw=] an "{{InvalidCharacterError}}" {{DOMException}}.
+
+ <li><p>If <var>context</var> is "<code>element</code>" and <var>localName</var> is not a
+ [=valid element local name=], then [=throw=] an "{{InvalidCharacterError}}" {{DOMException}}.
 
  <li><p>If <var>prefix</var> is non-null and <var>namespace</var> is null, then <a>throw</a> a
  "{{NamespaceError!!exception}}" {{DOMException}}.
@@ -5633,8 +5682,8 @@ method steps are to return the <a>list of elements with class names <var>classNa
   <p>When supplied, <var>options</var>'s {{ElementCreationOptions/is}} can be used to create a
   <a>customized built-in element</a>.
 
-  <p>If <var>localName</var> does not match the <code><a type>Name</a></code> production an
-  "{{InvalidCharacterError!!exception}}" {{DOMException}} will be thrown.
+  <p>If <var>localName</var> is not a <a>valid element local name</a> an
+  "{{InvalidCharacterError}}" {{DOMException}} will be thrown.
 
   <p>When both <var>options</var>'s {{ElementCreationOptions/customElementRegistry}} and
   <var>options</var>'s {{ElementCreationOptions/is}} are supplied, a
@@ -5654,8 +5703,9 @@ method steps are to return the <a>list of elements with class names <var>classNa
   <p>When supplied, <var>options</var>'s {{ElementCreationOptions/is}} can be used to create a
   <a>customized built-in element</a>.
 
-  <p>If <var>qualifiedName</var> does not match the <code><a type>QName</a></code> production an
-  "{{InvalidCharacterError!!exception}}" {{DOMException}} will be thrown.
+  <p>If <var>qualifiedName</var> is not a (possibly-prefixed)
+  <a>valid element local name</a> an "{{InvalidCharacterError}}" {{DOMException}} will be
+  thrown.
 
   <p>If one of the following conditions is true a "{{NamespaceError!!exception}}" {{DOMException}}
   will be thrown:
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -5231,17 +5260,17 @@ to return a new {{Comment}} <a for=/>node</a> whose <a for=CharacterData>data</a
...
 <dfn method for=Document><code>createProcessingInstruction(<var>target</var>, <var>data</var>)</code></dfn>
 method steps are:
 
 <ol>
- <li>If <var>target</var> does not match the
+ <li>If <var>target</var> is does not match the
  <!--<code data-anolis-type>PITarget</code>-->
- <code><a type>Name</a></code> production,
+ <code>[=XML/Name=]</code> production,
  then <a>throw</a> an "{{InvalidCharacterError!!exception}}" {{DOMException}}. <!-- DOM3 does not check for "xml" -->
```
annevk, line 5274: The added "is" here seems incorrect. Might also want to drop the dated comments while here.

annevk (commented):
Do we want to call out the `=` difference in a note? At least a comment would be good I think.

I'm not a big fan of adding "DOM API" to the naming. That makes more sense if this was defined outside of the DOM Standard itself. I think dropping it would still make everything work.

dom.bs
```diff
@@ -5135,8 +5171,7 @@ method steps are to return the <a>list of elements with class names <var>classNa
   <a for=/>node</a> whose
   <a for=ProcessingInstruction>target</a> is <var>target</var> and
   <a for=CharacterData>data</a> is <var>data</var>.
-  If <var>target</var> does not match the
-  <code><a type>Name</a></code> production an
+  If <var>target</var> is not a <a>valid basic name</a> an
```
annevk, line 5174: This changes the validation for `ProcessingInstruction` in a domintro box. That seems like an oversight.

annevk (commented):
Thanks, this looks good to me, modulo a small oversight. Anyone else that should review this? @mfreed7 perhaps?

dom.bs
```diff
@@ -209,27 +209,62 @@ against a <var>set</var>, run these steps:
...
 
 
-<h3 id=namespaces>Namespaces</h3>
+<h3 id=name-validation oldids=namespaces>Name validation</h3>
 
-<p>To <dfn export>validate</dfn> a <var>qualifiedName</var>, <a>throw</a> an
-"{{InvalidCharacterError!!exception}}" {{DOMException}} if <var>qualifiedName</var> does not match
-the <code><a type>QName</a></code> production.
+A [=string=] |string| is a <dfn>valid basic name</dfn> if it does not contain
+[=ASCII whitespace=], U+0000 NULL, U+002F (/), or U+003E (>).
```
mfreed7, line 215: Somehow `<` still makes me nervous, but as you point out, it's already allowed, so ok.

dom.bs
```diff
@@ -209,27 +209,62 @@ against a <var>set</var>, run these steps:
...
+<div class="note">
+ <p>This concept is used to validate [=attribute=] [=Attr/local names=] and
+ [=Attr/qualified names=], as well as both [=Element/namespace prefix|element=] and
+ [=Attr/namespace prefix|attribute=] namespace prefixes, in the cases where elements and attributes
+ are constructed with DOM APIs. Note that the prefixes additionally cannot contain U+003A (:) by
+ construction since they always result from the [=validate and extract=] algorithm.
+
+ <p>Note that this means [=attribute=] [=Attr/local names=] for attributes constructed with DOM APIs
+ can contain U+003D (=) in positions beyond the initial character, which is not possible when using
+ the HTML parser.
```
mfreed7, line 226: If we didn't think there would be a compat issue, it seems simpler to just outlaw `=` entirely here, rather than have this very special behavior. But also ok as-is.

mfreed7 (approved):
Looks good!

domenic:
Discussed equals sign at HTML triage meeting. Conclusion: disallow it in attributes everywhere. (Even though the parser allows it in the first-character position.)

Maufas:
> Cierr

> 
> * [ ] Al menos dos implementadores están interesados ​​(y ninguno se opone):
>   
>   * Gecko y Chromium se han involucrado en [caracteres válidos/no válidos en document.createElement() ](https://github.com/whatwg/dom/issues/849)[ #849](https://github.com/whatwg/dom/issues/849) pero aún no han firmado esta formulación exacta[](https://github.com/whatwg/dom/issues/849)
> * [ ]  [Las pruebas](https://github.com/web-platform-tests/wpt) son escritas y se pueden revisar y comentar en:
>   
>   * …
> * [ ]  [Los errores de implementación](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) se archivan:
>   
>   * Cromo: …
>   * Firefox: …
>   * Safari:…
> 
> (Consulte [Modo de trabajo WHATWG: Cambios](https://whatwg.org/working-mode#changes) para obtener más detalles).
> 
> Puntos para la discusión:
> 
> * No deseché `=`los nombres locales de los atributos internos. Tanto el analizador como las API DOM actualmente no los permiten, excepto que el analizador lo permite para el primer carácter. Estoy feliz de cambiar esto si la gente lo prefiere; Empecé con la versión más simple.
> * Esto no deshabilita sustitutos solitarios, el carácter de reemplazo de Unicode U+FFFD, comillas simples o < en cualquier posición, porque el analizador HTML ya permite introducirlos y parece más agradable alinearlos.
> * No cambié la validación para `createProcessingInstruction()`o `createDocumentType()`. Podríamos intentar simplificarlos también, tal vez después de investigar el comportamiento del analizador. Pero no parecían estar causando ningún problema real a los desarrolladores web, a diferencia de los elementos y los nombres locales, así que pensé que sería mejor dejarlos como están.
> 
> [Vista previa](https://whatpr.org/dom/1079.html) | [diferencia](https://whatpr.org/dom/1079/0bf93ba...738fce9.html)



> Cierra #849 .
> 
> * [ ] Al menos dos implementadores están interesados ​​(y ninguno se opone):
>   
>   * Gecko y Chromium se han involucrado en [caracteres válidos/no válidos en document.createElement() ](https://github.com/whatwg/dom/issues/849)[ #849](https://github.com/whatwg/dom/issues/849) pero aún no han firmado esta formulación exacta[](https://github.com/whatwg/dom/issues/849)
> * [ ]  [Las pruebas](https://github.com/web-platform-tests/wpt) son escritas y se pueden revisar y comentar en:
>   
>   * …
> * [ ]  [Los errores de implementación](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) se archivan:
>   
>   * Cromo: …
>   * Firefox: …
>   * Safari:…
> 
> (Consulte [Modo de trabajo WHATWG: Cambios](https://whatwg.org/working-mode#changes) para obtener más detalles).
> 
> Puntos para la discusión:
> 
> * No deseché `=`los nombres locales de los atributos internos. Tanto el analizador como las API DOM actualmente no los permiten, excepto que el analizador lo permite para el primer carácter. Estoy feliz de cambiar esto si la gente lo prefiere; Empecé con la versión más simple.
> * Esto no deshabilita sustitutos solitarios, el carácter de reemplazo de Unicode U+FFFD, comillas simples o < en cualquier posición, porque el analizador HTML ya permite introducirlos y parece más agradable alinearlos.
> * No cambié la validación para `createProcessingInstruction()`o `createDocumentType()`. Podríamos intentar simplificarlos también, tal vez después de investigar el comportamiento del analizador. Pero no parecían estar causando ningún problema real a los desarrolladores web, a diferencia de los elementos y los nombres locales, así que pensé que sería mejor dejarlos como están.
> 
> [Vista previa](https://whatpr.org/dom/1079.html) | [diferencia](https://whatpr.org/dom/1079/0bf93ba...738fce9.html)

domenic:
I think this is ready for re-review.

Potential issue: XML's definition of [Char](https://www.w3.org/TR/xml/#NT-Char) seems nonsensical (it excludes various Unicode characters below U+0020). And, its definition of the `[^#x00#x09#x0A#0x0Cx0D#x20/>]` syntax depends on that definition. Hmm.

domenic:
Refined to no longer use EBNF.

dom.bs
```diff
@@ -211,43 +211,25 @@ added.
...
-
-<dfn>BeyondHTMLParserName</dfn>          ::= <a>BeyondHTMLParserNameStartChar</a> (<a>BeyondHTMLParserNameChar</a>)*
-<dfn>BeyondHTMLParserNameStartChar</dfn> ::= ":" | "_" | [#x80-#x10FFFF]
-<dfn>BeyondHTMLParserNameChar</dfn>      ::= <a>BeyondHTMLParserNameStartChar</a> | [a-zA-Z] | "-" | "." | [0-9]
-</pre>
+* its [=string/length=] is at least 1;
+* its first [=code point=] is an [=ASCII alpha=], U+003A (:), U+005F (_), or in the range U+0080 to
+  U+10FFFF inclusive; and
+* its subsequent [=code points=], if any, are not [=ASCII whitespace=], U+0000 NULL, U+002F (/), or
+  U+003E (>).
```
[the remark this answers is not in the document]
domenic: Done, I think

annevk:
I don't think the EBNF allows for the second code point to be U+0001 when the first is `:`, for instance. At least the intent was to prevent that. Does EBNF work completely differently from ABNF in that `|` doesn't signify OR but instead "union"?

(I didn't see "An equivalent EBNF is the following" initially and I don't think what it states is correct.)

domenic:
I see, I did not capture that this was a branching scenario depending on the behavior of the first code point. And you addressed what harms names like that might hypothetically cause in https://github.com/whatwg/dom/issues/849#issuecomment-1058064183 .

I'll revise.

dom.bs
```diff
@@ -209,19 +209,59 @@ against a <var>set</var>, run these steps:
...
+HTML parser (the branch where the first [=code point=] is an [=ASCII alpha=]), plus some additional
+possibilities. For those additional possibilities, the ASCII range is restricted for historical
+reasons, but beyond ASCII anything is allowed.
+
+<div class=note>
+ <p>The following JavaScript-compatible regular expression is an implementation of the above
+ definition:
+
+ <pre class=lang-javascript>
+  /^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*)|(?:[:_\u0080-][A-Za-z-.:_\u0080-]*)$/u
```
annevk, line 256: @zcorpan or @mathiasbynens could you please review this?

annevk (approved):
Thanks, this looks accurate to me.

dom.bs
zcorpan: Missing ASCII digits.
```suggestion
  /^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*)|(?:[:_\u0080-][A-Za-z0-9-.:_\u0080-]*)$/u
```

mathiasbynens: I think JavaScript doesn’t support character class ranges without the end of the range being explicit, so I’d change both occurrences of `[\u0080-]` into `[\x80-\u{10FFFF}]`.

Is `[A-Za-z-.:_\u0080-]` meant to refer to step 4? In that case, it’s missing “ASCII digits”. Suggestion: `[A-Za-z0-9-.:_\x80-\u{10FFFF}]`.

- Valid namespace prefix: `[\0\t\n\f\r />]`
- Valid attribute local name: `[\0\t\n\f\r /=>]`
- Valid element local name, step 2.1: `[^\0\t\n\f\r />]`
- Valid element local name, step 3: `[:_\x80-\u{10FFFF}]`
- Valid element local name, step 4: `[A-Za-z0-9-.:_\x80-\u{10FFFF}]`

Piecing things together:

```js
/^(?:[A-Za-z][^\0\t\n\f\r\x20/>]*)|(?:[:_\x80-\u{10FFFF}][A-Za-z0-9-.:_\x80-\u{10FFFF}]*)$/u
```

mathiasbynens (approved):

annevk:
@josepharhar would you be interested in finishing this?

josepharhar:
Yes, I have started a WPT here: https://github.com/web-platform-tests/wpt/pull/38503

annevk:
\o/ I suspect that once you implement this and do a try run you'll find a lot of existing WPT tests that can be adjusted. There's probably no need for a new file, but maybe.

cdumez:
Any progress on this?

josepharhar:
Not recently, I have unfortunately been focused on other stuff.

josepharhar:
I came across this while implementing: https://github.com/whatwg/infra/issues/670

dom.bs
```diff
@@ -209,19 +209,59 @@ against a <var>set</var>, run these steps:
...
+
+   <li><p>Return true.
+  </ol>
+
+ <li><p>If |name|'s 0th [=code point=] is not U+003A (:), U+005F (_), or in the range U+0080
+ to U+10FFFF, inclusive, then return false.
+
+ <li><p>If |name|'s subsequent [=code points=], if any, are not [=ASCII alphas=], [=ASCII digits=],
+ U+002D (-), U+002E (.), U+003A (:), U+005F (_), or in the range U+0080 to U+10FFFF, inclusive, then
+ return false.
```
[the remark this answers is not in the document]
josepharhar: Ah nevermind, null characters are not valid as expected, I can't read sometimes

dom.bs
```diff
@@ -13,12 +13,12 @@ Indent: 1
 
 <pre class=anchors>
 urlPrefix: https://www.w3.org/TR/xml/#NT-
```
annevk, line 15: Should we add `spec: XML` while here?

dom.bs
```diff
@@ -177,19 +177,59 @@ first <a>following</a> <a for=tree>sibling</a> or null if it has no <a for=tree>
...
+possibilities. For those additional possibilities, the ASCII range is restricted for historical
+reasons, but beyond ASCII anything is allowed.
+
+<div class=note>
+ <p>The following JavaScript-compatible regular expression is an implementation of the above
+ definition:
+
+ <pre class=lang-javascript>
+  /^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*)|(?:[:_\u0080-\u{10FFFF}][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*)$/u
+ </pre>
```
annevk, line 225: Should this be `<pre><code class=lang-javascript>`?

dom.bs
```diff
@@ -177,19 +177,59 @@ first <a>following</a> <a for=tree>sibling</a> or null if it has no <a for=tree>
 added.
 
 
-<h3 id=namespaces>Namespaces</h3>
+<h3 id=name-validation oldids=namespaces>Name validation</h3>
```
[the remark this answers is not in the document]
domenic: I don't understand what this is suggesting. Leaving the heading the same?? It seems very incorrect now.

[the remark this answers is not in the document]
annevk: Oh my bad, I didn't see the change to the name. I prefer keeping the old ID as the sole ID, but the name change seems fine.

dom.bs
```diff
@@ -177,19 +177,59 @@ first <a>following</a> <a for=tree>sibling</a> or null if it has no <a for=tree>
...
+
+<p class=note>This concept is used to validate [=/element=] [=Element/local names=], when
+constructed by DOM APIs. The intention is to allow any name that is possible to construct using the
+HTML parser (the branch where the first [=code point=] is an [=ASCII alpha=]), plus some additional
+possibilities. For those additional possibilities, the ASCII range is restricted for historical
+reasons, but beyond ASCII anything is allowed.
+
+<div class=note>
+ <p>The following JavaScript-compatible regular expression is an implementation of the above
+ definition:
```
josepharhar, line 221: Just to be clear, instead of saying "the above definition" could you say "valid attribute local name"?

dom.bs
```diff
@@ -177,19 +177,59 @@ first <a>following</a> <a for=tree>sibling</a> or null if it has no <a for=tree>
...
+
+ <pre class=lang-javascript>
+  /^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*)|(?:[:_\u0080-\u{10FFFF}][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*)$/u
```
josepharhar, line 224: Thanks for making this regex! I think it's very useful for checking against my test and implementation.

I'm not sure that this regex is correct, or maybe my test and implementation is wrong, or maybe I'm using the regex incorrectly. Here is a test that I derived from the WPT I'm writing which shows which test cases I don't think match up: https://codepen.io/jarhar/pen/azzMgMP

dom.bs
josepharhar: Sorry meant "valid element local name". That's what this is for, right?

dom.bs
domenic: Thank you! I've fixed the regexp.

Re-running your test, it says

> Should have been valid element local name, regex said was invalid element local name:
>
> - codePoints: [98,105,103,69,109,111,106,105,127345,56689,65039,32], rendered string: bigEmoji🅱️ 

However I think this is correctly categorized by the regexp as invalid, because of the trailing code point 32 (U+0020, space). So something's slightly off in the rest of the test harness?

annevk (approved):
This still looks good. One thought that struck me is that in the future we can figure out which of those Char comments are still relevant and if we want to overtake any of the remaining XML-centric definitions.

dom.bs
josepharhar: Ah there wasn't supposed to be a space in there, I removed it and now it passes all the test cases. Thanks!

dom.bs
```diff
@@ -177,19 +177,59 @@ first <a>following</a> <a for=tree>sibling</a> or null if it has no <a for=tree>
...
 
- <li><p><a>Validate</a> <var>qualifiedName</var>.
+ <li>
+  <p>If |name|'s 0th [=code point=] is an [=ASCII alpha=], then:
+
+  <ol>
+   <li><p>If |name| contains [=ASCII whitespace=], U+0000 NULL, U+002F (/), or U+003E (>), then
+   return false.
+
+   <li><p>Return true.
```
josepharhar, line 200: So does this mean that characters past U+10FFFF are allowed? Like `a\u110000`?

josepharhar: I guess this might not be testable since `String.fromCodePoint(0x11000)` throws an exception, so maybe it doesn't matter?

mathiasbynens: > So does this mean that characters past U+10FFFF are allowed?

No. https://infra.spec.whatwg.org/#code-points limits code points to U+10FFFF.

domenic: Yeah, technically such things are not strings. It should be impossible for bytes above 0x10FFFF to appear inside the C++ string data structures that you're processing. You can't get them in their via either JavaScript APIs or the HTML parsing.

annevk: Yeah, it seems unlikely you'd need to hold a uint32 to implement this. (ICU has some suitable asserts if for some reason you do.)

josepharhar: Awesome, thanks! I was dealing with a C++ test that was manually setting the memory to funny things, sounds like this spec is good as-is.

dom.bs
```diff
@@ -6132,10 +6182,8 @@ interface DOMImplementation {
   Returns a <a>doctype</a>, with the given
   <var>qualifiedName</var>, <var>publicId</var>, and
   <var>systemId</var>. If <var>qualifiedName</var> does not
-  match the <code><a type>Name</a></code> production, an
-  "{{InvalidCharacterError!!exception}}" {{DOMException}} is thrown, and if it does not match the
-  <code><a type>QName</a></code> production, a
-  "{{NamespaceError!!exception}}" {{DOMException}} is thrown.
+  match the <code><a>QName</a></code> production, an
+  "{{InvalidCharacterError!!exception}}" {{DOMException}} is thrown.
```
josepharhar, line 6186: What if instead this called the new "validate and extract" algorithm?

josepharhar: In chromium the code for parsing this is shared between this and the attributeNS methods, so I'd have to split them in order to make this one keep using QName.

josepharhar: Actually this probably depends on the discussion here: https://github.com/whatwg/dom/issues/1373

domenic: Yep, this has been re-done.

mfreed7:
> * This does not disallow lone surrogates, the Unicode replacement character U+FFFD, single quotes, or < in any position, because the HTML parser allows introducing those already and it seems nicer to align.

As I was reviewing a Chromium CL that goes along with this PR, I became more worried about allowing `<` in a lot of places. This comment mentions that this was discussed and concluded - could you point me to that discussion? I'd like to (re-)read it.

My general concern is that allowing `<` would seem to cause trouble for things that are scanning for dangerous strings. E.g. `<<script>`  is now an element with tagname `<script`. Won't this potentially create bad situations somewhere? E.g. sanitizers? I can't immediately put my finger on the exact exploit, but it certainly feels like there must be one in there somewhere...

zcorpan:
I don't have a link to previous discussion, but a quick note:

> My general concern is that allowing `<` would seem to cause trouble for things that are scanning for dangerous strings. E.g. `<<script>` is now an element with tagname `<script`.

The first character needs to be ASCII alpha. But `<x<script>` would have tagname `x<script`, which matches the HTML parser.

mfreed7:
> The first character needs to be ASCII alpha. But `<x<script>` would have tagname `x<script`, which matches the HTML parser.

Right, ok. And I guess the important part is that if this is already allowed by the parser, then whatever problems this does cause must have already been handled. Or they're not problems.

Thanks.

domenic:
> Right, ok. And I guess the important part is that if this is already allowed by the parser, then whatever problems this does cause must have already been handled. Or they're not problems.

This is the main reason I'm not worried.

Regarding your question about history, I traced back through a variety of bugs; https://issues.chromium.org/issues/41276288#comment20 is one of the older nodes in the graph. I don't think `<` ever came up specifically though.

domenic:
Yes, thank you for the catch. I will update.

josepharhar:
Here is a new WPT: https://github.com/web-platform-tests/wpt/pull/38503

And here is another patch in progress to update existing WPTs, which doesn't have a WPT PR yet because we stopped generating those until the code review gets approved: https://chromium-review.googlesource.com/c/chromium/src/+/6570951

josepharhar (approved):

PR #1085, opened 2022-06-08
Add a fully active check for EventTarget in event listener inner invoke

### LINKED ISSUE #1084: Should event dispatch check that the event target is in a fully active realm?
I can't find in the DOM or WebIDL specs where such a check exists; apologies if I missed it.

The scenario is ([jsfiddle](https://jsfiddle.net/shaseley/up7c48va/6/)):
- A page has an iframe
- The iframe creates a new AbortController (`controller`) in its realm and hangs it off of its parent
- The parent page adds an event listener (added and defined in its realm) to `controller.signal`
- The iframe gets removed
- `controller.abort()` is called in the parent page's context

Should the event handler run?

This is similar to https://github.com/whatwg/webidl/issues/993, but it's the event target that is in a "detached realm", not the event listener or call to `controller.abort()`.

All the engines are consistent though &mdash; no one runs the event listener:
 - Blink and WebKit both early out in ~2.10 of [inner invoke](https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke). 
   -  For Blink specifically, we can't create the JS wrapper for the event because the [context is gone](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/bindings/core/v8/js_based_event_listener.cc;l=103;drc=8d6a246c9be4f6b731dc7f6e680b7d5e13a512b5;bpv=0;bpt=1).
 - Gecko skips handling the event during [EventTargetChainItem::HandleEvent](https://searchfox.org/mozilla-central/source/dom/events/EventDispatcher.cpp#318)() because the EventListenerManager is null (IIUC it got cleared during detach). I think this is during targeting, just prior to starting [inner invoke](https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke).

I was a bit surprised by the behavior, but given that no engine is running these, would it make sense to add an explicit check (assuming I didn't miss it) so the spec is in sync, and if so, thoughts on where?

shaseley:
This adds a fully active check in https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke that skips running event listeners if the EventTarget's realm's global's document is not fully active. Blink, Gecko, and WebKit were all skipping these event listeners, but the spec does not currently reflect this behavior.

I put the check where Blink and WebKit have it &mdash; just after the event listener was removed if it is a one-off. My understanding of Gecko's implementation is that the listeners are cleared on frame detach, so this should be compatable. The position of the check could be web observable if the document becomes fully active again, i.e. with bfcache, but Chromium at least [doesn't cache pages with opener references](https://web.dev/bfcache/#avoid-window.opener-references).

Closes #1084.

   * Chromium
   * …
   * https://chromium-review.googlesource.com/c/chromium/src/+/3691385
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/dom/1085.html" title="Last updated on Sep 15, 2026, 4:33 PM UTC (158be49)">Preview</a> | <a href="https://whatpr.org/dom/1085/a2331a4...158be49.html" title="Last updated on Sep 15, 2026, 4:33 PM UTC (158be49)">Diff</a>


dom.bs
```diff
@@ -1549,6 +1549,13 @@ and an optional <var>legacyOutputDidListenersThrowFlag</var>, run these steps:
    <!-- Do this before invocation to avoid reentrancy issues. No need to set removed to true since
         each listener in listeners is run once anyway. -->
 
+   <li><p>Let <var>targetGlobal</var> be <var>event</var>'s {{Event/currentTarget}}'s
+   <a>associated Realm</a>'s <a for=Realm>global object</a>.
```
annevk, line 1553: Nit:
```suggestion
   <a>relevant global object</a>.
```

annevk (commented):
@smaug---- what do you think of this?

Was this tested for various kind of events, including synthetic events?

shaseley:
I manually tested with [browser-initiated events](https://jsfiddle.net/shaseley/up7c48va/6/) via AbortController/AbortSignal and with [synthetic events](https://jsfiddle.net/wL5evuh3/3/), for both detaching a frame prior to and during dispatch. I observed the behavior to be consistent across browsers. I have WPT tests for synthetic events [here](https://chromium-review.googlesource.com/c/chromium/src/+/3691385) (not yet landed), and happy to add/run more tests if it makes sense.

smaug----:
Firefox has checks in the webidl callbacks layer to prevent calling listener in the globals which are "gone".

annevk:
That would suggest it needs to be added to https://webidl.spec.whatwg.org/#call-a-user-objects-operation although that would also affect other callers of that algorithm. (None come to mind at the moment.)

shaseley:
> Firefox has checks in the webidl callbacks layer to prevent calling listener in the globals which are "gone".

The global of the `EventTarget` or the listener? In this case, the checks are about the global of the `EventTarget`, i.e. the listener's global can still be attached, but the target's is not.

When I filed the issue, I did some debugging in FF code to see if I could figure out where the check was happening. I observed the following:

1. If the `EventTarget`'s global is detached when dispatching events, we hit an early out in [EventTargetChainItem::HandleEvent](https://searchfox.org/mozilla-central/source/dom/events/EventDispatcher.cpp#318)() because the `EventListenerManager` is null.
2. If the `EventTarget`'s global is detached during dispatch, I found iteration during `HandleEventInternal()` abruptly ends because the listeners got cleared. This happens, for example, if there are two event listeners and the first detaches the event target's global. I found [code](https://searchfox.org/mozilla-central/source/dom/base/nsIGlobalObject.cpp#65) that "disconnects" the event targets and clears listeners which I *think* caused this, but regardless I definitely observed iteration ending after the first event listener detached the event target's global.

I don't know this code nearly as well as y'all, but this is what I observed with some hacky printf debugging; hopefully that's helpful.

smaug----:
In Gecko if the global of the WebIDL callback itself isn't the current one anymore, the bindings layer prevents the call.
Certain EventTarget objects do indeed disconnect themselves too when closing the relevant global (the idea is roughly that things like XHR isn't really usable one the global is gone).

domenic:
It would be good to figure out something here, as right now there is a check in [the event handler processing algorithm](https://html.spec.whatwg.org/multipage/webappapis.html#the-event-handler-processing-algorithm) but not for event listeners in general.

This means that per spec, e.g., `el.onload = foo` will not execute `foo`, but `el.addEventListener("load", foo)` will execute it. Which is bizarre.

annevk:
I looked into resolving this and had AI write a bunch of tests as well as fixes for this PR and an HTML PR. It seems that the event handler attribute check was not aligned with what implementations were doing.

dom.bs
```diff
@@ -1707,25 +1707,35 @@ string <var>phase</var>, a boolean <var>invocationTargetInShadowTree</var>, and
    <a>remove an event listener</a> given <var>event</var>'s {{Event/currentTarget}} attribute value
    and <var>listener</var>.
 
-   <li><p>Let <var>global</var> be <var>listener</var> <a for="event listener">callback</a>'s
-   <a>associated realm</a>'s <a for=realm>global object</a>.
+   <li><p>Let <var>targetGlobal</var> be <var>event</var>'s {{Event/currentTarget}}'s
+   <a for=Node>node document</a>'s <a>relevant global object</a>, if <var>event</var>'s
```
zcorpan:
```suggestion
   <a for=Node>node document</a>'s <a>relevant global object</a> if <var>event</var>'s
```

zcorpan (approved):
nit

PR #1117, opened 2022-10-10
Add pop-up light dismiss logic to event dispatching

josepharhar:
This was moved from the HTML spec PR for the popup attribute based on this advice: https://github.com/whatwg/html/pull/8221#discussion_r987585621
TODO add a better description

   * Chrome
   * …
   * https://wpt.fyi/results/html/semantics/popups
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1307772
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/dom/1117.html" title="Last updated on Oct 27, 2022, 6:32 PM UTC (f4cef07)">Preview</a> | <a href="https://whatpr.org/dom/1117/bf5f6c2...f4cef07.html" title="Last updated on Oct 27, 2022, 6:32 PM UTC (f4cef07)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1393,7 +1393,7 @@ property of the event being dispatched.
      <li><p>Otherwise, set <var>event</var>'s {{Event/eventPhase}} attribute to
      {{Event/CAPTURING_PHASE}}.
 
-     <li><p>Run {light dismiss open pop-ups} given <var>struct</var>'s <a
+     <li><p>Run {light dismiss open pop-ups} given <var>event</var> and <var>struct</var>'s <a
      for=Event/path>invocation target</a>.
 
      <li><p><a>Invoke</a> with <var>struct</var>, <var>event</var>, "<code>capturing</code>", and
--- a/dom.bs
+++ b/dom.bs
@@ -1393,6 +1393,9 @@ property of the event being dispatched.
      <li><p>Otherwise, set <var>event</var>'s {{Event/eventPhase}} attribute to
      {{Event/CAPTURING_PHASE}}.
 
+     <li><p>Run {light dismiss open pop-ups} given <var>struct</var>'s <a
+     for=Event/path>invocation target</a>.
+
      <li><p><a>Invoke</a> with <var>struct</var>, <var>event</var>, "<code>capturing</code>", and
      <var>legacyOutputDidListenersThrowFlag</var> if given.
     </ol>


dom.bs
```diff
@@ -1393,6 +1393,9 @@ property of the event being dispatched.
      <li><p>Otherwise, set <var>event</var>'s {{Event/eventPhase}} attribute to
      {{Event/CAPTURING_PHASE}}.
 
+     <li><p>Run {light dismiss open pop-ups} given <var>struct</var>'s <a
+     for=Event/path>invocation target</a>.
```
annevk, line 1397: This needs some kind of note explaining the setup. This also seems like a layering violation. Is this really supposed to happen for all events or just certain UI events? Is it supposed to happen for synthetic events or just events dispatched by the user agent?

josepharhar: > This also seems like a layering violation.

Is HTML supposed to refer to DOM but not the other way around? I noticed that this spec refers to HTMLSlotElement from the HTML spec...

> Is this really supposed to happen for all events or just certain UI events?

Yes, it really happens for all events, at least in the [chromium implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/node.cc;l=2885;drc=b5666c58a9cf7539e6f57c3e6e957c8fd956c650).

> (If you rebase and force push the CI error will go away.)

done

annevk: It seems to me that the implementation only does something for mouse and keyboard input. Although maybe also for synthetic mouse and keyboard input? It still strikes me as somewhat wrong to have it as part of event dispatch. What's the design motivation for that?

josepharhar: There has been some discussion about the constraints of when to actually light dismiss based on the event here, hopefully the new changes address this concern: https://github.com/whatwg/html/issues/7785#issuecomment-1299086185

> It seems to me that the implementation only does something for mouse and keyboard input.

Do you mean that it should only work if the event is trusted? The implementation does not currently check isTrusted, and [mason's new patch](https://chromium-review.googlesource.com/c/chromium/src/+/3995612/1/third_party/blink/renderer/core/html/html_element.cc#1744) doesn't do so either.

> Although maybe also for synthetic mouse and keyboard input? It still strikes me as somewhat wrong to have it as part of event dispatch. What's the design motivation for that?

I don't know what the design motivation was, but if you have any alternative ideas I'm all ears

annevk:
I think this can be closed now right and instead closing of popovers happens at a higher-level?

domenic:
My understanding is that this is still needed for pointerdown/pointerup-based dismissal? https://whatpr.org/html/8221/popover.html#popover-light-dismiss

annevk:
I see, so why don't have this logic as part of the code that dispatches those events? Why does it need to be embedded in the dispatch logic of all events?

josepharhar:
> why don't have this logic as part of the code that dispatches those events?

Good idea, I'll try moving this to the code that creates pointer events and see what happens

josepharhar:
I have opened a PR on pointerevents here: https://github.com/w3c/pointerevents/pull/460
The chromium patch seems to be working ok: https://chromium-review.googlesource.com/c/chromium/src/+/4023021
I will tentatively close this PR since the pointerevents PR replaces this.

PR #1142, opened 2023-01-10
Make Document origin alias its policy container's origin.

wanderview:
This change depends on the changes in https://github.com/whatwg/html/pull/8447 which adds origin to policy container.

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …


dom.bs
```diff
@@ -4337,8 +4337,9 @@ dom-Range-extractContents, dom-Range-cloneContents -->
   <dl class=switch>
    <dt>{{Document}}
    <dd><p>Set <var>copy</var>'s <a for=Document>encoding</a>, <a for=Document>content type</a>,
-   <a for=Document>URL</a>, <a for=Document>origin</a>, <a for=Document>type</a>, and
-   <a for=Document>mode</a> to those of <var>node</var>.
+   <a for=Document>URL</a>, <a for=Document>policy container</a>'s
+   <a for="Policy Container">origin</a>, <a for=Document>type</a>, and <a for=Document>mode</a>
+   to those of <var>node</var>.
```
annevk, line 4342: I think now we're using nested data structures this seems less clear. Perhaps it's best to just write them all out explicitly?

> set copy's x to node's x, ...

annevk (commented):
"Policy Container" should be "policy container".

dom.bs
wanderview: I tried to break these out, but since this won't build before the other PR lands its difficult for me to see if it looks reasonable or not.

annevk: I was thinking a long inline paragraph, but we can wait until the HTML PR is ready.

Issue #878, opened 2020-07-13
Allow other specs to add event listeners ergonomically


domenic:
Background: https://github.com/whatwg/streams/pull/1053#discussion_r452433904, as well as the (on-pause) https://wicg.github.io/kv-storage/#add-a-simple-event-listener.

There are times when it is convenient for specs to listen to events. For Streams, we want to use the MessagePort infrastructure, and its message/messageerror events. @ricea states that there are tons of other places in Blink that similarly listen to events from C++.

Another instance of this is #493, where we decided to add a custom mechanism specific for `AbortSignal`. The downside noted there is that the ordering is not interleaved in an intuitive fashion with web-developer-added events. I'm not proposing we revisit that at the moment, although it might be a good idea after this issue settles.

From specs, adding such a listener is not currently easy, mainly because creating JS functions / Web IDL callbacks from algorithm steps is not currently easy. I think we should add something to make it easier.

Some options:

1. Make it easy to create Web IDL callbacks from specs (so, this would probably become a Web IDL issue). Then Streams would write something like "[Add an event listener](https://dom.spec.whatwg.org/#add-an-event-listener) with _target_ and an event listener whose callback is an `EventListener` [that performs the following steps]:".

    - This could clean up https://streams.spec.whatwg.org/#blqs-internal-slots

    - The main downside of this is that the general case will need you to supply a length and name for the function, which is undesirable overhead for the event listener case. We could have Web IDL make these optional, with the understanding that the caller has to be very careful to pass them when they're observable... but I don't like it much.

2. Make the DOM spec's event listener struct contain either an `EventListener` or some steps, and branch at the appropriate point.

3. Add a wrapper algorithm (or modify "add an event listener") which accepts steps instead of `EventListener` instances, and converts at the boundary, so that an "event listener" struct always contains an `EventListener`.

I'm open to any of these, although I disprefer (1). I'd be happy to send a pull request for any of them if I can get some editor direction.

annevk:
This seems to go somewhat counter to https://dom.spec.whatwg.org/#action-versus-occurance. I know that user agents have some internal infrastructure that looks like this, but it's not clear to me that it's also the right way to define this or even the right way to implement these things going forward.

domenic:
Hmm, I don't see the contradiction. Why would a user agent listening for a message event be worse than a web developer listening for a message event?

annevk:
Who has access to the event target?

domenic:
[...]

PR #1152, opened 2023-02-07
Add AbortSignal.any()

shaseley:
Hi DOM folks,

I've got a working implementation of [AbortSignal.any()](https://github.com/shaseley/abort-signal-any) with tentative WPT tests, so I wanted to send out a draft spec for thoughts/consideration. It's probably not ready as-is because it would break other specs (see below), but this is how I thought the end state might look.

### PR Description

This adds IDL and algorithms for [AbortSignal.any()](https://github.com/shaseley/abort-signal-any).

Implementation details:
 - This implements an optimization that puts all children on non-composite signals (i.e. those associated with a controller or timeout). This allows "intermediate" nodes (e.g. B in A follows B follows C) to be garbage collected if they are only being kept alive to propagate aborts.

 - This removes the follow algorithm, so callsites will need to be updated.

 - This makes "signal abort" private. This is done so we can make assumptions about when a signal can no longer abort. The few places that directly signal abort will be converted to use an AbortController (this also exposes an algorithm to "request abort" on a controller for this).

 - The "create a composite abort signal" algorithm takes an interface so that TaskSignal.any() can easily hook into it but create a TaskSignal.

 - Some algorithms that invoke "follow" create an AbortSignal in a particular realm. This enables doing that, but borrows some language from elsewhere in the spec w.r.t. doing the default thing. Neither of the other two static AbortSignal factory APIs specify a realm.

### Dependent / Follow-up work

As-is, this PR changes how other specs would interact with abort signals and controllers. The related work is to:
 1. Don't invoke "signal abort" outside of the DOM spec, use "request abort" on an abort controller instead: #1194.
 2. Don't invoke "follow" anywhere, use "create a composite abort signal" instead.
    1. https://github.com/whatwg/fetch/pull/1646
       1. https://github.com/w3c/ServiceWorker/pull/1678
    2. https://github.com/whatwg/streams/pull/1277
 3. (probably) Make sure algorithms are being removed when they will no longer have an effect, e.g. when the underlying operation is completed or failed: also #1194.

### Implementation Notes

There's an implementation of this in Chromium behind a flag in Canary (M112). The basic implementation was straightforward, but there was a bit of work to get the memory management implemented (what's described in the Garbage Collection section in the draft PR). Composite signals can't be GCed if they have abort algorithms (or abort event listeners) and there's a source signal that can still signal abort, but we weren't removing abort algorithms once they were "done", which prevented GC. I fixed this in Chromium, but we should probably also update specs too.

### PR Questions

   * Chromium
   * [Gecko](https://github.com/mozilla/standards-positions/issues/774)
   * https://github.com/web-platform-tests/wpt/pull/37434 (tentative WPT tests)
   * https://github.com/web-platform-tests/wpt/pull/39785 (rename from tentative)
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1323391
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1830781
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=256176
   * Deno (only for aborting and events): https://github.com/denoland/deno/issues/18944
   * Node.js (only for aborting and events): https://github.com/nodejs/node/issues/47811
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/mdn/issues/396

### Additional Info

 - [TAG Review](https://github.com/w3ctag/design-reviews/issues/737) (completed)

***
<a href="https://whatpr.org/dom/1152.html" title="Last updated on May 5, 2023, 5:06 AM UTC (4e2146c)">Preview</a> | <a href="https://whatpr.org/dom/1152/fb0bf26...4e2146c.html" title="Last updated on May 5, 2023, 5:06 AM UTC (4e2146c)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1812,7 +1812,7 @@ interface AbortSignal : EventTarget {
  <dt><code>AbortSignal . <a method for=AbortSignal lt=any(signals)>any</a>(<var>signals</var>)</code>
  <dd>Returns an {{AbortSignal}} instance which will be aborted once any of <var>signals</var> is
  aborted. Its <a for=AbortSignal>abort reason</a> will be set to whichever one of <var>signals</var>
- cause it to be aborted.
+ caused it to be aborted.
 
  <dt><code>AbortSignal . <a method for=AbortSignal lt=timeout(milliseconds)>timeout</a>(<var>milliseconds</var>)</code>
  <dd>Returns an {{AbortSignal}} instance which will be aborted in <var>milliseconds</var>
@@ -1844,11 +1844,11 @@ specified otherwise, its value is the empty set.
 initially false.
 
 <p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, which is a
-<a for=/>set</a> of weak references to {{AbortSignal}}s that the object is dependent on for its
-[=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
+<a for=/>set</a> of weak references to {{AbortSignal}} objects that the object is dependent on for
+its [=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
 
 <p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn>, which is a
-<a for=/>set</a> of weak references to {{AbortSignal}}s that are dependent on it for their
+<a for=/>set</a> of weak references to {{AbortSignal}} objects that are dependent on it for their
 [=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
 
 <p>To <dfn export for=AbortSignal>add</dfn> an algorithm <var>algorithm</var> to an {{AbortSignal}}
@@ -1970,10 +1970,6 @@ an optional <var>reason</var>, run these steps:
  <li><p><a for=set>For each</a> <var>dependentSignal</var> of <var>signal</var>'s
  [=AbortSignal/dependent signals=], [=AbortSignal/signal abort=] on <var>dependentSignal</var> with
  <var>signal</var>'s [=AbortSignal/abort reason=].
-
- <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>dependent signals</a>.
-
- <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>source signals</a>.
 </ol>
 
 <p>To <dfn export>create a composite abort signal</dfn> from a list of {{AbortSignal}} objects
@@ -2028,15 +2024,16 @@ interface that inherits from it, and an optional <var>realm</var>, run these ste
      [=AbortSignal/dependent signals=].
     </ol>
   </ol>
+
  <li>Return <var>resultSignal</var>.
 </ol>
 
 
 <h4 id=abort-signal-garbage-collection>Garbage collection</h4>
 
-<p>A [=AbortSignal/composite=] {{AbortSignal}} object must not be garbage collected while its
-[=AbortSignal/source signals=] is non-empty unless it has no registered event listeners for its
-{{AbortSignal/abort}} event and its [=AbortSignal/abort algorithms=] is empty.
+<p>A non-[=AbortSignal/aborted=] [=AbortSignal/composite=] {{AbortSignal}} object must not be
+garbage collected while its [=AbortSignal/source signals=] is non-empty and it has registered event
+listeners for its {{AbortSignal/abort}} event or its [=AbortSignal/abort algorithms=] is non-empty.
 
 
 <h3 id=abortcontroller-api-integration>Using {{AbortController}} and {{AbortSignal}} objects in
--- a/dom.bs
+++ b/dom.bs
@@ -30,6 +30,7 @@ type: interface
 urlPrefix: https://tc39.es/ecma262/#; spec: ECMASCRIPT
  text: Construct; url: sec-construct; type: abstract-op
  type: dfn
+  text: current realm; url: current-realm
   text: realm; url: realm
   text: surrounding agent; url: surrounding-agent
 urlPrefix: https://w3c.github.io/hr-time/#; spec: HR-TIME
@@ -1830,45 +1831,30 @@ interface AbortSignal : EventTarget {
  {{AbortController}} has signaled to abort; otherwise, does nothing.
 </dl>
 
-<p>An {{AbortSignal}} object has an associated <dfn export for=AbortSignal>abort reason</dfn>, which is a
-JavaScript value. It is undefined unless specified otherwise.
+<p>An {{AbortSignal}} object has an associated <dfn export for=AbortSignal>abort reason</dfn> (a
+JavaScript value), which is initially undefined.
 
-<p>An {{AbortSignal}} object is <dfn export for="AbortSignal">aborted</dfn> when its
-[=AbortSignal/abort reason=] is not undefined.
-
-<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>abort algorithms</dfn>, which is a
-<a for=/>set</a> of algorithms which are to be executed when it is [=AbortSignal/aborted=]. Unless
-specified otherwise, its value is the empty set.
-
-<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">composite</dfn> (a boolean), which is
-initially false.
-
-<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, which is a
-<a for=/>set</a> of weak references to {{AbortSignal}} objects that the object is dependent on for
-its [=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
-
-<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn>, which is a
-<a for=/>set</a> of weak references to {{AbortSignal}} objects that are dependent on it for their
-[=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
-
-<p>To <dfn export for=AbortSignal>add</dfn> an algorithm <var>algorithm</var> to an {{AbortSignal}}
-object <var>signal</var>, run these steps:
-
-<ol>
- <li><p>If <var>signal</var> is [=AbortSignal/aborted=], then return.
-
- <li><p><a for=set>Append</a> <var>algorithm</var> to <var>signal</var>'s
- <a for=AbortSignal>abort algorithms</a>.
-</ol>
-
-<p>To <dfn export for=AbortSignal>remove</dfn> an algorithm <var>algorithm</var> from an
-{{AbortSignal}} <var>signal</var>, <a for=set>remove</a> <var>algorithm</var> from
-<var>signal</var>'s <a for=AbortSignal>abort algorithms</a>.
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>abort algorithms</dfn>, (a
+<a for=/>set</a> of algorithms which are to be executed when it is [=AbortSignal/aborted=]),
+which is initially empty.
 
 <p class=note>The [=AbortSignal/abort algorithms=] enable APIs with complex
 requirements to react in a reasonable way to {{AbortController/abort()}}. For example, a given API's
 [=AbortSignal/abort reason=] might need to be propagated to a cross-thread environment, such as a
 service worker.
+
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">composite</dfn> (a boolean), which is
+initially false.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, (a weak
+<a for=/>set</a> of {{AbortSignal}} objects that the object is dependent on for its
+[=AbortSignal/aborted=] state), which is initially empty.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn> (a weak
+<a for=/>set</a> of {{AbortSignal}} objects that are dependent on the obejct for their
+[=AbortSignal/aborted=] state), which is initially empty.
+
+<hr>
 
 <p>The static <dfn method for=AbortSignal><code>abort(<var>reason</var>)</code></dfn> method steps
 are:
@@ -1909,7 +1895,7 @@ steps are:
 
 <p>The static <dfn method for=AbortSignal><code>any(<var>signals</var>)</code></dfn> method
 steps are to return the result of <a>creating a composite abort signal</a> from <var>signals</var>
-using {{AbortSignal}}.
+using {{AbortSignal}} and the <a>current realm</a>.
 
 <p>The <dfn attribute for=AbortSignal>aborted</dfn> getter steps are to return true if <a>this</a>
 is [=AbortSignal/aborted=]; otherwise false.
@@ -1951,8 +1937,27 @@ is [=AbortSignal/aborted=]; otherwise false.
 {{AbortController}} object, but an API observing the {{AbortSignal}} object can choose to ignore
 them. For instance, if the operation has already completed.
 
+<hr>
+
+<p>An {{AbortSignal}} object is <dfn export for="AbortSignal">aborted</dfn> when its
+[=AbortSignal/abort reason=] is not undefined.
+
+<p>To <dfn export for=AbortSignal>add</dfn> an algorithm <var>algorithm</var> to an {{AbortSignal}}
+object <var>signal</var>:
+
+<ol>
+ <li><p>If <var>signal</var> is [=AbortSignal/aborted=], then return.
+
+ <li><p><a for=set>Append</a> <var>algorithm</var> to <var>signal</var>'s
+ <a for=AbortSignal>abort algorithms</a>.
+</ol>
+
+<p>To <dfn export for=AbortSignal>remove</dfn> an algorithm <var>algorithm</var> from an
+{{AbortSignal}} <var>signal</var>, <a for=set>remove</a> <var>algorithm</var> from
+<var>signal</var>'s <a for=AbortSignal>abort algorithms</a>.
+
 <p>To <dfn for=AbortSignal>signal abort</dfn>, given an {{AbortSignal}} object <var>signal</var> and
-an optional <var>reason</var>, run these steps:
+an optional <var>reason</var>:
 
 <ol>
  <li><p>If <var>signal</var> is [=AbortSignal/aborted=], then return.
@@ -1974,16 +1979,12 @@ an optional <var>reason</var>, run these steps:
 
 <p>To <dfn export>create a composite abort signal</dfn> from a list of {{AbortSignal}} objects
 <var>signals</var>, using <var>signalInterface</var>, which must be either {{AbortSignal}} or an
-interface that inherits from it, and an optional <var>realm</var>, run these steps:
+interface that inherits from it, and a <var>realm</var>:
 
 <ol>
  <li>
   <p>Let <var>resultSignal</var> be a <a for=/>new</a> object implementing
-  <var>signalInterface</var> using <var>realm</var> if given, otherwise using the default behavior
-  defined in Web IDL.
-
-  <p class=XXX>As of the time of this writing Web IDL does not yet define any default behavior;
-  see <a href="https://github.com/whatwg/webidl/issues/135">whatwg/webidl#135</a>.
+  <var>signalInterface</var> using <var>realm</var>.
 
  <li><p>Set <var>resultSignal</var>'s [=AbortSignal/composite=] to true.
 
@@ -3539,6 +3540,7 @@ dictionary MutationObserverInit {
 <ul class=brief>
  <li>A <dfn for=MutationObserver id=concept-mo-callback>callback</dfn> set on creation.
  <li>A <dfn for=MutationObserver>node list</dfn> (a <a for=/>list</a> of weak references to
+
  <a for=/>nodes</a>), which is initially empty.
  <li>A <dfn for=MutationObserver id=concept-mo-queue>record queue</dfn> (a <a for=/>queue</a> of
  zero or more {{MutationRecord}} objects), which is initially empty.
--- a/dom.bs
+++ b/dom.bs
@@ -1850,7 +1850,7 @@ requirements to react in a reasonable way to {{AbortController/abort()}}. For ex
 [=AbortSignal/abort reason=] might need to be propagated to a cross-thread environment, such as a
 service worker.
 
-<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">composite</dfn> (a boolean), which is
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">dependent</dfn> (a boolean), which is
 initially false.
 
 <p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn> (a weak
@@ -1906,7 +1906,7 @@ steps are:
 
 <div algorithm>
 <p>The static <dfn method for=AbortSignal><code>any(<var>signals</var>)</code></dfn> method
-steps are to return the result of <a>creating a composite abort signal</a> from <var>signals</var>
+steps are to return the result of <a>creating a dependent abort signal</a> from <var>signals</var>
 using {{AbortSignal}} and the <a>current realm</a>.
 </div>
 
@@ -1997,7 +1997,7 @@ object <var>signal</var>:
 </div>
 
 <div algorithm>
-<p>To <dfn export>create a composite abort signal</dfn> from a list of {{AbortSignal}} objects
+<p>To <dfn export>create a dependent abort signal</dfn> from a list of {{AbortSignal}} objects
 <var>signals</var>, using <var>signalInterface</var>, which must be either {{AbortSignal}} or an
 interface that inherits from it, and a <var>realm</var>:
 
@@ -2009,14 +2009,14 @@ interface that inherits from it, and a <var>realm</var>:
  [=AbortSignal/aborted=], then set <var>resultSignal</var>'s [=AbortSignal/abort reason=] to
  <var>signal</var>'s [=AbortSignal/abort reason=] and return <var>resultSignal</var>.
 
- <li><p>Set <var>resultSignal</var>'s [=AbortSignal/composite=] to true.
+ <li><p>Set <var>resultSignal</var>'s [=AbortSignal/dependent=] to true.
 
  <li>
   <p><a for=list>For each</a> <var>signal</var> of <var>signals</var>:
 
   <ol>
    <li>
-    <p>If <var>signal</var>'s [=AbortSignal/composite=] is false, then:
+    <p>If <var>signal</var>'s [=AbortSignal/dependent=] is false, then:
 
     <ol>
      <li><p><a for=set>Append</a> <var>signal</var> to <var>resultSignal</var>'s
@@ -2032,7 +2032,7 @@ interface that inherits from it, and a <var>realm</var>:
 
     <ol>
      <li><p>Assert: <var>sourceSignal</var> is not [=AbortSignal/aborted=] and not
-     [=AbortSignal/composite=].
+     [=AbortSignal/dependent=].
 
      <li><p>If <var>resultSignal</var>'s [=AbortSignal/source signals=] <a for=set>contains</a>
      <var>sourceSignal</var>, then <a for=iteration>continue</a>.
@@ -2052,7 +2052,7 @@ interface that inherits from it, and a <var>realm</var>:
 
 <h4 id=abort-signal-garbage-collection>Garbage collection</h4>
 
-<p>A non-[=AbortSignal/aborted=] [=AbortSignal/composite=] {{AbortSignal}} object must not be
+<p>A non-[=AbortSignal/aborted=] [=AbortSignal/dependent=] {{AbortSignal}} object must not be
 garbage collected while its [=AbortSignal/source signals=] is non-empty and it has registered event
 listeners for its {{AbortSignal/abort}} event or its [=AbortSignal/abort algorithms=] is non-empty.
 
--- a/dom.bs
+++ b/dom.bs
@@ -1981,9 +1981,13 @@ an optional <var>reason</var>, run these steps:
 interface that inherits from it, and an optional <var>realm</var>, run these steps:
 
 <ol>
- <li><p>Let <var>resultSignal</var> be a <a for=/>new</a> object implementing
- <var>signalInterface</var> using <var>realm</var> if given, otherwise using the default behavior
- defined in Web IDL.
+ <li>
+  <p>Let <var>resultSignal</var> be a <a for=/>new</a> object implementing
+  <var>signalInterface</var> using <var>realm</var> if given, otherwise using the default behavior
+  defined in Web IDL.
+
+  <p class=XXX>As of the time of this writing Web IDL does not yet define any default behavior;
+  see <a href="https://github.com/whatwg/webidl/issues/135">whatwg/webidl#135</a>.
 
  <li><p>Set <var>resultSignal</var>'s [=AbortSignal/composite=] to true.
 
--- a/dom.bs
+++ b/dom.bs
@@ -1851,7 +1851,7 @@ initially false.
 [=AbortSignal/aborted=] state), which is initially empty.
 
 <p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn> (a weak
-<a for=/>set</a> of {{AbortSignal}} objects that are dependent on the obejct for their
+<a for=/>set</a> of {{AbortSignal}} objects that are dependent on the object for their
 [=AbortSignal/aborted=] state), which is initially empty.
 
 <hr>
@@ -3540,7 +3540,6 @@ dictionary MutationObserverInit {
 <ul class=brief>
  <li>A <dfn for=MutationObserver id=concept-mo-callback>callback</dfn> set on creation.
  <li>A <dfn for=MutationObserver>node list</dfn> (a <a for=/>list</a> of weak references to
-
  <a for=/>nodes</a>), which is initially empty.
  <li>A <dfn for=MutationObserver id=concept-mo-queue>record queue</dfn> (a <a for=/>queue</a> of
  zero or more {{MutationRecord}} objects), which is initially empty.
--- a/dom.bs
+++ b/dom.bs
@@ -1782,8 +1782,11 @@ constructor steps are:
 <a>this</a>'s <a for=AbortController>signal</a>.
 
 <p>The <dfn method for=AbortController><code>abort(<var>reason</var>)</code></dfn> method steps are
-to <a for=AbortSignal>signal abort</a> on <a>this</a>'s <a for=AbortController>signal</a> with
-<var>reason</var> if it is given.
+to <a for=AbortController>request abort</a> on <a>this</a> with <var>reason</var> if it is given.
+
+<p>To <dfn export for=AbortController>request abort</dfn> on an {{AbortController}} <var>controller</var>
+with an optional <var>reason</var>, <a for=AbortSignal>signal abort</a> on <var>controller</var>'s
+<a for=AbortController>signal</a> with <var>reason</var> if it is given.
 
 <h3 id=interface-AbortSignal>Interface {{AbortSignal}}</h3>
 
@@ -1792,6 +1795,7 @@ to <a for=AbortSignal>signal abort</a> on <a>this</a>'s <a for=AbortController>s
 interface AbortSignal : EventTarget {
   [NewObject] static AbortSignal abort(optional any reason);
   [Exposed=(Window,Worker), NewObject] static AbortSignal timeout([EnforceRange] unsigned long long milliseconds);
+  [NewObject] static AbortSignal _any(sequence&lt;AbortSignal> signals);
 
   readonly attribute boolean aborted;
   readonly attribute any reason;
@@ -1804,6 +1808,11 @@ interface AbortSignal : EventTarget {
  <dt><code>AbortSignal . <a method for=AbortSignal>abort</a>(<var>reason</var>)</code>
  <dd>Returns an {{AbortSignal}} instance whose <a for=AbortSignal>abort reason</a> is set to
  <var>reason</var> if not undefined; otherwise to an "{{AbortError!!exception}}" {{DOMException}}.
+
+ <dt><code>AbortSignal . <a method for=AbortSignal lt=any(signals)>any</a>(<var>signals</var>)</code>
+ <dd>Returns an {{AbortSignal}} instance which will be aborted if any of <var>signals</var> is
+ aborted. Its <a for=AbortSignal>abort reason</a> will be set to whichever one of <var>signals</var>
+ cause it to be aborted.
 
  <dt><code>AbortSignal . <a method for=AbortSignal lt=timeout(milliseconds)>timeout</a>(<var>milliseconds</var>)</code>
  <dd>Returns an {{AbortSignal}} instance which will be aborted in <var>milliseconds</var>
@@ -1830,6 +1839,17 @@ JavaScript value. It is undefined unless specified otherwise.
 <p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>abort algorithms</dfn>, which is a
 <a for=/>set</a> of algorithms which are to be executed when it is [=AbortSignal/aborted=]. Unless
 specified otherwise, its value is the empty set.
+
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">composite</dfn> (a boolean), which is
+initially false.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, which is a
+<a for=/>set</a> of {{AbortSignal}}s the object is dependent on for its [=AbortSignal/aborted=]
+state. Unless specified otherwise, its value is the empty set.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn>, which is a
+<a for=/>set</a> of {{AbortSignal}}s that are dependent on it for their [=AbortSignal/aborted=]
+state. Unless specified otherwise, its value is the empty set.
 
 <p>To <dfn export for=AbortSignal>add</dfn> an algorithm <var>algorithm</var> to an {{AbortSignal}}
 object <var>signal</var>, run these steps:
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -1789,6 +1792,7 @@ to <a for=AbortSignal>signal abort</a> on <a>this</a>'s <a for=AbortController>s
 interface AbortSignal : EventTarget {
   [NewObject] static AbortSignal abort(optional any reason);
   [Exposed=(Window,Worker), NewObject] static AbortSignal timeout([EnforceRange] unsigned long long milliseconds);
+  [NewObject] static AbortSignal _any(sequence&lt;AbortSignal> signals);
```
annevk, line 1795: I'm somewhat surprised that `Promise.any` takes an iterable and not rest parameters.

dom.bs
```diff
@@ -1779,8 +1779,11 @@ constructor steps are:
 <a>this</a>'s <a for=AbortController>signal</a>.
 
 <p>The <dfn method for=AbortController><code>abort(<var>reason</var>)</code></dfn> method steps are
-to <a for=AbortSignal>signal abort</a> on <a>this</a>'s <a for=AbortController>signal</a> with
-<var>reason</var> if it is given.
+to <a for=AbortController>request abort</a> on <a>this</a> with <var>reason</var> if it is given.
```
annevk, line 1782: Since it's not really a request, maybe just call this abort or also signal abort? (The for attribute helps with disambiguation and we can just not export the other one.)

dom.bs
```diff
@@ -1802,6 +1806,11 @@ interface AbortSignal : EventTarget {
  <dd>Returns an {{AbortSignal}} instance whose <a for=AbortSignal>abort reason</a> is set to
  <var>reason</var> if not undefined; otherwise to an "{{AbortError!!exception}}" {{DOMException}}.
 
+ <dt><code>AbortSignal . <a method for=AbortSignal lt=any(signals)>any</a>(<var>signals</var>)</code>
+ <dd>Returns an {{AbortSignal}} instance which will be aborted if any of <var>signals</var> is
```
annevk, line 1810: once any of*?

dom.bs
```diff
@@ -1921,10 +1945,10 @@ is [=AbortSignal/aborted=]; otherwise false.
 <dfn event for=AbortSignal><code>abort</code></dfn>.
 
 <p class=note>Changes to an {{AbortSignal}} object represent the wishes of the corresponding
-{{AbortController}} object, but an API observing the {{AbortSignal}} object can chose to ignore
+{{AbortController}} object, but an API observing the {{AbortSignal}} object can choose to ignore
 them. For instance, if the operation has already completed.
 
-<p>To <dfn export for=AbortSignal>signal abort</dfn>, given an {{AbortSignal}} object
+<p>To <dfn noexport for=AbortSignal>signal abort</dfn>, given an {{AbortSignal}} object
```
annevk, line 1951: Removing the export attribute should suffice.

dom.bs
```diff
@@ -1933,35 +1957,83 @@ them. For instance, if the operation has already completed.
...
+
+ <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>dependent signals</a>.
+
+ <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>source signals</a>.
 </ol>
 
-<p>A <var>followingSignal</var> (an {{AbortSignal}}) is made to
-<dfn export for=AbortSignal>follow</dfn> a <var>parentSignal</var> (an {{AbortSignal}}) by running
-these steps:
+<p>To <dfn export for=AbortSignal>create a composite abort signal</dfn> from a list of
```
annevk, line 1976: Let's drop the `for` attribute here. It's not really a member of an AbortSignal and it's unique enough.

dom.bs
```diff
@@ -1933,35 +1957,83 @@ them. For instance, if the operation has already completed.
...
+{{AbortSignal}} objects <var>signals</var> or a single {{AbortSignal}} <var>signal</var>, an
+optional <var>signalInterface</var>, which must be either {{AbortSignal}} or an interface that
+inherits from it, and an optional <var>realm</var>, run these steps:
```
annevk, line 1979: This is confusing. Let's always require a list. Folks can use the list syntax from Infra.

And do we want an interface or some kind of creation algorithm? (Also, if you want it to default to AbortSignal you should specify that here using the language from Infra, but please only do that if there are many callers. Otherwise it's not really worth it.)

dom.bs
```diff
@@ -1933,35 +1957,83 @@ them. For instance, if the operation has already completed.
...
- <li>
-  <p>Otherwise, <a for=AbortSignal lt=add>add the following abort steps</a> to
-  <var>parentSignal</var>:
+ <li><p>If <var>signal</var> was given, then set <var>signals</var> to a new <a for="/">list</a> and
+ <a for=list>append</a> <var>signal</var> to it.
+
+ <li><p><a for=list>For each</a> <var>signal</var> of <var>signals</var>: if <var>signal</var> is
+ [=AbortSignal/aborted=], then set <var>resultSignal</var>'s [=AbortSignal/abort reason=] to
+ <var>signal</var>'s [=AbortSignal/abort reason=] and return <var>resultSignal</var>.
+ </li>
```
annevk, line 1994: No need for `</li>` tags.

annevk (commented):
I've done a quick editorial take. This would probably benefit from review by @smaug----, @jakearchibald, and @domenic.

dom.bs
jakearchibald: I think the idea is array in, array out.

dom.bs
shaseley: Done.

dom.bs
shaseley: Done.

dom.bs
shaseley: That's fair - changed to requiring a list.

Regarding interface vs. creation algorithm: The main reason it has this form is to ensure the signal is newly created (i.e. not associated with a controller), since that's an assumption in the algorithm. The language ~= that of "create an event". There's another alternative where the exported algorithm just takes an optional realm and delegates to a private algorithm that takes a signal (to also be used by TaskSignal), but that's probably overkill?

I removed the default value since this likely won't have many callers callers (users of follow (i.e. fetch) and the TaskSignal specialization).

dom.bs
shaseley: Thanks, done.

dom.bs
```diff
@@ -1933,35 +1957,83 @@ them. For instance, if the operation has already completed.
...
 
+ <li><p><a for=list>For each</a> <var>signal</var> of <var>signals</var>:
```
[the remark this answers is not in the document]
shaseley: Thanks, fixed.

shaseley:
Thanks for taking a first pass, @annevk; comments addressed.

dom.bs
annevk: I don't think that's what `Promise.any` does, but maybe the other statics do and this matches those for consistency.

jakearchibald: oh yeah, you're right. I read `Promise.all`.

jakearchibald:
@shaseley can you provide a bit of detail on why the spec needs to make the garbage collection more explicit than other specs? Is it observable in some way?

My understanding (please correct me if I'm wrong @annevk) is that spec algorithms should be written as simply as possible, and UAs are invited to implement them in a more complex way that makes them work faster / more efficient as long as the observable behaviour is the same.

annevk:
Well, although we don't always do a good job at this, we do need to describe relationships between objects in a way that allows GC to eventually succeed. And I don't think we should accept "agent lifetime" as a way out, although sometimes that is acceptable.

I hadn't looked at the GC section here yet, I was hoping @smaug---- could take a look at that. But now that I have:

* Strong references are implied so don't need to be stated.
* That something is a weak reference should be mentioned as part of the definition, not in a separate section.
* The GC section should mainly state when something can be GC'd or ideally what prevents GC when GC would otherwise be possible. https://xhr.spec.whatwg.org/#garbage-collection is an example of that, although I think there's been a suggestion that user agents will never collect when the connection is still open either, so maybe that should be a must not instead.

(I hope that at some point JS/IDL formalize some of these aspects better so we can ground terminology like "weak reference" in terms of a larger web platform memory model.)

jakearchibald:
Fair enough. I guess it ensures we don't later break it in a way that prevents this particular kind of GC.

annevk:
Yeah and it also helps if we all think a bit about memory usage and leaks. (E.g., why you need to explicitly clone a response to read it twice, as I'm sure you're well aware.)

shaseley:
Yeah, I wasn't sure if this section actually belonged or not, but there is precedent in the DOM spec [here](https://dom.spec.whatwg.org/#garbage-collection). I included this because GC/memory management of composite signals was a concern in the [related issue](https://github.com/whatwg/dom/issues/920), so I wanted to include it to hopefully help guide other implementors.

FYI: there's one design decision related to the linking algorithm that helps support memory management and is observable: AbortSignal.any() always links the resulting signal to a non-composite signal (timeout or associated with a controller). This basically flattens the signal graph to a depth of 1, meaning intermediate composite signals don't need to be kept alive just to propagate abort. This affects the order events fire, which is covered in tests.

> Well, although we don't always do a good job at this, we do need to describe relationships between objects in a way that allows GC to eventually succeed. And I don't think we should accept "agent lifetime" as a way out, although sometimes that is acceptable.
> 
> I hadn't looked at the GC section here yet, I was hoping @smaug---- could take a look at that. But now that I have:
> 
> * Strong references are implied so don't need to be stated.
> * That something is a weak reference should be mentioned as part of the definition, not in a separate section.

FWIW I was using https://dom.spec.whatwg.org/#garbage-collection as an example, which should probably also get cleaned up then?:
 1. "Nodes have a strong reference to registered observers in their registered observer list." → strong is implied
 2. "Registered observers in a node’s registered observer list have a weak reference to the node." → this should probably move to the section where registered observers are defined? Or maybe this is about a mutation observer's node list? I can't tell exactly what that reference is.

> * The GC section should mainly state when something can be GC'd or ideally what prevents GC when GC would otherwise be possible. https://xhr.spec.whatwg.org/#garbage-collection is an example of that, although I think there's been a suggestion that user agents will never collect when the connection is still open either, so maybe that should be a must not instead.
>

Thanks, I'll have a look at that and clean this up.

> (I hope that at some point JS/IDL formalize some of these aspects better so we can ground terminology like "weak reference" in terms of a larger web platform memory model.)

shaseley:
Ok, I updated the GC section based on @annevk's [suggestions](https://github.com/whatwg/dom/pull/1152#issuecomment-1424411085). The idea is that source signals can store dependent signals weakly as long as dependent signals are kept alive as outlined in the GC section. Doing so prevents long-lived source signals (e.g. application-level signal) from holding onto composite signals indefinitely.

dom.bs
```diff
@@ -1921,48 +1945,93 @@ is [=AbortSignal/aborted=]; otherwise false.
...
+
+ <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>source signals</a>.
 </ol>
 
-<p>A <var>followingSignal</var> (an {{AbortSignal}}) is made to
-<dfn export for=AbortSignal>follow</dfn> a <var>parentSignal</var> (an {{AbortSignal}}) by running
-these steps:
+<p>To <dfn export>create a composite abort signal</dfn> from a list of {{AbortSignal}} objects
+<var>signals</var>, using <var>signalInterface</var>, which must be either {{AbortSignal}} or an
+interface that inherits from it, and an optional <var>realm</var>, run these steps:
```
smaug----, line 1978: Isn't it rather useless to say "or an interface that inherits from it"

dom.bs
```diff
@@ -1921,48 +1945,93 @@ is [=AbortSignal/aborted=]; otherwise false.
...
 
 <ol>
- <li><p>If <var>followingSignal</var> is [=AbortSignal/aborted=], then return.
+ <li><p>Let <var>resultSignal</var> be a <a for=/>new</a> object implementing
+ <var>signalInterface</var> using <var>realm</var> if given, otherwise using the default behavior
+ defined in Web IDL.
```
smaug----, line 1983: If we say something like "defined in Foo", I'd expect that to be then link to the definition. Especially since this is doing something unusual with realms, the default behavior should be clear too.

dom.bs
bathos: [“object implementing X”](https://webidl.spec.whatwg.org/#implements) is defined by Web IDL and includes objects implementing Y if Y extends X.

shaseley: But here the parameter is the interface itself and not a platform object, so I don't think that can be used (the definition requires an object)? The interface parameter is used below to create an object that implements it. (That the parameter is an interface and not object is why I don't think the phrasing is useless -- but it would be if it was an object.)

bathos: Ah, got it, I read that too quickly. That makes sense to me.

dom.bs
annevk: FWIW, in general realm-of-creation is a giant mess across all web platform standards. We've had a few attempts at trying to clean that up, but none successful thus far. There's a Web IDL issue open on defining some kind of implicit realm that we could use in most cases, but it hasn't been worked on.

dom.bs
```diff
@@ -1802,6 +1806,11 @@ interface AbortSignal : EventTarget {
  <dd>Returns an {{AbortSignal}} instance whose <a for=AbortSignal>abort reason</a> is set to
  <var>reason</var> if not undefined; otherwise to an "{{AbortError!!exception}}" {{DOMException}}.
 
+ <dt><code>AbortSignal . <a method for=AbortSignal lt=any(signals)>any</a>(<var>signals</var>)</code>
+ <dd>Returns an {{AbortSignal}} instance which will be aborted once any of <var>signals</var> is
+ aborted. Its <a for=AbortSignal>abort reason</a> will be set to whichever one of <var>signals</var>
+ cause it to be aborted.
```
annevk, line 1812: caused? (Not entirely sure.)

dom.bs
```diff
@@ -1828,6 +1837,17 @@ JavaScript value. It is undefined unless specified otherwise.
 <a for=/>set</a> of algorithms which are to be executed when it is [=AbortSignal/aborted=]. Unless
 specified otherwise, its value is the empty set.
 
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">composite</dfn> (a boolean), which is
+initially false.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, which is a
+<a for=/>set</a> of weak references to {{AbortSignal}}s that the object is dependent on for its
```
annevk, line 1844: I would prefer AbortSignal objects over pluralizing the name of the object. (Here and elsewhere.)

dom.bs
```diff
@@ -1921,48 +1945,97 @@ is [=AbortSignal/aborted=]; otherwise false.
...
 
  <li><p>[=Fire an event=] named {{AbortSignal/abort}} at <var>signal</var>.
+
+ <li><p><a for=set>For each</a> <var>dependentSignal</var> of <var>signal</var>'s
+ [=AbortSignal/dependent signals=], [=AbortSignal/signal abort=] on <var>dependentSignal</var> with
+ <var>signal</var>'s [=AbortSignal/abort reason=].
+
+ <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>dependent signals</a>.
+
+ <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>source signals</a>.
```
annevk, line 1973: So technically we don't have to do this, right? In https://github.com/whatwg/dom/issues/1159#issuecomment-1428011868 we "decided" to not do this for mutation observers so maybe we probably shouldn't do it here either.

dom.bs
```diff
@@ -1921,48 +1945,97 @@ is [=AbortSignal/aborted=]; otherwise false.
...
+     <var>sourceSignal</var>, then <a for=iteration>continue</a>.
+
+     <li><p><a for=set>Append</a> <var>sourceSignal</var> to <var>resultSignal</var>'s
+     [=AbortSignal/source signals=].
+
+     <li><p><a for=set>Append</a> <var>resultSignal</var> to <var>sourceSignal</var>'s
+     [=AbortSignal/dependent signals=].
+    </ol>
   </ol>
+ <li>Return <var>resultSignal</var>.
```
annevk, line 2028: Missing newline before this.

dom.bs
```diff
@@ -1921,48 +1945,97 @@ is [=AbortSignal/aborted=]; otherwise false.
...
 </ol>
 
 
+<h4 id=abort-signal-garbage-collection>Garbage collection</h4>
+
+<p>A [=AbortSignal/composite=] {{AbortSignal}} object must not be garbage collected while its
+[=AbortSignal/source signals=] is non-empty unless it has no registered event listeners for its
```
annevk, line 2035: Maybe "and it has registered ... or its abort algorithms is non-empty"?

dom.bs
shaseley: Arg, thanks. I think causes or caused both work (using caused).

dom.bs
shaseley: Sounds good; done.

dom.bs
shaseley: Ah yes, technically these can be removed, and consistency sgtm. But the GC section should probably be updated to reflect that we need only consider non-aborted signals (this was previously implied by source signals being empty for aborted signals).

I've removed clearing the sets and tweaked the GC section to say "A non-aborted composite signal..."

dom.bs
shaseley: Done.

dom.bs
shaseley: Yeah that's probably better. I think "unless ..." was me trying to make the "and ... or" unambiguous, but it's a bit awkward.

shaseley (commented):
Thanks for reviewing!

dom.bs
```diff
@@ -1921,48 +1945,94 @@ is [=AbortSignal/aborted=]; otherwise false.
...
 
 <p class=note>Changes to an {{AbortSignal}} object represent the wishes of the corresponding
-{{AbortController}} object, but an API observing the {{AbortSignal}} object can chose to ignore
+{{AbortController}} object, but an API observing the {{AbortSignal}} object can choose to ignore
 them. For instance, if the operation has already completed.
 
-<p>To <dfn export for=AbortSignal>signal abort</dfn>, given an {{AbortSignal}} object
-<var>signal</var> and an optional <var>reason</var>, run these steps:
+<p>To <dfn for=AbortSignal>signal abort</dfn>, given an {{AbortSignal}} object <var>signal</var> and
+an optional <var>reason</var>, run these steps:
```
domenic, line 1952: Nit: while here, you could remove `, run these steps`, as is the modern style. And also remove it from the below newly-added algorithm.

dom.bs
```diff
@@ -1921,48 +1945,94 @@ is [=AbortSignal/aborted=]; otherwise false.
...
+ <li>
+  <p>Let <var>resultSignal</var> be a <a for=/>new</a> object implementing
+  <var>signalInterface</var> using <var>realm</var> if given, otherwise using the default behavior
+  defined in Web IDL.
 
- <li><p>If <var>parentSignal</var> is [=AbortSignal/aborted=], then
- <a for=AbortSignal>signal abort</a> on <var>followingSignal</var> with <var>parentSignal</var>'s
- [=AbortSignal/abort reason=].
+  <p class=XXX>As of the time of this writing Web IDL does not yet define any default behavior;
+  see <a href="https://github.com/whatwg/webidl/issues/135">whatwg/webidl#135</a>.
```
domenic, line 1983: It might be better just to make the realm argument mandatory. For the `AbortSignal.any` caller, the correct realm is the current Realm.

dom.bs
```diff
@@ -1828,6 +1837,17 @@ JavaScript value. It is undefined unless specified otherwise.
...
+initially false.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, which is a
+<a for=/>set</a> of weak references to {{AbortSignal}} objects that the object is dependent on for
+its [=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn>, which is a
+<a for=/>set</a> of weak references to {{AbortSignal}} objects that are dependent on it for their
+[=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
+
```
domenic, line 1850: This section has gotten a little disorganized. E.g. add/remove are no longer next to "abort algorithms"; "is aborted" is in the middle of a bunch of field definitions; most field definitions use "unless specified otherwise", but one uses "initially".

Suggested ordering:

- Fields first. I like "initially" better than "unless specified otherwise".
- `<hr>`
- API definitions
- `<hr>`
- Supporting algorithms, including add/remove and "is aborted".

dom.bs
```diff
@@ -1828,6 +1837,17 @@ JavaScript value. It is undefined unless specified otherwise.
...
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">composite</dfn> (a boolean), which is
+initially false.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, which is a
+<a for=/>set</a> of weak references to {{AbortSignal}} objects that the object is dependent on for
+its [=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn>, which is a
+<a for=/>set</a> of weak references to {{AbortSignal}} objects that are dependent on it for their
+[=AbortSignal/aborted=] state. Unless specified otherwise, its value is the empty set.
```
domenic, line 1849: I think these should be "weak sets" instead of "sets of weak references". The difference is not behaviorally significant, but I think you'd need more verbiage to rigorously manage a set of weak references. E.g., instead of "Append _sourceSignal_ to ..." you'd need "Append a weak reference to _sourceSignal_ to...". And you'd need to dereference and null-check and so on.

domenic (approved):
Normatively LGTM, a few suggestions for making things a bit cleaner.

dom.bs
```diff
@@ -1821,35 +1831,31 @@ interface AbortSignal : EventTarget {
...
 
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">composite</dfn> (a boolean), which is
+initially false.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn>, (a weak
+<a for=/>set</a> of {{AbortSignal}} objects that the object is dependent on for its
+[=AbortSignal/aborted=] state), which is initially empty.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>dependent signals</dfn> (a weak
+<a for=/>set</a> of {{AbortSignal}} objects that are dependent on the obejct for their
```
domenic, line 1854:
```suggestion
<a for=/>set</a> of {{AbortSignal}} objects that are dependent on the object for their
```

dom.bs
```diff
@@ -3469,6 +3540,7 @@ dictionary MutationObserverInit {
 <ul class=brief>
  <li>A <dfn for=MutationObserver id=concept-mo-callback>callback</dfn> set on creation.
  <li>A <dfn for=MutationObserver>node list</dfn> (a <a for=/>list</a> of weak references to
+
```
domenic, line 3543: Probably this was added accidentally

domenic (approved):
Looks great!

dom.bs
shaseley: Yes, thanks.

dom.bs
shaseley: Fixed.

shaseley (commented):
Thanks Domenic!

dom.bs
```diff
@@ -1924,48 +1934,109 @@ is [=AbortSignal/aborted=]; otherwise false.
...
-{{AbortController}} object, but an API observing the {{AbortSignal}} object can chose to ignore
+{{AbortController}} object, but an API observing the {{AbortSignal}} object can choose to ignore
 them. For instance, if the operation has already completed.
 
-<p>To <dfn export for=AbortSignal>signal abort</dfn>, given an {{AbortSignal}} object
-<var>signal</var> and an optional <var>reason</var>, run these steps:
+<hr>
+
+<p>An {{AbortSignal}} object is <dfn export for="AbortSignal">aborted</dfn> when its
+[=AbortSignal/abort reason=] is not undefined.
```
smaug----, line 1943: Is this quite right? What if one aborts a signal using undefined

annevk:
@shaseley could you please file the implementation and MDN bugs and update OP? With Mozilla's support and WebKit's likely support I'm comfortable going ahead with this. Thanks for all your work on this!

shaseley:
> @shaseley could you please file the implementation and MDN bugs and update OP? With Mozilla's support and WebKit's likely support I'm comfortable going ahead with this. Thanks for all your work on this!

Done! Thanks so much to everyone for the reviews and help!

dom.bs
```diff
@@ -1924,46 +1947,114 @@ is [=AbortSignal/aborted=]; otherwise false.
...
 
 <ol>
- <li><p>If <var>followingSignal</var> is [=AbortSignal/aborted=], then return.
+ <li><p>Let <var>resultSignal</var> be a <a for=/>new</a> object implementing
+ <var>signalInterface</var> using <var>realm</var>.
 
- <li><p>If <var>parentSignal</var> is [=AbortSignal/aborted=], then
- <a for=AbortSignal>signal abort</a> on <var>followingSignal</var> with <var>parentSignal</var>'s
- [=AbortSignal/abort reason=].
+ <li><p>Set <var>resultSignal</var>'s [=AbortSignal/composite=] to true.
```
annevk, line 2008: Should we only do this after iteration and only if _signals_ is not the empty list?

Not sure why we would return a composite signal if any of the dependent signals is already aborted for instance or if there are no dependent signals.

annevk (commented):
When reviewing the commit message I realized this would remove "follow" and thus impact Fetch and Streams. I started looking at what this would mean for Fetch in https://github.com/whatwg/fetch/pull/1646 which also made me realize that maybe the internal composite variable is set too eagerly.

I suggest we work out the dependent PRs for Fetch and Streams to ensure we don't miss anything.

dom.bs
shaseley: I think setting it after checking aborted signals is probably okay, or at least functionally equivalent?

If it's an empty list, I think it should still be composite. If not, the signal could be added as a source for another composite signal, e.g.

```js
let signal1 = AbortSignal.any([]);
// If `signal1` is not composite, it would be added as a source for `signal2`, but
// it will never be aborted.
let signal2 = AbortSignal.any([signal1, realSignal]);
```

That would affect GC eligibility: `signal1` should be GC-able immediately and `signal2` would be GC-able when `realSignal` is GCed.

I think this maybe comes down to how SW/internal signals are handled. In https://github.com/whatwg/fetch/pull/1646#issuecomment-1532212040, I'm advocating for SW not creating a composite signal (or at least creating one with a source signal it controls) to avoid composite signals being aborted directly, which I think would be surprising (a hidden source!).

shaseley:
> When reviewing the commit message I realized this would remove "follow" and thus impact Fetch and Streams. I started looking at what this would mean for Fetch in [whatwg/fetch#1646](https://github.com/whatwg/fetch/pull/1646) which also made me realize that maybe the internal composite variable is set too eagerly.
> 
> I suggest we work out the dependent PRs for Fetch and Streams to ensure we don't miss anything.

Agreed! I started working on this yesterday but didn't upload a patch yet. I sent a PR for streams -- that just mentions "follow" in a comment but doesn't actually use it (it warns against specs invoking "follow" or "signal abort" directly on the stream signal).

benjamingr:
Any chance I can raise the order of signals/event listeners for a minor revision before this lands?

I think it's really confusing adding `"abort"` listeners manually and adding them implicitly through `AbortSignal.any` has different execution order like in this test:

```js
  test(t => {
    const controller = new controllerInterface();
    const signals = [];
    // The first event should be dispatched on the originating signal.
    signals.push(controller.signal);
    // All dependents are linked to `controller.signal` (never to another
    // composite signal), so this is the order events should fire.
    signals.push(signalInterface.any([controller.signal]));
    signals.push(signalInterface.any([controller.signal]));
    signals.push(signalInterface.any([signals[0]]));
    signals.push(signalInterface.any([signals[1]]));

    let result = "";
    for (let i = 0; i < signals.length; i++) {
      signals[i].addEventListener('abort', () => {
        result += i;
      });
    }
    controller.abort();
    assert_equals(result, "01234");
  }, `Abort events for ${desc} signals fire in the right order ${suffix}`);
```

Which effectively behaves like there is a capability letting you listen to abort in arbitrary order namely: my expectation above is the log to be "41230" and not "01234" since AbortSignal.any "listens" on that signal before the "abort" listener.

So when we do `controller.abort` in the above test I'd expect the AbortSignal.any to get the "abort" listener (conceptually, I get that's not the machinery) first and then to fire any listeners they have attached.

dom.bs
annevk: I see, that makes sense. In that case setting this state upfront is clearer.

And `Request` objects creating empty composite signals makes sense as well as there is not always something controlling them and this would make that more explicit. Though maybe that means composite isn't completely the right term? Perhaps they are "uncontrolled" / "unowned" or some such?

annevk:
Why would that be the expectation? If you had a similar situation with promises and the event dispatch happens at `resolve()` you'd get the exact same results.

benjamingr:
> Why would that be the expectation? If you had a similar situation with promises and the event dispatch happens at resolve() you'd get the exact same results.

I don't think that's really a fair analogy since promises settle synchronously and react asynchronously so you get a "microtick" between each place `.any` or `.all` would block if you "unleash zalgo" and have promises that call their `then` handlers synchronously I would absolutely expect `41230` and not 012345`.

Namely, consider:

```js
function any(promises) {
  return new Promise(r => promises.map(x => x.then(r)));
}
```

As a naive implementation of _any_ without correct error semantics - with this implementatiion the following:

```js
{
    let resolve;
    const promise = new Promise(r => resolve = r);
    const promises = [];

    promises.push(promise);

    promises.push(any([promise]));
    promises.push(any([promise]));
    promises.push(any([promises[0]]));
    promises.push(any([promises[1]]));
    console.log(promises)
    let result = "";
    for (let i = promises.length - 1; i >= 0; i--) {
      promises[i].then(() => {
        result += i;
      });
    }
    resolve();
    await Promise.all(promises);
    console.log(result);
}
```

would log `01234`  since each `.then` callback is deferred a micro-tick.

On the other hand if we consider signals, the "straightforward" implementation which is:

```js
function any(...signals) {
  const ac = new AbortController();
  // ignore case already aborted for example
  signals.forEach(s => s.addEventListener("abort", e => ac.abort(e));
  return ac;
}
```

Would log 41230 since handlers use the same machinery as regular event listeners (addEventListeners).

Basically, using the existing machinery for promises leads to this order but using EventTargets does not.

This change fundamentally uses a way to subscribe to signals that is _different_ from what is accessible to userland and userland can't reproduce in a spec-complaint way on a regular EventTarget. This is unlike promises where the order is coonsistent.

benjamingr:
> I think I see now. That seems hard to reconcile with the GC goals. Do you have an idea for that?

Sure - instead of signals following other signals internally we use the same machinery of `addEventListener` which would make the spec much simpler - we instead absorb the complexity in EventTarget that accepts an (internal only) "weak" flag to tell the implementation not to hold the listener strongly (in spec speak this wouldn't go through the public addEventListener but through "add an event listener" directly).

That would unify the underlying mechanism and this would remain "just another EventTarget" instead of having a separate subscription mechanism and instead we gain the ability to have weak event listeners (only internally for now).

The cool bit is that this would also enable weak event listeners in the future more generally if we want to expose those to user-land.

benjamingr:
(To make sure this is clear - I am very much in favor of this API)

annevk:
We quite deliberately don't use event listeners internally because of ordering concerns. That's why we have "abort algorithms". Perhaps those could be weakly held as well though, not sure.

benjamingr:
@annevk interesting, I didn't know that (or didn't remember) can you elaborate on that?  What is the ordering concern in this case?

shaseley:
Part of the reason for the current ordering is that dependent signals are always linked to root source signals, which flattens the dependency tree so it has a depth of at most 1. This cuts out intermediate nodes so they aren't being kept alive just to propagate abort, e.g. if you have a long chain of 'signal following signal following signal, etc.' I think even with weak event listeners, we'd need to keep intermediate signals alive to get the order you want?

dom.bs
shaseley: Moved setting the flag until after checking the input signals for being aborted.

Not sure this is the right framing, but I've been thinking of such signals as being 'composed of nothing' or 'dependent on the empty set (of signals)', which is inert but still composite/dependent (note: signals can reach this state if source signals are GCed). Would 'dependent' is better, or does that have the same problem? 'dependent' could be nice because it matches the 'dependent signals' variable.

I have a slight preference for dependent/composite because it mentions a relationship to the input signals, but I don't feel strongly. unowned/uncontrolled could work too, but I think an "create an uncontrolled abort signal given [empty set]" would be less clear/readable than "create an dependent abort signal given [empty set]".

annevk:
@benjamingr see https://github.com/whatwg/dom/issues/878. You wouldn't want a userland-added event listener to be able to stop your listener from running, for instance.

dom.bs
annevk: Dependent seems reasonable. Thanks!

annevk:
An earlier event listener can call `stopImmediatePropagation()`.

andreubotella:
I'm confused by the fact that dependent signals is a weak set, because it seems like it might prevent the `abort` event from being fired in cases like this:

```js
const ac = new AbortController();

AbortSignal.any([ac.signal]).onabort = () => {
  doSomething();
};

// The signal returned from AbortSignal.any could be collected at some
// point between the end of this script and the click event.

buttonEl.onclick = () => ac.abort();
```

shaseley:
> Yeah though I suspect very long chains with that sort of dependencies aren't common/practical anyway? Any chance you can provide an example? The motivating examples I've used in the past is "signal for server + signal for each request" in Node and "tree of components where cancellation propagates" in web components/browsers.
> 
> In what case is an action hierarchical but the intermediate steps go "out of scopes" before the last steps?

Maybe tree-of-components like thing happening async, e.g. everything is scheduled (child tasks run after/are scheduled from parent tasks)? Abstract, but internally someone brought up something like task A spawns B spawns C etc. where each has a signal argument that gets cloned using the API.

I'm not sure how realistic such situations will be. Intuitively, I thought (think) hierarchical abort made sense: source is aborted completely, then dependents, rather than things being interleaved. In that case, this optimization still affects the order that things fire (hence speccing it), but it holds the invariant that sources are aborted before dependents.

> This change fundamentally uses a way to subscribe to signals that is _different_ from what is accessible to userland and userland can't reproduce in a spec-complaint way on a regular EventTarget. This is unlike promises where the order is coonsistent.

I think you could monkeypatch `AbortSignal.prototype.abort` and `AbortSignal.timeout` to get the proposed ordering? Here's a [proof-of-concept](https://gist.github.com/shaseley/f16655b5a1e2ddcb2341d39d6ea5f2e8) polyfill that provides the spec ordering for the test in question.

shaseley:
> Edit: Oh, I see that the garbage collection explicitly covers this case. My bad. IIUC an implementation might fulfill that requirement by making the dependent signals a strong set, and removing signals from it when they are aborted. But I guess that is an implementation detail.

Yep, that's right. When implementing the prototype, I started with strong sets to make sure everything was working, then implemented memory management/weakness on top.

dom.bs
shaseley: Great, done.

annevk:
I updated OP with the follow-up issues and PRs. I will merge this Monday and the follow-up PRs likely on Tuesday.

benjamingr:
I'm more confused than when I started, I feel strongly that the execution order should be consistent and there should be a safe variant for "abort" that triggers all listeners but I don't have capacity to deal with either. 

I don't think we've justified the added complexity of not using "abort" events and the argument for not using "abort" because it's not safe basically leaves userland without a solution.

However I don't understand this well enough to support/object to either order. So once this lands Node.js will implement whatever you pick I'm just concerned.

shaseley:
@annevk re: ordering, "follow" used abort algorithms (so yet another ordering), and I just realized changing fetch to use the new algorithm would be observable (e.g. the order is swapped in [this test](https://github.com/web-platform-tests/wpt/blob/master/fetch/api/abort/general.any.js#L569) when replacing "follow").

I'm not sure if this would cause problems or not, but it's observable. If we wanted to maintain compat with "follow" and use a common algorithm (this part's a goal), I think we'd need to use abort algorithms, which would complicate this somewhat (e.g. we'd probably need weak abort algorithms).

benjamingr:
Or we don't flatten the tree and use `addEventListener` (and add a weak flag to that). Then solve the issue with `stopImmediatePropagation` preventing abort later in a separate PR?

annevk:
@shaseley I guess it's worth investigating how complex using abort algorithms would be to not change that test, but if it ends up being unwieldy I think changing the expectation should be safe enough.

@benjamingr `addEventListener()` would change ordering as well and disallowing `stopImmediatePropagation()` does not appear to have agreement. #1195 investigates a better userland pattern.

shaseley:
> @shaseley I guess it's worth investigating how complex using abort algorithms would be to not change that test, 

When creating a dependent signal, we'd need to add an algorithm to each signal provided (and not flatten the dependency graph since that affects order). The algorithms would need to be held weakly, tied to the dependent signal's lifetime in some way. Since IIUC abort algorithms aren't GCed objects, probably the way to do this would be to make "abort algorithms" a weak map using a unique GCed object as the key, either provided or created by "add" (we actually do this in blink to prevent algorithms from leaking).

But I think this gets a bit unwieldy because we'd need to either:
 1. Always hold algorithms weakly, changing all the specs that use AbortSignal to hold onto the key (or provide one). And also change add, remove, etc.
 1. Support both weak and strong algorithms, with dependent signals managing the "any" algorithm key objects and the target signal managing other algorithm keys. This is possible, but seems like an unfortunate bit of complexity.

> but if it ends up being unwieldy I think changing the expectation should be safe enough.
>

Yeah, I think changing the expectation should be reasonably safe, and I think that makes sense to do that. I don't think it's worth the complexity unless we want to always hold algorithms weakly (and update all the specs).

dom.bs
```diff
@@ -1821,35 +1838,32 @@ interface AbortSignal : EventTarget {
...
- <a for=AbortSignal>abort algorithms</a>.
-</ol>
+<p>An {{AbortSignal}} object has an associated <dfn export for=AbortSignal>abort reason</dfn> (a
+JavaScript value), which is initially undefined.
 
-<p>To <dfn export for=AbortSignal>remove</dfn> an algorithm <var>algorithm</var> from an
-{{AbortSignal}} <var>signal</var>, <a for=set>remove</a> <var>algorithm</var> from
-<var>signal</var>'s <a for=AbortSignal>abort algorithms</a>.
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>abort algorithms</dfn>, (a
+<a for=/>set</a> of algorithms which are to be executed when it is [=AbortSignal/aborted=]),
```
cdumez, line 1845: Shouldn't this be an ordered set to maintain ordering?

dom.bs
```diff
@@ -1821,35 +1838,32 @@ interface AbortSignal : EventTarget {
...
 <p class=note>The [=AbortSignal/abort algorithms=] enable APIs with complex
 requirements to react in a reasonable way to {{AbortController/abort()}}. For example, a given API's
 [=AbortSignal/abort reason=] might need to be propagated to a cross-thread environment, such as a
 service worker.
 
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">dependent</dfn> (a boolean), which is
+initially false.
+
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>source signals</dfn> (a weak
+<a for=/>set</a> of {{AbortSignal}} objects that the object is dependent on for its
```
cdumez, line 1857: ditto.

dom.bs
cdumez: The reason I am asking is that one of the WPT subtests (last one) seems to rely on a particular ordering.

shaseley: Yes, ordered sets are necessary here and below, but per Infra, all sets are ordered sets (see the note in https://infra.spec.whatwg.org/#ordered-set). The link produced by that reference directs to the ordered set definition.

dom.bs
```diff
@@ -1821,35 +1838,32 @@ interface AbortSignal : EventTarget {
...
+<p>An {{AbortSignal}} object has associated <dfn for=AbortSignal>abort algorithms</dfn>, (a
+<a for=/>set</a> of algorithms which are to be executed when it is [=AbortSignal/aborted=]),
+which is initially empty.
 
 <p class=note>The [=AbortSignal/abort algorithms=] enable APIs with complex
 requirements to react in a reasonable way to {{AbortController/abort()}}. For example, a given API's
 [=AbortSignal/abort reason=] might need to be propagated to a cross-thread environment, such as a
 service worker.
 
+<p>An {{AbortSignal}} object has a <dfn for="AbortSignal">dependent</dfn> (a boolean), which is
```
cdumez, line 1853: Do we really need this flag? Why can't we rely on `source signals` being non-empty?

dom.bs
```diff
@@ -1924,46 +1947,114 @@ is [=AbortSignal/aborted=]; otherwise false.
...
+     <li><p><a for=set>Append</a> <var>resultSignal</var> to <var>signal</var>'s
+     [=AbortSignal/dependent signals=].
+    </ol>
+
+   <li>
+    <p>Otherwise, <a for=list>for each</a> <var>sourceSignal</var> of <var>signal</var>'s
+    [=AbortSignal/source signals=]:
+
+    <ol>
+     <li><p>Assert: <var>sourceSignal</var> is not [=AbortSignal/aborted=] and not
```
cdumez, line 2034: It is not clear to me what ensure that the second part of the assertion holds. How do we know for sure that AbortSignal.any() cannot be called with an AbortSignal parameter that has it's `dependent` flag set?

Maybe we should recursively go through source signals?

dom.bs
shaseley: The difference has to do with empty signals and GC eligibility.

```js
const signal = AbortSignal.any([]);
…
const signal2 = AbortSignal.any([signal]);
```

`signal` is an "empty dependent signal", and such signals can never abort. When creating `signal2`, the "create a dependent signal" algorithm will not add `signal` as a source signal because its sources are empty;  if we remove the "dependent" flag, it will. It's a minor difference that won't affect correctness, but it could affect GC eligibility.

dom.bs
shaseley: It can be called with dependent signals. When it encounters one in the outer for-loop, it iterates over the dependent signal's source signals and links those. And by always linking to a dependent's source signals, dependent signals are never added as sources. So this is saying "a dependent signal's source signal cannot be a dependent signal." 

The idea is something like this:

```js
const root = someController.signal;

// Non-dependent signal passed, simple case (append it directly):
//   root.__dependent_signals = [signal1];
//   signal1.__source_signals = [root];
const signal1 = AbortSignal.any([someController.signal]);

// Dependent signal passed, iterate over its source signals and link those:
//   root.__dependent_signals = [signal1, signal2];
//   signal2.__source_signals = [root]; (note: root is linked, not signal1).
const signal2 = AbortSignal.any([signal1]);

// Same as above:
// root.__dependent_signals = [signal1, signal2, signal3];
// signal3.__source_signals = [root];
const signal3 = AbortSignal.any([signal2]);
```

This is by design to "flatten" the signal dependency graph by linking directly to signals that can abort directly (timeouts or controller signals), which helps optimize GC since intermediates aren't kept alive to propagate abort. (Note also that new sources are fixed at construction time).

PR #1160, opened 2023-02-13
Correct MutationObserver ownership story

### LINKED ISSUE #713: MutationObserver invocation order
Based on: https://github.com/jsdom/jsdom/pull/2398#discussion_r230243035

While implementing MutationObserver in jsdom, I wasn't able to make the following test pass while following the spec: https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/slotchange-event.html#L540-L594

The test run the following steps:
1. Adds 2 mutation observers (one on `id` attribute and another one on `title`) and a slot with a `slotchange` event handler.
2. Update the `id` attribute and the append a child in the light dom of the host. We then enqueue a microtask to notify the listeners. At this point the `id` attribute mutation observer contains 1 mutation record, and the signal slot list contains the slot.
3. The [notifiy mutation observer](https://dom.spec.whatwg.org/#notify-mutation-observers) algorithm is executed later on:
	1. The mutation observer list is copied, it contains the 2 mutation observers, and it iterates over the 2 mutation observers
	2. The callback for the `id` mutation observer is executed and in it's callback it updates title attribute and append a new element the light dom of the host. By updating the `title` attribute a new mutation record is enqueued but this time for the `title` mutation observer.
	3. The callback for the `title` mutation observer is invoked because it has a mutation record. And the **test throws here since it expects that the `slotchange` event is invoked first**.

[Chrome](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/mutation_observer.cc?sq=package:chromium&g=0&l=349-350) and [Webkit](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/MutationObserver.cpp#L273) don't run into the same issue since they keep track of an active mutation observer list instead of iterating over all the mutation observers.

### LINKED ISSUE #720: "Append mo to unit of related similar-origin bro..."
https://dom.spec.whatwg.org/commit-snapshots/23ec6421f47ab629bd14dfc6d4910414dc443d0b/#ref-for-list-append⑦

> Append mo to unit of related similar\-origin browsing contexts' mutation observer list\. 

If my understanding is correct, this should be held as a weak pointer, not the traditional strong pointer. Otherwise, `MutationObserver` objects can never be garbage collected.

This probably deserves at least a note in the spec. Or, you could change the list to "weak set", as is done for https://html.spec.whatwg.org/multipage/webappapis.html#outstanding-rejected-promises-weak-set . Or we could start formalizing this concept of reference graphs in Infra.

### LINKED ISSUE #1159: MutationObserver GC section looks bogus
As discovered in https://github.com/whatwg/dom/pull/1152#issuecomment-1424788407 by @shaseley.

1. Strong references are implied.
2. The weak reference is about a concept the specification doesn't acknowledge at all. I vaguely recall that it might have been refactored away.

annevk:
The existing design had some leaks and was also different from implementations. In particular, assume you have two mutation observers and a change triggered the first one as well as a slot change. Then if in the reaction to the first one a further change triggers the second one it would run before the slot change per the existing specification, but that didn't match implementations and more importantly didn't allow for garbage collection.

Test: https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/slotchange-event.html#L540-L594.

Fixes #713. Fixes #1159. Closes #720.

***
<a href="https://whatpr.org/dom/1160.html" title="Last updated on Feb 17, 2023, 9:12 AM UTC (5d7c322)">Preview</a> | <a href="https://whatpr.org/dom/1160/0111a95...5d7c322.html" title="Last updated on Feb 17, 2023, 9:12 AM UTC (5d7c322)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3467,12 +3467,12 @@ dictionary MutationObserverInit {
 
 <p>Each {{MutationObserver}} object has these associated concepts:
 
-<ul>
- <li>A <dfn noexport for=MutationObserver id=concept-mo-callback>callback</dfn> set on creation.
- <li>A <dfn noexport for=MutationObserver>node list</dfn> (a <a for=/>list</a> of weak references to
+<ul class=brief>
+ <li>A <dfn for=MutationObserver id=concept-mo-callback>callback</dfn> set on creation.
+ <li>A <dfn for=MutationObserver>node list</dfn> (a <a for=/>list</a> of weak references to
  <a for=/>nodes</a>), which is initially empty.
- <li>A <dfn export for=MutationObserver id=concept-mo-queue>record queue</dfn> (a <a for=/>queue</a>
- of zero or more {{MutationRecord}} objects), which is initially empty.
+ <li>A <dfn for=MutationObserver id=concept-mo-queue>record queue</dfn> (a <a for=/>queue</a> of
+ zero or more {{MutationRecord}} objects), which is initially empty.
 </ul>
 
 <dl class=domintro>
@@ -3609,7 +3609,7 @@ method steps are:
    <a for="registered observer">options</a> is <var>options</var> to <var>target</var>'s
    <a>registered observer list</a>.
 
-   <li><p><a for=list>Append</a> <var>target</var> to <a>this</a>'s
+   <li><p><a for=list>Append</a> a weak reference to <var>target</var> to <a>this</a>'s
    <a for=MutationObserver>node list</a>.
   </ol>
 </ol>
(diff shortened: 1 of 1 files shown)


annevk:
@rniwa @smaug---- @domenic @shaseley @FremyCompany @pmdartus I would appreciate your review!

shaseley:
LGTM. This matches what Blink's doing with the pending mutation observers.

annevk:
Thanks! I will merge this somewhere next week assuming Bikeshed issues have been resolved and there's no further feedback.

PR #1161, opened 2023-02-15
Prohibit ServiceWorker event handler update after the initialization.

yoshisatoyanagisawa:
This is a very early stage proposal on the specification change.  I hope to listen to what you think of the change.
I will update the following checklist afterwords.

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1121,8 +1121,7 @@ participate in a tree structure.</p>
  <a for="service worker">script resource</a>'s
  <a for="script resource">has ever been evaluated flag</a> is set, and <var>listener</var>'s
  <a for="event listener">type</a> matches the {{Event/type}} attribute value of any of the
- <a>service worker events</a>, then <a>report a warning to the console</a> that this might not give
- the expected results. [[!SERVICE-WORKERS]]
+ <a>service worker events</a>, then throws INVALID_MODIFICATION_ERROR exception. [[!SERVICE-WORKERS]]
 
  <li><p>If <var>listener</var>'s <a for="event listener">signal</a> is not null and is
  [=AbortSignal/aborted=], then return.


yoshisatoyanagisawa:
@jakearchibald @LeszekSwirski
I have drafted the specification update upon the discussion.  Please let me know your thoughts.
Thank you.

annevk:
Did you look into the history as to why this was not throwing? It's not clear to me why we are revisiting that.

yoshisatoyanagisawa:
Maybe https://github.com/whatwg/dom/issues/371 ?

As far as I checked the history, this used to throw but changed to just warn.
https://github.com/whatwg/dom/commit/7a48d64b41c586afc2e57d25eab99e8cfba5de1a

jakearchibald:
Past discussion is here https://github.com/whatwg/dom/issues/371#issuecomment-378220178

If the developer adds a service worker global event for "install", "activate", "fetch", "push" after the initial execution of the service worker, they're going to get behavior they don't expect. Worse, it might work as they expect in testing, but sometimes go wrong for real users.

The goal here is to avoid this case. Right now we show a console warning, which is ok, but could it be worth tightening the rules here so issues are found in development?

## Option 1: Throw

I'd forgotten that we'd discussed that previously. We could limit the throwing to a [subset of event types](https://w3c.github.io/ServiceWorker/#service-worker-events), but it seems that folks we still be against that.

## Option 2: Ignore

Instead of throwing, the call could be ignored. This would cause it to fail in dev, and (hopefully) cause the developer to see the console warning, and amend their code. Again, this could be limited to particular event types.

In both cases, the behaviour should be applied to both event listeners and event handlers.

annevk:
I prefer not changing this. If we got to do this again I'd probably push harder for service workers not using events, but now that they do use events they should stick to the event model as close as possible. Having certain known "types" be ignored or throw exceptions very much goes against that.

Contrived, I know, but you might very well have a custom `push` event you dispatch on the global or any other kind of known "type" we might add to the platform.

PR #1169, opened 2023-03-17
Export "event listener/type".

yoshisatoyanagisawa:
To be used in the service worker specification, let me export "event listener/type".
This is not a significant specification change but just export a term to be used from other specifications.

***
<a href="https://whatpr.org/dom_standard/1169.html" title="Last updated on Mar 17, 2023, 9:53 AM UTC (c18e7aa)">Preview</a> | <a href="https://whatpr.org/dom/1169/4db09dd...c18e7aa.html" title="Last updated on Mar 17, 2023, 9:53 AM UTC (c18e7aa)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -959,7 +959,7 @@ when something has occurred.
 <a>event</a> and consists of:
 
 <ul class=brief>
- <li><dfn for="event listener">type</dfn> (a string)
+ <li><dfn export for="event listener">type</dfn> (a string)
  <li><dfn for="event listener">callback</dfn> (null or an {{EventListener}} object)
  <li><dfn for="event listener">capture</dfn> (a boolean, initially false)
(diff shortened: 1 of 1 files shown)


yoshisatoyanagisawa:
@jakearchibald Will you take a look?
As you requested in https://github.com/w3c/ServiceWorker/pull/1672, I have drafted PR to export a term.

jakearchibald (changes requested):
Actually, you don't need this change, since the parent term is already exported.

Just reference it properly in the service worker PR 😄

yoshisatoyanagisawa:
Thank you for the comment.  Let me withdraw this PR, then.
And I noticed that I need to write the case like <a spec="dom">...</a> because of ambiguity not because was not exported.

PR #1170, opened 2023-03-20
Export terms needed for ServiceWorker/1671

jakearchibald:
This exports the terms needed for https://github.com/w3c/ServiceWorker/pull/1672 (now https://github.com/w3c/ServiceWorker/pull/1674).

However, this includes exporting a term that was marked "Intentionally not exported". Soooooo, what now @annevk? 😄

We need to be able to look at the listeners to determine they're all no-op. I could add an algorithm to the DOM spec that gets the listeners of type "fetch" for a given service worker global. That would avoid exporting the whole concept.


annevk:
That sounds preferable. Preferably with a "legacy" prefix to prevent copycats. "legacy-obtain service-worker fetch event listeners" or some such that takes a service worker global and returns that subset of listeners.

jakearchibald:
@annevk done! PTAL

dom.bs
```diff
@@ -1243,6 +1243,25 @@ property of the event being dispatched.
 <p>Ideally, any new event APIs are defined such that they do not need this property. (Use
 <a href="https://github.com/whatwg/dom/issues">whatwg/dom</a> for discussion.)
 
+<p>To <dfn export>legacy-obtain service worker fetch event listeners</dfn> given a
+{{ServiceWorkerGlobalScope}} <var>global</var>, run these steps. They return a <a for=/>list</a> of
+{{EventListener}} objects.
+
+<ol>
+ <li><p>Let <var>eventListeners</var> be a new <a for=/>list</a>.
```
annevk, line 1251:
```suggestion
 <li><p>Let <var>callbacks</var> be « ».
```

annevk (commented):
Thanks, looks good modulo nits.

jakearchibald:
@annevk feedback addressed!

jakearchibald:
@annevk happy to land this?

dom.bs
```diff
@@ -1243,6 +1243,25 @@ property of the event being dispatched.
...
+
+ <ol>
+  <li><p>If <var>listener</var>'s <a for="event listener">type</a> is "<code>fetch</code>", and
+  <var>listener</var>'s <a for="event listener">callback</a> is not null, then
+  <a for=list>append</a> <var>listener</var>'s <a for="event listener">callback</a> to
+  <var>callbacks</var>.
+ </ol>
+
+ <li><p>Return <var>callbacks</var>.
+</ol>
```
jakearchibald, line 1264:
```suggestion
</ol>

```

jakearchibald:
Cheers!

PR #1176, opened 2023-03-25
Handle attribute changes after changing attribute

josepharhar:
At least in chromium, synchronous internal listeners for attribute changes which are run as "attribute change steps" are run after the attribute is actually changed, not before.

This affects popover attribute related algorithms in HTML: https://github.com/whatwg/html/pull/9048#discussion_r1141122175

   * Chromium
   * Gecko
   * WebKit
   * The tests in this HTML PR effectively test this: https://github.com/whatwg/html/pull/9048
   * Chromium: N/A
   * Gecko: N/A
   * WebKit: N/a

***
<a href="https://whatpr.org/dom/1176.html" title="Last updated on Jun 8, 2023, 3:38 PM UTC (40d679d)">Preview</a> | <a href="https://whatpr.org/dom/1176/90691ba...40d679d.html" title="Last updated on Jun 8, 2023, 3:38 PM UTC (40d679d)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6185,10 +6185,10 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 <a>attribute</a> <var>attribute</var> to <var>value</var>, run these steps:
 
 <ol>
+ <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>value</var>.
+
  <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
  <a for=Attr>element</a>, <var>attribute</var>'s <a for=Attr>value</a>, and <var>value</var>.
-
- <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>value</var>.
 </ol>
 
 <p>To <dfn export id=concept-element-attributes-append lt="append an attribute">append</dfn> an
--- a/dom.bs
+++ b/dom.bs
@@ -6209,13 +6209,15 @@ steps:
 <a>attribute</a> <var>attribute</var>, run these steps:
 
 <ol>
- <li><a for=list>Remove</a> <var>attribute</var> from <var>attribute</var>'s
- <a for=Attr>element</a>'s <a for=Element>attribute list</a>.
+ <li><p>Let <var>element</var> be <var>attribute</var>'s <a for=Attr>element</a>.</p></li>
+
+ <li><a for=list>Remove</a> <var>attribute</var> from <var>element</var>'s <a for=Element>attribute
+ list</a>.
 
  <li><p>Set <var>attribute</var>'s <a for=Attr>element</a> to null.
 
- <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
- <a for=Attr>element</a>, <var>attribute</var>'s <a for=Attr>value</a>, and null.
+ <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>element</var>,
+ <var>attribute</var>'s <a for=Attr>value</a>, and null.
 </ol>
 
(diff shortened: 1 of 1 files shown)


annevk (commented):
@josepharhar you want to change all call sites of "Handle attribute changes" presumably.

@smaug---- @rniwa I assume this is the same everywhere and this was just a bug.

smaug----:
In Gecko one can handle attribute changes before the value has been added/changed/removed or after, depending on the case. And both are useful. The reason is mostly that BeforeSetAttr can still access the old value (which can be used for unregistering element from some data structures or cancelling loads etc.), and AfterSetAttr can then trigger whatever the new value should do. But I'm not sure if this is all just implementation detail. Would need to read through how `attribute change steps ` is used.

And at least the pr would make the spec inconsistent. Why only `change` is modified, why not also `append` and others?

annevk:
Right, I agree that we should change all of them together.

josepharhar:
> Right, I agree that we should change all of them together.

Ok, I changed all of them

dom.bs
```diff
@@ -6185,54 +6185,54 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
- <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
- <a for=Attr>element</a>, <var>attribute</var>'s <a for=Attr>value</a>, and null.
-
  <li><a for=list>Remove</a> <var>attribute</var> from <var>attribute</var>'s
  <a for=Attr>element</a>'s <a for=Element>attribute list</a>.
 
  <li><p>Set <var>attribute</var>'s <a for=Attr>element</a> to null.
+
+ <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
+ <a for=Attr>element</a>, <var>attribute</var>'s <a for=Attr>value</a>, and null.
```
annevk, line 6218: Presumably we need to cache attribute's element beforehand?

Also, it's surprising that for element removal it runs before, but for attribute removal it's after.

Did you double check what implementations do here?

dom.bs
```diff
@@ -6185,54 +6185,54 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
  <li><p><a for=list>Replace</a> <var>oldAttr</var> by <var>newAttr</var> in <var>oldAttr</var>'s
  <a for=Attr>element</a>'s <a for=Element>attribute list</a>.
 
  <li><p>Set <var>newAttr</var>'s <a for=Attr>element</a> to <var>oldAttr</var>'s
  <a for=Attr>element</a>.
 
  <li><p>Set <var>oldAttr</var>'s <a for=Attr>element</a> to null.
+
+ <li><p><a>Handle attribute changes</a> for <var>oldAttr</var> with <var>oldAttr</var>'s
+ <a for=Attr>element</a>, <var>oldAttr</var>'s <a for=Attr>value</a>, and <var>newAttr</var>'s
```
emilio, line 6234: You need to handle attribute changes with `newAttr`'s element, etc. `oldAttr`'s element was assigned to null the line before.

dom.bs
josepharhar: > Presumably we need to cache attribute's element beforehand?

Done

> Also, it's surprising that for element removal it runs before, but for attribute removal it's after.
> 
> Did you double check what implementations do here?

There's things that run before and after element removal in chromium: https://github.com/whatwg/dom/pull/1185#issuecomment-1520926675

As for attribute changes, I just did a quick look and it does not look like there is anything which runs prior to an attribute change in chromium.

dom.bs
josepharhar: done

annevk (approved):
This looks great, thanks!

@emilio did you want to have a final look at this as well?

emilio (approved):

PR #1185, opened 2023-04-17
Add before removing steps

josepharhar:
The existing "removing steps" allow other specs to run algorithms after a node has been removed from the tree, but there is no way to run steps before a node has been removed from the tree.

This patch adds new "before removing steps" to allow other specs to run algorithms before the node has been removed.

This is needed for this HTML issue:
https://github.com/whatwg/html/issues/9161

   * Chrome
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2754,10 +2754,11 @@ within a <var>parent</var>, run these steps:
 </ol>
 
 
-<p><a lt="Other applicable specifications">Specifications</a> may define
-<dfn export id=concept-node-remove-ext>removing steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed <var ignore>removedNode</var>, and optionally <var ignore>oldParent</var>, as
-indicated in the <a for=/>remove</a> algorithm below.
+<p><a lt="Other applicable specifications">Specifications</a> may define <dfn export
+id=concept-node-remove-ext>removing steps</dfn> and <dfn export
+id-concept-node-before-remove-ext>before removing steps</dfn> for all or some <a for=/>nodes</a>.
+These algorithms are passed <var ignore>removedNode</var>, and optionally <var
+ignore>oldParent</var>, as indicated in the <a for=/>remove</a> algorithm below.
 
 <p>To <dfn export id=concept-node-remove>remove</dfn> a <var>node</var>, with an optional
 <i>suppress observers flag</i>, run these steps:
@@ -2766,6 +2767,12 @@ indicated in the <a for=/>remove</a> algorithm below.
  <li><p>Let <var>parent</var> be <var>node</var>'s <a for=tree>parent</a>.
 
  <li><p>Assert: <var>parent</var> is non-null.
+
+ <li><p>Run the <a>before removing steps</a> with <var>node</var> and <var>parent</var>.
+
+ <li><p>For each <a>shadow-including descendant</a> <var>descendant</var> of <var>node</var>, in
+ <a>shadow-including tree order</a>, run the <a>before removing steps</a> with
+ <var>descendant</var>.</p></li>
 
  <li><p>Let <var>index</var> be <var>node</var>'s <a for=tree>index</a>.
 


annevk (commented):
Do implementations even run after steps or is this essentially the same problem as the attribute notification?

josepharhar:
Yes, implementations run steps after as well.
Things chromium does before removal:
- Hides popovers
- Fires sync mutation events
- Cleans up Range instances

Things chromium does after removal:
- Clears some style invalidation bits
- Removes the associated computed style
- Removes references to the element's id and name attributes from the document
- Removes the element from the document's top layer

I'm not sure how many of these are specced or not, but chromium definitely does stuff before and after removal and these are just a few points that I could easily explain by skimming the code - there are a lot more things that happen in both cases.

annevk:
Would this end up running script in the case of `hidePopover()`? Or would that always be prevented here?

josepharhar:
When the removal steps call hidePopover, the fireEvents flag is set to false which prevents any script from being run.

mbrodesser-Igalia:
> Yes, implementations run steps after as well.
> Things chromium does before removal:
> 
>     Hides popovers
>     Fires sync mutation events
>     Cleans up Range instances
> 
> Things chromium does after removal:
> 
>     Clears some style invalidation bits
>     Removes the associated computed style
>     Removes references to the element's id and name attributes from the document
>     Removes the element from the document's top layer

That implies, the "hide a popover" algorithm (https://html.spec.whatwg.org/multipage/popover.html#hide-popover-algorithm) needs to be adapted too, since the element is removed from the document's top layer in that algorithm.

annevk:
> Things chromium does before removal

Except for popover we handle all of those in DOM directly. Should we really make that an extension point? Is there some other way to design popover?

annevk:
That the DOM Standard handles them in its own algorithms without providing hooks. (Technically not true for mutation events, but there's an open issue for that and if we ever did specify those we wouldn't do it through a hook.)

josepharhar:
> Except for popover we handle all of those in DOM directly. Should we really make that an extension point? Is there some other way to design popover?

We could call hide popover from the DOM spec instead of exposing an extension point.

I did [some investigation](https://chromium-review.googlesource.com/c/chromium/src/+/4470968) and it actually turns out that chromium doesn't exactly run hide popover before removal, but does it at an intermediate stage where the state looks like this:
- The popover element's [parent](https://dom.spec.whatwg.org/#concept-tree-parent) is null.
- The popover element's [connected flag](https://dom.spec.whatwg.org/#connected) is true.

I'm not sure if this state can exist in the spec because step 11 of the [removing steps](https://dom.spec.whatwg.org/#concept-node-remove) probably atomically removes the parent and sets the connected flag to false - chromium has separate state for these.

The reason that chromium does this is because it turns out that we don't truly have a hook for before removing steps on any node, at least not yet. Mason told me that this is basically close enough to truly running hide popover before removal though, and I agree. If that's not the case, then we can change it in chromium.

I tried [calling hide popover after removal instead of before](https://chromium-review.googlesource.com/c/chromium/src/+/4577762), but it caused 7 different WPTs to fail. I think that if we shouldn't add an extension, then we should call hide popover from the DOM spec at the right place before removal.

annevk:
I guess what I'm wondering is whether we could do it in some other way without modifying the DOM specification.

josepharhar:
I suppose that if we added a parameter to the hide popover algorithm which tells it that the element is disconnected and we should hide it anyway, and also add that parameter to check popover validity, then it could theoretically be possible to hide after disconnecting and therefore not change the DOM spec. We would probably also need to pass in a document since I assume that the element would no longer have a "node document" after being disconnected

annevk:
That sounds pretty good to me, thanks for investigating! @rwlbuis @nt1m?

mbrodesser-Igalia:
An alternative idea: when https://html.spec.whatwg.org/#hide-popover-algorithm is invoked, instead of invoking https://html.spec.whatwg.org/#check-popover-validity from there, check only if https://html.spec.whatwg.org/#popover-visibility-state is "hidden" and if so return early.
Be aware that https://html.spec.whatwg.org/#popover-visibility-state is set to "hidden" only in https://html.spec.whatwg.org/#hide-popover-algorithm.

https://html.spec.whatwg.org/#dom-hidepopover would need to be extended to check whether the popover attribute is not present and throw in that case. Similar adaptations might be needed for other callers of https://html.spec.whatwg.org/#hide-popover-algorithm, but it might overall be a simpler solution than https://github.com/whatwg/dom/pull/1185#issuecomment-1579186990.

nt1m:
Before removing steps is how it's done in WebKit, but I'm open to different solution too.

josepharhar:
> An alternative idea: when https://html.spec.whatwg.org/#hide-popover-algorithm is invoked, instead of invoking https://html.spec.whatwg.org/#check-popover-validity from there, check only if https://html.spec.whatwg.org/#popover-visibility-state is "hidden" and if so return early.

If we don't call check popover validity as the first thing in hide popover algorithm, then we would be skipping opportunities to throw exceptions. For example, this should throw an exception:
```javascript
const div = document.createElement('div');
// don't add the popover attribute
div.hidePopover(); // throws because there is no popover attribute
```

> https://html.spec.whatwg.org/#dom-hidepopover would need to be extended to check whether the popover attribute is not present and throw in that case. Similar adaptations might be needed for other callers of https://html.spec.whatwg.org/#hide-popover-algorithm, but it might overall be a simpler solution than https://github.com/whatwg/dom/pull/1185#issuecomment-1579186990.

I see, yeah I guess that in the case where we want to hide a popover regardless of the dom state, the only relevant check in the check popover algorithm is whether or not the popover is in the showing or hidden state, so we could conditionally replace calls to the check popover algorithm with checks for the popover visibility state. Sounds good to me, and it could also be used to fix https://github.com/whatwg/html/issues/9367

mbrodesser-Igalia:
> I see, yeah I guess that in the case where we want to hide a popover regardless of the dom state, the only relevant check in the check popover algorithm is whether or not the popover is in the showing or hidden state, so we could conditionally replace calls to the check popover algorithm with checks for the popover visibility state. Sounds good to me, and it could also be used to fix https://github.com/whatwg/html/issues/9367

Keep in mind:
```
<div>X</div>
<script>
let d = document.querySelector("div");
d.hidePopover(); // Should throw according to spec and does in Chrome dev edition.
</script>
```
However, https://jsfiddle.net/hyzLu4nk/1/ doesn't throw in Chrome dev edition. Adapting the spec to let https://html.spec.whatwg.org/#dom-hidepopover throw only when https://html.spec.whatwg.org/#popover-showing-or-hiding is `false` would cover that.

mbrodesser-Igalia:
> However, https://jsfiddle.net/hyzLu4nk/1/ doesn't throw in Chrome dev edition. 

Given https://github.com/whatwg/html/pull/9142#issuecomment-1538596372, it seems to make sense to specify that. @josepharhar WDYT?
@nt1m would you support specifying this?

Locally I've a WPT for that scenario which I'd request to pull once the above questions are clarified.

mbrodesser-Igalia:
Given https://github.com/whatwg/dom/pull/1176 was merged, this can be closed?

annevk:
@mbrodesser-Igalia that's about attribute mutations, this is about tree mutations, so no?

mbrodesser-Igalia:
> @mbrodesser-Igalia that's about attribute mutations, this is about tree mutations, so no?

Yes. Wanted to post that in another ticket, sorry.

josepharhar:
I opened an HTML PR which should solve the bug without the need for adding the DOM hook this PR adds: https://github.com/whatwg/html/pull/9457

PR #1205, opened 2023-06-01
Fix typo AbortSignal _any -> any

Jamesernator:
This is just a typo in the IDL, everywhere else in the spec refers to the method as `any` as it should be.


annevk:
This is not a typo. `any` is a reserved identifier in Web IDL.

PR #1223, opened 2023-09-04
Add flat tree parent definition

josepharhar:
This can be used in the HTML spec as suggested in this issue: https://github.com/whatwg/html/issues/8970

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5956,6 +5956,20 @@ is an object or one of its <a>shadow-including ancestors</a>.
 <p class=note>The <a>retargeting</a> algorithm is used by <a lt=dispatch>event dispatch</a> as well
 as other specifications, such as <cite>Fullscreen</cite>. [[FULLSCREEN]]
 
+<p>To get the <dfn export>flat tree parent</dfn> of a <var>node</var>:</p>
+
+<ol>
+ <li><p>Let <var>parent</var> be <var>node</var>'s <a for=tree>parent</a>.</p></li>
+
+ <li><p>If <var>parent</var> is not a <a for=Element>shadow host</a>, then return
+ <var>node</var>.</p></li>
+
+ <li><p>If <var>node</var>'s <a for=slottable>assigned slot</a> is null, then return null.</p></li>
+
+ <li><p>Return <var>node</var>'s <a for=slottable>assigned slot</a>'s <a
+ for=tree>parent</a>.</p></li>
+</ol>
+
 
 <h3 id=interface-element>Interface {{Element}}</h3>
 


dom.bs
```diff
@@ -5956,6 +5956,23 @@ is an object or one of its <a>shadow-including ancestors</a>.
...
+ <li><p>If <var>parent</var> is a <a for=/>shadow root</a>, then return <var>parent</var>'s <a
+ for=DocumentFragment>host</a>.</p></li>
+
+ <li><p>If <var>parent</var> is not a <a for=Element>shadow host</a>, then return
+ <var>node</var>.</p></li>
+
+ <li><p>If <var>node</var>'s <a for=slottable>assigned slot</a> is null, then return null.</p></li>
+
+ <li><p>Return <var>node</var>'s <a for=slottable>assigned slot</a>'s <a
+ for=tree>parent</a>.</p></li>
```
josepharhar, line 5973: @mfreed7 do I need to run anything recursive in the last step here? I briefly tried to make a demo with a slot assigned into another slot, but it didn't seem to work.

josepharhar: I guess I at least need to check if the parent is a shadow host, right?

dom.bs
```diff
@@ -5956,6 +5956,23 @@ is an object or one of its <a>shadow-including ancestors</a>.
...
+<p>To get the <dfn export>flat tree parent</dfn> of a <var>node</var>:</p>
+
+<ol>
+ <li><p>Let <var>parent</var> be <var>node</var>'s <a for=tree>parent</a>.</p></li>
+
+ <li><p>If <var>parent</var> is a <a for=/>shadow root</a>, then return <var>parent</var>'s <a
+ for=DocumentFragment>host</a>.</p></li>
+
+ <li><p>If <var>parent</var> is not a <a for=Element>shadow host</a>, then return
+ <var>node</var>.</p></li>
```
annevk, line 5968: You want to return parent here, presumably.

dom.bs
```diff
@@ -5956,6 +5956,23 @@ is an object or one of its <a>shadow-including ancestors</a>.
 <p class=note>The <a>retargeting</a> algorithm is used by <a lt=dispatch>event dispatch</a> as well
 as other specifications, such as <cite>Fullscreen</cite>. [[FULLSCREEN]]
 
+<p>To get the <dfn export>flat tree parent</dfn> of a <var>node</var>:</p>
+
+<ol>
+ <li><p>Let <var>parent</var> be <var>node</var>'s <a for=tree>parent</a>.</p></li>
+
+ <li><p>If <var>parent</var> is a <a for=/>shadow root</a>, then return <var>parent</var>'s <a
```
annevk, line 5964: No wrapping in phrasing elements.

dom.bs
```diff
@@ -5956,6 +5956,23 @@ is an object or one of its <a>shadow-including ancestors</a>.
 <p class=note>The <a>retargeting</a> algorithm is used by <a lt=dispatch>event dispatch</a> as well
 as other specifications, such as <cite>Fullscreen</cite>. [[FULLSCREEN]]
 
+<p>To get the <dfn export>flat tree parent</dfn> of a <var>node</var>:</p>
```
annevk, line 5959: What about children of a slot element?

annevk (commented):
I was hoping this definition could mostly stay in CSS and in DOM we could use the composed tree (which we use for events).

Perhaps having both concepts in DOM is useful, but early on we were fairly deliberate on the split and I feel like we lost some of the knowledge as to why we did things that way.

cc @rniwa @hayatoito @domenic @smaug----

Issue #583, opened 2018-03-05
Allow custom "get the parent" algorithms for EventTargets.


zandaqo:
> In the future we could allow custom get the parent algorithms. Let us know if this would be useful for your programs. For now, all author-created EventTargets do not participate in a tree structure.

I'd like to see that future. That is, the ability to create a tree structure for author-created EventTargets like the one we have with Nodes, with bubbling and capturing, and so forth. Two points in support of this:
1. It would just make the platform more consistent when all EventTargets behave similar or have ways to do so. Right now we have a "weird" limitation to keep in mind when we work with custom EventTargets.
2. Listening just to a parent instead of each of its children simplifies our code when we work with Nodes, and same can apply to author-created EventTargets.

An example use case:
Let's take the classic Todo app example. Imagine that each todo item is a custom EventTarget emitting "change" events, and the whole list is an array. Now, if we need to respond to any change in the list, say, recalculate the number of items, we have to add event listeners to each item. However, if we could define our array as the parent for each item where events could bubble to (or be captured on) the parent, we could get away with just one listener on the array. Since the items and the array already have strong coupling, adding an extra "parent" pointer wouldn't make difference in terms of the design.

annevk:
To what extent are you interested in also having shadow tree capabilities for your custom event targets? E.g., being able to hide event targets from `composedPath()` and such.

zandaqo:
@annevk I cannot think of a use case for it. If we are building tree structures out of these custom EventTargets solely to propagate events, having a capability of hiding/skipping certain nodes seems superfluous--one can just avoid putting those hidden nodes into the said tree structure.

aayushch:
We are building web extensions which would like to implement events, to propagate critical occurrences from the browser api event listeners back to the extension business logic layer.

We would want to leverage the EventTarget implementation to build propagation, rather doing it in an ad-hoc way.

The ability to bubble the events is going to be a huge advantage given we can override/implement a custom 'get the parent' algorithm.
The shadow tree capabilities are not really necessary for author created event targets.

clshortfuse:
I'm currently using a polyfill for `EventTarget` in NodeJS environments. I see `Deno` will support `EventTarget` out of the box, so at least we're going to get nice usage from `EventTarget`.

But, for my polyfill, I'm creating a custom "get the parent" implementation via `WeakMap` as to not pollute the the property list.

````
/**
 * @typedef {Object} EventDispatcherEventTargetPrivateFields
[...]

PR #1230, opened 2023-10-10
Allow for customisation of the "get the parent" algorithm

keithamus:
This modifies the EventTarget IDL to allow assignment of a parent EventTarget to an EventTarget instance, while also modifying the "get the parent" algorithm to default to returning that instance.

It also modifies the Event Dispatch algorithm to ensure that custom parent chains cannot cause loops.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …


dom.bs
```diff
@@ -939,6 +946,34 @@ when something has occurred.
 <a for=/>list</a> of zero or more <a>event listeners</a>). It is initially the empty list.
 <!-- Intentionally not exported. -->
 
+<p>Each {{EventTarget}} object has an associated <dfn for=EventTarget>attached internals</dfn> (null
+or an {{EventTargetInternals}} object), initially null.
+<!-- Intentionally not exported. -->
```
annevk, line 951: No need for this comment, but I agree!

dom.bs
```diff
@@ -939,6 +946,34 @@ when something has occurred.
...
+
+<p>When a <dfn export for=Event id=concept-eventtarget-constructor>constructor</dfn> of the {{EventTarget}}
+interface, or of an interface that inherits from the {{EventTarget}} interface, is invoked, these steps
+must be run, given the arguments <var>eventTargetCallback</var>:
```
annevk, line 955: You need to write this similar to https://dom.spec.whatwg.org/#dom-abortcontroller-abortcontroller. So something like:

> The new EventTarget(callback) constructor steps are:

The argument name should match the argument name used in IDL.

dom.bs
```diff
@@ -939,6 +946,34 @@ when something has occurred.
...
+
+<ol>
+ <li><p>Let <var>eventTarget</var> be the result of creating a new object using this interface.
```
annevk, line 958: With the new-style constructor steps, creation already happened. You just need to manipulate "this".

dom.bs
```diff
@@ -939,6 +946,34 @@ when something has occurred.
...
+
+ <li><p>Initialize <var>eventTarget</var>'s {{EventTarget/event listener list}} attribute to the empty list.
+
+ <li><p>Initialize <var>eventTarget</var>'s {{EventTarget/parent}} attribute to null.
```
annevk, line 962: Aren't these the defaults already?

dom.bs
```diff
@@ -939,6 +946,34 @@ when something has occurred.
...
+
+ <li><p>Let <var>eventTargetInternals</var> a new {{EventTargetInternals}} instance.
```
annevk, line 964: Only create this when the callback is given.

dom.bs
```diff
@@ -939,6 +946,34 @@ when something has occurred.
...
+
+ <li><p>If <var>eventTargetCallback</var> is not null, then <a
+ for=/invoke-a-callback-function>invoke</a> <var>eventTargetCallback</var> with «
+  <var>eventTargetInternals</var> » and with <var>eventTarget</var> aas the <a>callback this
```
annevk, line 968: as*

annevk (commented):
Some initial feedback. Mainly around how to organize this. Overall I think this works, but would love to hear from @rniwa and @smaug---- especially.

dom.bs
```diff
@@ -956,8 +991,9 @@ when something has occurred.
 object, an <a>event listener</a> is a broader concept as can be seen above.
 
 <p>Each {{EventTarget}} object also has an associated <dfn export>get the parent</dfn> algorithm,
-which takes an <a>event</a> <var>event</var>, and returns an {{EventTarget}} object. Unless
-specified otherwise it returns null.
+which takes an <a>event</a> <var>event</var>, and returns an {{EventTarget}}. Unless otherwise
+specified otherwise it returns the associated <a for=EventTarget>attached internals</a><a
+for=EventTargetInternals">parent</a> attribute.
```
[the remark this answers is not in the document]
keithamus: Okay I've added some steps which I think resolve this but let me know, I'm sure it could use some refinement.

dom.bs
```diff
@@ -939,6 +946,50 @@ when something has occurred.
...
+
+<ol>
+  <li>If <var>theParent</var> is equal to <var>internals</var>' <a
+  for=EventTargetInternals>eventTarget</a> then <a>throw</a> a
+   "{{HierarchyRequestError!!exception}}" {{DOMException}}.
+
+  <li>If <var>theParent</var>'s <a for=EventTarget>attached internals</a>
+  {{EventTargetInternals/parent}} is equal to <var>internals</var>' <a
+  for=EventTargetInternals>eventTarget</a> then<a>throw</a> a "{{HierarchyRequestError!!exception}}"
+  {{DOMException}}.
```
keithamus, line 985: By providing steps to set the parent I also squeezed in a couple more early checks where we can assert that there aren't circular references going on.

dom.bs
```diff
@@ -939,6 +946,50 @@ when something has occurred.
...
+
+  <li>Set <var>internal</var>'s <a for=EventTargetInternals>eventTarget</a> <a>get the parent</a>
```
keithamus, line 987: I couldn't find precedent for setting an algorithm so this is just made up, hopefully it suffices?

dom.bs
```diff
@@ -1322,6 +1373,9 @@ property of the event being dispatched.
     <p>While <var>parent</var> is non-null:</p>
 
     <ol>
+     <li>If the <var>event</var>'s <a for=Event>path</a> <a for=set>contains</a> <var>parent</var>
+     then <a>throw</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}}.
+
```
smaug----, line 1378: This would be slow. I don't think we can do this in normal case when dispatching events to dom tree. The relevant code in implementations is so hot that extra virtual c++ calls show up, as an example.
We need to ensure that there are now loops when setting up the parents. That is already done for DOM nodes anyhow - DOM tree doesn't contain loops.

dom.bs
```diff
@@ -908,9 +908,14 @@ for historical reasons.
 <h3 id=interface-eventtarget>Interface {{EventTarget}}</h3>
 
 <pre class=idl>
+[Exposed=*]
+interface EventTargetInternals {
+  attribute EventTarget parent;
+};
+
```
smaug----, line 915: This feels a bit cumbersome API, one needs to pass a callback and then use its param to change a property value.

smaug---- (changes requested):

dom.bs
keithamus: Would you prefer us to check the chain during setting of the parent? Would that suffice?

dom.bs
keithamus: Would you prefer an init dictionary? One downside to this, as explained in https://github.com/whatwg/dom/issues/583#issuecomment-1754494834 is that an init dictionary cannot allow setting the parent post-construction.

dom.bs
annevk: I think it would, yeah.

dom.bs
annevk: Yeah, I suppose we could also do what `new Promise()` does, but it doesn't seem as extensible:
```js
new EventTarget(setParent => {
  setParent(someParent);
}
```
I think as long as the capability is only granted to whoever constructed the `EventTarget` object I'm okay, so alternative proposals welcome.

dom.bs
keithamus: Okay I think I've resolved this

dom.bs
lukewarlow: > I think as long as the capability is only granted to whoever constructed the EventTarget object l'm okay, so alternative proposals welcome.

Out of interest, why? Dom nodes can have their parent changed by someone other than the creator (by just moving the node in the Dom), so why does event target need to be protected?

annevk: We don't want the event path of arbitrary nodes to be changeable. And I doubt anyone adopting `EventTarget` would want that. (And if they do they can make a public API I suppose.)

PR #1244, opened 2024-01-09
Editorial: remove redundant contains check in AbortSignal.any()

vinhill:
Step 4.2.2. of [create a dependent abort signal](https://dom.spec.whatwg.org/#create-a-dependent-abort-signal) prevents duplicated signals in [source signals](https://dom.spec.whatwg.org/#abortsignal-source-signals) for example in the following snippet
```
const signal = AbortSignal.any([ controller.signal, AbortSignal.any([controller.signal]) ]);
```
so maybe the same should hold for 
```
const signal2 = AbortSignal.any([ AbortSignal.any([controller.signal]), controller.signal ]);
const signal2 = AbortSignal.any([ controller.signal, controller.signal ]);
```
To my knowledge, duplicated signals in source signals should not have any impact due to step 1 of [signal abort](https://dom.spec.whatwg.org/#abortsignal-signal-abort) ensuring signals are only aborted once. So this change should cause no behavior change.

### PR Questions

   * …
- [ ] [Tests](https://github.com/web-platform-tests/wpt):
   * There already is https://github.com/web-platform-tests/wpt/blob/2fcb60d7e9f294400c6274d964428621ca6cd3d2/dom/abort/resources/abort-signal-any-tests.js#L81C1-L88C72
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/dom/1244.html" title="Last updated on Jan 16, 2024, 8:52 AM UTC (c79e137)">Preview</a> | <a href="https://whatpr.org/dom/1244/da484da...c79e137.html" title="Last updated on Jan 16, 2024, 8:52 AM UTC (c79e137)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2079,9 +2079,6 @@ interface that inherits from it, and a <var>realm</var>:
      <li><p>Assert: <var>sourceSignal</var> is not [=AbortSignal/aborted=] and not
      [=AbortSignal/dependent=].
 
-     <li><p>If <var>resultSignal</var>'s [=AbortSignal/source signals=] <a for=set>contains</a>
-     <var>sourceSignal</var>, then <a for=iteration>continue</a>.
-
      <li><p><a for=set>Append</a> <var>sourceSignal</var> to <var>resultSignal</var>'s
      [=AbortSignal/source signals=].
 


vinhill:
My bad, it does. Would it make sense then to remove step 4.2.2.?

annevk:
@shaseley might know. Given that it's followed by two appends it's not immediately clear at least.

shaseley:
4.2.2 should be safe to remove since both 4.2.3 and 4.2.4 would be no-ops for a duplicate source. I don't think the difference was intentional (and I remember purposely omitting the check in 4.1 because of set behavior); there might have been a previous version of the `any()` PR where skipping the steps mattered, but that's not the case now -- so removing to match 4.1 SGTM.

vinhill:
Is it fine to do the change in this pull request, or should I open a new one? Can I omit implementation bugs given this change should have no impact?

annevk:
It's fine here. And yes, no need for bugs on editorial changes.

We'll have to prefix it with "Editorial: ", but I can do that when landing.

PR #1247, opened 2024-01-23
Draft integration with Trusted Types, take 2.

koto:
This calls the get Trusted Types-compliant attribute value algorithm from Trusted Types (https://github.com/w3c/trusted-types/pull/418) from attribute's change, append, and replace.

Changed the signature of `setAttribute` and `setAttributeNS` to accept Trusted Types as values. The underlying Attr node's values continue to be `DOMString`, so moving nodes across elements or adding standalone attributes to elements can cause TT violations. This matches WPT tests and the Chromium's implementation.

See  and #789. Supercedes PR #809.

   * Gecko
   * Chromium (already implemented)
   * https://github.com/web-platform-tests/wpt/tree/master/trusted-types
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=739170

***
<a href="https://whatpr.org/dom/1247.html" title="Last updated on Feb 26, 2024, 4:21 PM UTC (10ce041)">Preview</a> | <a href="https://whatpr.org/dom/1247/6a9b5a2...10ce041.html" title="Last updated on Feb 26, 2024, 4:21 PM UTC (10ce041)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -52,7 +52,8 @@ spec:html; type:element
 <p>This specification depends on the Infra Standard. [[!INFRA]]
 
 <p>Some of the terms used in this specification are defined in <cite>Encoding</cite>,
-<cite>Selectors</cite>, <cite>Trusted Types</cite>, <cite>Web IDL</cite>, <cite>XML</cite>, and <cite>Namespaces in XML</cite>.
+<cite>Selectors</cite>, <cite>Trusted Types</cite>, <cite>Web IDL</cite>, <cite>XML</cite>, and
+<cite>Namespaces in XML</cite>.
 [[!ENCODING]]
 [[!SELECTORS4]]
 [[!TRUSTED-TYPES]]
@@ -6358,7 +6359,8 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 <ol>
  <li><p>Let <var>oldValue</var> be <var>attribute</var>'s <a for=Attr>value</a>.</p></li>
 
- <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>, with <var>attribute</var>'s <a for=Attr>element</a>.
+ <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>, with
+ <var>attribute</var>'s <a for=Attr>element</a>.
 
  <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
  <a for=Attr>element</a>, <var>oldValue</var>, and <var>value</var>.
@@ -6410,13 +6412,17 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
-<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn> <var>value</var> for an <a>attribute</a> <var>attribute</var>, with <a>element</a> <var>element</var>, run these steps:
+<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn>
+<var>value</var> for an <a>attribute</a> <var>attribute</var>, with <a for=/>element</a>
+<var>element</var>, run these steps:
 
- <ol>
- <li><p>Let <var>validValue</var> be the result of calling <a>Get Trusted Types-compliant attribute value</a>
-  for <var>attribute</var>, with <var>element</var> and <var>value</var>.[[!TRUSTED-TYPES]]
+<ol>
+ <li><p>Let <var>validValue</var> be the result of calling
+ <a>Get Trusted Types-compliant attribute value</a> for <var>attribute</var>, with
+ <var>element</var> and <var>value</var>. [[!TRUSTED-TYPES]]
+
  <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>validValue</var>.
- </ol>
+</ol>
 
 <hr>
 
@@ -6480,7 +6486,8 @@ string <var>namespace</var> (default null):</p>
 
  <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
 
- <li><p><a>Validate and set attribute value</a> <var>newAttr</var>'s <a for="Attr">value</a> for <var>newAttr</var> with <var>element</var>.
+ <li><p><a>Validate and set attribute value</a> <var>newAttr</var>'s <a for="Attr">value</a> for
+ <var>newAttr</var> with <var>element</var>.
 
  <li><p>If <var>oldAttr</var> is non-null, then <a lt="replace an attribute">replace</a>
  <var>oldAttr</var> with <var>attr</var>.
@@ -6493,9 +6500,9 @@ string <var>namespace</var> (default null):</p>
 
 <div algorithm>
 <p>To <dfn export id=concept-element-attributes-set-value>set an attribute value</dfn> given an
-<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string or <a>TrustedType</a> <var>value</var>,
-an optional null or string <var>prefix</var> (default null), and an optional null or string
-<var>namespace</var> (default null):
+<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string or {{TrustedType}}
+<var>value</var>, an optional null or string <var>prefix</var> (default null), and an optional null
+or string <var>namespace</var> (default null):
 
 <ol>
  <li>Let <var>attribute</var> be the result of
@@ -6512,12 +6519,14 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
  <ol>
   <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>namespace</a> is
   <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
-  <a for=Attr>local name</a> is <var>localName</var> and
-  <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>.
+  <a for=Attr>local name</a> is <var>localName</var> and <a for=Node>node document</a> is
+  <var>element</var>'s <a for=Node>node document</a>.
 
-  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with <var>element</var>.
+  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with
+  <var>element</var>.
 
-  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>
+  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>.
+
   <li><p>Return.
  </ol>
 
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -6354,6 +6355,10 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 <a>attribute</a> <var>attribute</var> to <var>value</var>, run these steps:
 
 <ol>
+<li><p>Set <var>value</var> to the result of calling <a>Get Trusted Types-compliant attribute
```
annevk, line 6358: Nit: indentation is wrong. And DOM doesn't do newlines in phrasing-level elements (i.e., wrap before `<a>`).

annevk: Question: this ordering means that if _value_ is changed by TT the mutation record reflects that. Does that have test coverage?

dom.bs
```diff
@@ -6481,7 +6496,7 @@ string <var>namespace</var> (default null):</p>
 
 <div algorithm>
 <p>To <dfn export id=concept-element-attributes-set-value>set an attribute value</dfn> given an
-<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string <var>value</var>,
+<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string or <a>TrustedType</a> <var>value</var>,
```
annevk, line 6499: This needs to be `{{TrustedType}}` I think to get the markup correct. `<a>` doesn't add code.

dom.bs
```diff
@@ -6490,12 +6505,27 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
  <a lt="get an attribute by namespace and local name">getting an attribute</a> given
  <var>namespace</var>, <var>localName</var>, and <var>element</var>.
 
- <li>If <var>attribute</var> is null, create an <a>attribute</a> whose <a for=Attr>namespace</a> is
- <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
- <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is <var>value</var>, and
- <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>, then
- <a lt="append an attribute">append</a> this <a>attribute</a> to <var>element</var>, and then
- return.
+ <li><p>If <var>attribute</var> is null, then:
```
annevk, line 6508: Nit: indentation is wrong here and below. README has some instructions. `<p>` should be on a newline and indented one, for instance.

dom.bs
```diff
@@ -6760,7 +6790,7 @@ method steps are:
 
  <li><p>If <var>attribute</var> is null, create an <a>attribute</a> whose
  <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
- <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
+ stringified <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
```
annevk, line 6793: As noted elsewhere, this doesn't work as-is.

dom.bs
```diff
@@ -6490,12 +6505,27 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
...
+ <var>value</var>. [[!TRUSTED-TYPES]]
+
+ <li><p><a for=list>Append</a> <var>attribute</var> to <var>element</var>'s
+ <a for=Element>attribute list</a>.
+
+ <li><p>Set <var>attribute</var>'s <a for=Attr>element</a> to <var>element</var>.
+
+ <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>element</var>, null, and
+ <var>attribute</var>'s <a for=Attr>value</a>.
+ <li><p>Return.
```
annevk, line 6527: I don't understand why the changes here (which are for setAttributeNS() and prolly some other callers) are much more elaborate than they are for setAttribute().

Why can't "append an attribute" handle this in the way it already does per the changes above?

annevk (commented):
I see it's still a draft, but I thought I'd post some initial feedback.

dom.bs
koto: Fixed the formatting, thanks for pointers. You're correct re: MutationObserver.

@mbrodesser-Igalia, we need to add a MutationObserver test (https://jsfiddle.net/014ze36t/2/) to WPT.

dom.bs
koto: I actually rewrote all this from scratch, adding an algorithm that validates with TT and sets the attr value (a string at that point). There's no need for stringification in DOM now, PTAL.

dom.bs
koto: Append might be too late, at that point Attr value (a string) is set, so the checks bubbled up to the calling algorithms. Check the current version, I think I missed one callsite in the one you reviewed. Now it should be OK, with the intentional omission of `clone`.

dom.bs
```diff
@@ -6408,6 +6411,18 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
+<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn>
+<var>value</var> for an <a>attribute</a> <var>attribute</var>, with <a for=/>element</a>
```
annevk, line 6415: Nit: type of value should be specified.

dom.bs
```diff
@@ -6408,6 +6411,18 @@ steps:
...
+<var>element</var>, run these steps:
```
annevk, line 6416: Modern style is to leave out "run these steps" when the sentence starts with "To" (and maybe in some other cases).
```suggestion
<var>element</var>:
```

dom.bs
```diff
@@ -6758,10 +6790,20 @@ method steps are:
  and null otherwise.
  <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
 
- <li><p>If <var>attribute</var> is null, create an <a>attribute</a> whose
- <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
- <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
- then <a lt="append an attribute">append</a> this <a>attribute</a> to <a>this</a>, and then return.
+ <li><p>If <var>attribute</var> is null, then:
```
annevk, line 6793: Wrapping is still wrong here. The `<li>` has multiple children so the `<p>` needs to be newlined and indented. The `<ol>` below doesn't seem indented enough either. Remember that it's a child of the `<li>`.

dom.bs
```diff
@@ -6497,6 +6515,20 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
...
+  <a for=Attr>local name</a> is <var>localName</var> and <a for=Node>node document</a> is
+  <var>element</var>'s <a for=Node>node document</a>.
+
+  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with
+  <var>element</var>.
+
+  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>.
+
+  <li><p>Return.
+ </ol>
```
annevk, line 6530: This seems wrong. This should be behind the "If attribute is null" check above, presumably?

annevk: However, I'm also not sure why this cannot be part of "append an attribute" which would give us some deduplication. If you do the validation before the attribute is appended to the element, it seems fine? What am I missing?

annevk: In particular because in these cases we end up creating a new attribute and so if something throws, it would simply be as if that attribute was not created, right?

And it seems that the current algorithms miss the `toggleAttribute()` endpoint.

dom.bs
annevk: https://github.com/whatwg/dom/pull/1247#discussion_r1466326599

mbrodesser-Igalia:
Is it possible to get the preview linked at https://github.com/whatwg/dom/pull/1247#issue-2096528268 fixed? Would simplify reading the diff.

dom.bs
```diff
@@ -6481,21 +6499,30 @@ string <var>namespace</var> (default null):</p>
...
+ <li>
+  <p>If <var>attribute</var> is null, then:
+  <ol>
+    <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>namespace</a> is
+    <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
+    <a for=Attr>local name</a> is <var>localName</var> and <a for=Node>node document</a> is
+    <var>element</var>'s <a for=Node>node document</a>.
+
+    <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with
+    <var>element</var>.
```
smaug----, line 6520: This should probably clarify that validation may throw an exception. What should happen in that case?

dom.bs
```diff
@@ -6481,21 +6499,30 @@ string <var>namespace</var> (default null):</p>
...
+
+    <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>.
```
smaug----, line 6522: This is problematic. Validation may have run scripts, and scripts may have already added another attribute with same name. That can't be allowed.

dom.bs
```diff
@@ -6758,10 +6785,21 @@ method steps are:
...
+
+  <ol>
+   <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>local name</a> is
+   <var>qualifiedName</var> and <a for=Node>node document</a> is <a>this</a>'s
+   <a for=Node>node document</a>.
+
+   <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>,
+   with <a>this</a>.
+
+   <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <a>this</a>.
```
smaug----, line 6799: This has also the problem that since validation may run scripts, the attribute list may now already have attribute with the same name. And validation may throw an exception.


(but yeah, in general these checks do need to happen very early when we're about to set an attribute)

mbrodesser-Igalia:
@koto can the PR preview please be fixed? The changes themselves are sufficiently complicated, so let's make our life easier by simplifying reviewing.

dom.bs
```diff
@@ -6408,6 +6411,18 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
+<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn>
+{{TrustedType}} or a string <var>value</var> for an <a>attribute</a> <var>attribute</var>, with
+<a for=/>element</a> <var>element</var>:
+
+<ol>
+ <li><p>Let <var>validValue</var> be the result of calling
+ <a>Get Trusted Types-compliant attribute value</a> for <var>attribute</var>, with
```
lukewarlow, line 6420:
```suggestion
 [$get Trusted Types-compliant attribute value$] for <var>attribute</var>, with
```
This fixes the bikeshed error (it's exported with a lowercase g)

lukewarlow:
I might be missing something but this doesn't seem to account for event handler attributes? I know there's separate handling for them in TT but wont this change block them being set? Or do they go down a different codepath?

koto:
> @koto can the PR preview please be fixed? The changes themselves are sufficiently complicated, so let's make our life easier by simplifying reviewing.

This is done, I'll proceed with resolving all the comments.

lukewarlow:
@koto what remaining steps need to be taken on this?

dom.bs
lukewarlow: Having tested this in Chrome it results in some rather weird behaviour (the DOM apis for getting the attribute return the value of the attribute set inside the TT validation steps (from default policy in this case). But the HTML markup in dev tools AND the actual action that occurs is the one set via the original call to setAttribute.

I'm thinking we could check after the validate call if the attribute exists, and if it does we can either throw or skip out to the process that happens when the attribute already exists?

dom.bs
lukewarlow: If we edit append an attribute we might end up with duplicated calls to validate.

For example in the screenshot below you'll see that we need to validate before step 5, but then step 6 which appends would also call validate?

Screenshot 2024-04-10 at 16 12 14

lukewarlow: Having said that we can reorder these steps so that step 4 is inside of 5's if so it becomes 5.1 and 5.2 is to replace?

And then we can append attr and handle validation inside of append?

(I've also just spotted it refers to a newAttr which doesn't exist in this algorithm I'll make sure to fix that)

lukewarlow: Is `toggleAttribute()` actually an issue? Don't you still need to go through one of the other algorithms to actually set a value?

dom.bs
lukewarlow: Addressed in https://github.com/whatwg/dom/pull/1268 by rechecking the attribute state and throwing an exception if the default policy has done something funky.

dom.bs
lukewarlow: Addressed in https://github.com/whatwg/dom/pull/1268

PR #1258, opened 2024-02-22
Integration with trusted types inside of node conversion

lukewarlow:
Fixes https://github.com/w3c/trusted-types/issues/438

   * Chromium (already implemented)
   * Gecko
   * https://wpt.fyi/trusted-types already includes a number of tests for these changes.
   * Chromium: Shipped
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=270435

***
<a href="https://whatpr.org/dom/1258.html" title="Last updated on Apr 18, 2024, 2:14 PM UTC (6b615a7)">Preview</a> | <a href="https://whatpr.org/dom/1258/77cd75e...6b615a7.html" title="Last updated on Apr 18, 2024, 2:14 PM UTC (6b615a7)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3048,14 +3048,14 @@ standards that want to define APIs shared between <a for=/>documents</a> and
  <li><p>Let <var>isScriptElement</var> be true if <var>parent</var> is an {{HTMLScriptElement}}
  <a for=/>node</a>; otherwise false.
 
- <li><p>Let <var>newNodes</var> be the empty list.
+ <li><p>Let <var>newNodes</var> be « ».
 
  <li>
   <p><a for=list>For each</a> <var>value</var> of <var>nodes</var>:
 
   <ol>
    <li>
-    <p>If <var>value</var> is a {{Text}} <a for=/>node</a>, and <var>isScriptElement</var>
+    <p>If <var>value</var> is a {{Text}} <a for=/>node</a> and <var>isScriptElement</var>
     is true:
 
     <ol>
@@ -3073,14 +3073,14 @@ standards that want to define APIs shared between <a for=/>documents</a> and
 
     <ol>
      <li><p>Let <var>newValue</var> be a new {{Text}} <a for=/>node</a> whose
-     <a for=CharacterData>data</a> is <var>value</var>'s <a for=TrustedScript>\[[Data]]</a>
-     internal slot value and <a for=Node>node document</a> is <var>document</var>.
+     <a for=CharacterData>data</a> is <var>value</var>'s <a for=TrustedScript>data</a>
+     and <a for=Node>node document</a> is <var>document</var>.
 
      <li><p><a for=list>Append</a> <var>newValue</var> to <var>newNodes</var>.
     </ol>
 
    <li><p>Otherwise, <a for=list>append</a> <var>value</var> to <var>newNodes</var>.
- </ol>
+  </ol>
 
  <li><p>If <var>newNodes</var> contains one <a for=/>node</a>, then set <var>node</var> to
  <var>newNodes</var>[0].
--- a/dom.bs
+++ b/dom.bs
@@ -3165,7 +3165,7 @@ the number of <a for=tree>children</a> of <a>this</a> that are <a for=/>elements
 
 <ol>
  <li><p>Let <var>node</var> be the result of <a>converting nodes into a node</a> given <a>this</a>,
- <var>nodes</var> and <a>this</a>'s <a for=Node>node document</a>.
+ <var>nodes</var>, and <a>this</a>'s <a for=Node>node document</a>.
 
  <li><p><a>Pre-insert</a> <var>node</var> into <a>this</a> before <a>this</a>'s
  <a for=tree>first child</a>.
@@ -3175,7 +3175,7 @@ the number of <a for=tree>children</a> of <a>this</a> that are <a for=/>elements
 
 <ol>
  <li><p>Let <var>node</var> be the result of <a>converting nodes into a node</a> given <a>this</a>,
- <var>nodes</var> and <a>this</a>'s <a for=Node>node document</a>.
+ <var>nodes</var>, and <a>this</a>'s <a for=Node>node document</a>.
 
  <li><p><a>Append</a> <var>node</var> to <a>this</a>.
 </ol>
@@ -3185,7 +3185,7 @@ are:
 
 <ol>
  <li><p>Let <var>node</var> be the result of <a>converting nodes into a node</a> given <a>this</a>,
- <var>nodes</var> and <a>this</a>'s <a for=Node>node document</a>.
+ <var>nodes</var>, and <a>this</a>'s <a for=Node>node document</a>.
 
  <li><p><a>Ensure pre-insertion validity</a> of <var>node</var> into <a>this</a> before
  null.
@@ -3290,7 +3290,7 @@ CharacterData includes ChildNode;
  <a for=tree>sibling</a> not in <var>nodes</var>; otherwise null.
 
  <li><p>Let <var>node</var> be the result of <a>converting nodes into a node</a>, given
- <var>parent</var>, <var>nodes</var> and <a>this</a>'s <a for=Node>node document</a>.
+ <var>parent</var>, <var>nodes</var>, and <a>this</a>'s <a for=Node>node document</a>.
 
  <li><p>If <var>viablePreviousSibling</var> is null, then set it to <var>parent</var>'s
  <a for=tree>first child</a>; otherwise to <var>viablePreviousSibling</var>'s
@@ -3311,7 +3311,7 @@ CharacterData includes ChildNode;
  <a for=tree>sibling</a> not in <var>nodes</var>; otherwise null.
 
  <li><p>Let <var>node</var> be the result of <a>converting nodes into a node</a>, given
- <var>parent</var>, <var>nodes</var> and <a>this</a>'s <a for=Node>node document</a>.
+ <var>parent</var>, <var>nodes</var>, and <a>this</a>'s <a for=Node>node document</a>.
 
  <li><p><a>Pre-insert</a> <var>node</var> into <var>parent</var> before
  <var>viableNextSibling</var>.
@@ -3328,7 +3328,7 @@ CharacterData includes ChildNode;
  <a for=tree>sibling</a> not in <var>nodes</var>; otherwise null.
 
  <li><p>Let <var>node</var> be the result of <a>converting nodes into a node</a>, given
- <var>parent</var>, <var>nodes</var> and <a>this</a>'s <a for=Node>node document</a>.
+ <var>parent</var>, <var>nodes</var>, and <a>this</a>'s <a for=Node>node document</a>.
 
  <li>
   <p>If <a>this</a>'s <a for=tree>parent</a> is <var>parent</var>, <a>replace</a> <a>this</a> with
--- a/dom.bs
+++ b/dom.bs
@@ -37,8 +37,6 @@ urlPrefix: https://w3c.github.io/hr-time/#; spec: HR-TIME
  type:typedef; urlPrefix: dom-; text: DOMHighResTimeStamp
  type:dfn; text: current high resolution time; url: dfn-current-high-resolution-time
  type:dfn; text: relative high resolution coarse time; url: dfn-relative-high-resolution-coarse-time
-urlPrefix: https://w3c.github.io/trusted-types/dist/spec/#; spec: TRUSTED-TYPES
- type:dfn; text: [[Data]]; for: TrustedScript; url: trustedscript-data
 </pre>
 
 <pre class=link-defaults>
--- a/dom.bs
+++ b/dom.bs
@@ -3056,31 +3056,26 @@ standards that want to define APIs shared between <a for=/>documents</a> and
     <p>If <var>value</var> is a <a for=/>node</a>, then:
 
     <ol>
-     <li><p>If <var>isScriptElement</var> is false, then <a for=iteration>continue</a>.
-
-     <li><p>If <var>value</var> is not a {{Text}} <a for=/>node</a>, then <a for=iteration>continue</a>.
-
-     <li><p>Let <var>text</var> be the result of calling <a abstract-op>Get Trusted Type compliant string</a>,
-     with {{TrustedScript}}, <var>document</var>'s <a>relevant global object</a>,
-     <var>value</var>'s <a for=CharacterData>data</a>, "<code>HTMLScriptElement text</code>",
-     "<code>script</code>".
-
-     <li><p>Set <var>value</var>'s <a for=CharacterData>data</a> to <var>text</var>.
+     <li><p>If <var>value</var> is a {{Text}} <a for=/>node</a>, and <var>isScriptElement</var>
+     is false, then set <var>value</var>'s <a for=CharacterData>data</a> to <var>text</var> the
+     result of calling <a abstract-op>Get Trusted Type compliant string</a>, with {{TrustedScript}},
+     <var>document</var>'s <a>relevant global object</a>, <var>value</var>'s
+     <a for=CharacterData>data</a>, "<code>HTMLScriptElement text</code>", and "<code>script</code>".
     </ol>
 
-    <li>
-     <p>If <var>value</var> is a {{TrustedScript}}, then:
+   <li>
+    <p>If <var>value</var> is a {{TrustedScript}}, then:
 
-     <ol>
-      <li><p>Let <var>text</var> be <var>value</var>'s <a for=TrustedScript>\[[Data]]</a>
-      internal slot value.
+    <ol>
+     <li><p>Let <var>text</var> be <var>value</var>'s <a for=TrustedScript>\[[Data]]</a>
+     internal slot value.
 
-      <li><p>Let <var>newValue</var> be a new {{Text}} <a for=/>node</a> whose
-      <a for=CharacterData>data</a> is <var>text</var> and <a for=Node>node document</a> is
-      <var>document</var>.
+     <li><p>Let <var>newValue</var> be a new {{Text}} <a for=/>node</a> whose
+     <a for=CharacterData>data</a> is <var>text</var> and <a for=Node>node document</a> is
+     <var>document</var>.
 
-      <li><p>Replace <var>value</var> with <var>newValue</var>.
-     </ol>
+     <li><p>Replace <var>value</var> with <var>newValue</var>.
+    </ol>
  </ol>
 
  <li><p>If <var>nodes</var> contains one <a for=/>node</a>, then set <var>node</var> to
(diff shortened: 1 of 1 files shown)


lukewarlow:
Bolt on to #1247

mbrodesser-Igalia:
@lukewarlow you mentioned somewhere you'd appreciate a review in order to simplify the change. Since it's not primarily about correctness, how helpful would an imminent review be?

dom.bs
```diff
@@ -3057,9 +3096,9 @@ interface mixin ParentNode {
   readonly attribute Element? lastElementChild;
   readonly attribute unsigned long childElementCount;
 
-  [CEReactions, Unscopable] undefined prepend((Node or DOMString)... nodes);
-  [CEReactions, Unscopable] undefined append((Node or DOMString)... nodes);
-  [CEReactions, Unscopable] undefined replaceChildren((Node or DOMString)... nodes);
+  [CEReactions, Unscopable] undefined prepend((Node or DOMString or TrustedScript)... nodes);
+  [CEReactions, Unscopable] undefined append((Node or DOMString or TrustedScript)... nodes);
+  [CEReactions, Unscopable] undefined replaceChildren((Node or DOMString or TrustedScript)... nodes);
```
mbrodesser-Igalia, line 3101: For the record: these methods should be annotated with a note for web-developers that passed text nodes and strings are internally passed to trusted types' default policy, if needed.

dom.bs
```diff
@@ -9101,7 +9140,7 @@ and Opera Next 12.00 alpha all do. -->
 
 <p>The <dfn method for=Range><code>intersectsNode(<var>node</var>)</code></dfn> method steps are:
 <!-- Supported by Chrome 17 dev and Opera Next 12.00 alpha, but not IE9 or
-Firefox 12.0a1. -->
+Firefox 12.0a<li>-->
```
lukewarlow, line 9143: revert

dom.bs
```diff
@@ -3031,15 +3031,54 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
-<var>nodes</var> and <var>document</var>, run these steps:
+<var>parent</var>, <var>nodes</var> and <var>document</var>, run these steps:
 
 <ol>
+ <li><p>Let <var>isScriptElement</var> be <var>parent</var> is a {{HTMLScriptElement}}.
+ <li><p>For each <var>value</var> in <var>nodes</var>:
+  <ol>
+   <li>If <var>value</var> is a <a for=/>node</a>, then:
+    <ol>
+     <li>If <var>isScriptElement</var> is false, skip to the next <var>value</var>.
```
annevk, line 3042: then continue* (see also Infra)

dom.bs
```diff
@@ -3031,15 +3031,54 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+      <li>If <var>value</var>'s {{nodeType}} is not {{TEXT_NODE}}, skip to the next <var>value</var>.
```
annevk, line 3043: Please compare the class instead as we do elsewhere in this standard. That would also encompass CDATASection.

annevk (commented):
This looks complicated. Any reason we cannot convert to Text nodes first and then potentially throw for Text nodes if they pose some kind of danger in combination with a `script` element?

lukewarlow:
> This looks complicated. Any reason we cannot convert to Text nodes first and then potentially throw for Text nodes if they pose some kind of danger in combination with a script element?

Ah yeah that's a good idea, can process text nodes and raw strings at the same time them and then can just handle the trusted script after the other processing (because it needs to hold onto the fact it's safe). I knew there was probably a nicer way.

dom.bs
```diff
@@ -3040,6 +3040,34 @@ standards that want to define APIs shared between <a for=/>documents</a> and
  <a for=CharacterData>data</a> is the string and <a for=Node>node document</a> is
  <var>document</var>.
 
+ <li><p>Let <var>isScriptElement</var> be <var>parent</var> is a {{HTMLScriptElement}}.
+
+ <li><p><a for=set>For each</a> <var>value</var> in <var>nodes</var>:
```
annevk, line 3045: The formatting here is wrong. Please ensure it's consistent with the rest of the document. The README should have instructions or a pointer to them.

dom.bs
```diff
@@ -3040,6 +3040,34 @@ standards that want to define APIs shared between <a for=/>documents</a> and
  <a for=CharacterData>data</a> is the string and <a for=Node>node document</a> is
  <var>document</var>.
 
+ <li><p>Let <var>isScriptElement</var> be <var>parent</var> is a {{HTMLScriptElement}}.
```
annevk, line 3043: This reads wonky. I think it needs to be something like:

> Let x be true if y; otherwise false.

Also, it's "an" HTMLScriptElement and that should probably read "an HTMLScriptElement node" or perhaps "a script element". (I forgot if the DOM Standard has precedent for this.)

dom.bs
annevk: You resolved this, but it doesn't appear to be resolved.

dom.bs
```diff
@@ -3040,6 +3040,34 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+   <li>If <var>value</var> is a <a for=/>node</a>, then:
+    <ol>
+     <li>If <var>isScriptElement</var> is false, skip to the next <var>value</var>.
+     <li>If <var>value</var> is not a {{Text}} <a for=/>node</a>, then <a for=iteration>continue</a>.
+     <li>Let <var>text</var> be the result of executing the [$Get Trusted Type compliant string$] algorithm, with the following arguments:
+        * {{TrustedScript}} as <span>expectedType</span>,
+        * <var>document</var>'s [=relevant global object=] as <span>global</span>,
+        * <var>value</var>'s <a for=CharacterData>data</a> as <span>input</span>,
+        * `HTMLScriptElement text` as <span>sink</span>,
+        * `script` as <span>sinkGroup</span>.
```
annevk, line 3056: This invocation style is not described by Infra. It needs adjusting. We also don't use `<span>` in DOM in this way.

dom.bs
```diff
@@ -3040,6 +3040,34 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+
+        <p>If the algorithm threw an error, rethrow the error.
```
annevk, line 3058: This is implied, see Infra.

dom.bs
```diff
@@ -3040,6 +3040,34 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+     <li>Set <var>value</var>'s <a for=CharacterData>data</a> to <var>text</var>.
+    </ol>
+
+    <li>If <var>value</var> is a {{TrustedScript}}, then:
+     <ol>
+      <li>Let <var>text</var> be <var>value</var>'s Data internal slot value.
```
annevk, line 3064: "Data internal slot value" needs some cross-referencing.

dom.bs
```diff
@@ -3040,6 +3040,34 @@ standards that want to define APIs shared between <a for=/>documents</a> and
  <a for=CharacterData>data</a> is the string and <a for=Node>node document</a> is
  <var>document</var>.
 
+ <li><p>Let <var>isScriptElement</var> be <var>parent</var> is a {{HTMLScriptElement}}.
+
+ <li><p><a for=set>For each</a> <var>value</var> in <var>nodes</var>:
+  <ol>
+   <li>If <var>value</var> is a <a for=/>node</a>, then:
```
annevk, line 3047: All `<li>` should have a nested `<p>` for any text.

dom.bs
lukewarlow: Apologies idk what happened there, I addressed it below this line must have missed this. Will do that now

dom.bs
lukewarlow: This should hopefully be sorted. Not entirely sure what the "correct" way is.

dom.bs
lukewarlow: Hopefully this is fixed.

lukewarlow:
@annevk is there anything left that's needed for this PR? Wanna make sure I'm not holding anything up.

dom.bs
```diff
@@ -37,6 +37,8 @@ urlPrefix: https://w3c.github.io/hr-time/#; spec: HR-TIME
  type:typedef; urlPrefix: dom-; text: DOMHighResTimeStamp
  type:dfn; text: current high resolution time; url: dfn-current-high-resolution-time
  type:dfn; text: relative high resolution coarse time; url: dfn-relative-high-resolution-coarse-time
+urlPrefix: https://w3c.github.io/trusted-types/dist/spec/#; spec: TRUSTED-TYPES
+ type:dfn; text: [[Data]]; for: TrustedScript; url: trustedscript-data
```
annevk, line 41: This shouldn't be needed? Also why is this using square brackets?

dom.bs
```diff
@@ -3040,6 +3044,44 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+
+    <ol>
+     <li><p>If <var>isScriptElement</var> is false, then <a for=iteration>continue</a>.
+
+     <li><p>If <var>value</var> is not a {{Text}} <a for=/>node</a>, then <a for=iteration>continue</a>.
+
+     <li><p>Let <var>text</var> be the result of calling <a abstract-op>Get Trusted Type compliant string</a>,
+     with {{TrustedScript}}, <var>document</var>'s <a>relevant global object</a>,
+     <var>value</var>'s <a for=CharacterData>data</a>, "<code>HTMLScriptElement text</code>",
+     "<code>script</code>".
```
annevk, line 3065: Missing "and" in the argument list.

dom.bs
```diff
@@ -3040,6 +3044,44 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+
+     <li><p>Set <var>value</var>'s <a for=CharacterData>data</a> to <var>text</var>.
```
annevk, line 3067: Why define _text_ if we use it once? I would inline this step into the one above.

dom.bs
```diff
@@ -3040,6 +3044,44 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+    </ol>
+
+    <li>
```
annevk, line 3070: This cannot have the same indentation level as `</ol>`.

dom.bs
annevk: In fact, would this work:

> If _value_ is a Text node and isScriptElement is true, then set _value_'s data to ...

?

dom.bs
```diff
@@ -3040,6 +3044,44 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+
+     <ol>
+      <li><p>Let <var>text</var> be <var>value</var>'s <a for=TrustedScript>\[[Data]]</a>
+      internal slot value.
+
+      <li><p>Let <var>newValue</var> be a new {{Text}} <a for=/>node</a> whose
+      <a for=CharacterData>data</a> is <var>text</var> and <a for=Node>node document</a> is
+      <var>document</var>.
+
+      <li><p>Replace <var>value</var> with <var>newValue</var>.
```
annevk, line 3081: "Replace" is not defined here. I guess you are attempting to modify _nodes_ directly? That makes me worried about https://github.com/whatwg/infra/pull/451.

dom.bs
lukewarlow: We use the internal slot so it's referenced here. Also it uses brackets because that's how the slot is defined?

annevk: Let me clarify:

1. It should be exported by Trusted Types and since Trusted Types is indexed(?), we shouldn't have to list it explicitly.
2. We generally don't define internal slots this way in web platform specifications and definitely not for any nodes as far as I know.

lukewarlow: https://html.spec.whatwg.org/#shared-internal-slot:-crossoriginpropertydescriptormap plus many other examples exist in the HTML spec of internal slots being defined with the [[slot]] syntax.

lukewarlow: You're correct I don't need to define the anchor I think that must have been from before I exported the definition in TT left over erroneously.

dom.bs
lukewarlow: Yeah, and tbf both WebKit and Chromium actually make a new list object that they append to, so perhaps that'd be better to spec like that? MIght also handle the fact that the trusted types check can do script execution a bit better too? (Though looping through a list and can potentially be modified might be a seperate issue?)

annevk: That would sidestep this issue so that's probably better, yeah.

dom.bs
```diff
@@ -3040,6 +3042,39 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+ <li>
+  <p><a for=set>For each</a> <var>value</var> of <var>nodes</var>:
+
+  <ol>
+   <li>
+    <p>If <var>value</var> is a <a for=/>node</a>, then:
+
+    <ol>
+     <li><p>If <var>value</var> is a {{Text}} <a for=/>node</a>, and <var>isScriptElement</var>
+     is false, then set <var>value</var>'s <a for=CharacterData>data</a> to <var>text</var> the
```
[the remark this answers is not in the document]
lukewarlow: 🤦🏻 yep it should be true. Sorry.

dom.bs
```diff
@@ -3040,6 +3042,39 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+     <var>document</var>'s <a>relevant global object</a>, <var>value</var>'s
+     <a for=CharacterData>data</a>, "<code>HTMLScriptElement text</code>", and "<code>script</code>".
+    </ol>
+
+   <li>
+    <p>If <var>value</var> is a {{TrustedScript}}, then:
+
+    <ol>
+     <li><p>Let <var>text</var> be <var>value</var>'s <a for=TrustedScript>\[[Data]]</a>
+     internal slot value.
```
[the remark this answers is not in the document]
lukewarlow: What's the best syntax for this? Just `<a for=TrustedScript>Data</a> internal slot value.`? I'll make a PR to fix this in the TT spec and then update it here.

dom.bs
```diff
@@ -3040,6 +3042,39 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
 
+ <li><p>Let <var>isScriptElement</var> be true if <var>parent</var> is an {{HTMLScriptElement}}
+ <a for=/>node</a>; otherwise false.
+
+ <li>
+  <p><a for=set>For each</a> <var>value</var> of <var>nodes</var>:
+
+  <ol>
+   <li>
+    <p>If <var>value</var> is a <a for=/>node</a>, then:
```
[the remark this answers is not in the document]
lukewarlow: Yeah you're right, I think I wrote it this way originally to make the steps clearer but now the inner bit is 1 step it might aswell just be pulled out.

dom.bs
lukewarlow: Should this be a set or a list? Based on what infra says I *think* maybe a set? But then the nodes parameter comes from a JS union so is that maybe a list?

dom.bs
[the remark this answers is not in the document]
annevk: > Let text be value's data.

And in TT:

> A TrustedScript object has an associated [type] data. It's initially [a value].

dom.bs
```diff
@@ -3040,12 +3042,49 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+    <ol>
+     <li><p>Let <var>newValue</var> be a new {{Text}} <a for=/>node</a> whose
+     <a for=CharacterData>data</a> is <var>value</var>'s <a for=TrustedScript>\[[Data]]</a>
+     internal slot value and <a for=Node>node document</a> is <var>document</var>.
+
+     <li><p><a for=list>Append</a> <var>newValue</var> to <var>newNodes</var>.
+    </ol>
+
+   <li><p>Otherwise, <a for=list>append</a> <var>value</var> to <var>newNodes</var>.
+ </ol>
```
annevk, line 3080: Indentation is wrong here. I think this `</ol>` just needs an additional space?

dom.bs
```diff
@@ -3040,12 +3042,49 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+ <a for=/>node</a>; otherwise false.
+
+ <li><p>Let <var>newNodes</var> be the empty list.
+
+ <li>
+  <p><a for=list>For each</a> <var>value</var> of <var>nodes</var>:
+
+  <ol>
+   <li>
+    <p>If <var>value</var> is a {{Text}} <a for=/>node</a>, and <var>isScriptElement</var>
```
annevk, line 3055:
```suggestion
    <p>If <var>value</var> is a {{Text}} <a for=/>node</a> and <var>isScriptElement</var>
```

dom.bs
```diff
@@ -3040,12 +3042,49 @@ standards that want to define APIs shared between <a for=/>documents</a> and
  <a for=CharacterData>data</a> is the string and <a for=Node>node document</a> is
  <var>document</var>.
 
- <li><p>If <var>nodes</var> contains one <a for=/>node</a>, then set <var>node</var> to
- <var>nodes</var>[0].
+ <li><p>Let <var>isScriptElement</var> be true if <var>parent</var> is an {{HTMLScriptElement}}
+ <a for=/>node</a>; otherwise false.
+
+ <li><p>Let <var>newNodes</var> be the empty list.
```
annevk, line 3048:
```suggestion
 <li><p>Let <var>newNodes</var> be « ».
```

dom.bs
lukewarlow: Ah yeah apologies, I keep getting confused because the lack of closing li and p tags (which I hope is correct for the Dom spec?)

annevk: That is correct.

dom.bs
```diff
@@ -3040,12 +3042,49 @@ standards that want to define APIs shared between <a for=/>documents</a> and
...
+
+     <li><p><a for=list>Append</a> <var>newValue</var> to <var>newNodes</var>.
+    </ol>
+
+   <li>
+    <p>Otherwise, if <var>value</var> is a {{TrustedScript}}:
+
+    <ol>
+     <li><p>Let <var>newValue</var> be a new {{Text}} <a for=/>node</a> whose
+     <a for=CharacterData>data</a> is <var>value</var>'s <a for=TrustedScript>data</a>
```
lukewarlow, line 3073: `<a for=TrustedScript>data</a>` definitely should be linkifying but it's not. Possibly bikeshed hasn't reindexed it yet?

annevk: Likely, yeah.

lukewarlow: Seems to work locally now after a bikeshed update.

dom.bs
annevk: I don't think it matters here and using list syntax is nice.

annevk (approved):

lukewarlow:
LGTM @annevk thanks for your help getting this across the line.

fred-wang:
For the record, this was reverted in https://github.com/whatwg/dom/commit/809bfa24535b1d55d761d8add353e1c3fde05801

Issue #808, opened 2019-12-09
Side effects due to tree insertion or removal (script, iframe)


annevk:
Elements that trigger a script-executing side effect due to insertion or removal need to do so using some kind of queue if we ever want to remove mutation events fully (see #305). In particular, I think it's important that tree mutation succeeds and that any script is executed afterwards (but before custom element reactions, see https://github.com/web-platform-tests/wpt/pull/20660).

I'd like to use this thread for discussion, since currently it's spread among various places (https://github.com/whatwg/dom/issues/575, https://github.com/whatwg/dom/pull/732, https://github.com/whatwg/html/pull/4354, https://github.com/whatwg/html/issues/4611, https://github.com/whatwg/html/issues/4965).

Chrome and Firefox have this model for `script` insertion. (Though there are differences, if two scripts are inserted and the first removes the second, Firefox will execute the second, Chrome will not. They also react differently to changes to children of the script element, with Chrome only caring about insertions there.)

Chrome has this model for `iframe` insertion as well, but not removal. Firefox only has it for removal (insertion cannot trigger script in Firefox as it does not have a synchronous `load` event; note that this means that Chrome and Firefox create nested browsing contexts at different times).

Firefox has this model for `style` insertion, but since `style` cannot trigger script I don't think we should follow that.

@nox @bzbarsky @tkent-google @rniwa thoughts? Do we all share the overall goal of executing script at a "safe" point? How would you like to see this broken down further?

bzbarsky:
@emilio has thought about this some as well, I think.

emilio:
Gecko has a slightly different (and somewhat weird) model for `<iframe>`, but it has _some_ sort of model.

Gecko loads iframes [from the outermost call `Document::EndUpdate`](https://searchfox.org/mozilla-central/rev/d24696b5abaf9fb75f7985952eab50d5f4ed52ac/dom/base/Document.cpp#6913), which gets called after ever DOM mutation and such. If it's safe to run script there, then it does it synchronously, otherwise it adds to the same queue `<style>` uses (via `AddScriptRunner`).

annevk:
As noted by @nox `<meta http-equiv=default-style>` is also deferred in Chrome (and Safari it seems). Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1602476 on Firefox not handling it dynamically.

annevk:
@tkent-google @rniwa @cdumez ping.

tkent-google:
[...]

PR #1261, opened 2024-02-28
Introduce DOM post-connection steps

domfarolino:
This PR introduces a node's set of _post-connection steps_. For any given [#concept-node-insert](https://dom.spec.whatwg.org/#concept-node-insert) operation, these steps run for each inserted node synchronously _after_ all DOM insertions are complete. This PR is meant to supersede the similar https://github.com/whatwg/dom/pull/732.

The goal here is to separate the following:
 1. Script-observable but not-script-executing insertion side effects
    - This includes synchronously applying styles to the document, etc...
 2. Script-executing insertion side effects
    - This includes creating an iframe's document and [synchronously firing its load event](https://html.spec.whatwg.org/C#the-iframe-element:iframe-load-event-steps)

For any given call to [#concept-node-insert](https://dom.spec.whatwg.org/#concept-node-insert), the above model allows us to keep all of (1) running synchronously after each node's insertion (as part of its [_insertion steps_](https://dom.spec.whatwg.org/#concept-node-insert-ext)), while pushing all script-executing (or DOM tree-modifying or frame tree-modifying etc.) side effects to the new set of post-connection steps, which run synchronously during insertion, but _after all_ nodes finish their insertion.

This essentially makes insertions "atomic" from the perspective of script, since script will not run until a given batch of DOM insertions are complete. This two-stage approach aligns the spec with a model most similar to [Blink](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/node.h;l=782-812;drc=f4a00cc248dd2dc8ec8759fb51620d47b5114090) & Gecko's implementation, and fixes https://github.com/whatwg/dom/issues/808. This PR also helps out with https://github.com/whatwg/html/issues/1127 and https://github.com/whatwg/dom/issues/575 (per https://github.com/whatwg/dom/pull/732#issuecomment-467403090).

To accomplish, this we audited all insertion side effects on the web platform in https://docs.google.com/document/d/1Fu_pgSBziVIBG4MLjorpfkLTpPD6-XI3dTVrx4CZoqY/edit#heading=h.q06t2gg4vpw, and catalogued whether they have script-observable side-effects, whether they invoke script, whether we have tests for them, and how each implementation handles them. This gave us a list of present "insertion steps" that should move to the "post-connection steps", because they invoke script and therefore prevent insertions from being "atomic". This PR is powerless without counterpart changes to HTML, which will actually _use_ the post-connection steps for all current insertion steps that invoke script or modify the frame tree. The follow-up HTML work is tracked, at the time of writing this:

 - https://github.com/whatwg/html/pull/10188
 - https://github.com/whatwg/html/issues/10241
 - `<another incoming PR will be made for iframes>`

   * ✅ Chrome (mostly implements this model; fails at least one ["nested `script` ordering" test](https://wpt.live/dom/nodes/insertion-removing-steps/Node-appendChild-script-in-script.tentative.html))
   * ✅ Firefox (mostly implements this model; fails a few tests that rely on iframe `load` event being fired synchronously, since I guess it fires this event asynchronously..)
   * 🟠 Safari: @annevk would you be able to provide a WebKit signal here or should I file a standards position issue?
   * [WPT results can be seen here](https://wpt.fyi/results/dom/nodes/insertion-removing-steps?label=master&label=experimental&aligned)
   * https://github.com/web-platform-tests/wpt/pull/44308
   * https://github.com/web-platform-tests/wpt/pull/44658
   * https://github.com/web-platform-tests/wpt/pull/44774
   * https://github.com/web-platform-tests/wpt/pull/44906
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed: N/A given https://github.com/whatwg/dom/pull/1261#issuecomment-2037596814.

----

Two things need more discussion:
 - **Script child element mutation**:
    - https://github.com/whatwg/dom/pull/732#pullrequestreview-328249015 and https://github.com/whatwg/html/pull/4354#issuecomment-476038918 discuss Chromium _& WebKit's_ model of not executing scripts when their children are modified except for insertion.
        - **Regarding insertion only**: All browsers are aligned on executing a script (that hasn't ["already started"](https://html.spec.whatwg.org/C#script-processing-model:already-started-3)) when a single child node is inserted. For insertion multiple child nodes atomically, see [this WPT](https://wpt.fyi/results/dom/nodes/insertion-removing-steps/Node-appendChild-text-in-script.tentative.html) which Chrome & Firefox pass, and WebKit barely misses due to the post-connection steps[^1]. (Related: _all_ browsers are consistent (see [this test case](https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=12435)) about _not_ executing newly-inserted text nodes in a script that has "already started")
        - **Regarding child removal/change**: As a follow-up to https://github.com/whatwg/dom/pull/732#pullrequestreview-328249015 and https://github.com/whatwg/html/pull/4354#issuecomment-476038918, I've investigated this a little more and found that actually _all_ browsers follow the Chromium behavior here. See [**this WPT**](https://wpt.fyi/results/dom/nodes/insertion-removing-steps/script-does-not-run-on-child-removal.window.html?label=pr_head&max-count=1&pr=45085) (from https://github.com/web-platform-tests/wpt/pull/45085), where all browsers never trigger the execution of a script when a child node is removed. To accomplish this with the spec, we should probably just add an argument to the child changed steps, but that can be done in a follow-up PR to https://github.com/whatwg/html/pull/10188, since it is _slightly_ orthogonal
    - The only oddity I can find with scripts in Chromium's model is that we don't follow [this rule exactly](https://html.spec.whatwg.org/multipage/scripting.html#prepare-the-script-element:~:text=after%20any%20script%20elements%20inserted%20at%20that%20time), which mandates that when a script (that hasn't already started yet) gets a child `<script>` inserted into it, it must execute _after_ that child script does. Chromium does not do that; the child mutation of the outer script (that inserts the inner `<script>`) triggers the execution of the outer script in its post-connection steps, and then after, the execution of the inner script during _its_ post-connection steps. WebKit is aligned with Chrome on this and Firefox is the odd one out. Maybe we're OK aligning with Chrome/Firefox in this case?
 - **Dealing with "removal" steps**
    - In general, removal is distinct from insertion, and I think has no bearing on the decisions we make in this PR and its corresponding HTML one. However, through some offline discussion, we didn't want to commit to this insertion model without having done some investigation into the removal side of things. I've started this in [this doc](https://docs.google.com/document/d/1Fu_pgSBziVIBG4MLjorpfkLTpPD6-XI3dTVrx4CZoqY/edit#heading=h.n7us367xam1v), but see [**the summary here**](https://gist.github.com/domfarolino/94167cef258c7de96b933ac9b825cfd9). Basically there are only **two** places during Node removal where browsers ever synchronously execute script: (1) `pagehide` which all browsers do on same-origin iframes, and (2) `blur` events (only Chrome does this).
    - Chrome is committed to experimenting with *not* firing in either of these scenarios. Along with the deprecation of mutation events, this should mark the end of all synchronous script execution on Node removal

[^1]: The only reason WebKit fails this is because due to the lack of post-connection steps, the script executes after the first text node is appended, so by the time the second text node gets attached, the script is [already started](https://html.spec.whatwg.org/C#script-processing-model:already-started-3)

***
<a href="https://whatpr.org/dom/1261.html" title="Last updated on Jun 3, 2024, 10:26 AM UTC (fb21a6d)">Preview</a> | <a href="https://whatpr.org/dom/1261/68df785...fb21a6d.html" title="Last updated on Jun 3, 2024, 10:26 AM UTC (fb21a6d)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2671,25 +2671,25 @@ do these things asynchronously, however.
  <p>The script in the above example prints <code class=lang-javascript>'rgb(255, 0, 0)'</code>
  because the following happen in order:
 
-  <ol>
-   <li>
-    <p>The <a for=/>insert</a> algorithm runs, which will insert the <{script}> and <code><a element
-    spec=HTML>style</a></code> elements in order.
+ <ol>
+  <li>
+   <p>The <a for=/>insert</a> algorithm runs, which will insert the <{script}> and <code><a element
+   spec=HTML>style</a></code> elements in order.
 
-    <ol>
-     <li>The HTML Standard's <a>insertion steps</a> run for the <{script}> element; they do nothing.
-     [[!HTML]]
+   <ol>
+    <li>The HTML Standard's <a>insertion steps</a> run for the <{script}> element; they do nothing.
+    [[!HTML]]
 
-     <li>The HTML Standard's <a>insertion steps</a> run for the <code><a element
-     spec=HTML>style</a></code> element; they immediately apply its style rules to the document.
-     [[!HTML]]
+    <li>The HTML Standard's <a>insertion steps</a> run for the <code><a element
+    spec=HTML>style</a></code> element; they immediately apply its style rules to the document.
+    [[!HTML]]
 
-     <li>The HTML Standard's <a>post-insertion steps</a> run for the <{script}> element; they run
-     the script, which immediately observes the style rules that were applied in the above step.
-     [[!HTML]]
-    </ol>
-   </li>
-  </ol>
+    <li>The HTML Standard's <a>post-insertion steps</a> run for the <{script}> element; they run
+    the script, which immediately observes the style rules that were applied in the above step.
+    [[!HTML]]
+   </ol>
+  </li>
+ </ol>
 </div>
 
 <!-- See https://github.com/whatwg/dom/issues/34#issuecomment-125571750 for why we might need to
--- a/dom.bs
+++ b/dom.bs
@@ -2706,7 +2706,7 @@ that <var ignore>connectedNode</var> <a>participates</a> in, create <a for=/>bro
 or otherwise execute JavaScript. These steps allow a batch of <a>nodes</a> to be
 <a for=/>inserted</a> <em>atomically</em> with respect to script, with all major side effects
 occurring <em>after</em> the batch insertions into the <a>node tree</a> is complete. This ensures
-that all pending <a>node tree</a> insertions completely finish before more insertions can occur. 
+that all pending <a>node tree</a> insertions completely finish before more insertions can occur.
 
 <p><a lt="other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-children-changed-ext>children changed steps</dfn> for all or some
--- a/dom.bs
+++ b/dom.bs
@@ -2652,8 +2652,8 @@ events</a>, or otherwise execute JavaScript. These steps may [=queue a global ta
 do these things asynchronously, however.
 
 <div class=example id=example-foo-what-do-i-put-here>
- <p>While the <a>insertion steps</a> cannot execute JavaScript (among other things), it is expected
- that they will indeed have script-observable consequences. Consider the below example:
+ <p>While the <a>insertion steps</a> cannot execute JavaScript (among other things), they will
+ indeed have script-observable consequences. Consider the below example:
 
  <pre class=lang-javascript><code>
  const h1 = document.querySelector('h1');
@@ -2706,9 +2706,8 @@ for=/>nodes</a> to perform any insertion-related operations that modify the <a>n
 <var ignore>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>, or
 otherwise execute JavaScript. These steps allow a batch of <a>nodes</a> to be <a for=/>inserted</a>
 <i>atomically</i> with respect to script, with all major side effects occurring <i>after</i> the
-batch insertions into the <a>node tree</a> is complete, but before <a lt="notify mutation
-observers"><code>MutationObserver</code>s are notified</a>. This ensures that all pending <a>node
-tree</a> insertions completely finish before more insertions can occurr.
+batch insertions into the <a>node tree</a> is complete. This ensures that all pending <a>node
+tree</a> insertions completely finish before more insertions can occur.
 
 
 <p><a lt="other applicable specifications">Specifications</a> may define
@@ -2784,25 +2783,26 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
     <p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
     <var>node</var>, in <a>shadow-including tree order</a>:
 
-    <ol>
-     <li><p>Run the <a>insertion steps</a> with <var>inclusiveDescendant</var>.
+     <ol>
+      <li><p>Run the <a>insertion steps</a> with <var>inclusiveDescendant</var>.
 
-     <li>
-      <p>If <var>inclusiveDescendant</var> is <a>connected</a>, then:
+      <li>
+       <p>If <var>inclusiveDescendant</var> is <a>connected</a>, then:
 
-      <ol>
-       <li><p>If <var>inclusiveDescendant</var> is <a for=Element>custom</a>, then
-       <a>enqueue a custom element callback reaction</a> with <var>inclusiveDescendant</var>,
-       callback name "<code>connectedCallback</code>", and an empty argument list.
+       <ol>
+        <li><p>If <var>inclusiveDescendant</var> is <a for=Element>custom</a>, then
+        <a>enqueue a custom element callback reaction</a> with <var>inclusiveDescendant</var>,
+        callback name "<code>connectedCallback</code>", and an empty argument list.
 
-       <li>
-        <p>Otherwise, <a lt="try to upgrade an element">try to upgrade</a>
-        <var>inclusiveDescendant</var>.
+        <li>
+         <p>Otherwise, <a lt="try to upgrade an element">try to upgrade</a>
+         <var>inclusiveDescendant</var>.
 
-        <p class=note>If this successfully upgrades <var>inclusiveDescendant</var>, its
-        <code>connectedCallback</code> will be enqueued automatically during the
-        <a>upgrade an element</a> algorithm.
-      </ol>
+         <p class=note>If this successfully upgrades <var>inclusiveDescendant</var>, its
+         <code>connectedCallback</code> will be enqueued automatically during the
+         <a>upgrade an element</a> algorithm.
+       </ol>
+      </p>
      </li>
     </ol>
    </li>
--- a/dom.bs
+++ b/dom.bs
@@ -2648,7 +2648,8 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
 algorithm is passed <var>insertedNode</var>, as indicated in the <a for=/>insert</a> algorithm
 below. These steps must not modify the <a>node tree</a> that <var>insertedNode</var>
 <a>participates</a> in, create <a for=/>browsing contexts</a>, <a lt="fire an event">fire
-events</a>, or otherwise execute JavaScript.
+events</a>, or otherwise execute JavaScript. These steps may [=queue a global task|queue tasks=] to
+do these things asynchronously, however.
 
 <div class=example id=example-foo-what-do-i-put-here>
  <p>While the <a>insertion steps</a> cannot execute JavaScript (among other things), it is expected
@@ -2703,9 +2704,10 @@ for=/>insert</a> algorithm below.
 <p class=note>The purpose of the <a>post-insertion steps</a> is to provide an opportunity for <a
 for=/>nodes</a> to perform any insertion-related operations that modify the <a>node tree</a> that
 <var ignore>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>, or
-otherwise execute JavaScript. These steps allow all nodes that will be inserted by a given <a
-for=/>insert</a> operation to by inserted <i>atomically</i>, with all major side effects occurring
-<i>after</i> <a>node tree</a> insertion is complete.
+otherwise execute JavaScript. These steps allow a batch of <a>nodes</a> to be <a for=/>inserted</a>
+<i>atomically</i> with respect to script, with all major side effects occurring <i>after</i> the
+batch insertions into the <a>node tree</a> is complete, but before <a lt="notify mutation
+observers"><code>MutationObserver</code>s are notified</a>.
 
 
 <p><a lt="other applicable specifications">Specifications</a> may define
--- a/dom.bs
+++ b/dom.bs
@@ -2647,9 +2647,9 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
 <dfn export id=concept-node-insert-ext>insertion steps</dfn> for all or some <a for=/>nodes</a>. The
 algorithm is passed <var>insertedNode</var>, as indicated in the <a for=/>insert</a> algorithm
 below. These steps must not modify the <a>node tree</a> that <var>insertedNode</var>
-<a>participates</a> in, create <a for=/>browsing contexts</a>, <a lt="fire an event">fire
-events</a>, or otherwise execute JavaScript. These steps may [=queue a global task|queue tasks=] to
-do these things asynchronously, however.
+<a>participates</a> in, create <a for=/>browsing contexts</a>,
+<a lt="fire an event">fire events</a>, or otherwise execute JavaScript. These steps may
+[=queue a global task|queue tasks=] to do these things asynchronously, however.
 
 <div class=example id=example-foo-what-do-i-put-here>
  <p>While the <a>insertion steps</a> cannot execute JavaScript (among other things), they will
@@ -2695,18 +2695,18 @@ do these things asynchronously, however.
      adjust this further based on the requirements of the script element. There might be other ways
      to define that though as Olli suggests, so leaving that out for now. -->
 
-<p><a lt="Other applicable specifications">Specifications</a> may also define <dfn export
-id=concept-node-post-connection-ext>post-connection steps</dfn> for all or some <a for=/>nodes</a>.
-The algorithm is passed <var ignore>connectedNode</var>, as indicated in the <a for=/>insert</a>
-algorithm below.
+<p><a lt="Other applicable specifications">Specifications</a> may also define
+<dfn export id=concept-node-post-connection-ext>post-connection steps</dfn> for all or some
+<a for=/>nodes</a>. The algorithm is passed <var ignore>connectedNode</var>, as indicated in the
+<a for=/>insert</a> algorithm below.
 
-<p class=note>The purpose of the <a>post-connection steps</a> is to provide an opportunity for <a
-for=/>nodes</a> to perform any connection-related operations that modify the <a>node tree</a> that
-<var ignore>connectedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>, or
-otherwise execute JavaScript. These steps allow a batch of <a>nodes</a> to be <a for=/>inserted</a>
-<em>atomically</em> with respect to script, with all major side effects occurring <em>after</em> the
-batch insertions into the <a>node tree</a> is complete. This ensures that all pending <a>node
-tree</a> insertions completely finish before more insertions can occur.
+<p class=note>The purpose of the <a>post-connection steps</a> is to provide an opportunity for
+<a for=/>nodes</a> to perform any connection-related operations that modify the <a>node tree</a>
+that <var ignore>connectedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>,
+or otherwise execute JavaScript. These steps allow a batch of <a>nodes</a> to be
+<a for=/>inserted</a> <em>atomically</em> with respect to script, with all major side effects
+occurring <em>after</em> the batch insertions into the <a>node tree</a> is complete. This ensures
+that all pending <a>node tree</a> insertions completely finish before more insertions can occur. 
 
 <p><a lt="other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-children-changed-ext>children changed steps</dfn> for all or some
@@ -2813,11 +2813,12 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
  <li>
   <p>Let <var>staticNodeList</var> be a <a for=/>list</a> of <a for=/>nodes</a>, initially « ».</p>
 
-  <p class="note">We collect all <a for=/>nodes</a> <em>before</em> calling the <a>post-connection
-  steps</a> on any one of them, instead of calling the <a>post-connection steps</a> <em>while</em>
-  we're traversing the <a>node tree</a>. This is because the <a>post-connection steps</a> can modify
-  the tree's structure, making live traversal unsafe, possibly leading to the <a>post-connection
-  steps</a> being called multiple times on the same <a>node</a>.</p>
+  <p class="note">We collect all <a for=/>nodes</a> <em>before</em> calling the
+  <a>post-connection steps</a> on any one of them, instead of calling the
+  <a>post-connection steps</a> <em>while</em> we're traversing the <a>node tree</a>. This is because
+  the <a>post-connection steps</a> can modify the tree's structure, making live traversal unsafe,
+  possibly leading to the <a>post-connection steps</a> being called multiple times on the same
+  <a>node</a>.</p>
  </li>
 
  <li>
--- a/dom.bs
+++ b/dom.bs
@@ -2822,7 +2822,7 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
  </li>
 
  <li>
-  <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
+  <p>For each <var>node</var> of <var>nodes</var>, in <a>tree order</a>:
 
   <ol>
    <li><p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
--- a/dom.bs
+++ b/dom.bs
@@ -2812,14 +2812,28 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
  <li><p>Run the <a>children changed steps</a> for <var>parent</var>.
 
  <li>
+  <p>Let <var>staticNodeList</var> be a <a for=/>list</a> of <a for=/>nodes</a>, initially
+  empty.</p>
+
+  <p class="note">We collect all <a for=/>nodes</a> <i>before</i> calling the <a>post-insertion
+  steps</a> on any one of them, instead of calling the <a>post-insertion steps</a> <i>while</i>
+  we're traversing the <a>node tree</a>. This is because the <a>post-insertion steps</a> can
+  modify the tree's structure, making live traversal unsafe, possibly leading to the
+  <a>post-insertion steps</a> being called multiple times on the same <a>node</a>.</p>
+ </li>
+
+ <li>
   <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
 
   <ol>
    <li><p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
-   <var>node</var>, in <a>shadow-including tree order</a>, run the <a>post-insertion steps</a> with
-   <var>inclusiveDescendant</var>.
+   <var>node</var>, in <a>shadow-including tree order</a>, <a for=list>append</a>
+   <var>inclusiveDescendant</var> to <var>staticNodeList</var>.
   </ol>
  </li>
+
+ <li><p><a for=list>For each</a> <var>node</var> in <var>staticNodeList</var>, run the
+ <a>post-insertion steps</a> with <var>node</var>.
 </ol>
 
 
--- a/dom.bs
+++ b/dom.bs
@@ -2662,30 +2662,29 @@ do these things asynchronously, however.
  const script = fragment.appendChild(document.createElement('script'));
  const style = fragment.appendChild(document.createElement('style'));
 
- script.innerText= 'console.log(getComputedStyle(h1).color)'; // Prints 'rgb(255, 0, 0)'
+ script.innerText= 'console.log(getComputedStyle(h1).color)'; // Logs 'rgb(255, 0, 0)'
  style.innerText = 'h1 {color: rgb(255, 0, 0);}';
 
  document.body.append(fragment);
  </code></pre>
 
- <p>The script in the above example prints <code class=lang-javascript>'rgb(255, 0, 0)'</code>
- because the following happen in order:
+ <p>The script in the above example logs <code class=lang-javascript>'rgb(255, 0, 0)'</code> because
+ the following happen in order:
 
  <ol>
-  <li>
-   <p>The <a for=/>insert</a> algorithm runs, which will insert the <{script}> and <code><a element
-   spec=HTML>style</a></code> elements in order.
+  <li><p>The <a for=/>insert</a> algorithm runs, which will insert the <{script}> and <code>
+  <a element spec=HTML>style</a></code> elements in order.
 
    <ol>
-    <li>The HTML Standard's <a>insertion steps</a> run for the <{script}> element; they do nothing.
-    [[!HTML]]
+    <li><p>The HTML Standard's <a>insertion steps</a> run for the <{script}> element; they do
+    nothing. [[!HTML]]
 
-    <li>The HTML Standard's <a>insertion steps</a> run for the <code><a element
-    spec=HTML>style</a></code> element; they immediately apply its style rules to the document.
-    [[!HTML]]
+    <li><p>The HTML Standard's <a>insertion steps</a> run for the <code>
+    <a element spec=HTML>style</a></code> element; they immediately apply its style rules to the
+    document. [[!HTML]]
 
-    <li>The HTML Standard's <a>post-connection steps</a> run for the <{script}> element; they run the
-    script, which immediately observes the style rules that were applied in the above step.
+    <li><p>The HTML Standard's <a>post-connection steps</a> run for the <{script}> element; they run
+    the script, which immediately observes the style rules that were applied in the above step.
     [[!HTML]]
    </ol>
   </li>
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -2645,12 +2645,69 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
 
 <p><a lt="Other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-insert-ext>insertion steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed <var ignore>insertedNode</var>, as indicated in the <a for=/>insert</a>
-algorithm below.
+algorithm is passed <var>insertedNode</var>, as indicated in the <a for=/>insert</a> algorithm
+below. These steps must not modify the <a>node tree</a> that <var>insertedNode</var>
+<a>participates</a> in, create <a for=/>browsing contexts</a>, <a lt="fire an event">fire
+events</a>, or otherwise execute JavaScript.
```
noamr, line 2651: Can we clarify to say that these steps can't do these things synchronously, but they can queue a task/microtask that does them.

dom.bs
```diff
@@ -2645,12 +2645,69 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+id=concept-node-post-insert-ext>post-insertion steps</dfn> for all or some <a for=/>nodes</a>. The
+algorithm is passed <var ignore>insertedNode</var>, as indicated in the <a
+for=/>insert</a> algorithm below.
+
+<p class=note>The purpose of the <a>post-insertion steps</a> is to provide an opportunity for <a
+for=/>nodes</a> to perform any insertion-related operations that modify the <a>node tree</a> that
+<var ignore>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>, or
+otherwise execute JavaScript. These steps allow all nodes that will be inserted by a given <a
+for=/>insert</a> operation to by inserted <i>atomically</i>, with all major side effects occurring
+<i>after</i> <a>node tree</a> insertion is complete.
```
noamr, line 2708: But before mutation observers I guess?

dom.bs
```diff
@@ -2645,12 +2645,69 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
 
 <p><a lt="Other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-insert-ext>insertion steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed <var ignore>insertedNode</var>, as indicated in the <a for=/>insert</a>
-algorithm below.
+algorithm is passed <var>insertedNode</var>, as indicated in the <a for=/>insert</a> algorithm
```
noamr, line 2648: Would be good to add a bit of a motivation for the separation into two sets of steps

dom.bs
domfarolino: Yes. I've clarified this in the latest commit.

dom.bs
domfarolino: Done!

dom.bs
domfarolino: Hmm you mean describe the motivation for having both `#concept-node-insert` and `#concept-node-insert-ext`, or having both `#concept-node-insert-ext` and `#concept-node-post-insert-ext`? Assuming you mean the latter, do you mean just clarifying that we don't want script to run in between individual node insertions, because it can modify the DOM tree and complicate the ongoing insertion?

noamr: Yes, the latter.  A motivation for the whole thing of running some of the insertion steps after completing the whole mutation

domfarolino: I've elaborated on this a bit in the post-insertion steps note. Let me know if you had something else in mind.

domfarolino:
PTAL @annevk! I'll prepare an HTML PR and commit message for this, as well as file the appropriate browser bugs. But I'd love your thoughts on the meat of this so far, since it should be good to once those particulars are done.

dom.bs
```diff
@@ -2752,6 +2812,16 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+ <li>
+  <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
+
+  <ol>
+   <li><p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
+   <var>node</var>, in <a>shadow-including tree order</a>, run the <a>post-insertion steps</a> with
+   <var>inclusiveDescendant</var>.
+  </ol>
+ </li>
```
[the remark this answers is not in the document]
domfarolino: > Is this the correct order relative to mutation observers? This should be observable using `takeRecords()`.

Good catch. Yes, Chrome and Firefox both can observe the _queued_ record this way. I am not actually sure the WPTs assert this, so let me modify a few to include this check.

> I also still worry about this not being entirely consistent with the removal case.

Hmm, could you elaborate on this concern? I don't know what is inconsistent about it. Assuming we have no opportunities to execute script during removal, we shouldn't need to move anything to an equivalent post-removal steps. And post-removal steps seem risky anyways, since we'd be operating exclusively on detached elements. The true equivalent for removal would be to let each node run their own script-observable non-script-executing side-effects one-by-one before any of them get removed, but I feel like as long as none of them are executing script, then that shouldn't actually be observable? And we certainly don't want to run script for each element just immediately before its removal, since that's the worst part of mutation events.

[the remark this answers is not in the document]
noamr: Currently only chromium fires `blur` events when removing, and we're checking if that's safe to fix. The only other existing script side effect is `pagehide` in same-process iframes. 

Both are not in the spec - if just reading the standard and not implementation, there are 0 script side effects to removing nodes (unless we missed a spot).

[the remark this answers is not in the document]
domfarolino: Yeah, please see [this part of the OP](https://github.com/whatwg/dom/pull/1261#:~:text=Dealing%20with%20%22removal%22%20steps) of this PR that discusses the current state of removal. I promised we wouldn't proceed with insertion without taking a reasonable look at the removal story!

domfarolino:
FWIW, even though much of the prior discussion about this and #808 makes it seem like WebKit doesn't implement the post-insertion steps that this PR introduces, the only reason that Blink implements this correctly is because it inherited the code from WebKit!

See https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/ContainerNodeAlgorithms.cpp#L54 in WebKit, where we track which elements have post-insertion steps to call later (and the equivalent Blink code), _as well as_ the and the equivalent code in Blink https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/container_node.cc;l=1095-1096;drc=54b5f6dcc0e51751aec186771b4012ef663b538b.

Also see https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/ContainerNode.cpp#L302-L307 in WebKit, where we _run_ each eligible element's post-insertion steps after the children change steps, and the equivalent code in Blink https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/container_node.cc;l=1071-1077;drc=9ab453c8e7d336f430c50488e8638ca802b69d80.

For script execution, WebKit even [requests that post-insertion steps are used](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/ScriptElement.h#L99-L104), I just don't really understand why WebKit fails tests like this then: https://wpt.fyi/results/dom/nodes/insertion-removing-steps/Node-appendChild-script-and-button-from-div.tentative.html?label=experimental&label=master&aligned. Someone like @rniwa probably knows though.

dom.bs
```diff
@@ -2645,12 +2645,72 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+algorithm is passed <var ignore>insertedNode</var>, as indicated in the <a
+for=/>insert</a> algorithm below.
+
+<p class=note>The purpose of the <a>post-insertion steps</a> is to provide an opportunity for <a
+for=/>nodes</a> to perform any insertion-related operations that modify the <a>node tree</a> that
+<var ignore>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>, or
+otherwise execute JavaScript. These steps allow a batch of <a>nodes</a> to be <a for=/>inserted</a>
+<i>atomically</i> with respect to script, with all major side effects occurring <i>after</i> the
+batch insertions into the <a>node tree</a> is complete, but before <a lt="notify mutation
+observers"><code>MutationObserver</code>s are notified</a>. This ensures that all pending <a>node
```
smaug----, line 2710: Why is this talking about notifying MutationObservers? That happens at microtask checkpoint which is way later.

domfarolino: I added this in https://github.com/whatwg/dom/pull/1261/commits/164eae9b4bc2ea64a8a7426d6ac82e55c9904639, in response to https://github.com/whatwg/dom/pull/1261#discussion_r1520204437 which mentioned MutationObservers. Idk, for this non-normative text it seemed reasonable to just drive home the point that this all happens technically after insertion, but still very much synchronously, and therefore earlier than the next-soonest reaction timing thing, which is MutationObserver firing. If you feel strongly, I'm happy to remove if Anne agrees.

dom.bs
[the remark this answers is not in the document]
smaug----: Gecko triggers script execution after all the insertions have been done. 
But there should be just one MutationRecord ,  that should be covered by DOM spec when dealing with adding DocumentFragments and such. MutationRecord has all the inserted and removed nodes in separate lists.

dom.bs
smaug----: CE Reaction or what? That isn't when MutationObservers get notified.

dom.bs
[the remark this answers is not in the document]
smaug----: Hmm, this model isn't exactly what Gecko has. We use script runners to run stuff ,which might run scripts, at safe time. I wonder if that model would be easier here.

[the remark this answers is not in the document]
domfarolino: I don't know what a script runner is. But if you have a concrete test case that shows how Gecko deviates from the spec here, that would be great. Two points though:

 - This PR doesn't actually make changes to script execution. That would be https://github.com/whatwg/html/pull/10188, which I'd be grateful for your review on as well!
 - The combined behavior change of both this PR and https://github.com/whatwg/html/pull/10188 is captured by [these tests](https://wpt.fyi/results/dom/nodes/insertion-removing-steps?label=experimental&label=master&aligned). We'd love for you to weigh on the test expectations over in https://github.com/whatwg/html/pull/10188, to see if any of the significant Firefox failures are a result of inappropriate test expectations.

dom.bs
domfarolino: Ah, well we also have tests for [custom element reactions](https://wpt.fyi/results/dom/nodes/insertion-removing-steps/Node-appendChild-script-and-custom-from-fragment.tentative.html?label=experimental&label=master&aligned). In any case, my point is that the non-normative text here is just to drive home the point that the post-insertion steps run synchronously after insertion and before anything. Do you think I should remove it?

rniwa:
> Also see https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/ContainerNode.cpp#L302-L307 in WebKit, where we _run_ each eligible element's post-insertion steps after the children change steps, and the equivalent code in Blink https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/container_node.cc;l=1071-1077;drc=9ab453c8e7d336f430c50488e8638ca802b69d80.
> 
> For script execution, WebKit even [requests that post-insertion steps are used](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/ScriptElement.h#L99-L104), I just don't really understand why WebKit fails tests like this then: https://wpt.fyi/results/dom/nodes/insertion-removing-steps/Node-appendChild-script-and-button-from-div.tentative.html?label=experimental&label=master&aligned. Someone like @rniwa probably knows though.

It's because in WebKit, updating of element.form also happens in the post insertion steps.

domfarolino:
> It's because in WebKit, updating of element.form also happens in the post insertion steps.

Good to know. Do you think WebKit would be willing to change that? It's not a decision that needs to block _this PR_, since this PR just gives us the primitives available for HTML to do whatever we decide. But so far it seems 2/3 impls agree on that specific case.

rniwa:
> > It's because in WebKit, updating of element.form also happens in the post insertion steps.
> 
> Good to know. Do you think WebKit would be willing to change that? It's not a decision that needs to block _this PR_, since this PR just gives us the primitives available for HTML to do whatever we decide. But so far it seems 2/3 impls agree on that specific case.

Unlikely. There are some design & performance constraints about this.

domfarolino:
> Unlikely. There are some design & performance constraints about this.

I'd appreciate some elaboration on this, just to understand why other browsers seem to be OK not doing this. However, it is a low priority discussion relative to this PR, since this PR doesn't change anything WRT that scenario specifically. But if you'd be willing to review this PR that would be great.

domfarolino:
Gentle ping for reviews (and for a response to https://github.com/whatwg/dom/pull/1261#issuecomment-2022663904).

dom.bs
smaug----: I think it is very confusing to talk about MutationObservers (microtasks) here.

dom.bs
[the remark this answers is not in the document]
smaug----: Script runners are like the nano tasks which were discussed years ago. Basically a thing which runs something which may trigger script execution, and runs at safe time. Very similar to post-insertion steps. But the key difference is that those aren't bound strictly to any node.  And you may create script runners (and often do) during insertion steps.

dom.bs
domfarolino: @annevk what do you think? I think mentioning the fact that post-insertion steps run synchronously before any microtasks run is useful, but if you agree with Olli I'm happy to nix this bit.

dom.bs
```diff
@@ -2645,12 +2645,72 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+ script.innerText= 'console.log(getComputedStyle(h1).color)'; // Prints 'rgb(255, 0, 0)'
+ style.innerText = 'h1 {color: rgb(255, 0, 0);}';
+
+ document.body.append(fragment);
+ </code></pre>
+
+ <p>The script in the above example prints <code class=lang-javascript>'rgb(255, 0, 0)'</code>
+ because the following happen in order:
+
+  <ol>
```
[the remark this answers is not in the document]
domfarolino: I think you're right. I've indented the `<old>` and added the end `<p>` after it.

dom.bs
```diff
@@ -2645,12 +2645,71 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+for=/>insert</a> algorithm below.
+
+<p class=note>The purpose of the <a>post-insertion steps</a> is to provide an opportunity for <a
+for=/>nodes</a> to perform any insertion-related operations that modify the <a>node tree</a> that
+<var ignore>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>, or
+otherwise execute JavaScript. These steps allow a batch of <a>nodes</a> to be <a for=/>inserted</a>
+<i>atomically</i> with respect to script, with all major side effects occurring <i>after</i> the
+batch insertions into the <a>node tree</a> is complete. This ensures that all pending <a>node
+tree</a> insertions completely finish before more insertions can occur.
+
```
annevk, line 2711: Why a double newline here?

dom.bs
```diff
@@ -2724,25 +2783,26 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
     <p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
     <var>node</var>, in <a>shadow-including tree order</a>:
 
-    <ol>
-     <li><p>Run the <a>insertion steps</a> with <var>inclusiveDescendant</var>.
+     <ol>
```
annevk, line 2786: This is also wrong.

dom.bs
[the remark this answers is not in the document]
domfarolino: Oof, I'm sorry! I think I fixed the indenting now. Although I feel like we're back where we were at the first comment here? Not sure.

dom.bs
domfarolino: Removed!

dom.bs
domfarolino: Because this snippit is outdated I can't quite tell if it is still relevant, but I think it is fixed. Can you take a look?

dom.bs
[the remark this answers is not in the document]
annevk: No, this one is still wrong. I think you ended up changing a different list and you did revert that one now.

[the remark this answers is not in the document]
domfarolino: Should be good now. Apologies for the churn.

annevk:
I don't think we need implementation bugs for this PR given its largely non-normative nature. Can you update the proposed commit message so it's clearer where the corresponding HTML work happens? I found these two so far:

- https://github.com/whatwg/html/pull/10188
- https://github.com/whatwg/html/issues/10241

Given that we have a concrete HTML PR I think we can merge this, but I'd like to give folks a bit more time to comment. So let's say I merge this next week Wednesday, barring any other feedback.

dom.bs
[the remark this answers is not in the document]
smaug----: Script runners are just like tasks which are run when there is first chance to run scripts.  The nice thing there is that there can be many of them, and they don't need to be bound to any node. It is unclear to me how "post insertion steps" would work if there are multiple things which need to be done at that time.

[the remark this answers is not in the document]
noamr: Given that we don't often add new post-insertion steps/script runners, and the fact that there aren't that many of those today, is there an observable difference between that and post-insertion steps? To me it doesn't seem material whether we spec it this way or that way, it's a way to express that iframes & scripts never run scriptable side-effects in the middle of the DOM update.

annevk:
Let's see if we can briefly discuss this tomorrow. I do see the appeal of Gecko's script runner approach as it avoids the ordering issue. On the other hand, thus far the ordering issue is largely theoretical given the limited number of things we expect will need to be changed.

dom.bs
```diff
@@ -2752,6 +2810,16 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+ <li>
+  <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
+
+  <ol>
+   <li><p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
+   <var>node</var>, in <a>shadow-including tree order</a>, run the <a>post-insertion steps</a> with
+   <var>inclusiveDescendant</var>.
+  </ol>
+ </li>
```
smaug----, line 2822: Adding the same comment here what I added to #whatwg matrix channel
" I don't know how the insertion step model works in case of mutations. Say, you have an element and it has 3 child elements. That subtree root element is inserted to document tree. Now child 1 has insertion steps which run some script which removes the child and adds it to be the last of the child nodes. So child list is now 2, 3, 1. Do the insertion steps get run on 1 again, because it is now after 2 and 3?"

smaug----: Unless I'm missing something, the pr doesn't match what Blink does.  Isn't there a queue here https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/container_node.cc;l=430;bpv=0;bpt=0 ?
So the difference between ScriptRunner and that approach is just that ScriptRunner queues operations and NodeVector is for queuing nodes which have post insertion stuff to run.

domfarolino: The queue mechanism here and the one in Blink have no observable differences. In Blink, all elements with post-insertion steps end up in that queue of nodes to get their steps run. In this PR, all elements with post-insertion steps also run after insertion. Should be no behavioral difference.

I can’t speak to script runners at all since there is no proposal for them, and I don’t really understand them.

smaug----: There is difference. Blink can deal with DOM mutations if they happen while going through post insertion steps. The PR does not. Blink implementation has that explicit queue of nodes to go through, the PR just vaguely talks about descendants (and doesn't explain what happens if DOM is mutated).

smaug----: And ScriptRunner is like Blink's implementation, but instead of queuing nodes, one would queue operations (kind of like "nano tasks").

domfarolino: OK thanks for the discussion @smaug----. Anne and I (among others) also discussed this a bit at the last WHATNOT meeting. The latest commit adds a static node list that we populate with all of the inserted nodes. Then we call the post-insertion steps on each of them in that list (if they are connected), so that we don't perform any live tree traversal operations in between post-insertion steps. PTAL!

dom.bs
```diff
@@ -2752,6 +2810,30 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+  <ol>
+   <li><p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
+   <var>node</var>, in <a>shadow-including tree order</a>, <a for=list>append</a>
+   <var>inclusiveDescendant</var> to <var>staticNodeList</var>.
+  </ol>
+ </li>
+
+ <li><p><a for=list>For each</a> <var>node</var> in <var>staticNodeList</var>, if <var>node</var> is
+ <a>connected</a>, then run the <a>post-insertion steps</a> with <var>node</var>.
```
smaug----, line 2836: Why do we have the requirement for the node to be connected? (It wasn't clear to me either when looking at blink's source code, but I didn't look at the blame for that part).

domfarolino: The idea is that post-insertion steps should not run on elements that aren't connected to the DOM. For those elements, we shouldn't be performing actions that modify the frame tree or load resources or anything else the post-insertion steps are responsible for outside of the insertion steps.

We could do this implicitly / vaguely which is kind of how HTML does it now, by specifying something like ["becomes connected"](https://html.spec.whatwg.org/multipage/infrastructure.html#becomes-connected) steps (but for post-insertion steps), which supposedly run "during" the "insertion steps" (without any ordering specified) with a "is connected" precondition built in. Or we could do it slightly more explicitly by not having an "is connected" gate specified in DOM, and relying on each element's post-insertion steps to have a precondition that early-returns without a DOM connection. That doesn't seem better though. So I think the condition in this PR is the most straightforward way of achieving this. We have tests too that ensure e.g., scripts don't run when disconnected after insertion (by other post-insertion steps, in the trickiest of cases, for example).

smaug----: I'd expect the algorithm be generic, and then each post-insertion step implementation to check if the node is connected (if needed). Nothing vague there.

If we go with the approach the pr has, at least the algorithm name should hint that it isn't about generic insertion step, but it is explicitly about inserting to document. Post-connected steps?
(insertion steps algorithm has nothing to do with nodes being connected or not)

smaug----, line 2836: (just wondering what should happen, and what browsers do if  some earlier post-connected step moves a node, which hasn't get change to run post-connected steps to be elsewhere in the document - so its post-connected steps get run twice. That might be a case to handle separately in each case - but if that is needed, it might be easy to just add is-connected check to the element specific algorithm)

domfarolino: Great scenario. So to make the case more concrete, you're envisioning something like:

1. Insert **(a)** script and **(b)** iframe at the same time.
    - Both are now in the DOM, but we have not yet made it to the step where we iterate over them and run their post-insertion steps (i.e., the spec step you're commenting on above)
1. **(a)** Run the script's post-insertion steps: this moves the iframe around in the DOM. This both:
    - Re-inserts the iframe into the DOM
    - Runs the iframe's post-insertion steps
1. **(b)** Re-run the iframe's post-insertion steps, even though it exists elsewhere in the DOM now.

Chromium currently runs the post-insertion steps the second time. I think WebKit does the same from looking at their code, but I can't quite re-produce.

Chromium has code that handles this [explicitly for iframes here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/html_frame_element_base.cc;l=192-195;drc=df5e5c413d71fd4d130e837dc5bd7bd0c7ff3cf5) so that we don't try and re-create an iframe's initial document after it has been inserted. And for scripts, I think you would just rely on the [already started boolean](https://html.spec.whatwg.org/C#already-started) to not re-execute a script that has already executed due to insertion ([Chromium code here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script/script_loader.cc;l=571-573;drc=b74be264a68fdb02cb216ad60ea610562d2f800d)).

I lean towards DOM being OK running the post-insertion steps multiple times, and then having per-element checks do the right thing if they don't want to run multiple times. The reason I lean towards this is because I can construct a case where we actually rely on the _second_ invocation of the post-insertion steps doing stuff.

Imagine a case where you insert two scripts at the same time, i.e., `document.body.append(script1, script2)`. `script2` has an invalid `type` and therefore is not executable as-is. `script1` does two things:
1. Moves `script2` around in the DOM (thus invoking the second script's post-insertion steps for the first time)
2. Removes `script2`'s invalid `type` attribute, making it finally valid (but not immediately executing it)

At least in Chromium, when `script2`'s post-insertion steps run for the second time (as a result of the outermost `document.body.append(script1, script2)` call finally invoking `script2`'s post-insertion steps), `script2` will finally execute. Here's the code snippet:

```js
const script1 = document.createElement('script');
script1.innerText = `
  const div = document.querySelector('div');
  div.append(script2);
  script2.removeAttribute('type');
`;

const script2 = document.createElement('script');
script2.id = 'script2';
// Makes the script invalid so that it does not run upon insertion.
script2.type = 'foo';
script2.innerText = 'console.log("grazie");';

document.body.append(script1, script2);
```

For some reason WebKit does not run the script, which I think indicates that the post-insertion steps do not run the second time for `script2`. I would love to know why; @rniwa might know. I might've investigated this over in https://github.com/whatwg/html/pull/10188 but I can't remember off the top of my head. Either way, hopefully this sheds some light on this scenario. WDYT @smaug----

domfarolino: Hmm, I see the point. For selfish implementer reasons I would gently push back just because both the WebKit and Blink use the word "insertion" in their representations of these steps, so reducing cognitive overhead for implementers seems like a good thing to me. But I am not married to it. If we really wanted to change this, I would!

annevk: If the semantics really are post-connected in Chromium and Chromium they should just rename. I agree with Olli that we should attempt to have accurate names.

domfarolino: I'm assuming you mean Chromium and WebKit? Either way, yes that sounds good to me. Latest commit here renames everything accordingly. I went with `post-connection steps` instead of `post-connected` steps, since `post-connection` is similar to `post-insertion` which is what this PR had before the rename (i.e., we didn't choose `post-inserted steps`). I'm indifferent; if you think it should be renamed to `post-connected` let me know!

dom.bs
```diff
@@ -2752,6 +2810,30 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
  <var>parent</var> with <var>nodes</var>, « », <var>previousSibling</var>, and <var>child</var>.
 
  <li><p>Run the <a>children changed steps</a> for <var>parent</var>.
+
+ <li>
+  <p>Let <var>staticNodeList</var> be a <a for=/>list</a> of <a for=/>nodes</a>, initially
+  empty.</p>
```
annevk, line 2816:
```suggestion
  « ».</p>
```

dom.bs
```diff
@@ -2752,6 +2810,30 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+  <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
+
+  <ol>
+   <li><p>For each <a>shadow-including inclusive descendant</a> <var>inclusiveDescendant</var> of
+   <var>node</var>, in <a>shadow-including tree order</a>, <a for=list>append</a>
+   <var>inclusiveDescendant</var> to <var>staticNodeList</var>.
+  </ol>
+ </li>
+
+ <li><p><a for=list>For each</a> <var>node</var> in <var>staticNodeList</var>, if <var>node</var> is
```
annevk, line 2835: For each X _of_ Y

dom.bs
annevk: This needed a follow-up issue I think as discussed in WHATNOT. Can you file that?

dom.bs
domfarolino: I think this can be resolved now after the latest commits, so I'll do just that.

dom.bs
domfarolino: Ah, yeah. I just copied the line above which I guess it outdated maybe.

dom.bs
```diff
@@ -2645,12 +2645,70 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
 
 <p><a lt="Other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-insert-ext>insertion steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed <var ignore>insertedNode</var>, as indicated in the <a for=/>insert</a>
-algorithm below.
+algorithm is passed <var>insertedNode</var>, as indicated in the <a for=/>insert</a> algorithm
+below. These steps must not modify the <a>node tree</a> that <var>insertedNode</var>
+<a>participates</a> in, create <a for=/>browsing contexts</a>, <a lt="fire an event">fire
```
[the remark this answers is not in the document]
domfarolino: OK before I address this I want to make sure I know what phrasing-level means. Do you mean I should wrap after `,` and before `<a lt="fire an event"`, as to not interrupt any part of the element or its inner text, with a wrap?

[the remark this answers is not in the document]
annevk: Yeah that's correct.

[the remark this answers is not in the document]
annevk: https://html.spec.whatwg.org/multipage/dom.html#phrasing-content

dom.bs
domfarolino: Done. https://github.com/whatwg/dom/issues/1291

dom.bs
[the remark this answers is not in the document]
domfarolino: OK this and all instances should be taken care of now.

dom.bs
```diff
@@ -2752,6 +2809,30 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+  <p class="note">We collect all <a for=/>nodes</a> <em>before</em> calling the
+  <a>post-connection steps</a> on any one of them, instead of calling the
+  <a>post-connection steps</a> <em>while</em> we're traversing the <a>node tree</a>. This is because
+  the <a>post-connection steps</a> can modify the tree's structure, making live traversal unsafe,
+  possibly leading to the <a>post-connection steps</a> being called multiple times on the same
+  <a>node</a>.</p>
+ </li>
+
+ <li>
+  <p>For each <var>node</var> in <var>nodes</var>, in <a>tree order</a>:
```
annevk, line 2825:
```suggestion
  <p>For each <var>node</var> of <var>nodes</var>, in <a>tree order</a>:
```

annevk (approved):
I plan on merging this next week Wednesday assuming there's no further comments.

PR #1267, opened 2024-03-27
Record timing info for event handling.

noamr:
This feeds into "Long Animation Frame" timing.

   * Chromium (implemented)
   * Gecko (https://github.com/mozilla/standards-positions/issues/929)
   * See https://wpt.fyi/results/long-animation-frame/tentative/l
   * Chromium: already implemented
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1348405
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=270913
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/mdn/issues/535

***
<a href="https://whatpr.org/dom/1267.html" title="Last updated on Apr 15, 2024, 4:54 PM UTC (6902a5d)">Preview</a> | <a href="https://whatpr.org/dom/1267/68df785...6902a5d.html" title="Last updated on Apr 15, 2024, 4:54 PM UTC (6902a5d)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1625,13 +1625,13 @@ and an optional <var>legacyOutputDidListenersThrowFlag</var>, run these steps:
 
      <li><p>If <var>invocationTargetInShadowTree</var> is false, then set <var>global</var>'s
      <a for=Window>current event</a> to <var>event</var>.
-
-     <li><p><a>Record timing info for event listener</a> given <var>event</var> and
-     <var>listener</var>.
     </ol>
 
    <li><p>If <var>listener</var>'s <a for="event listener">passive</a> is true, then set
    <var>event</var>'s <a>in passive listener flag</a>.
+
+   <li><p>If <var>global</var> is a {{Window}} object, then
+   <a>record timing info for event listener</a> given <var>event</var> and <var>listener</var>.
 
    <li>
     <p><a>Call a user object's operation</a> with <var>listener</var>'s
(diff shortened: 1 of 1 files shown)


annevk:
Why is it called "event handler" when you're in fact creating something for each listener? Event handlers and event listeners are distinct.

noamr:
> Why is it called "event handler" when you're in fact creating something for each listener? Event handlers and event listeners are distinct.

Good point, will rename this.

noamr:
> > Why is it called "event handler" when you're in fact creating something for each listener? Event handlers and event listeners are distinct.
> 
> Good point, will rename this.

Renamed, though would need to wait for the refs to update before this builds successfully.

annevk (commented):
I wonder if this should happen after setting event's in passive listener flag. At least it seems the intent is that this happens close to actually calling the callback and where it is placed now it seems like it could easily move further away from that.

You'll have the duplicate global conditional, but seems acceptable?

noamr:
> I wonder if this should happen after setting event's in passive listener flag. At least it seems the intent is that this happens close to actually calling the callback and where it is placed now it seems like it could easily move further away from that.
> 
> You'll have the duplicate global conditional, but seems acceptable?

Yes, that's ok

noamr:
> > I wonder if this should happen after setting event's in passive listener flag. At least it seems the intent is that this happens close to actually calling the callback and where it is placed now it seems like it could easily move further away from that.
> > You'll have the duplicate global conditional, but seems acceptable?
> 
> Yes, that's ok

Done

annevk (approved):

PR #1268, opened 2024-03-27
Trusted types attributes

lukewarlow:
This calls the get Trusted Types-compliant attribute value algorithm from Trusted Types (https://github.com/w3c/trusted-types/pull/418) from attribute's change, append, and replace.

Changed the signature of setAttribute and setAttributeNS to accept Trusted Types as values. The underlying Attr node's values continue to be DOMString, so moving nodes across elements or adding standalone attributes to elements can cause TT violations. This matches WPT tests and the Chromium's implementation.

See and https://github.com/whatwg/dom/issues/789. Supercedes PR https://github.com/whatwg/dom/pull/809.

   * Chromium
   * Gecko
   * https://github.com/web-platform-tests/wpt/tree/master/trusted-types
   * Chromium:https://bugs.chromium.org/p/chromium/issues/detail?id=739170
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=266630

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -52,9 +52,10 @@ spec:html; type:element
 <p>This specification depends on the Infra Standard. [[!INFRA]]
 
 <p>Some of the terms used in this specification are defined in <cite>Encoding</cite>,
-<cite>Selectors</cite>, <cite>Web IDL</cite>, <cite>XML</cite>, and <cite>Namespaces in XML</cite>.
+<cite>Selectors</cite>, <cite>Trusted Types</cite>, <cite>Web IDL</cite>, <cite>XML</cite>, and <cite>Namespaces in XML</cite>.
 [[!ENCODING]]
 [[!SELECTORS4]]
+[[!TRUSTED-TYPES]]
 [[!WEBIDL]]
 [[!XML]]
 [[!XML-NAMES]]
@@ -6032,8 +6033,8 @@ interface Element : Node {
   sequence&lt;DOMString> getAttributeNames();
   DOMString? getAttribute(DOMString qualifiedName);
   DOMString? getAttributeNS(DOMString? namespace, DOMString localName);
-  [CEReactions] undefined setAttribute(DOMString qualifiedName, DOMString value);
-  [CEReactions] undefined setAttributeNS(DOMString? namespace, DOMString qualifiedName, DOMString value);
+  [CEReactions] undefined setAttribute(DOMString qualifiedName, (TrustedType or DOMString) value);
+  [CEReactions] undefined setAttributeNS(DOMString? namespace, DOMString qualifiedName, (TrustedType or DOMString) value);
   [CEReactions] undefined removeAttribute(DOMString qualifiedName);
   [CEReactions] undefined removeAttributeNS(DOMString? namespace, DOMString localName);
   [CEReactions] boolean toggleAttribute(DOMString qualifiedName, optional boolean force);
@@ -6354,6 +6355,10 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 <a>attribute</a> <var>attribute</var> to <var>value</var>, run these steps:
 
 <ol>
+<li><p>Set <var>value</var> to the result of calling <a>Get Trusted Types-compliant attribute
+ value</a> for <var>attribute</var>, with <var>attribute</var>'s <a for=Attr>element</a> and
+ <var>value</var>. [[!TRUSTED-TYPES]]
+
  <li><p>Let <var>oldValue</var> be <var>attribute</var>'s <a for=Attr>value</a>.</p></li>
 
  <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>value</var>.
@@ -6367,6 +6372,11 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 steps:
 
 <ol>
+<li><p>Set <var>attribute</var>'s
+ <a for=Attr>value</a> to the result of calling <a>Get Trusted Types-compliant attribute value</a>
+ for <var>attribute</var>, with <var>element</var> and <var>attribute</var>'s <a for=Attr>value</a>.
+ [[!TRUSTED-TYPES]]
+
  <li><p><a for=list>Append</a> <var>attribute</var> to <var>element</var>'s
  <a for=Element>attribute list</a>.
 
@@ -6395,6 +6405,11 @@ steps:
 <a>attribute</a> <var>oldAttr</var> with an <a>attribute</a> <var>newAttr</var>, run these steps:
 
 <ol>
+ <li><p>Set <var>newAttr</var>'s
+ <a for=Attr>value</a> to the result of calling <a>Get Trusted Types-compliant attribute value</a>
+ for <var>newAttr</var>, with <var>oldAttr</var>'s <a for=Attr>element</a> and <var>newAttr</var>'s
+ <a for=Attr>value</a>.[[!TRUSTED-TYPES]]
+
  <li><p><a for=list>Replace</a> <var>oldAttr</var> by <var>newAttr</var> in <var>oldAttr</var>'s
  <a for=Attr>element</a>'s <a for=Element>attribute list</a>.
 
@@ -6481,7 +6496,7 @@ string <var>namespace</var> (default null):</p>
 
 <div algorithm>
 <p>To <dfn export id=concept-element-attributes-set-value>set an attribute value</dfn> given an
-<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string <var>value</var>,
+<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string or <a>TrustedType</a> <var>value</var>,
 an optional null or string <var>prefix</var> (default null), and an optional null or string
 <var>namespace</var> (default null):
 
@@ -6492,7 +6507,7 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
 
  <li>If <var>attribute</var> is null, create an <a>attribute</a> whose <a for=Attr>namespace</a> is
  <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
- <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is <var>value</var>, and
+ <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is stringified <var>value</var>, and
  <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>, then
  <a lt="append an attribute">append</a> this <a>attribute</a> to <var>element</var>, and then
  return.
@@ -6760,7 +6775,7 @@ method steps are:
 
  <li><p>If <var>attribute</var> is null, create an <a>attribute</a> whose
  <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
- <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
+ stringified <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
  then <a lt="append an attribute">append</a> this <a>attribute</a> to <a>this</a>, and then return.
 
  <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
--- a/dom.bs
+++ b/dom.bs
@@ -6412,8 +6412,8 @@ steps:
 </ol>
 
 <p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn>
-<var>value</var> for an <a>attribute</a> <var>attribute</var>, with <a for=/>element</a>
-<var>element</var>, run these steps:
+{{TrustedType}} or a string <var>value</var> for an <a>attribute</a> <var>attribute</var>, with
+<a for=/>element</a> <var>element</var>:
 
 <ol>
  <li><p>Let <var>validValue</var> be the result of calling
@@ -6508,26 +6508,21 @@ or string <var>namespace</var> (default null):
  <a lt="get an attribute by namespace and local name">getting an attribute</a> given
  <var>namespace</var>, <var>localName</var>, and <var>element</var>.
 
- <li>If <var>attribute</var> is null, create an <a>attribute</a> whose <a for=Attr>namespace</a> is
- <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
- <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is <var>value</var>, and
- <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>, then
- <a lt="append an attribute">append</a> this <a>attribute</a> to <var>element</var>, and then
- return.
+ <li>
+  <p>If <var>attribute</var> is null, then:
+  <ol>
+    <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>namespace</a> is
+    <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
+    <a for=Attr>local name</a> is <var>localName</var> and <a for=Node>node document</a> is
+    <var>element</var>'s <a for=Node>node document</a>.
 
- <ol>
-  <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>namespace</a> is
-  <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
-  <a for=Attr>local name</a> is <var>localName</var> and <a for=Node>node document</a> is
-  <var>element</var>'s <a for=Node>node document</a>.
+    <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with
+    <var>element</var>.
 
-  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with
-  <var>element</var>.
+    <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>.
 
-  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>.
-
-  <li><p>Return.
- </ol>
+    <li><p>Return.
+  </ol>
 
  <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
 </ol>
@@ -6790,20 +6785,21 @@ method steps are:
  and null otherwise.
  <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
 
- <li><p>If <var>attribute</var> is null, then:
+ <li>
+  <p>If <var>attribute</var> is null, then:
 
- <ol>
-  <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>local name</a> is
-  <var>qualifiedName</var> and <a for=Node>node document</a> is <a>this</a>'s
-  <a for=Node>node document</a>.
+  <ol>
+   <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>local name</a> is
+   <var>qualifiedName</var> and <a for=Node>node document</a> is <a>this</a>'s
+   <a for=Node>node document</a>.
 
-  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>,
-  with <a>this</a>.
+   <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>,
+   with <a>this</a>.
 
-  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <a>this</a>.
+   <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <a>this</a>.
 
-  <li><p>Return.
- </ol>
+   <li><p>Return.
+  </ol>
 
  <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
 </ol>
--- a/dom.bs
+++ b/dom.bs
@@ -52,7 +52,8 @@ spec:html; type:element
 <p>This specification depends on the Infra Standard. [[!INFRA]]
 
 <p>Some of the terms used in this specification are defined in <cite>Encoding</cite>,
-<cite>Selectors</cite>, <cite>Trusted Types</cite>, <cite>Web IDL</cite>, <cite>XML</cite>, and <cite>Namespaces in XML</cite>.
+<cite>Selectors</cite>, <cite>Trusted Types</cite>, <cite>Web IDL</cite>, <cite>XML</cite>, and
+<cite>Namespaces in XML</cite>.
 [[!ENCODING]]
 [[!SELECTORS4]]
 [[!TRUSTED-TYPES]]
@@ -6357,7 +6358,8 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 <ol>
  <li><p>Let <var>oldValue</var> be <var>attribute</var>'s <a for=Attr>value</a>.</p></li>
 
- <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>, with <var>attribute</var>'s <a for=Attr>element</a>.
+ <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>, with
+ <var>attribute</var>'s <a for=Attr>element</a>.
 
  <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
  <a for=Attr>element</a>, <var>oldValue</var>, and <var>value</var>.
@@ -6409,13 +6411,17 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
-<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn> <var>value</var> for an <a>attribute</a> <var>attribute</var>, with <a>element</a> <var>element</var>, run these steps:
+<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn>
+<var>value</var> for an <a>attribute</a> <var>attribute</var>, with <a for=/>element</a>
+<var>element</var>, run these steps:
 
- <ol>
- <li><p>Let <var>validValue</var> be the result of calling <a>Get Trusted Types-compliant attribute value</a>
-  for <var>attribute</var>, with <var>element</var> and <var>value</var>.[[!TRUSTED-TYPES]]
+<ol>
+ <li><p>Let <var>validValue</var> be the result of calling
+ <a>Get Trusted Types-compliant attribute value</a> for <var>attribute</var>, with
+ <var>element</var> and <var>value</var>. [[!TRUSTED-TYPES]]
+
  <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>validValue</var>.
- </ol>
+</ol>
 
 <hr>
 
@@ -6479,7 +6485,8 @@ string <var>namespace</var> (default null):</p>
 
  <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
 
- <li><p><a>Validate and set attribute value</a> <var>newAttr</var>'s <a for="Attr">value</a> for <var>newAttr</var> with <var>element</var>.
+ <li><p><a>Validate and set attribute value</a> <var>newAttr</var>'s <a for="Attr">value</a> for
+ <var>newAttr</var> with <var>element</var>.
 
  <li><p>If <var>oldAttr</var> is non-null, then <a lt="replace an attribute">replace</a>
  <var>oldAttr</var> with <var>attr</var>.
@@ -6492,9 +6499,9 @@ string <var>namespace</var> (default null):</p>
 
 <div algorithm>
 <p>To <dfn export id=concept-element-attributes-set-value>set an attribute value</dfn> given an
-<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string or <a>TrustedType</a> <var>value</var>,
-an optional null or string <var>prefix</var> (default null), and an optional null or string
-<var>namespace</var> (default null):
+<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string or {{TrustedType}}
+<var>value</var>, an optional null or string <var>prefix</var> (default null), and an optional null
+or string <var>namespace</var> (default null):
 
 <ol>
  <li>Let <var>attribute</var> be the result of
@@ -6511,12 +6518,14 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
  <ol>
   <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>namespace</a> is
   <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
-  <a for=Attr>local name</a> is <var>localName</var> and
-  <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>.
+  <a for=Attr>local name</a> is <var>localName</var> and <a for=Node>node document</a> is
+  <var>element</var>'s <a for=Node>node document</a>.
 
-  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with <var>element</var>.
+  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with
+  <var>element</var>.
 
-  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>
+  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>.
+
   <li><p>Return.
  </ol>
 
@@ -6782,13 +6791,19 @@ method steps are:
  <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
 
  <li><p>If <var>attribute</var> is null, then:
-   <ol>
-     <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose
-       <a for=Attr>local name</a> is <var>qualifiedName</var> and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>.
-     <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>, with <a>this</a>.
-     <li><a lt="append an attribute">Append</a> <var>attribute</var> to <a>this</a>.
-     <li><p>Return.
-   </ol>
+
+ <ol>
+  <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>local name</a> is
+  <var>qualifiedName</var> and <a for=Node>node document</a> is <a>this</a>'s
+  <a for=Node>node document</a>.
+
+  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>,
+  with <a>this</a>.
+
+  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <a>this</a>.
+
+  <li><p>Return.
+ </ol>
 
  <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
 </ol>
--- a/dom.bs
+++ b/dom.bs
@@ -6505,12 +6505,27 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
  <a lt="get an attribute by namespace and local name">getting an attribute</a> given
  <var>namespace</var>, <var>localName</var>, and <var>element</var>.
 
- <li>If <var>attribute</var> is null, create an <a>attribute</a> whose <a for=Attr>namespace</a> is
- <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
- <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is stringified <var>value</var>, and
- <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>, then
- <a lt="append an attribute">append</a> this <a>attribute</a> to <var>element</var>, and then
- return.
+ <li><p>If <var>attribute</var> is null, then:
+
+ <ol>
+  <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>namespace</a> is
+  <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
+  <a for=Attr>local name</a> is <var>localName</var> and
+  <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>.
+
+  <li><p>Set <a>attribute</a>'s <var>value</var> to the result of calling <a>Get Trusted Types-compliant attribute
+ value</a> for <var>attribute</var>, with <var>element</var> and
+ <var>value</var>. [[!TRUSTED-TYPES]]
+
+ <li><p><a for=list>Append</a> <var>attribute</var> to <var>element</var>'s
+ <a for=Element>attribute list</a>.
+
+ <li><p>Set <var>attribute</var>'s <a for=Attr>element</a> to <var>element</var>.
+
+ <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>element</var>, null, and
+ <var>attribute</var>'s <a for=Attr>value</a>.
+ <li><p>Return.
+ </ol>
 
  <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
 </ol>
--- a/dom.bs
+++ b/dom.bs
@@ -6355,13 +6355,9 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 <a>attribute</a> <var>attribute</var> to <var>value</var>, run these steps:
 
 <ol>
-<li><p>Set <var>value</var> to the result of calling <a>Get Trusted Types-compliant attribute
- value</a> for <var>attribute</var>, with <var>attribute</var>'s <a for=Attr>element</a> and
- <var>value</var>. [[!TRUSTED-TYPES]]
-
  <li><p>Let <var>oldValue</var> be <var>attribute</var>'s <a for=Attr>value</a>.</p></li>
 
- <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>value</var>.
+ <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>, with <var>attribute</var>'s <a for=Attr>element</a>.
 
  <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
  <a for=Attr>element</a>, <var>oldValue</var>, and <var>value</var>.
@@ -6372,11 +6368,6 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
 steps:
 
 <ol>
-<li><p>Set <var>attribute</var>'s
- <a for=Attr>value</a> to the result of calling <a>Get Trusted Types-compliant attribute value</a>
- for <var>attribute</var>, with <var>element</var> and <var>attribute</var>'s <a for=Attr>value</a>.
- [[!TRUSTED-TYPES]]
-
  <li><p><a for=list>Append</a> <var>attribute</var> to <var>element</var>'s
  <a for=Element>attribute list</a>.
 
@@ -6405,11 +6396,6 @@ steps:
 <a>attribute</a> <var>oldAttr</var> with an <a>attribute</a> <var>newAttr</var>, run these steps:
 
 <ol>
- <li><p>Set <var>newAttr</var>'s
- <a for=Attr>value</a> to the result of calling <a>Get Trusted Types-compliant attribute value</a>
- for <var>newAttr</var>, with <var>oldAttr</var>'s <a for=Attr>element</a> and <var>newAttr</var>'s
- <a for=Attr>value</a>.[[!TRUSTED-TYPES]]
-
  <li><p><a for=list>Replace</a> <var>oldAttr</var> by <var>newAttr</var> in <var>oldAttr</var>'s
  <a for=Attr>element</a>'s <a for=Element>attribute list</a>.
 
@@ -6422,6 +6408,14 @@ steps:
  <a for=Attr>element</a>, <var>oldAttr</var>'s <a for=Attr>value</a>, and <var>newAttr</var>'s
  <a for=Attr>value</a>.
 </ol>
+
+<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn> <var>value</var> for an <a>attribute</a> <var>attribute</var>, with <a>element</a> <var>element</var>, run these steps:
+
+ <ol>
+ <li><p>Let <var>validValue</var> be the result of calling <a>Get Trusted Types-compliant attribute value</a>
+  for <var>attribute</var>, with <var>element</var> and <var>value</var>.[[!TRUSTED-TYPES]]
+ <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>validValue</var>.
+ </ol>
 
 <hr>
 
@@ -6485,6 +6479,8 @@ string <var>namespace</var> (default null):</p>
 
  <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
 
+ <li><p><a>Validate and set attribute value</a> <var>newAttr</var>'s <a for="Attr">value</a> for <var>newAttr</var> with <var>element</var>.
+
  <li><p>If <var>oldAttr</var> is non-null, then <a lt="replace an attribute">replace</a>
  <var>oldAttr</var> with <var>attr</var>.
 
@@ -6505,7 +6501,12 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
  <a lt="get an attribute by namespace and local name">getting an attribute</a> given
  <var>namespace</var>, <var>localName</var>, and <var>element</var>.
 
- <li><p>If <var>attribute</var> is null, then:
+ <li>If <var>attribute</var> is null, create an <a>attribute</a> whose <a for=Attr>namespace</a> is
+ <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
+ <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is <var>value</var>, and
+ <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>, then
+ <a lt="append an attribute">append</a> this <a>attribute</a> to <var>element</var>, and then
+ return.
 
  <ol>
   <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose <a for=Attr>namespace</a> is
@@ -6513,18 +6514,10 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
   <a for=Attr>local name</a> is <var>localName</var> and
   <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>.
 
-  <li><p>Set <a>attribute</a>'s <var>value</var> to the result of calling <a>Get Trusted Types-compliant attribute
- value</a> for <var>attribute</var>, with <var>element</var> and
- <var>value</var>. [[!TRUSTED-TYPES]]
+  <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var> with <var>element</var>.
 
- <li><p><a for=list>Append</a> <var>attribute</var> to <var>element</var>'s
- <a for=Element>attribute list</a>.
-
- <li><p>Set <var>attribute</var>'s <a for=Attr>element</a> to <var>element</var>.
-
- <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>element</var>, null, and
- <var>attribute</var>'s <a for=Attr>value</a>.
- <li><p>Return.
+  <li><p><a lt="append an attribute">Append</a> <var>attribute</var> to <var>element</var>
+  <li><p>Return.
  </ol>
 
  <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
@@ -6788,10 +6781,14 @@ method steps are:
  and null otherwise.
  <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
 
- <li><p>If <var>attribute</var> is null, create an <a>attribute</a> whose
- <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
- stringified <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
- then <a lt="append an attribute">append</a> this <a>attribute</a> to <a>this</a>, and then return.
+ <li><p>If <var>attribute</var> is null, then:
+   <ol>
+     <li><p>Set <var>attribute</var> to a new <a>attribute</a> whose
+       <a for=Attr>local name</a> is <var>qualifiedName</var> and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>.
+     <li><p><a>Validate and set attribute value</a> <var>value</var> for <var>attribute</var>, with <a>this</a>.
+     <li><a lt="append an attribute">Append</a> <var>attribute</var> to <a>this</a>.
+     <li><p>Return.
+   </ol>
 
  <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
 </ol>


lukewarlow:
This is a clone of #1247 where I will finish any outstanding work to get this across the line

lukewarlow:
See also https://github.com/whatwg/dom/pull/1258 which is another integration point with the DOM spec that we need for TT.

dom.bs
```diff
@@ -6484,8 +6514,19 @@ string <var>namespace</var> (default null):</p>
 
  <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
 
- <li><p>If <var>oldAttr</var> is non-null, then <a lt="replace an attribute">replace</a>
- <var>oldAttr</var> with <var>attr</var>.
+ <li>
+  <p>If <var>oldAttr</var> is non-null, then:</p>
```
otherdaniel, line 6518: Would this condition result in the TT-check being skipped if the old attribute didn't exist?

lukewarlow: In that case append an attribute is called which runs validate and set as it's first step, so TT check still happens

lukewarlow: I need to check if replace is used elsewhere, if it's not I might move the validate and set call into it, to be consistent with other places.

dom.bs
```diff
@@ -6484,8 +6514,19 @@ string <var>namespace</var> (default null):</p>
...
+
+  <ul>
+   <li><p><a>Validate and set attribute value</a> <var>attr</var>'s <a for="Attr">value</a> for
+   <var>attr</var> with <var>element</var>.
+
+   <li><p>If <var>element</var> <a lt="has an attribute">has</a> an <a>attribute</a>
```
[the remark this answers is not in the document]
lukewarlow: I've read through this spec path and I believe that, if the default policy removes the existing attribute node, then this will call replace an attribute, which in turn calls replace a list item, which results in a no-op because the old value no longer exists to be replaced.

So I think in this situation you're probably right the spec doesn't need to do anything, will make that change.

dom.bs
lukewarlow: I don't think replace is used anywhere else so will move the validation step a level down.

dom.bs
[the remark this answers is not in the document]
lukewarlow: Having said that I'm slightly uneasy about stuff like attribute change steps still firing and how that all works spec wise.

Both Chromium and WebKit don't actually follow the spec 100% here and so I think would actually result in a different behaviour to this spec (the index lookup happens after the TT call) so that's also not ideal.

lukewarlow:
I think I've addressed all the comments from #1247.

I do want to point out Chrome and WebKit don't (or at least not in a way obvious to me) 100% follow the flow of the spec and as a result this may result in differences specifically in weird cases with attribute mutation.

So that bit especially it would be good to get feedback on.

It's also worth being aware that like Chromium's implementation this spec means that certain ways to update a nodes value don't work with a trusted type object as a user might expect. (e.g. `iframe.getAttributeNode('srcdoc').value = trustedHTMLObj;` will throw unless allowed by a default policy). I think in pratice this will be fine, and in some cases is the only real option we have without nodes keeping track of whether they're trusted or not (which would add lots of complexity)

dom.bs
```diff
@@ -4279,7 +4281,7 @@ empty string instead, and then do as described below, switching on the interface
 
 <dl class=switch>
  <dt>{{Attr}}
- <dd><p><a>Set an existing attribute value</a> with <a>this</a> and the given value.
+ <dd><p><a>Set an existing attribute value</a> with <a>this</a>, the given value.
```
annevk, line 4284: I don't understand this change.

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
 
 <p>To <dfn export id=concept-element-attributes-append lt="append an attribute">append</dfn> an
-<a>attribute</a> <var>attribute</var> to an <a for=/>element</a> <var>element</var>, run these
-steps:
+<a>attribute</a> <var>attribute</var> to an <a for=/>element</a> <var>element</var>, with optional
+boolean <var>validate</var> (default true), run these steps:
 
 <ol>
+ <li>
+  <p>If <var>validate</var>, then:
```
annevk, line 6396:
```suggestion
  <p>If <var>validate</var> is true, then:
```

annevk (commented):
This change is either incomplete or makes many cosmetic changes that would be best proposed separately as they confuse me quite a bit.

dom.bs
```diff
@@ -6875,11 +6915,14 @@ method steps are to return the result of
...
 
-<p>The <dfn method for=Element><code>setAttributeNode(<var>attr</var>)</code></dfn> and
-<dfn method for=Element><code>setAttributeNodeNS(<var>attr</var>)</code></dfn> methods steps are to
+<p>The <dfn method for=Element><code>setAttributeNode(<var>attr</var>)</code></dfn> methods steps are to
 return the result of <a lt="set an attribute">setting an attribute</a> given <var>attr</var> and
 <a>this</a>.
 
+<p>The <dfn method for=Element><code>setAttributeNodeNS(<var>attr</var>)</code></dfn> methods steps are to
+return the result of <a lt="set an attribute">setting an attribute</a> given <var>attr</var>,
+<a>this</a>.
```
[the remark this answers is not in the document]
lukewarlow: Spotted this earlier, have fixed.

dom.bs
lukewarlow: apologies some of this is because stuff was added and then removed and I've clearly not reverted stuff very well

dom.bs
```diff
@@ -6422,6 +6440,18 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
+<p>To <dfn id=concept-element-attributes-validate-and-set-value>validate and set attribute value</dfn>
+{{TrustedType}} or a string <var>value</var> for an <a>attribute</a> <var>attribute</var>, with
+<a for=/>element</a> <var>element</var>:
```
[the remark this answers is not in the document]
lukewarlow: Changed this to Verify

lukewarlow:
> This change is either incomplete or makes many cosmetic changes that would be best proposed separately as they confuse me quite a bit.

Apologies I thought I'd reverted all of these changes but I missed a few, it's because I changed stuff and then changed it back and this led to some wonky diffs. Have hopefully reverted all of these unnecessary changes

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
- <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>value</var>.
+ <li><p><a>Verify and set attribute value</a> <var>value</var> for <var>attribute</var>, with
+ <var>attribute</var>'s <a for=Attr>element</a>.
 
- <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
- <a for=Attr>element</a>, <var>oldValue</var>, and <var>value</var>.
+ <li><p>If <var>attribute</var>'s <a for=Attr>element</a> <a lt="has an attribute">has</a>
+ an <a>attribute</a> <var>attribute</var>, then <a>handle attribute changes</a> for
+ <var>attribute</var> with <var>attribute</var>'s <a for=Attr>element</a>, <var>oldValue</var>, and
+ <var>value</var>.
```
annevk, line 6385: Should this be _value_ or _attribute_'s value? They can be different, right?

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
+
+ <li><p>Otherwise, throw an "{{InvalidStateError!!exception}}" {{DOMException}}.
```
annevk, line 6387: Does this match the exception Trusted Types throws by default?

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
  <li><p>Let <var>oldValue</var> be <var>attribute</var>'s <a for=Attr>value</a>.</p></li>
 
- <li><p>Set <var>attribute</var>'s <a for=Attr>value</a> to <var>value</var>.
+ <li><p><a>Verify and set attribute value</a> <var>value</var> for <var>attribute</var>, with
+ <var>attribute</var>'s <a for=Attr>element</a>.
 
- <li><p><a>Handle attribute changes</a> for <var>attribute</var> with <var>attribute</var>'s
- <a for=Attr>element</a>, <var>oldValue</var>, and <var>value</var>.
+ <li><p>If <var>attribute</var>'s <a for=Attr>element</a> <a lt="has an attribute">has</a>
+ an <a>attribute</a> <var>attribute</var>, then <a>handle attribute changes</a> for
```
annevk, line 6383: "has an attribute" should be the linking text I think. Not sure why we'd mention the type as its own thing here.

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
+ <var>value</var>.
+
+ <li><p>Otherwise, throw an "{{InvalidStateError!!exception}}" {{DOMException}}.
 </ol>
 
 <p>To <dfn export id=concept-element-attributes-append lt="append an attribute">append</dfn> an
-<a>attribute</a> <var>attribute</var> to an <a for=/>element</a> <var>element</var>, run these
-steps:
+<a>attribute</a> <var>attribute</var> to an <a for=/>element</a> <var>element</var>, with optional
+boolean <var>validate</var> (default true), run these steps:
```
annevk, line 6392: an optional boolean*

You can remove "run these steps" as well. We tend to leave that out these days when the algorithm is introduced with "To" and no return value is to be listed.

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
 
 <ol>
+ <li>
+  <p>If <var>validate</var> is true, then:
+  <ol>
```
annevk, line 6397: Need a newline before `<ol>`.

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
 
 <p>To <dfn export id=concept-element-attributes-append lt="append an attribute">append</dfn> an
-<a>attribute</a> <var>attribute</var> to an <a for=/>element</a> <var>element</var>, run these
-steps:
+<a>attribute</a> <var>attribute</var> to an <a for=/>element</a> <var>element</var>, with optional
+boolean <var>validate</var> (default true), run these steps:
 
 <ol>
+ <li>
+  <p>If <var>validate</var> is true, then:
```
annevk, line 6396:
```suggestion
  <p>If <var>validate</var> is true:
```

dom.bs
```diff
@@ -6373,17 +6376,32 @@ given a <var>document</var>, <var>localName</var>, <var>namespace</var>, and opt
...
+  <ol>
+   <li><p><a>Verify and set attribute value</a>
+   <var>attribute</var>'s <a for="Attr">value</a> for <var>attribute</var> with <var>element</var>.
```
annevk, line 6399: The text isn't consistent around comma usage when invoking this algorithm.

dom.bs
```diff
@@ -6425,6 +6446,18 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
+<p>To <dfn id=concept-element-attributes-verify-and-set-value>verify and set attribute value</dfn>
+{{TrustedType}} or string <var>value</var> for an <a>attribute</a> <var>attribute</var>, with
+<a for=/>element</a> <var>element</var>:
+
+<ol>
+ <li><p>Let <var>validValue</var> be the result of calling
```
annevk, line 6454: verifiedValue, presumably?

dom.bs
```diff
@@ -6425,6 +6446,18 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
+<p>To <dfn id=concept-element-attributes-verify-and-set-value>verify and set attribute value</dfn>
```
annevk, line 6449: Remove the id attribute here.

dom.bs
```diff
@@ -6425,6 +6446,18 @@ steps:
...
+{{TrustedType}} or string <var>value</var> for an <a>attribute</a> <var>attribute</var>, with
+<a for=/>element</a> <var>element</var>:
```
annevk, line 6451: Perhaps:

> given a X x, Y y, and Z z:

as style instead?

dom.bs
```diff
@@ -6485,8 +6518,6 @@ string <var>namespace</var> (default null):</p>
  <var>attr</var>'s <a for=Attr>namespace</a>, <var>attr</var>'s <a for=Attr>local name</a>, and
  <var>element</var>.
 
- <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
```
annevk, line 6488: This seems like it would change the mutation records for existing applications even when they don't use Trusted Types.

dom.bs
lukewarlow: No Trusted Types throws a TypeError by default, not sure what's best to put here.

dom.bs
lukewarlow: This doesn't seem intentional (likely a result of changes that have subsequently been partially reverted), I'll add it back

dom.bs
lukewarlow: I think that's done

annevk:
Attributes are stored in a list and those do have indices per Infra. What am I missing?

lukewarlow:
Sorry I mean algorithmically the spec and implementations don't follow the same flow. So it's trickier to reason between the spec and implementation. This might just be my lack of familiarity with these APIs too.

annevk:
The path of least resistance is prolly matching Chromium. Introducing new paths that throw is always risky. If you are looking for guidance as to how, I'd need quite a bit more context to provide helpful suggestions.

dom.bs
lukewarlow: We no longer throw

lukewarlow:
> The path of least resistance is prolly matching Chromium. Introducing new paths that throw is always risky. If you are looking for guidance as to how, I'd need quite a bit more context to provide helpful suggestions.

I've partially matched Chromium but gone a bit further with the error correction as Chromium still has some bugs wrt mutating attributes within the default policy.

dom.bs
```diff
@@ -6530,23 +6545,39 @@ string <var>namespace</var> (default null):</p>
 
 <div algorithm>
 <p>To <dfn export id=concept-element-attributes-set-value>set an attribute value</dfn> given an
```
lukewarlow, line 6547: This algorithm covers setAttributeNS

dom.bs
```diff
@@ -6807,12 +6838,29 @@ method steps are:
  and null otherwise.
  <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
 
- <li><p>If <var>attribute</var> is null, create an <a>attribute</a> whose
- <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
- <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
- then <a lt="append an attribute">append</a> this <a>attribute</a> to <a>this</a>, and then return.
+ <li><p>Let <var>attributeExists</var> be false if <var>attribute</var> is null, and true otherwise.
```
lukewarlow, line 6841: This algorithm covers setAttribute

dom.bs
```diff
@@ -7382,7 +7430,13 @@ string <var>value</var>, run these steps:
  <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
  <a for=Attr>value</a> to <var>value</var>.
 
- <li><p>Otherwise, <a lt="change an attribute">change</a> <var>attribute</var> to <var>value</var>.
+ <li><p>Otherwise:
```
lukewarlow, line 7433: This covers attr.nodeValue, attr.value, and attr.textContent.

dom.bs
```diff
@@ -6519,6 +6529,11 @@ string <var>namespace</var> (default null):</p>
 
  <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
 
+ <li><p>Let <var>verifiedValue</var> be the result of calling <a>verify attribute value</a>
```
lukewarlow, line 6532: This covers, setAttributeNode, setAttributeNodeNS, setNamedItem and setNamedItemNS

dom.bs
```diff
@@ -6870,11 +6918,17 @@ method steps are:
   <p>If <var>attribute</var> is null, then:
 
   <ol>
-   <li><p>If <var>force</var> is not given or is true, create an <a>attribute</a> whose
+   <li><p>Set <var>attribute</var> to an <a>attribute</a> whose
```
lukewarlow, line 6921: This covers toggleAttribute. I need to double check this because I think it could still lead to a weird state in it's current form.

lukewarlow: Have updated this to work how chrome appears to

otherdaniel: The TT-check in Chromium's `toggleAttribute` was unintentional. It came about as a side effect of how `toggleAttribute` was implemented on top of other methods. We can't come up with any XSS based on `toggleAttribute`. It'd probably be best to remove it from the PR, and that we fix our implementation on our end.

lukewarlow: I've reverted this now so there's no enforcement in the spec. I will do a WebKit PR to remove the enforcement from there. And the existing chromium issue that discussed some of this can be used to remove it from Chromium.

annevk:
This now seems to invoke TT even when we have specifications manipulate attributes, or when `element.id`'s setter is used. Is that intentional?

For review it really helps to have an up-to-date commit message that clearly describes the intent. That's why we made it one of the PR checkboxes.

It also seems that it would cause a change in ordering between enforcing this in IDL and in the algorithms as now certain exceptions are thrown first. That will need test coverage.

lukewarlow:
> This now seems to invoke TT even when we have specifications manipulate attributes, 

I don't understand sorry could you explain what's changed since before?

> or when element.id's setter is used. Is that intentional?

That's not intentional? I don't see what I've changed that would lead that to happen?

In all cases if TT's algorithm is invoked it just returns the value you provide it, unless it's one of a select few atttributes on a select few elements (or an event handler).

I will write an overview first thing tomorrow and update the PR to provide that.

lukewarlow:
# Trusted Types integration with DOM

The DOM spec already contains partial integration with Trusted Types this was included in https://github.com/whatwg/dom/pull/1258

This PR updates the DOM functions related to attributes.

The following functions' IDL are adjusted, to accept a TrustedType (TrustedHTML or TrustedScript or TrustedScriptURL) or a string.

**Element.setAttribute**
**Element.setAttributeNS**

The above two functions are the "sanctioned" methods for updating the value of an attribute, any other way relies on default policy else they'll fail TT checks.

The algorithm for setAttribute function (and the "set an attribute value" algorithm used by setAttributeNS) is updated to call into the Trusted Types such that if the attribute being set is one of those covered by Trusted Types,
the value being set is verified (potentially being replaced with a different value).

These algorithms include some handling for if the trusted types default policy changes whether the attribute being set exists or not.

**Element.setAttributeNode** and **Element.setAttributeNodeNS** both, along with **NamedNodeMap.setNamedItem** and **NamedNodeMap.setNamedItemNS** make use of the "setting an attribute" algorithm, which is similarly updated to call into Trusted Types.

**Attr.value**, **Node.textContent**, and **Node.nodeValue** setters call "set an existing attribute value", this algorithm is updated such that for attributes attached to an element, trusted types is called into and verifies the value.

In all these cases the changes only have any impact for a select few attributes (e.g. event handlers, iframe srcdoc attribute)

lukewarlow:
The change to `set an attribute value` breaks iframe.srcdoc property setting currently (it'll do a double check).

dom.bs
```diff
@@ -6519,6 +6529,11 @@ string <var>namespace</var> (default null):</p>
 
  <li><p>If <var>oldAttr</var> is <var>attr</var>, return <var>attr</var>.
 
+ <li><p>Let <var>verifiedValue</var> be the result of calling <a>verify attribute value</a>
+ <var>attr</var>'s <a for=Attr>value</a> for <var>attr</var>, with <var>element</var>.
+
+ <li><p>Set <var>attr</var>'s <a for=Attr>value</a> to <var>verifiedValue</var>.
+
```
smaug----, line 6536: Hmm, so doing the verification may run scripts. And that means oldAttr might not be anymore in the element it used to be.   Could that cause issues? Could the value be validated first for certain kind of element but then used on some other kind of element?

dom.bs
```diff
@@ -7382,7 +7437,13 @@ string <var>value</var>, run these steps:
  <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
  <a for=Attr>value</a> to <var>value</var>.
 
- <li><p>Otherwise, <a lt="change an attribute">change</a> <var>attribute</var> to <var>value</var>.
+ <li><p>Otherwise:
+  <ol>
+    <li><p>Let <var>verifiedValue</var> be the result of calling <a>verify attribute value</a>
+    <var>value</var> for <var>attribute</var>, with <a>this</a>.
+
+   <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>verifiedValue</var>.
```
smaug----, line 7445: Doesn't this lead to effectively null pointer crashes in algorithms if attribute has been moved out from an element by the tt callback?

smaug---- (commented):
Some questions. I could be missing something in TT spec which somehow covers the problematic cases, but please at least explain and ask review again.

dom.bs
lukewarlow: The default policy doesn't provide context about which element an attribute is set on only the name of the attribute.  In this case this algorithm is triggered by APIs such as setAttributeNode or setNamedItem.

So I don't think there's anything that can happen here that's too bad. Also any mechanism you use inside of the default policy will itself trigger the default policy so it should be fine?

dom.bs
lukewarlow: I guess potentially if some sub algorithm of change an attribute is assuming that element must exist? Fwiw in practice this doesn't cause issues from what I can see.

Not sure what the best fix here would be?

dom.bs
```diff
@@ -6457,6 +6457,16 @@ steps:
...
 </ol>
 
+<p>To <dfn>verify attribute value</dfn>
+{{TrustedType}} or string <var>value</var> for an <a>attribute</a> <var>attribute</var>, given an
+<a for=/>Element</a> <var>element</var>:
+
+<ol>
+ <li><p>Return the result of calling
+ <a abstract-op>get Trusted Types-compliant attribute value</a> for <var>attribute</var>, with
+ <var>element</var>, <var>value</var>. [[!TRUSTED-TYPES]]
```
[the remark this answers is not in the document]
lukewarlow: They will end up rethrowing the exception which is what's expected.

annevk:
Now that 809bfa24535b1d55d761d8add353e1c3fde05801 is merged this should add the Infrastructure changes back in again as this now introduces the dependency on Trusted Types.

dom.bs
```diff
@@ -7435,7 +7492,13 @@ string <var>value</var>, run these steps:
  <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
  <a for=Attr>value</a> to <var>value</var>.
 
- <li><p>Otherwise, <a lt="change an attribute">change</a> <var>attribute</var> to <var>value</var>.
+ <li><p>Otherwise:
+  <ol>
+    <li><p>Let <var>verifiedValue</var> be the result of calling <a>verify attribute value</a>
+    <var>value</var> for <var>attribute</var>, with <a>this</a>.
+
+   <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>verifiedValue</var>.
```
domfarolino, line 7500: Indenting of this line is wrong. Either that, or the first `<li>` above is indented wrong.

domfarolino (commented):
So I came across this in my inbox, very belatedly, but the PR seems to have been in a lull for a bit. What is the current status? Seems there are a few comment threads that are open and unresolved, but no major blockers identified I guess? Let me know if I can help.

mbrodesser-Igalia:
https://github.com/w3c/trusted-types/issues/438 was reverted and additionally mentions `ChildNode`'s `before`, `after` and `replaceWith`.

https://www.w3.org/TR/trusted-types/#integration-with-dom refers to this PR. So should this PR cover `ChildNode`'s methods too?

CC @fred-wang

mbrodesser-Igalia:
https://www.w3.org/TR/trusted-types/#integration-with-dom links to this ticket.
So this ticket should either cover the APIs mentioned at https://github.com/w3c/trusted-types/issues/438 too, or they should be dealt with in a separate ticket.

CC @fred-wang

fred-wang:
>  Tests are written and can be reviewed and commented upon at:
   https://github.com/web-platform-tests/wpt/tree/master/trusted-types

On https://github.com/whatwg/dom/pull/1268#issuecomment-2165953464 you mentioned the affected APIs are

- setAttribute
- setAttributeNS
- Element.setAttributeNode
- Element.setAttributeNodeNS
- NamedNodeMap.setNamedItem
- NamedNodeMap.setNamedItemNS
- Attr.value
- Node.textContent
- Node.nodeValue

Is that an exhautive list?

I think we would need tests for each of the affected API to check:
- That attribute change is blocked by the default policy.
- That a non-trivial default policy is applied before attribute change.

I found the following tests:
- block-string-assignment-to-attribute-via-attribute-node.html
- block-string-assignment-to-Element-setAttribute.html
- testing/web-platform/tests/trusted-types/TrustedType-AttributeNodes.html

But they don't cover e.g. setAttributeNodeNS or setNamedItemNS and the last two are a bit messy (https://phabricator.services.mozilla.com/D227943 was enough to make the last one pass in Firefox, which seems dubious to me).

dom.bs
smaug----: The callback may have added the new attr to be in some other element's attribute list

fred-wang: FWIW, in https://bugzilla.mozilla.org/show_bug.cgi?id=1944511 I'm adding new tests and will try to make sure this case and the other ones mentioned below are covered.

But my guess would be we should just throw the same exception as in step 1 in that case:

> 1 If attr’s [element](https://dom.spec.whatwg.org/#concept-attribute-element) is neither null nor element, [throw](https://webidl.spec.whatwg.org/#dfn-throw) an "[InUseAttributeError](https://webidl.spec.whatwg.org/#inuseattributeerror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

mozfreddyb: The default policy only getting attribute information is an issue elsewhere, btw. https://github.com/w3c/trusted-types/issues/569 - Would be useful to add element information imho.

lukewarlow: Good catch, I've added the fix suggested by Fred as that seems the most logical solution. I'll make sure to add a test against this when I fix it in WebKit.

dom.bs
lukewarlow: I've ran specfmt on this PR, hopefully that address this?

dom.bs
fred-wang: I haven't had time to do much progress since my previous comments here, so this is mostly guesses based on the change to dom/base/Attr.cpp from https://phabricator.services.mozilla.com/D236107. Mozilla's "change the attribute" is indeed calling a SetAttr method on the element and I see GetTrustedTypesCompliantAttributeValue could:

1. Detach the attribute node: in that the attribute element's is null and probably we should change the spec to do a no-op like in "If attribute's element is null, then set attribute's value to value" above.
2. Destroy the original element: this means implementations need to hold a strong reference to the element or skip the SetAttr call. However, I believe the attribute somehow keeps a strong reference to the element it is attached to, so this is probably covered by 1 already.
3. Move the attribute to another element. Again, I believe this will effectively detached the attribute node from the original element, so covered by 1 already.

dom.bs
fred-wang: Just for the record, this is https://phabricator.services.mozilla.com/D236107?id=977466#inline-1311431 and I agree with @smaug---- this is a problem in my current patch. I also started to add test for this and other cases in https://phabricator.services.mozilla.com/D236161#change-XQGcauHgW6CO

dom.bs
fred-wang: Sorry, not a "no-op" I mean when the attribute's element is null then Mozilla does not need to call Element:SetAttr, just to set Attr::mValue directly. So probably we don't need a spec change here, but implementations need to properly deal with the different cases mentioned and we need tests for them.

fred-wang: I see Chromium / WebKit just call their element's "set attribute", which takes care of applying TT policy default:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/attr.cc;drc=003b43cb850a2fb81bb482cd95381338b7205933;l=73
https://searchfox.org/wubkat/rev/446d00e44d23eba2d8c1c1c92bf0171472dee385/Source/WebCore/dom/Attr.cpp#97

They both store the element in a strong reference (RefPtr and Member) so it won't get destroyed by the callback. And they will modify the attribute value of that element.

But the spec does not check whether the attribute's element become null. And if it is attached it to a different element, then it will just run the "handle attribute changes" on that new element:
https://dom.spec.whatwg.org/#concept-element-attributes-change
https://dom.spec.whatwg.org/#handle-attribute-changes

otherdaniel:
Hi all, sorry I'm a bit late with feedback. It turns out, what Chromium does here is buggy since we are running structure checks before the Trusted Type checks and then getting inconsistent results when a default policy modifies the attribute that is being checked. Thanks for pointing this out.

My (new) preference would be to always run TT checks before any other steps; here and elsewhere. We mostly already do this. That would provide a strict sequencing of TT before any other DOM checks or DOM manipulations, and should IMHO solve this class of problems entirely. I'd be happier if we had a general rule here, and I'd rather do some more work on our end than risk us overlooking another issue like this.

For testing, I'd propose to add WPT tests with competing error conditions and checking for the correct exception being thrown. (I.e., setting an attribute node that's already connected should throw the TT-style TypeException, not the InUseAttributeError DOMException). Something like the following seems to work when I try it locally:

```js
<div onclick="1+1;" id="test"></div>
<script>
test(t => {
  var attribute = document.getElementById("test").getAttributeNode("id");
  assert_throws_dom("InUseAttributeError", _ => {
    document.body.setAttributeNode(attribute);
  }, "Expect InUseAttributeError DOMException.");

  attribute = document.getElementById("test").getAttributeNode("onclick");
  assert_throws_js(TypeError, _ => {
    document.body.setAttributeNode(attribute);
  }, "Expect a Trusted-Types style TypeError, not InUseAttributeError DOMException.");
}, "Element.setAttribute competing exceptions");
</script>
```

fred-wang:
Just to be more explicit, I believe we have three categories of APIs:

- Element.setAttribute/Element.setAttributeNS: It seems browsers agree to always successfully set the attribute in any case. If the attribute node is moved to another element in the default's policy callback, we end up with the attribute set on two different elements.

- Element.setAttributeNode, Element.setAttributeNodeNS, NamedNodeMap.setNamedItem, NamedNodeMap.setNamedItemNS: If the attribute node is moved to another element in the default's policy callback, WebKit and Chromium may end up associating the attribute node to two different element's attribute maps. My proposal is to throw an InUseAttributeError error if the attribute node was moved to another element (which is what would be happening if we re-run https://dom.spec.whatwg.org/#concept-element-attributes-set without TT check).

- Attr.value, Attr.nodeValue, Attr.textContent: These APIs essentially operate on an attribute. If the attribute's element is non-null, then WebKit and Chromium just calls setAttributeNS on that element and no new issue arises. The spec instead performs the TT check before setting the element's attribute but does not say what to do when the attribute's element changes (i.e. attribute is detached from the element or moved to another element). My proposal is again to just re-run current https://dom.spec.whatwg.org/#set-an-existing-attribute-value without TT check. This would differ from WebKit/Chromium's current implementation.

I updated https://phabricator.services.mozilla.com/D236161 to match this proposal ; and https://phabricator.services.mozilla.com/D236107 makes all these new tests pass. I'm happy to change to something else depending on what we decide at the end.

otherdaniel:
The solution outlined here is correct & self-consistent, and if all implementations agree I'll be happy.

I still think we can be simpler, though: The default policy introduces a seemingly concurrent operation. All operations, on all three categories of APIs outlined above, can be explained by serializing these operations.

Consider setAttributeNode:

As spec'ed here: First check structure; then run TT check; then re-check structure.
    - TT error => throws TypeException
    - structure error but no TT error => DOMException AttributeInUseError
    - structure error created in TT callback => DOMException AttributeInUseError
    - TT error *and* structure error =>TypeException
 
If instead, we spec it like so:
1. Let verifiedValue be value.
2. If verify is true: [... TT check; assign to verifiedValue ...]
3. If attr’s element is neither null nor element, throw an "InUseAttributeError" DOMException
4. [... old spec, but verifiedValue instead of value ...]

I think this would get us all of the same error conditions, but fewer operations. And IMHO easier to understand. And I like it, because explaining this behaviour in terms of serializing the checks and running the TT check gives us a guideline which we can apply to all similar cases.

There is one observable difference I can think of: If we have a structure error, and if the TT policy itself throws an exception, then in the first case we'd get AttributeInUseError, and in the second we'd get TypeException. I think that's okay.

annevk:
Any updates here? I rather like @otherdaniel's suggestion.

lukewarlow:
If others are happy with that idea I'm happy to look into updating this PR to match.

smaug----:
I think that would work for setNameItem* and setAttributeNode*.
Current code in Gecko does "If attr’s element is neither null nor element" before and after, and maybe doing it only after is good enough. right @fred-wang ?

Attr.value/.nodeValue/.textContent should do validation (if there is element) too before modifying the value.
(chromium and webkit seem to do something else, assuming I'm reading the code correctly)

Element.setAttribute/Element.setAttributeNS  in the PR needs some tweaking too. Implementations don't necessarily create Attr nodes when those APIs are used, but if they do, what happens if TT callback moves the Attr away. As fred-wang mentioned earlier, implementations already just create a new attribute on the element.

dom.bs
lukewarlow: Having discussed this with Fred we've got an idea on what to do here. The spec will say that if the element is changed by the TT callback that we early return (perhaps throw in which case we need to decide an exception).

dom.bs
```diff
@@ -6589,6 +6591,14 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
+<p>To <dfn>verify attribute value</dfn> {{TrustedType}} or string <var>value</var> for an
+<a>attribute</a> <var>attribute</var>, given an <a for=/>Element</a> <var>element</var>:
+
+<ol>
+ <li><p>Return the result of calling <a abstract-op>get Trusted Types-compliant attribute value</a>
+ for <var>attribute</var>, with <var>element</var>, <var>value</var>. [[!TRUSTED-TYPES]]
```
annevk, line 6599:
```suggestion
 for <var>attribute</var>, with <var>element</var> and <var>value</var>. [[!TRUSTED-TYPES]]
```
Although maybe we can use the more standard "given attribute, element, and value"?

dom.bs
```diff
@@ -6641,6 +6651,9 @@ string <var>namespace</var> (default null):</p>
 <a for=/>attribute</a> <var>attr</var> and an <a for=/>element</a> <var>element</var>:
 
 <ol>
+ <li><p>Let <var>verifiedValue</var> be the result of calling <a>verify attribute value</a>
+ <var>attr</var>'s <a for=Attr>value</a> for <var>attr</var>, with <var>element</var>.
```
annevk, line 6655:
```suggestion
 <li><p>Let <var>verifiedValue</var> be the result of <a>verifying attribute value</a> given
 <var>attr</var>'s <a for=Attr>value</a>, <var>attr</var>, and <var>element</var>.
```
However, this doesn't match the signature above. What gives?

dom.bs
```diff
@@ -6656,29 +6669,53 @@ string <var>namespace</var> (default null):</p>
...
  <a lt="get an attribute by namespace and local name">getting an attribute</a> given
  <var>namespace</var>, <var>localName</var>, and <var>element</var>.
 
- <li>If <var>attribute</var> is null, create an <a>attribute</a> whose <a for=Attr>namespace</a> is
- <var>namespace</var>, <a for=Attr>namespace prefix</a> is <var>prefix</var>,
- <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is <var>value</var>, and
- <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node document</a>, then
- <a lt="append an attribute">append</a> this <a>attribute</a> to <var>element</var>, and then
- return.
+ <li><p>Let <var>attributeExists</var> be false if <var>attribute</var> is null, and true otherwise.
```
annevk, line 6690:
```suggestion
 <li><p>Let <var>attributeExists</var> be false if <var>attribute</var> is null; otherwise true.
```

dom.bs
```diff
@@ -6656,29 +6669,53 @@ string <var>namespace</var> (default null):</p>
...
+
+ <li><p>If <var>attributeExists</var> is false, set <var>attribute</var> to an <a>attribute</a>
+ whose <a for=Attr>namespace</a> is <var>namespace</var>, <a for=Attr>namespace prefix</a> is
+ <var>prefix</var>, <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is
+ <var>value</var>, and <a for=Node>node document</a> is <var>element</var>'s <a for=Node>node
+ document</a>.
+
+ <li><p>Let <var>verifiedValue</var> be <var>value</var>.
+
+ <li><p>If <var>verify</var> is true:
```
annevk, line 6700: `<p>` on its own line if the `<li>` has multiple children.

dom.bs
```diff
@@ -6656,29 +6669,53 @@ string <var>namespace</var> (default null):</p>
...
+   <li><p>Set <var>attributeExists</var> to true if <var>element</var> <a lt="has an attribute">has
+   an attribute</a> <var>attribute</var>; otherwise false.
+  </ol>
+ </li>
 
- <li><p><a lt="change an attribute">Change</a> <var>attribute</var> to <var>value</var>.
+ <li><p>If <var>attributeExists</var> is true, <a lt="change an attribute">change</a>
+ <var>attribute</var> to <var>verifiedValue</var>.
+
+ <li><p>Otherwise:
```
annevk, line 6713: `<p>` on its own line as `<li>` has multiple children.

dom.bs
```diff
@@ -6656,29 +6669,53 @@ string <var>namespace</var> (default null):</p>
...
+ document</a>.
+
+ <li><p>Let <var>verifiedValue</var> be <var>value</var>.
+
+ <li><p>If <var>verify</var> is true:
+  <ol>
+   <li><p>Set <var>verifiedValue</var> to the result of calling <a>verify attribute value</a>
+   <var>value</var> for <var>attribute</var>, with <var>element</var>.
+
+   <li><p>Set <var>attributeExists</var> to true if <var>element</var> <a lt="has an attribute">has
```
annevk, line 6705: No newlines in phrasing level elements.

dom.bs
```diff
@@ -6939,12 +6976,27 @@ method steps are:
  and null otherwise.
  <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
 
- <li><p>If <var>attribute</var> is null, create an <a>attribute</a> whose
- <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
- <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>,
- then <a lt="append an attribute">append</a> this <a>attribute</a> to <a>this</a>, and then return.
+ <li><p>Let <var>attributeExists</var> be false if <var>attribute</var> is null, and true otherwise.
```
annevk, line 6979:
```suggestion
 <li><p>Let <var>attributeExists</var> be false if <var>attribute</var> is null; otherwise true.
```

dom.bs
```diff
@@ -7514,7 +7566,13 @@ string <var>value</var>, run these steps:
  <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
  <a for=Attr>value</a> to <var>value</var>.
 
- <li><p>Otherwise, <a lt="change an attribute">change</a> <var>attribute</var> to <var>value</var>.
+ <li><p>Otherwise:
+  <ol>
+    <li><p>Let <var>verifiedValue</var> be the result of calling <a>verify attribute value</a>
```
annevk, line 7571: Single-space indentation.

annevk (commented):
I did a quick editorial review. Please double check the comments throughout as some apply several times.

dom.bs
lukewarlow: I think I've addressed all of these, will check the github diff to make sure

dom.bs
lukewarlow: I've updated this and the signature to hopefully be clearer. I think it does match the signature it was just written rather poorly. I've changed the signature to a more convetional one and updated the call sites accordingly.

dom.bs
lukewarlow: Apologies Idk if this specfmt got lost by a rebase but I think things have been corrected by my recent change.

dom.bs
lukewarlow: This is fixed and has a test, per Daniels' suggestion and my recent spec change this should be even cleaner. It will require a test change to verify the order of exceptions (the TT TypeError / default policy exception is now the first exception thrown)

dom.bs
```diff
@@ -6656,29 +6669,57 @@ string <var>namespace</var> (default null):</p>
...
  <li><p>Return <var>oldAttr</var>.
 </ol>
 </div>
 
 <div algorithm>
-<p>To <dfn export id=concept-element-attributes-set-value>set an attribute value</dfn> given an
-<a for=/>element</a> <var>element</var>, a string <var>localName</var>, a string <var>value</var>,
-an optional null or string <var>prefix</var> (default null), and an optional null or string
-<var>namespace</var> (default null):
+<p>To <dfn export id=concept-element-attributes-set-value>set an attribute value</dfn> given an <a
```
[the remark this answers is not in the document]
domfarolino: This was probably done by `specfmt`, as it basically follows the HTML wrapping rules where this is allowed. Would you prefer we not use it on non-HTML specs for now? I recall you being sympathetic to allowing this kind of wrapping outside of HTML, but maybe you'd prefer to do that only if we update the entire spec to that model.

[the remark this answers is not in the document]
annevk: I think I said that I only want it to be used once we know it can be used in automation and then I want it enforced by automation. Perhaps I've been inconsistent though.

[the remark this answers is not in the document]
lukewarlow: I'll stop using it for Dom and fix these up manually then.

dom.bs
```diff
@@ -7514,7 +7572,23 @@ string <var>value</var>, run these steps:
...
+  <p>Otherwise:
+
+  <ol>
+   <li><p>Let <var>originalElement</var> be <var>attribute</var>'s <a for=Attr>element</a>.
+
+   <li><p>Let <var>verifiedValue</var> be the result of calling <a>verify attribute value</a>
+   given <var>value</var>, <var>attribute</var>, and <a>this</a>.
+
+   <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
+   <a for=Attr>value</a> to <var>value</var>, and return.
```
otherdaniel, line 7585: Should this set Attr/value to `verifiedValue` (instead of `value`)?

WPT trusted-types/set-attributes-mutations-in-callback.tentative.html seems to expect verifiedValue here, and I get a number test failures if I implement this as written.

lukewarlow: Cc @fred-wang can you remember what we discussed. I seem to recall that maybe the tests for this specific function still need some work?

verifiedValue would make sense on the surface but there's also maybe a reason why it doesn't.

lukewarlow: (Apologies for the delay)

I just looked into this and I think it should indeed be verifiedValue. I'll update the spec PR now. Just as an fyi that test specifically is out of date with the spec currently so you'll probably still gets some failures.

dom.bs
```diff
@@ -6939,12 +6982,30 @@ method steps are:
...
+ <li><p>If <var>attributeExists</var> is false, then set <var>attribute</var> to an <a>attribute</a>
+ whose <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
+ <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>.
+
+ <li><p>Let <var>verifiedValue</var> be the result of
+ <a lt="verify attribute value">verifying an attribute value</a> given <var>value</var>,
+ <var>attribute</var>, and <a>this</a>.
+
+ <li><p>Set <var>attributeExists</var> to true if <a>this</a>
+ <a lt="has an attribute">has an attribute</a> <var>attribute</var>; otherwise false.
```
smaug----, line 6996: As I said on Matrix, I wish we wouldn't need to do identity checks on attributes. That would be a bit tricky, or at least slow in Gecko, and I think also in other engines.

lukewarlow: per matrix discussion I've updated this and the setAttributeNS equivalent.

dom.bs
```diff
@@ -6667,18 +6683,32 @@ an optional null or string <var>prefix</var> (default null), and an optional nul
...
- return.
+ <li><p>If <var>attribute</var> is null, then set attribute to an <a>attribute</a> whose
+ <a for=Attr>namespace</a> is <var>namespace</var>, <a for=Attr>namespace prefix</a> is
+ <var>prefix</var>, <a for=Attr>local name</a> is <var>localName</var>, <a for=Attr>value</a> is
+ <var>value</var>, and <a for=Node>node document</a> is <var>element</var>'s
+ <a for=Node>node document</a>.
+
+ <li><p>Let <var>verifiedValue</var> be the result of
+ <a lt="verify attribute value">verifying an attribute value</a> given <var>value</var>,
+ <var>attribute</var>, and <var>element</var>.
```
smaug----, line 6698: Hmm, why do we need this algorithm to verify the value? I'd expect the relevant webidl attribute setters to be modified, similarly to set HTMLScriptElement.src.

In fact, HTMLIFrameElement.srcdoc setter already does TT check to get a reasonable string and this would do it again? Or am I missing something.

lukewarlow: But this is for setAttributeNS()? Does this also get used by property setters?

lukewarlow: Ah yeah you're right this is shared by element reflection...

So yeah this will need moving out to the method algorithm. I think in this case it's probably different enough to just duplicate it a bit.

dom.bs
```diff
@@ -6946,6 +6948,14 @@ steps:
  <a for=Attr>value</a>.
 </ol>
 
+<p>To <dfn>verify attribute value</dfn> given a {{TrustedType}} or string <var>value</var>, an
```
annevk, line 6951:
```suggestion
<p>To <dfn>verify an attribute value</dfn> given a {{TrustedType}} or string <var>value</var>, an
```
Although why do we even have this algorithm as it directly calls into the Trusted Types algorithm.

dom.bs
```diff
@@ -7294,14 +7310,26 @@ method steps are:
  <li><p>Let <var>attribute</var> be the first <a>attribute</a> in <a>this</a>'s
  <a for=Element>attribute list</a> whose <a for=Attr>qualified name</a> is <var>qualifiedName</var>,
  and null otherwise.
- <!-- This is step 2 of "get an attribute by name", modified as appropriate -->
+
+ <li><p>If <var>attribute</var> is null, then set <var>attribute</var> to an <a>attribute</a>
+ whose <a for=Attr>local name</a> is <var>qualifiedName</var>, <a for=Attr>value</a> is
+ <var>value</var>, and <a for=Node>node document</a> is <a>this</a>'s <a for=Node>node document</a>.
```
annevk, line 7316: Is the idea here to create a dummy attribute that you don't append in order to calculate the verified value?

Why is that done this way? This is quite confusing to read through.

annevk (commented):
I would love to have a more detailed description of what we are trying to accomplish here. I don't think the resulting algorithms are very clear. Have you looked at whether it's possible to abstract certain bits?

dom.bs
lukewarlow: Yeah dropping it and calling TT directly makes sense.

dom.bs
```diff
@@ -7013,6 +7020,8 @@ string <var>namespace</var> (default null):</p>
 
  <li><p>Otherwise, <a lt="append an attribute">append</a> <var>attr</var> to <var>element</var>.
 
+ <li><p>Set <var>attr</var>'s <a for=Attr>value</a> to <var>verifiedValue</var>.
```
lukewarlow: Need to check if this is in the right place in the algorithm.

cc @fred-wang

dom.bs
lukewarlow: Apologies for the slow reply on here I thought I replied before but it must have got lost.

TLDR: I agree with you I can't really remember why it's done this way. I updated the TT spec to just take the local name and namespace directly rather than an attribute object. As part of doing this change I realised these algorithms were overly complicated for what was actually needed.

Hopefully now the integration points should be much cleaner and less invasive.

lukewarlow:
@otherdaniel I believe this spec PR fully matches your proposals to call TT as early as possible now. If you've got time I'd be thankful for a review to get another pair of eyes on this.

dom.bs
smaug----: (1) So in case oldAttr == attr, its value isn't updated.


(2) 'append an attribute' calls 'Handle attribute changes', and verifiedValue is set to attr after that. That doesn't seem right. Shouldn't value be set before?

lukewarlow: 1) So in case oldAttr == attr, its value isn't updated.

Yes, I think that makes sense? If the attribute hasn't changed does it make sense to put it through Trusted Types default policy? I'll double check what browsers actually do in this case though.

2) 'append an attribute' calls 'Handle attribute changes', and verifiedValue is set to attr after that. That doesn't seem right. Shouldn't value be set before?

Ah yeah, you're right, step 7 should be moved to step 5.

I've fixed 2 now.

lukewarlow: So I double checked and for point 1 there's actually an interop bug here, WebKit (and Gecko) doesn't throw a TT error. Chromium currently does, so we can choose whichever behaviour we think is best here.

cc @annevk  and @otherdaniel for your opinions

TLDR should doing

`element.setAttributeNode(element.attributes[1])` - throw a TT error when TT is enforced (and where element.attributes[1] is something like an onclick attribute?

annevk: I think we either have to throw or do nothing at all, right? The latter is very weird as even no-op mutations result in a mutation record being queued. And queuing a mutation record for trusted type attributes without validation having occurred seems bad, so I think we have to throw.

dom.bs
```diff
@@ -7891,7 +7908,25 @@ string <var>value</var>, run these steps:
  <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
  <a for=Attr>value</a> to <var>value</var>.
 
- <li><p>Otherwise, <a lt="change an attribute">change</a> <var>attribute</var> to <var>value</var>.
+ <li>
+  <p>Otherwise:
+
+  <ol>
+   <li><p>Let <var>originalElement</var> be <var>attribute</var>'s <a for=Attr>element</a>.
```
```suggestion
   <li><p>Let <var>element</var> be <var>attribute</var>'s <a for=Attr>element</a>.
```

dom.bs
```diff
@@ -7891,7 +7908,25 @@ string <var>value</var>, run these steps:
...
+
+   <li><p>Let <var>verifiedValue</var> be the result of calling <a abstract-op>get
+   Trusted Types-compliant attribute value</a> with <var>attribute</var>'s
+   <a for=Attr>local name</a>, <var>attribute</var>'s <a for=Attr>namespace</a>, <a>this</a>,
```
annevk: There is no this here.

dom.bs
```diff
@@ -7891,7 +7908,25 @@ string <var>value</var>, run these steps:
...
+   and <var>value</var>. [[!TRUSTED-TYPES]]
+
+   <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
+   <a for=Attr>value</a> to <var>verifiedValue</var>, and return.
+
+   <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is not <var>originalElement</var>, then
+   return.
```
annevk: I think it would be nicer if we throw for both of these cases as they result from the policy changing aspects of the attribute that it probably shouldn't? But maybe that's not compatible with what Chromium does.

Also, the other changes are made directly in `setAttribute` and `setAttributeNS`. Do we end up invoking the policy early enough here for the various call sites? It seems like there would be a difference in exception order.

annevk (commented):
This is indeed a whole bit more straightforward to read through.

dom.bs
lukewarlow: cc @fred-wang I know we had a discussion about whether this should throw exceptions or not, but it's been so long I cannot remember what the decisions were.

Also cc @otherdaniel for if you have opinions on the throwing vs not throwing.

lukewarlow: Regarding does this get called early enough.

"set an existing attribute value" is the first step invoked by Attr's value setter steps, Attr's textContent setter steps, and Attr's nodeValue setter steps.

NamedNodeMap's setNamedItem and setNamedItemNS, along with Element's setAttributeNode and setAttributeNodeNS both call "set an attribute" as their first steps.

Then setAttribute and setAttributeNs both have their own dedicated steps.

dom.bs
```diff
@@ -7115,6 +7117,11 @@ string <var>namespace</var> (default null):</p>
 <a for=/>attribute</a> <var>attr</var> and an <a for=/>element</a> <var>element</var>:
 
 <ol>
+ <li><p>Let <var>verifiedValue</var> be the result of calling <a abstract-op>get
+ Trusted Types-compliant attribute value</a> with <var>attr</var>'s <a for=Attr>local name</a>,
```
annevk: I don't understand why this is an abstract-op. Generally outside of TC39 we should not be using that.

The wrapping is also wrong for DOM (applies several times). There's no wrapping inside of phrasing-level elements to make searching easier.

annevk (commented):
@fred-wang @otherdaniel if you could reply to @lukewarlow's comment above that would help a lot, thanks! @smaug---- did you want to have another look as well?

dom.bs
```diff
@@ -7436,8 +7449,12 @@ method steps are:
  [=validate and extract|validating and extracting=] <var>namespace</var> and
  <var>qualifiedName</var> given "<code>element</code>".
 
- <li><p><a>Set an attribute value</a> for <a>this</a> using <var>localName</var>, <var>value</var>,
- and also <var>prefix</var> and <var>namespace</var>.
+ <li><p>Let <var>verifiedValue</var> be the result of calling <a abstract-op>get
+ Trusted Types-compliant attribute value</a> with <var>localName</var>, <var>namespace</var>,
+ <var>element</var>, and <var>value</var>. [[!TRUSTED-TYPES]]
```
tete17:
```suggestion
 <var>this</var>, and <var>value</var>. [[!TRUSTED-TYPES]]
```

annevk: This should use this indeed, but not as a variable.

dom.bs
lukewarlow: Idk why it's an abstract-op either possibly how it always was or maybe I mistakenly changed it.

Trusted Types spec PR to drop it https://github.com/w3c/trusted-types/pull/602 I've pushed the change to this spec PR too (might not build till the TT Pr is merged)

dom.bs
```diff
@@ -7115,6 +7117,11 @@ string <var>namespace</var> (default null):</p>
 <a for=/>attribute</a> <var>attr</var> and an <a for=/>element</a> <var>element</var>:
 
 <ol>
+ <li><p>Let <var>verifiedValue</var> be the result of calling
+ <a>get trusted type compliant attribute value</a> with <var>attr</var>'s
+ <a for=Attr>local name</a>, <var>attr</var>'s <a for=Attr>namespace</a>, <var>element</var>, and
+ <var>attr</var>'s <a for=Attr>value</a>. [[!TRUSTED-TYPES]]
+
```
smaug----, line 7124: This is a bit nitpicking, but I think I'd prefer still that passing existing attribute node here would be no-op.
Gecko does attr.element == some other element -> throw, attr == existingAttr return; also before TT check.

dom.bs
```diff
@@ -8028,7 +8045,25 @@ string <var>value</var>, run these steps:
...
+   <li><p>Let <var>verifiedValue</var> be the result of calling
+   <a>get trusted type compliant attribute value</a> with <var>attribute</var>'s
+   <a for=Attr>local name</a>, <var>attribute</var>'s <a for=Attr>namespace</a>,
+   <var>element</var>, and <var>value</var>. [[!TRUSTED-TYPES]]
+
+   <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is null, then set <var>attribute</var>'s
+   <a for=Attr>value</a> to <var>verifiedValue</var>, and return.
+
+   <li><p>If <var>attribute</var>'s <a for=Attr>element</a> is not <var>element</var>, then
+   return.
```
smaug----: Gecko doesn't have this check. Hmm, and actually, the API becomes a bit surprising with this check. One is trying to change the value, but nothing happens. No exception, and value isn't changed.

smaug----: So, I would drop this check. no-op isn't good, and if TT callback has moved the attribute, then it has. We need to trust TT callbacks in that sense.

dom.bs
smaug----: ...but I can also live with this proposed algorithm.

dom.bs
lukewarlow: I've gone ahead and removed that early return. After our discussions. https://github.com/web-platform-tests/wpt/pull/55762 - test updated here.

dom.bs
lukewarlow: After our discussion I'm going to keep it how it is, tests to ensure Firefox behaviour fails is added here https://github.com/web-platform-tests/wpt/pull/55762

dom.bs
lukewarlow: Regarding throwing we've actually decided to remove the early return here.

dom.bs
lukewarlow: cc @otherdaniel just as an fyi there is a minor change to make to the Attr.nodeValue case.

fred-wang:
:tada:

annevk:
Indeed, thanks a lot @lukewarlow for finishing this and @smaug---- for his careful reviews! And of course everyone else that helped improve Trusted Types integration over the years. 🎉

PR #1279, opened 2024-04-18
Editorial: explain the possible primary interfaces for nodes

domenic:
I was trying to check whether a switch statement was exhaustive and had to do a bit of mental juggling to get there. I thought I'd write out the results for future reference.

***
<a href="https://whatpr.org/dom/1279.html" title="Last updated on Apr 22, 2024, 2:55 AM UTC (9fae4af)">Preview</a> | <a href="https://whatpr.org/dom/1279/77cd75e...9fae4af.html" title="Last updated on Apr 22, 2024, 2:55 AM UTC (9fae4af)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2252,6 +2252,10 @@ can be used to explore this matter in more detail.
  {{ProcessingInstruction}}, or {{Comment}}.
 
  <p>Objects that implement {{Text}} sometimes implement {{CDATASection}}.
+
+ <p>Thus, every [=node=]'s [=primary interface=] is one of: {{Attr}}, {{CDATASection}}, {{Comment}},
+ {{Document}}, {{DocumentFragment}}, {{DocumentType}}, {{Element}} or an inherited interface of it,
+ {{ProcessingInstruction}}, {{ShadowRoot}}, or {{Text}}.
 </div>
 
 <p>For brevity, this specification refers to an object that <a>implements</a> {{Node}} and an
--- a/dom.bs
+++ b/dom.bs
@@ -2253,9 +2253,9 @@ can be used to explore this matter in more detail.
 
  <p>Objects that implement {{Text}} sometimes implement {{CDATASection}}.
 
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -2252,6 +2252,10 @@ can be used to explore this matter in more detail.
  {{ProcessingInstruction}}, or {{Comment}}.
 
  <p>Objects that implement {{Text}} sometimes implement {{CDATASection}}.
+
+ <p>Thus, every [=node=]'s [=primary interface=] is one of: {{Attr}}, {{CDATASection}}, {{Comment}},
+ {{Document}}, {{DocumentFragment}}, {{DocumentType}}, {{Element}} or an inherited interface of it,
```
annevk, line 2257:
```suggestion
 {{Document}}, {{DocumentFragment}}, {{DocumentType}}, or {{Element}}; or an inherited interface of it:
```

annevk: Although I'm not sure how this is correct. Why is ProcessingInstruction an inherited interface and Comment is not?

Also, why would ShadowRoot not be a primary interface? Isn't that the most derived? And I think both Text and CDATASection are primary interfaces as they are the most derived on the objects they occur on. Just when CDATASection is the primary interface, it's also an inherited interface of Text.

domenic: The suggested change seems incorrect, and then your further comment seems to be based on confusion about what the spec text would say after that incorrect change.

I'll rephrase the original in another way. Every node's primary interface is one of:

- Attr
- CDATASection
- Comment
- Document
- DocumentFragment
- DocumentType
- Element
- HTMLElement
- HTMLDivElement
- ...
- ProcessingInstruction
- ShadowRoot
- Text

annevk: I see, that makes sense. Can we put them in the same order as the document outline does? I think I generally try to be consistent about that node order throughout the document.

domenic: Done!

PR #1295, opened 2024-07-15
Mark dependent abort signals as aborted before firing events

### LINKED ISSUE #1293: AbortSignal.any() assertion failure

In Gecko [Bug 1903676](https://bugzilla.mozilla.org/show_bug.cgi?id=1903676), a test case was found where the assertion in [create a dependent abort signal](https://dom.spec.whatwg.org/#create-a-dependent-abort-signal) step 4.2.1. does not hold. I believe this is a spec issue.

Here is the example
```
let a = new AbortController();
let b = AbortSignal.any([a.signal]);
a.signal.addEventListener("abort", () => {
  AbortSignal.any([b]);
  console.log(b.aborted, a.signal.aborted);  // false, true
})
a.abort();
```
During the abort event of `a`, we create a dependent signal from `b`, which is not yet aborted. If `b` is not aborted and dependent on `a`, the spec expects `a` to be not aborted too and asserts this. The problem is that [signal abort](https://dom.spec.whatwg.org/#abortsignal-signal-abort) fires the event before aborting the dependent signals, steps 5 and 6 of signal abort could be switched to resolve this issue.

@shaseley

shaseley:
The assert in 4.2.1 of "create a dependent abort signal" fails when creating a dependent signal while dispatching abort events or running abort algorithms if abort had not yet been propagated to one of the sources.

This fix splits "signal abort" into two phases: first, set the abort reason on the signal being aborted and all of its unaborted dependents; next, run the abort algorithms and dispatch events for the signal and those same dependents. Note that:
 1. Dependent signals do not themselves have dependent signals, which means it's unnecessary to recursively call "signal abort"
 2. This approach retains the existing event dispatch order, while ensuring the abort state is synced before any JS runs

This fixes #1293.

---

   * Chromium
   * WebKit
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at: https://github.com/web-platform-tests/wpt/pull/47653
   * Chromium: https://issues.chromium.org/issues/361129283
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1903676
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=278159
   * Deno (only for aborting and events): https://github.com/denoland/deno/issues/25118
   * Node.js (only for aborting and events): https://github.com/nodejs/node/issues/54466
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A, bug fix

***
<a href="https://whatpr.org/dom/1295.html" title="Last updated on Aug 20, 2024, 8:57 PM UTC (0d05525)">Preview</a> | <a href="https://whatpr.org/dom/1295/7c494e5...0d05525.html" title="Last updated on Aug 20, 2024, 8:57 PM UTC (0d05525)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2033,16 +2033,42 @@ an optional <var>reason</var>:
  <li><p>Set <var>signal</var>'s [=AbortSignal/abort reason=] to <var>reason</var> if it is given;
  otherwise to a new "{{AbortError!!exception}}" {{DOMException}}.
 
+ <li><p>Let <var>dependentSignalsToAbort</var> be a new <a for=/>list</a>.
+
+ <li>
+  <p><a for=set>For each</a> <var>dependentSignal</var> of <var>signal</var>'s
+  [=AbortSignal/dependent signals=]:
+
+  <ol>
+   <li>
+    <p>If <var>dependentSignal</var> is not [=AbortSignal/aborted=], then:
+
+    <ol>
+     <li><p>Set <var>dependentSignal</var>'s [=AbortSignal/abort reason=] to <var>signal</var>'s
+     [=AbortSignal/abort reason=].
+
+     <li><p><a for=list>Append</a> <var>dependentSignal</var> to
+     <var>dependentSignalsToAbort</var>.
+    </ol>
+  </ol>
+
+ <li><p><a>Run the abort steps</a> for <var>signal</var>.
+
+ <li><p><a for=set>For each</a> <var>dependentSignal</var> of <var>dependentSignalsToAbort</var>,
+ <a>run the abort steps</a> for <var>dependentSignal</var>.
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn>run the abort steps</dfn> for an {{AbortSignal}} <var>signal</var>:
+
+<ol>
  <li><p><a for=set>For each</a> <var>algorithm</var> of <var>signal</var>'s
  [=AbortSignal/abort algorithms=]: run <var>algorithm</var>.
 
  <li><p><a for=set>Empty</a> <var>signal</var>'s <a for=AbortSignal>abort algorithms</a>.
 
  <li><p>[=Fire an event=] named {{AbortSignal/abort}} at <var>signal</var>.
-
- <li><p><a for=set>For each</a> <var>dependentSignal</var> of <var>signal</var>'s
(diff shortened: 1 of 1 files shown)


annevk:
I wonder if @domenic / @jakearchibald / @jasnell are willing to review this?

Does this require implementation changes or a test?

jasnell:
Seems fine to me

shaseley:
> Does this require implementation changes or a test?

Yes to both. WIP tests are [here](https://chromium-review.googlesource.com/c/chromium/src/+/5664649/1/third_party/blink/web_tests/external/wpt/dom/abort/resources/abort-signal-any-tests.js#193). I was holding off on landing that and filing bugs until someone had a chance to take a first pass on the PR.

domenic (approved):

annevk:
Since this is fixing a bug, feel free to count WebKit as interested.

@vinhill thoughts?

maraisr:
Just created related WebKit bug issue: https://bugs.webkit.org/show_bug.cgi?id=278159. Will be ready to land if not landed over the weekend.

vinhill:
I'm no longer employed at Mozilla, but they are likely interested due to [Bug 1903676](https://bugzilla.mozilla.org/show_bug.cgi?id=1903676)
@saschanaz

annevk:
@shaseley from my perspective this is good to land, modulo checking the checkboxes in OP.

saschanaz:
LGTM from Mozilla. We can reuse https://bugzilla.mozilla.org/show_bug.cgi?id=1903676 as an implementation bug.

shaseley:
@annevk bugs filed and checkboxes updated!

domfarolino:
I just came across this after looking into the AbortSignal algorithms while working on Observables, and I don't understand this part:

> Dependent signals do not themselves have dependent signals, which means it's unnecessary to recursively call "signal abort"

Why is this true? In Observables, we were using dependent `AbortSignal`s for a while, for each signal that was used to subscribe to upstream Observables, and I believe we were able to create a long chain of `AbortSignal` dependencies such that aborting some initial signal led to the recursive invocation of (old version of) the "**signal abort**" algorithm, on a whole deep chain of signals. Why does this PR say that is not possible?

Note: We are no longer chaining dependent signals like this in Observables, because the timing of aborting parent signals before dependents did not work for us (https://github.com/WICG/observable/pull/154), but still I think it is very possible to do, no?

PR #1306, opened 2024-08-21
Corrected the variable's name from 'element' to 'result'.

pwombwell:
Matches the surroundings - there is no variable 'element'.


annevk:
Thanks!

PR #1307, opened 2024-08-26
Introduce `moveBefore()` state-preserving atomic move API

### LINKED ISSUE #1255: Atomic move operation for element reparenting & reordering

Chrome (@domfarolino, @noamr, @mfreed7) is interested in pursuing the addition of an atomic move primitive in the DOM Standard. This would allow an element to be re-parented or re-ordered _without_ today's side effects of first being removed and then inserted.

Here are all of the prior issues/PRs I could find related to this problem space:
 - https://github.com/whatwg/html/issues/5484
 - https://github.com/whatwg/dom/issues/808
 - https://github.com/whatwg/dom/issues/880
 - https://github.com/whatwg/dom/issues/891
 - https://github.com/whatwg/dom/pull/1185
 - https://github.com/whatwg/dom/pull/732
 - https://github.com/whatwg/html/pull/4354
 - https://github.com/web-platform-tests/wpt/pull/44308
 - https://github.com/web-platform-tests/wpt/pull/15264

**Problem**

Without an atomic move operation, re-parenting or re-ordering elements involves [first removing them](https://dom.spec.whatwg.org/#concept-node-adopt:~:text=node%20document.-,If%20node%E2%80%99s%20parent%20is%20non%2Dnull%2C%20then%20remove%20node,-.) and then re-inserting them. With the DOM Standard's current removal/insertion model, this resets lots of state on various elements, including iframe document state, selection/focus on `<input>`s, and more. See @josepharhar's [reparenting demo](https://reparent.jarhar.com/) for a more exhaustive list of state that gets reset.

This causes lots of developer pain, as recently voiced [on X](https://twitter.com/nomsternom/status/1755275696434286831) by frameworks [like HTMX](https://twitter.com/htmx_org/status/1755280639962591704), and other companies such as Wix, Microsoft, and internally at Google.

This state-resetting is in part caused by the DOM Standard's current insertion & removal model. While [well-defined](https://github.com/whatwg/dom/issues/808#issuecomment-1913887358), its model of [insertion](https://dom.spec.whatwg.org/#ref-for-concept-node-insert-ext) and [removal](https://dom.spec.whatwg.org/#ref-for-concept-node-remove-ext%E2%91%A0) steps has two issues, both captured by https://github.com/whatwg/dom/issues/808:

 1. **Undesirable model**: The current DOM Standard allows for the non-atomic insertion of multiple nodes at a time. In practice, this means when appending e.g., a DocumentFragment, script can run in between each individual child insertion, thus observing DOM state before the entire fragment insertion is complete.
 2. **Interop issues**: While Safari matches the spec, Chromium & Gecko have a model that ensures all DOM mutations are synchronously performed before any script runs as a result of the mutations.

One very limited partial solution that does not actually involve any DOM tree manipulation, is this shadow DOM example that @emilio had posted a while back: https://github.com/whatwg/html/issues/5484#issuecomment-620481794 (see my brief recreation of it below).

Screen Recording 2024-01-29 at 5 00 26 PM

But as mentioned, this does not seem to perform any real DOM mutations; rather, the slot mutation seems to just visually compose the element in the right place. Throughout this example, the iframe's actual [parent](https://dom.spec.whatwg.org/#concept-tree-parent) does not change.

----

Otherwise, we know there is some historical precedent for trying to solve this problem with WebKit's since-rolled-back "magic iframes". See https://github.com/whatwg/html/issues/5484#issuecomment-619642936 and https://bugs.webkit.org/show_bug.cgi?id=13574#c12. We believe that the concerns from that old approach can be ameliorated by:
 - Fixing #808, as to not leave elements in a disconnected state _at all_ during the atomic portion of the move (i.e., the re-parenting/re-ordering)
 - Not allowing atomic moves _across documents_, which should greatly simplify the security story of this work

### How would you solve it?

**Solution**

To lay the groundwork for an atomic move primitive in the DOM Standard, we plan on resolving https://github.com/whatwg/dom/issues/808 by introducing a model desired by @annevk, @domfarolino, @noamr, and @mfreed7, that resembles Gecko & Chromium's model of handling all script-executing insertion/removal side-effects _after_ all DOM mutations are done, for any given insertion.

With this in place, we believe it will be much easier to separate out the cases where we can simply _**skip**_ the invocation of insertion/removal side-effects for nodes that are atomically moved in the DOM. This will make us, and implementers, confident that there won't be any way to observe an inconsistent DOM state while atomically moving an element, or experience other nasty unknown side-effects.

The API shape for this new primitive is an open question. Below are a few ideas:
 - **A new DOM API** like replaceChildAtomic()/replaceChildrenAtomic() that can take a [connected](https://dom.spec.whatwg.org/#connected) node and atomically re-parent it without removal/insertion side-effects.
    - One limitation here is that we'd have to pick and choose which existing DOM APIs we want to mirror with atomic counterparts. For example, if we ever wanted append() or appendChild() to ever be able to _also_ atomically move already-connected nodes, we'd have to introduce appendAtomic() and appendChildAtomic(), and so on.
 - A **setting for existing DOM APIs**, e.g., `append(node, {atomic: true})`, `replaceChild(node, {atomic: true})`
 - A **scoped, declarative attribute** that changes the behavior of DOM mutation APIs in a subtree
    - This could be an element attribute that makes all existing DOM mutation APIs behave "atomically" when operating on already-connected nodes under the element's subtree
    - This could also be a property on the document overall, set via a header/meta tag, or some other mechanism

----

Compatibility issues here take the form relying on insertion/removal side-effects which no longer happen during an atomic move. They vary depending on the shape of our final design.

1. With a new DOM API/setting that developers have to affirmatively opt-into, you could atomically move fragments/subtrees constructed by _other library code that's unaware it's being atomically moved_. Those fragments may be built in a way that relies on non-atomic move side-effects (though we haven't heard of such concerns directly yet).
1. Consider an element attribute that changes the behavior of all DOM mutation APIs to behave atomically on already-connected nodes in its subtree. You could minimize compat concerns by externally-constructed portions of the subtree to opt-_out_ of atomic moves with the same attribute. But what would that mean exactly, to have part of a subtree move atomically and part of it not?

A non-exhaustive list of additional complexities that would be nice to track/discuss before a formal design:
 - How to handle mutation events? There was [discussion at the TPAC 2023](https://github.com/whatwg/meta/issues/284#:~:text=Domenic%3A%20If%20all%20else%20fails%2C%20we%20could%20say%20that%20mutation%20events%20don%27t%20fire%20if%20any%20new%20feature%20is%20used) about suppressing mutation events when new-ish DOM features are used, so we could probably get away with simply suppressing mutation events whenever an atomic move is being performed??
 - Handling things like focus/selection properly (need to land on desired behavior)
 - Fixing up things like [live ranges](https://dom.spec.whatwg.org/#concept-live-range); the way DOM handles this today might already be suitable for atomic moves, but unclear

### Anything else?

_No response_

### LINKED ISSUE #1335: moveBefore() cosmetic concerns
While addressing some more nits in #1307 I realized that `moveBefore()` returning a node is a pattern we generally don't follow for new APIs: https://github.com/w3ctag/design-principles/issues/286.

I'm not sure if that deviates it enough from `insertBefore()` to warrant a new name.

I'm also wondering if it should be `[Unscopable]`. I think we generally agreed to do that for new APIs, but it's been hit or miss as to whether we remember. E.g., `getRootNode()` does not have this. (This is probably why it should have been an inverted flag from the beginning.)

cc @domfarolino @noamr @domenic @smaug----

domfarolino:
This PR introduces a new DOM API on the `ParentNode` mixin: `moveBefore()`. It mirrors `insertBefore()` in shape, but defers to a new DOM manipulation primitive that this PR adds in service of this new API: the "move" primitive. The move primitive contains some of the DOM tree bookkeeping steps from the remove primitive, as well as the insert primitive, and does three more interesting things:
 1. Calls the [moving steps hook](https://whatpr.org/dom/1307.html#concept-moving-steps-ext) with the moved node and the old parent (possibly null, just like the removing steps)
 2. Queues a custom element callback reaction for the `connectedMoveCallback()`
 3. Queues two back-to-back mutation record tasks: one for removal from the old parent; one for insertion into the new one

The power of the move primitive comes from the fact that the algorithm does not defer to the traditional insert and removal primitives, and therefore does not invoke the [removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext) and [insertion steps](https://dom.spec.whatwg.org/#concept-node-insert-ext). This allows most state to be preserved by default (i.e., we don't tear down iframes, or close dialogs). Sometimes, the insertion/removing step overrides in other specifications have steps that do need to be performed during a move anyways. These specifications are expected to override the moving steps hook and perform the necessary work accordingly. See https://github.com/whatwg/html/pull/10657 for HTML.

Closes #1255. Closes #1335.

Remaining tasks (some will be PRs in other standards):

- [x] Custom element integration
- [x] Keep popovers open
- [x] Don't call post-connection steps if state-preserving atomic move is in progress
- [x] Don't call becomes connected / becomes browsing-context
- [x] Only disconnect subframes on removal when state-preserving atomic move is **not** in progress
- [x] Keep dialogs open: see [removing steps](https://html.spec.whatwg.org/C#the-dialog-element:html-element-removing-steps)
- [x] img/source: this shouldn't count as a [relevant mutation](https://html.spec.whatwg.org/C#relevant-mutations)
- [x] Preserve fullscreen 
- [x] Preserve focus 
    - Need to resolve `focusin` event semantics
- ~[ ] Don't reset animations / transitions. See [here](https://drafts.csswg.org/css-transitions/#starting) 
    - Maybe nothing needs to be done here. Given [how element removals are handled](https://drafts.csswg.org/css-transitions/#transition-cancel:~:text=If%20an%20element%20is%20no%20longer%20in%20the%20document%2C%20implementations%20must%20cancel%20any%20running%20transitions%20on%20it%20and%20remove%20transitions%20on%20it%20from%20the%20completed%20transitions), the spec does NOT require transitions to be removed from the UA's set of _running transitions_ for moved nodes since they are never removed from the Document.~
- ~[ ] Preserve text-selection. See [set the selection range](https://html.spec.whatwg.org/C#set-the-selection-range). Edit: Nothing needs to be done here. Selection metadata (i.e., `selectionStart` and kin) is preserved by default in browsers, consistent with HTML (no action is taken on removal). The UI behavior of the selection not being highlighted is a side-effect of the element losing focus~
- [ ] Selection API: don't reset the Document's [selection](https://w3c.github.io/selection-api/#dfn-selection)
    - Updates to the selection range should happen according to how the DOM Standard primitives update ranges. The Selection API specification admits as much, by [deferring to the](https://w3c.github.io/selection-api/#responding-to-dom-mutations) insert and removal algorithms. Therefore, we should reference the move primitive from the Selection API specification, and ensure that the move primitive in _this DOM Standard PR_ updates live ranges correctly: https://github.com/w3c/selection-api/pull/341
    - `selectionchange` event: We've decided to allow `selectionchange` event to still fire, [since it is queued in a task](https://w3c.github.io/selection-api/#ref-for-dfn-schedule-a-selectionchange-event-1). No changes for this part are required.
- [x] Pointer event state reset: see [here](https://w3c.github.io/pointerevents/#suppressing-a-pointer-event-stream)
    - The specification does not reset an element's previously-set pointer capture upon DOM node insertion or removal. While this seems like a gap in that specification (https://github.com/w3c/pointerevents/issues/526), it appears that nothing needs to be done in that specification to satisfy this preservation semantic.
    - Test: https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/moveBefore/tentative/pointer-events.html
- [x] Hide input/select picker: [here](https://html.spec.whatwg.org/C#show-the-picker%2C-if-applicable)
    - Nothing in the specification seems to hook into DOM for removal when the picker dialog is shown, so no work needs to be done in this PR. See https://github.com/whatwg/html/issues/10743 for a follow-up.
- [x] Preserve pointer lock: [here](https://w3c.github.io/pointerlock/#requestPointerLock)
    - When a ["pointer-lock target becomes disconnected"](https://w3c.github.io/pointerlock/#ref-for-dfn-pointer-lock-target-12), pointer lock is exited. Since the "becomes disconnected" hook is not run during an atomic move, pointer lock is preserved naturally.
- [ ] Containment: keep last remembered size(see [here](https://drafts.csswg.org/css-sizing-4/#last-remembered))

   * Chromium: https://issues.chromium.org/issues/40150299
   * WebKit: https://github.com/WebKit/standards-positions/issues/375
   * Gecko: https://github.com/mozilla/standards-positions/issues/1053
   * https://github.com/web-platform-tests/wpt/tree/master/dom/nodes/moveBefore/tentative (need to mark as non-tentative)
   * Chromium: https://issues.chromium.org/issues/40150299
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1923880
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=281223
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/mdn/issues/613 & the `impacts-documentation` label

***
<a href="https://whatpr.org/dom/1307.html" title="Last updated on Feb 27, 2025, 4:30 PM UTC (bfbcfd9)">Preview</a> | <a href="https://whatpr.org/dom/1307/b64559c...bfbcfd9.html" title="Last updated on Feb 27, 2025, 4:30 PM UTC (bfbcfd9)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2886,13 +2886,13 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run 
 algorithm is passed a <a for=/>node</a> <var ignore>movedNode</var>, and a <a for=/>node</a>-or-null
 <var ignore>oldParent</var> as indicated in the <a for=/>move</a> algorithm below. Like the
 <a>insertion steps</a>, these steps must not modify the <a>node tree</a> that
-<var>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>,
+<var>movedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>,
 <a lt="fire an event">fire events</a>, or otherwise execute JavaScript. These steps may queue tasks
 to do these things asynchronously, however.
 
 
 <p>To <dfn>move</dfn> a <a for=/>node</a> <var>node</var> into a <a for=/>node</a>
-<var>newParent</var> before a <a for=/>node</a> <var>child</var>:
+<var>newParent</var> before a <a for=/>node</a>-or-null <var>child</var>:
 
 <ol>
  <!-- Start pre-move validity checks -->
@@ -2945,8 +2945,8 @@ to do these things asynchronously, however.
  <var>node</var>'s <a>assigned slot</a>.
 
  <li><p>If <var>oldParent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
- <var>oldParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> is the empty list, then
- run <a>signal a slot change</a> for <var>oldParent</var>.
+ <var>oldParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a>
+ <a for=list>is empty</a>, then run <a>signal a slot change</a> for <var>oldParent</var>.
 
  <li>
   <p>If <var>node</var> has an <a>inclusive descendant</a> that is a <a>slot</a>:
@@ -2956,7 +2956,6 @@ to do these things asynchronously, however.
 
    <li><p>Run <a>assign slottables for a tree</a> with <var>node</var>.
   </ol>
- </li>
 
  <!-- Start insertion-related bookkeeping steps -->
  <li>
@@ -2986,8 +2985,8 @@ to do these things asynchronously, however.
  <a>slottable</a>, then <a>assign a slot</a> for <var>node</var>.
 
  <li><p>If <var>newParent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
- <var>newParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> is the empty list,
- then run <a>signal a slot change</a> for <var>newParent</var>.
+ <var>newParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a>
+ <a for=list>is empty</a>, then run <a>signal a slot change</a> for <var>newParent</var>.
 
  <li><p>Run <a>assign slottables for a tree</a> with <var>node</var>'s <a for=tree>root</a>.
 
@@ -3364,10 +3363,11 @@ Element includes ParentNode;
  <dd>
   <p>Moves, without first removing, <var>movedNode</var> into <var>node</var> after <var>child</var>
   if <var>child</var> is non-null; otherwise after the <a>last child</a> of <var>node</var>. This
-  methods preserves state associated with <var>movedNode</var> so that it persists after the move.
+  method preserves state associated with <var>movedNode</var>.
 
   <p><a>Throws</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}} if the constraints of
-  the <a>node tree</a> are violated.
+  the <a>node tree</a> are violated, or the state associated with the moved node cannot be
+  preserved.
   <!-- "NotFoundError" is impossible -->
 
  <dt><code><var>node</var> . <a method for=ParentNode lt="querySelector()">querySelector</a>(<var>selectors</var>)</code>


dom.bs
```diff
@@ -4986,6 +5033,27 @@ within <a>this</a>.
...
+ <li><p>Let <var>perform state-preserving atomic move</var> be true if <a>this</a> is
+ <a>connected</a>, <var>node</var> is <a>connected</a>, <a>this</a>'s <a for=Node>node
+ document</a> is <a>fully active</a>, and <a>this</a>'s <a for=/>root</a> is the same as
+ <var>node</var>'s <a for=/>root</a>.</p></li>.
+
+ <li><p>If <var>perform state-preserving atomic move</var> is true, then set <a>this</a>'s <a
+ for=Node>node document</a>'s <a>state-preserving atomic move in progress</a> to true.</p></li>
+
+ <li><p>Let <var>return node</var> be the result of <a>pre-inserting</a> <var>node</var> into
+ <a>this</a> before <var>child</var>.</p></li>
```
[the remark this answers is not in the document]
domfarolino: Yeah I think initially I saw it as a separate primitive too, but the more I looked at it, the more the difference between the two seemed really subtle. It mostly has to do with MutationObservers (and half of the relevant logic here is tucked away in the "remove" primitive) and *not* running the post-connection steps. So I feel like we'd end up with a near line-by-line copy of "insert", modulo one or two small differences. I'll take another look to see if my intuition is accurate, but I do kinda suspect this is where we'd end up.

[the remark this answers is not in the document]
annevk: It'd be good to have more details. Because you can't move a `DocumentFragment` so a vast number of checks don't apply. Adopt won't ever run.

There might be a number of range and shadow tree checks that end we end up duplicating, but perhaps that calls for abstracting those instead. At least to me a state flag seems rather unappealing.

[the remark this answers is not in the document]
domfarolino: It's possible there is enough from "insert" that we'd omit in "move", to make it creation worth it, sure.

> At least to me a state flag seems rather unappealing.

With a separate "move" primitive _and_ the decision to throw when we can't "move", we could get rid of the state variable that's currently in "insert" (i.e., `<var>statePreservingAtomicMoveInProgress</var>`). But the state bool on _`Document`_ is what other specifications will refer to in their removal steps to react to a move appropriately, so I'm not sure we can get rid of _that_ one.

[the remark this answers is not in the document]
annevk: Hmm, wouldn't we give other specifications "move steps"?

smaug----:
> * old target
> 
> * target
> 
> * moved node (I'm not sure you can ever move multiple at this point, but maybe we should allow for it in the mutation record design?)

Shouldn't the target node be all the time the same, it is just the siblings which change.
So we'd need only oldPreviousSibling and oldNextSibling. Oh, hmm, this isn't only about moving children but moving anything.

If this is really just remove and add back elsewhere, we could just reuse the existing childList MutationRecords, one for remove, one for adding node back, and possibly just add a flag to MutationRecord that it was about move.


(movedNodes is a bit confusing, since it seems to depend on the connectedness of the relevant nodes and it is apparently empty for the removal part. And it is unclear to me why we need the connectedness check. This is about basic DOM tree operations, and I'd assume those to work the same way whether or not the node is connected)

annevk:
Creating two separate mutation records that a consumer would have to merge to (fully) understand it's a move seems suboptimal?

I agree that it should probably work for disconnected nodes as well, but I don't think we want to support a case where the shadow-including root changes.

ajklein:
It's been a long time since I've thought about this stuff, but I'm inclined to agree with @smaug---- that creating a new type of `MutationRecord` feels unnecessary. Users of `MutationObserver` already have to do coalescing if they want to make sense of the stream of changes they observe. There are already other move-like operations, such as appending a child that's already somewhere else in the tree, which today generates two records.

dom.bs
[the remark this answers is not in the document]
domfarolino: I thought about this a bit more. I think we're happy to create a new "move" primitive algorithm following the "insert" and "remove" algorithms. As you mentioned, it will be a simpler algorithm than "insert" since we'll avoid the `DocumentFragment` stuff — oh and also the post-connection stuff, among maybe more things.

There's still the discussion around introducing either (**a**) new "move" steps/hook that other specs provide to react to moves, vs (**b**) a Document-scoped boolean that the new "move" algorithm sets, for existing insertion/removal steps to use, to not reset certain state inside those steps.

Option (b) is the direction of the current PR, which we think makes sense after implementing this in Chromium. I think this is because so far we've found no behavior that we specifically want to do during a move that isn't done by insertion/removal — so any "move steps" we provided would just be empty. **Does that make sense?**

Now, I think the real question here is whether we want to:

1. Keep running the [insertion](https://dom.spec.whatwg.org/#concept-node-insert-ext)/[removal](https://dom.spec.whatwg.org/#concept-node-remove-ext) steps during a "move" operation, and just pepper `if during atomic move is true, don't run some of these substeps!` wherever state would ordinarily be reset
    - Cancel out [the iframe removal steps](https://html.spec.whatwg.org/C#the-iframe-element:html-element-insertion-steps)
    - Cancel [img relevant mutations](https://html.spec.whatwg.org/C#reacting-to-dom-mutations:html-element-insertion-steps), to avoid re-firing load events
1. Simply not run insertion/removal hooks _at all_ during a move. This is only feasible if we determine that there are absolutely no steps in any spec's existing insertion/removal hooks that need to run during a move. We'll have to do an audit to determine this, which I'm starting now.

So the way I see it, those are probably our options, as opposed to creating new move steps which I think from our findings so far, would be unused. (We'll course-correct if this finding changes).

dom.bs
```diff
@@ -4986,6 +5033,27 @@ within <a>this</a>.
...
 return the result of <a>pre-removing</a> <var>child</var> from <a>this</a>.
 
+<p>The <dfn method for=Node><code>moveBefore(<var>node</var>, <var>child</var>)</code></dfn>
+method steps are:
+
+<ol>
+ <li><p>Let <var>perform state-preserving atomic move</var> be true if <a>this</a> is
+ <a>connected</a>, <var>node</var> is <a>connected</a>, <a>this</a>'s <a for=Node>node
+ document</a> is <a>fully active</a>, and <a>this</a>'s <a for=/>root</a> is the same as
+ <var>node</var>'s <a for=/>root</a>.</p></li>.
```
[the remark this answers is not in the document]
domfarolino: Done!

dom.bs
```diff
@@ -2652,6 +2652,50 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
  <!-- Technically this is post-insert. -->
 </ol>
 
+<p>To <dfn export for=Node id=concept-node-ensure-pre-move-validity>ensure pre-move validity</dfn>
+of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run these steps:
+
+<ol>
+ <li><p>If either <var>parent</var> or <var>node</var> are not <a>connected</a>, then
+ <a>throw</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}}.</p></li>
```
smaug----, line 2660: Hmm, why do we have this limitation. Why not allow move within the same tree? Basically why the condition about shadow-including root isn't enough?

dom.bs
```diff
@@ -2652,6 +2652,50 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+
+ <li><p>If <var>parent</var>'s <a for=/>shadow-including root</a> is not the same as
+ <var>node</var>'s <a for=/>shadow-including root</a>, then <a>throw</a> a
+ "{{HierarchyRequestError!!exception}}" {{DOMException}}.</p></li>
+
+ <li><p>If <var>parent</var> is not an {{Element}} <a for=/>node</a>, then <a>throw</a> a
+ "{{HierarchyRequestError!!exception}}" {{DOMException}}.
```
smaug----, line 2667: Why moves in DocumentFragments aren't possible? Especially, what if parent is a ShadowRoot?

dom.bs
```diff
@@ -2845,6 +2889,125 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+ <li>
+  <p>For each <a>live range</a> whose <a for=range>start node</a> is <var>parent</var> and
+  <a for=range>start offset</a> is greater than <var>index</var>, decrease its
+  <a for=range>start offset</a> by 1.</p>
+
+  <p class="note">Note that unlike the traditional <a for=/ lt="remove">removal</a> case, we do not
+  need to update <a>live range</a> state when their <a for=range>start node</a> or
+  <a for=range>end node</a> is an <a>inclusive descendant</a> of the <var>node</var>. This is
+  because said <a>nodes</a> do not get removed from their <a>tree</a>, so ranges associated with
+  them stay intact.</p>
```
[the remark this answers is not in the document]
domfarolino: I don't think so. It should be equivalent to when a range's start container is set (to the first time) to a node after the end container as well. Here are tests for this, which include `intersectsNode` assertions to make sure the Range's state is not peculiar: https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/moveBefore/tentative/live-range-updates.html.

dom.bs
domfarolino: DocumentFragments are always disconnected, right? We require that the moved node, and its new parent, are both connected to the DOM.

dom.bs
domfarolino: I believe we discussed this at TPAC. The idea is that `moveBefore()` is used to preserve state, so in cases where state cannot be preserved, we should not just fall back to usual `insertBefore()`-style behavior. Instead, we'll throw and let the developer handle this case themselves. Basically we let `moveBefore()` succeed only when we're sure we can perform a state-preserving move, and when disconnected destinations or origins are involved, the state cannot be preserved.

dom.bs
smaug----: Shadowroot is a DocumentFragment. I would expect moves under ShadowRoot being possible.
(I wasn't thinking the connectedness case here, only that one should be able reorder the children of a ShadowRoot)

dom.bs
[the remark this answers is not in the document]
smaug----: Ok, so something else still updates Range boundary points so that start is before end point? The note is then a bit confusing.

dom.bs
domfarolino: I didn't realize `ShadowRoot` was a `DocumentFragment`! Thanks for pointing this out. OK I've changed the spec to enable this (we can discuss connected-ness separately), and I've added a test for this in https://crrev.com/c/6012161. I'll resolve this now.

dom.bs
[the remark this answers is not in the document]
domfarolino: > Ok, so something else still updates Range boundary points so that start is before end point?

No, a Range's start does not have to be before its end, in DOM tree order. Just like if you create a new Range whose start comes after its end in DOM tree order, the start's [position](https://dom.spec.whatwg.org/#concept-range-bp-position) will be "after" its end, which is observable via `intersectsNode()` since nodes that are after end but before start, will no longer `intersectsNode()` the Range. The same goes for when `moveBefore()` moves things around such that the range's start is after its end, in tree order.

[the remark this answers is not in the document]
smaug----: Wait, how do you create a Range which has its start after end? setStart/End etc will modify the other boundary point if the order isn't right.

dom.bs
smaug----: I would definitely expect a move operation to succeed if one stays within a same tree, whether or not your connected to document.

dom.bs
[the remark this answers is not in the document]
domfarolino: Ooof, OK then the spec indeed does leave ranges in a bad state, since what this PR allows (wrt `Range` + `moveBefore()`) is not possible with `Range` today. Great catch, @smaug----. In that case, I think the best way to deal with this would be to, during `moveBefore()`, detect if the moved node is the inclusive ancestor of a start or end boundary point of any range, and if that boundary point crosses its corresponding start/end, and collapse the node at the other boundary point. How does that sound to you?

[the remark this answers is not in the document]
smaug----: That wouldn't be consistent with other tree operations. I think I'd prefer a model where ranges were treated as if a node was removed and then added back to a different place in tree. 
(Kind of like how MutationRecords are handled)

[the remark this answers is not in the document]
domfarolino: In that case, we wouldn't be able to preserve selection with `moveBefore()`, which is one of the pieces of state we were really hoping to preserve with this API. This is especially annoying when you `moveBefore()` the selection range's start container to somewhere earlier in the doc, or the end container somewhere later. The range is still totally valid, so it seems like we should be able to preserve selection. What do you think?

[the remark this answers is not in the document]
domfarolino: Another option for partial preservation (discussed with @noamr) would be to preserve the selection if and only if the moved node is an inclusive ancestor of _both_ the start and end containers of the range. That way if the selection (or other range) is entirely contained within the moved node, we don't lose it.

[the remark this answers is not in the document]
noamr: @smaug---- perhaps a useful middle ground can be that we preserve the selection only if the moved node is an ancestor of the entire range. This way if the range is moved while "intact", the selection would be preserved, but if parts of it are moved it's considered "broken" and it's dissolved. This is useful when moving whole elements which might have bits of text selected inside them.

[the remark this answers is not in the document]
smaug----: Not sure what  you mean with "wouldn't be able to preserve selection". Using the normal removal/addition rules  would preserve selection the usual way. The range just gets modified.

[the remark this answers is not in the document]
smaug----: Container approach (not modifying the range if full contained) would be still problematic if the container was moved from light dom to shadow dom, and the range was a range which one can access from selection.getRangeAt().  Selection should not reveal nodes in shadow dom.
Also it might be a bit surprising if selection is preserved in some cases but not in others.

Maybe, unrelated to moveBefore, there should be some helper methods to create a StaticRange from Range, and then update selection later, if/when needed, using that StaticRange. Hmm, though StaticRange isn't quite enough, since one needs also direction.

dom.bs
annevk: I don't see how connected <> disconnected would work. It would need a completely different code path, no?

I can see connected <> connected and disconnected <> disconnected. Someone (either Noam or Dominic) argued against disconnected <> disconnected earlier, but it seems reasonable to allow as well.

dom.bs
```diff
@@ -4986,6 +5151,16 @@ within <a>this</a>.
 <p>The <dfn method for=Node><code>removeChild(<var>child</var>)</code></dfn> method steps are to
 return the result of <a>pre-removing</a> <var>child</var> from <a>this</a>.
 
+<p>The <dfn method for=Node><code>moveBefore(<var>node</var>, <var>child</var>)</code></dfn>
+method steps are:
+
+<ol>
+ <li><p>Let <var>return node</var> be the result of <a>pre-moving</a> <var>node</var> into
+ <a>this</a> before <var>child</var>, rethrowing any exceptions.</p></li>
```
annevk, line 5159: We don't say rethrow.

dom.bs
noamr: I'm ok with both. One can say that disconnected<>disconnected doesn't have any state to preserve, so whether we throw or do the same as insertBefore is a judgement call.

annevk: We should not do the same as insertBefore. It should run the moving steps (whether anyone cares about those for disconnected is a different question).

noamr: Sure, I mean that it should be observably the same in the disconnected<>disconnected case.

dom.bs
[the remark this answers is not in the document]
domfarolino: OK the latest commit reverts special range handling, treating ranges just like we would in a normal remove situation. I'll resolve this.

dom.bs
domfarolino: I've made this PR support both the `connected->connected` and `disconnected->disconnected` case, as of https://github.com/whatwg/dom/pull/1307/commits/2bbc8346167310cc3e7f8d2ea65e5151c9cc219f. I think that's sufficient, and we've heard that developers want this too, so I think that's enough to resolve things here.

tjallingt: Disconnected -> connected could theoretically also been seen as stated preserving since the disconnected node didn't have any previous state so there is nothing to preserve.

Connected -> disconnected moves could destroy the state though so should definitely not be allowed.

domfarolino: > Disconnected -> connected could theoretically also been seen as stated preserving since the disconnected node didn't have any previous state so there is nothing to preserve.

I think I disagree. The state could be null `contentDocument`, lack of focus, no form owner, and more etc. That "no state" is still a kind of state, and it'd either be:
 - **Changed** by becoming connected into a document, and running the many hooks that the platform does during that milestone — some of which initialize iframe setup, run script, etc. In this case, move just becomes insertion, which we don't want to do because the two are distinct primitives.
 - **Not changed**, by just inserting the element into the DOM and leaving it kind of dead/void. This would be completely new and I think kind of weird. For example `parent.moveBefore(disconnectedIframe, null)` would make `disconnectedIframe` finally connected, but with a null content document because we never initialized the setup. In that case, it feels like there was no reason to *move* it into the DOM in the first place. Same goes for other things with form owners, executable `<script>`s, etc. I'm not sold that this is a real use case.

sorvell:
Would it be possible to have this method fallback to `insertBefore` if the conditions for an atomic move are not met? 

Without this, in practice it seems likely that users will need to write a boilerplate function that tests the 2 nodes for `isConnected` before calling. 

```js
const moveOrInsertNode = (container, node, ref_node = null) => {
 const canMove =  (container.isConnected && node.isConnected && (!ref_node || ref_node.parentNode === container);
  return canMove ? container.moveBefore(node, ref_node) : container.insertBefore(node, ref_node);
}
```

WebReflection:
FWIWI I fully agree with @sorvell ... the whole point of this proposal was to simplify DOM manipulation, not to complicate it even further with tons of repeated checks or `try/catch` around nodes that might not be fully owned by the underlying logic. The fallback to `insertBefore` when it fails is inevitable, as the intent is "*to move that node in that place*", nothing else, so it'd be counter-productive to have many checks around such intent + a try/catch toplay it safely plus the inevitable `insertBefore` on the catch.

domfarolino:
I'm personally pretty sympathetic to the arguments above. It was the original direction we went in, and I do think it is simpler to use, however the discussion we had at TPAC resulted in editors pushing back, on account of "move" being a fundamentally different primitive from "insert". @annevk are you open to revisiting this?

tjallingt:
"Move" would still be a fundamentally different primitive if it tried to preserve state where possible since the "insert" guarantees a default state. I understand that from a spec perspective this makes "move" a superset of "insert" given that it may need to run initialization hooks but I think that is a more usable primitive for developers.

domfarolino:
I definitely understand that. `moveBefore()` appears to be easier to use if it didn't throw any new errors that `insertBefore()`  doesn't already throw. At the same time, there is a predictability cost to this. As I [mentioned over in the Blink I2S](https://groups.google.com/a/chromium.org/g/blink-dev/c/YE_xLH6MkRs/m/ba-KZPUHAQAJ), it might  be weird if `moveBefore()` sometimes had huge side effects like running script, initializing iframes, applying styles, etc., yet other times no side effects at all (because the "atomic" move was carried out).

That said, if 99% of `moveBefore()` uses are from a generic place like a framework, where you _always_ want to move the node from A->B (just atomically when possible!) then the error-throwing behavior becomes cumbersome, because then 99% of users have to call `insertBefore()` from a catch block just to use the API correctly.

I guess it comes down to how often we think people actually care to know whether a truly "atomic" move can be performed, and might act differently if they find out it can't. That's where the error-throwing behavior becomes really useful; without it, if you want to know whether `moveBefore()` will actually move atomically, you have to remember to check [all of the pre-conditions](https://whatpr.org/dom/1307.html#concept-node-ensure-pre-move-validity) before calling `moveBefore()`.

noamr:
I think that the following scenario is plausible:

```js
try {
  element.moveBefore(newNode, refNode);
} catch {
  element.insertBefore(newNode, refNode);
}  finally {
  // check whether we need to do something extra - e.g. the focused element reparented
  // and some library relies on the side effects
}
```

It's likely that the `finally` clause would be empty in the vast majority of cases, however it's debatable whether it's the platform's job to hide this difference for the sake of ergonomics in a somewhat low-level API like DOM.

dead-claudia:
I wonder if that choice of error tolerance could be moved to a boolean argument. Like, `parent.moveBefore(child, refNode, true)` to tolerate failure to move (like a missing custom element move callback) and `parent.moveBefore(child, refNode, false)` to throw on failure, or vice versa. That could be an acceptable compromise.

Also, how does this handle the case of `parent` and `child` having different `Document`s? If I'm understanding correctly, an exception is thrown, but that's just a guess that checking "shadow-including root" equality implies that.

noamr:
> > however it's debatable whether it's the platform's job to hide this difference for the sake of ergonomics in a somewhat low-level API like DOM.
> 
> @noamr There _is_ some precedent for light ergonomics in the DOM already: `child.before`, `child.after`, `parent.append`, and `parent.replaceChildren` all four accepting strings and auto-converting them into text nodes. I feel this could qualify as such "light ergonomics".

Yes, and we can certainly consider adding more ergonomic variants in the future on top of the `moveBefore` behavior, the same way `appendChild` is an ergonomic version of `insertBefore`.
However, `moveBefore` itself, being the first primitive to support the "move" operation, is not that - there is value in making it as single-purpose as possible: it only moves. It doesn't fall back to something else, but fails and notifies the author.

noamr:
> I wonder if that choice of error tolerance could be moved to a boolean argument. Like, `parent.moveBefore(child, refNode, true)` to tolerate failure to move (like a missing custom element move callback) and `parent.moveBefore(child, refNode, false)` to throw on failure, or vice versa. That could be an acceptable compromise.

This particular API shape is a boolean trap, but I think we can can consider variants in the future that have an 'move if you can insert if you can't' behavior. But this is a slightly higher level behavior than moving. As I said, there is value in having at least some DOM APIs that are as primitive as possible.

> Also, how does this handle the case of `parent` and `child` having different `Document`s? If I'm understanding correctly, an exception is thrown, but that's just a guess that checking "shadow-including root" equality implies that.

It throws when moving across documents.

WebReflection:
> it might be weird if moveBefore() sometimes had huge side effects like running script, initializing iframes, applying styles, etc., yet other times no side effects at all (because the "atomic" move was carried out).

if everyone will inevitably end up using a `try/catch` that fallback to `insertBefore` this API throwing doesn't help at all + all checks [mentioned in here](https://github.com/whatwg/dom/pull/1307#issuecomment-2491906616) are not obvious, developer friendly, or performance friendly.

When a developer does not want to have side-effects it can perform those checks manually, the rest 99% of the world will just `moveBefore` and fallback to `insertBefore`, including all frameworks, hence my question: who is this throwing behavior helping and why an alternative solution such as `parent.canMoveNode(node)` has not been discussed, as that would do surely the right thing internally as opposite of asking developers to know what are all conditions to check before a node can be moved or not?

Please note I am not suggesting `canMoveNode` together with a throwing `moveBefore`, I am suggesting a `moveBefore` that tries to do the right thing and fallback to `insertBefore` without throwing, plus a `canMoveNode` for anyone interested in that detail (which is niece, or tests based things, or perf based things, or debugging based things).

In short, if try/catch is the workaround, everyone will use it and all the reasons it couldn't be done behind the scene in the name of "*it's better for developers*" will be futile, I hope we can agree on that.

annevk:
The reason it throws is because in the future we might become more ambitious and also tackle those scenarios. The reason we don't have various accompanying APIs such as `canMoveBefore()` is because we tend to develop APIs in baby steps. Doing it all at once typically leads to regrets. Please continue this discussion in #1255 as otherwise this PR becomes unmanageable.

annevk (changes requested):
The aspects shared between insert and move, and remove and move, need to be shared. We don't want to have to maintain identical range mutation in multiple places.

dom.bs
```diff
@@ -2652,6 +2652,63 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+  to both be <a>connected</a>, or both be disconnected. Otherwise, state preservation cannot be
+  guaranteed during the move operation.</p>
+ </li>
+
+ <li><p>If <var>parent</var>'s <a for=/>shadow-including root</a> is not the same as
+ <var>node</var>'s <a for=/>shadow-including root</a>, then <a>throw</a> a
+ "{{HierarchyRequestError!!exception}}" {{DOMException}}.</p></li>
+
+ <li><p>If <var>parent</var> is not an {{Element}} or {{DocumentFragment}} <a for=/>node</a>, then
+ <a>throw</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}}.
```
annevk, line 2680: The reason this restriction is okay is because documents cannot have multiple element children and you can only use `insertBefore()` at the root when the current element is removed. So it really relies on the connected checks above. That's very subtle.

However, when looking at non-elements it's quite a bit more restrictive than what it probably should be. I don't see why we'd forbid moving a comment to be a child of a document. So I would suggest we do this differently, even though in practice there's no benefit to moving comments and the like.

Ideally we'd align this with "pre-insertion validity" or share as much as possible.

dom.bs
```diff
@@ -2845,6 +2902,124 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+   name "<code>connectedMoveCallback</code>", and an empty argument list.
+
+   <li>
+    <p>Otherwise, <a lt="try to upgrade an element">try to upgrade</a>
+    <var>inclusiveDescendant</var>.
+
+    <p class=note>If this successfully upgrades <var>inclusiveDescendant</var>, its
+    <code>connectedCallback</code> will be enqueued automatically during the
+    <a>upgrade an element</a> algorithm.
+   </li>
```
annevk, line 3011: How could this happen due to a move? That seems incorrect?

dom.bs
```diff
@@ -3977,8 +4152,8 @@ method steps are:
  <li><p>Assert: either <var>addedNodes</var> or <var>removedNodes</var> <a for=set>is not empty</a>.
 
  <li><p><a>Queue a mutation record</a> of "<code>childList</code>" for <var>target</var> with
- null, null, null, <var>addedNodes</var>, <var>removedNodes</var>, <var>previousSibling</var>,
- and <var>nextSibling</var>.
+ null, null, null, <var>addedNodes</var>, <var>removedNodes</var>, <var>previousSibling</var>, and
+ <var>nextSibling</var>.
```
annevk, line 4156: Let's not change this in this PR.

dom.bs
```diff
@@ -4118,6 +4293,8 @@ interface Node : EventTarget {
   [CEReactions] Node appendChild(Node node);
   [CEReactions] Node replaceChild(Node node, Node child);
   [CEReactions] Node removeChild(Node child);
+
```
annevk, line 4296: I don't think we need this newline here.

dom.bs
noamr: I guess if we had scoped element registries and we moved across trees?

domfarolino: OK yeah I had to get educated a bit more on exactly how custom elements work and upgrade, but I agree this is not relevant for this PR. It's now removed; resolving.

dom.bs
```diff
@@ -2652,6 +2652,63 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+<ol>
+ <li>
+  <p>If any of the following conditions are true</p>
+
+  <ul>
+   <li><p><var>parent</var> is <a>connected</a> and <var>node</var> is not <a>connected</a>; or</p></li>
+
+   <li><p><var>parent</var> is not <a>connected</a> and <var>node</var> is <a>connected</a>,</p></li>
+  </ul>
```
smaug----, line 2666: Why do we need this when we have " If parent ’s shadow-including root is not the same as node ’s shadow-including root " check below?

dom.bs
smaug----: Yeah, I'd expect this to handle basically ParentNode mixin.

domfarolino: Hmm, OK. Would you prefer we simply make this condition perfectly match the pre-insertion validity steps? I.e., expanding it to include Document parents too? I'm fine with that. Let me know what you think @annevk.

> Yeah, I'd expect this to handle basically ParentNode mixin.

Hmm, I'm not exactly sure what this means. But does what I recommend above satisfy what you're looking for?

dom.bs
domfarolino: So, I think back when we only supported the connected->connected case, the "shadow-including root matches" check was right, because "shadow-including root" was essentially the same as "node document", in the connected case.

But now that we support the disconnected->disconnected case, I actually think the "shadow-including root" check is too strict, right? For example, imagine:

```js
const divA = document.createElement('div');
const p = divA.appendChild(document.createElement('p'));
const divB = document.createElement('div');

// Throws because divB's shadow-including root does not match p's, since they're both disconnected.
divB.moveBefore(p, null);
```

That is, in the disconnected->disconnected case, it is way more likely that the target node and destination container's shadow-including roots will not match, since in the disconnected case, a shadow-including root just tops-out at the nearest disconnected parent. So my proposal is to simply change the shadow-including root comparison to instead compare node documents. I'll do that now, and please let me know if you think that sounds good.

dom.bs
annevk: Yeah, I think that would be good, although I have not thought through all implications of such a change.

domfarolino: OK I've done that in https://github.com/whatwg/dom/pull/1307/commits/90c9a943a859267acd05b37c48120427e497c7e2, and changed the Chromium implementation and added a test in https://crrev.com/c/6081493. I agree this is a pretty minor change, especially since Documents can still only have a single Element child, and it is nice to have consistency with the pre-insertion validity conditions. I'll resolve this now.

domfarolino:
All comments appear to be addressed at this point. PTAL!

annevk:
I've reopened the validity topic. I think you now allow elements to be moved into document's children without any kind of additional checks. The checks are also in a different order. And given how many are shared it probably warrants more abstraction. (Also, "pre-insertion", but "pre-move" is weird. Maybe "pre-insertion" is wrong?)

domfarolino:
> I think you now allow elements to be moved into document's children without any kind of additional checks.

OK, I've fixed this, adding the relevant pre-insertion conditions to the pre-move algorithm. I'll do the same for Chromium and WPTs.

> The checks are also in a different order.

Fixed. The pre-move-specific checks still come first, but all **relevant** pre-insert conditions come after, and in the same order as the appear in the pre-insertion algorithm.

> And given how many are shared it probably warrants more abstraction.

Hmm, I could go either way. There are not that many conditions in pre-move that are carbon copies of their corresponding conditions pre-insert conditions. Perhaps the first 3 can be cleanly lifted out now, but the remaining are modified enough to probably not warrant abstraction. Would you prefer I lift the first 3 conditions out into a common sub algorithm?

> (Also, "pre-insertion", but "pre-move" is weird. Maybe "pre-insertion" is wrong?)

I've renamed all "pre-insertion" instances to "pre-insert".

dom.bs
smaug----: Well, I don't want to support move between otherwise unrelated subtrees. This all becomes simpler if we don't have that. No need to check same document or anything like that.

domfarolino: > Well, I don't want to support move between otherwise unrelated subtrees.

What are "otherwise unrelated subtrees"? We certainly want to support the code example I wrote above, no?

annevk: I think Olli always wants there to be a shared ancestor.

domfarolino: Yeah, I personally don’t think I see the value in enforcing such a requirement. Do you feel strongly about this? I think folks should be able to create detached DocumentFragments and move things around from them to possibly other detached subtrees, and mostly I don’t see the value of enforcing a common ancestor between old and new parent. What do you think @annevk ?

dom.bs
```diff
@@ -2845,6 +2911,101 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+    <p class="note">Because the <a>move</a> algorithm is a separate primitive from
+    <a for=/>insert</a> and <a for=/>remove</a>, it does not invoke the traditional
+    <a>insertion steps</a> or <a>removing steps</a> for <var>inclusiveDescendant</var>.
+   </li>
+
+   <li><p>If <var>inclusiveDescendant</var> is <a for=Element>custom</a> and <var>newParent</var> is
+   <a>connected</a>, then <a>enqueue a custom element callback reaction</a> with
+   <var>inclusiveDescendant</var>, callback name "<code>connectedMoveCallback</code>", and an empty
+   argument list.
```
annevk, line 2997:
```suggestion
   <var>inclusiveDescendant</var>, callback name "<code>connectedMoveCallback</code>", and « ».
```
We also have to clean this up elsewhere, but that can be done separately. New algorithms should do it right from the start though.

dom.bs
```diff
@@ -8163,6 +8320,31 @@ but not its <a for=range>end node</a>, or vice versa.
  </ul>
 </div>
 
+<p>The <dfn>live range pre-removing steps</dfn> given a <a>node</a> <var>node</var>, are as follows:
```
annevk, line 8323: Should this not be pre-remove steps?

dom.bs
```diff
@@ -4118,6 +4264,7 @@ interface Node : EventTarget {
   [CEReactions] Node appendChild(Node node);
   [CEReactions] Node replaceChild(Node node, Node child);
   [CEReactions] Node removeChild(Node child);
+  [CEReactions] Node moveBefore(Node node, Node? child);
```
annevk, line 4267: I think we should put this method on `ParentNode`. That would also allow us to simplify the validity checks a bit.

annevk: It also needs a "web developer" intro box.

dom.bs
annevk: I'm not sure I hold a strong opinion, but I do like the symmetry Olli is suggesting.

dom.bs
```diff
@@ -2845,6 +2911,101 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+passed a <a for=/>node</a> <var ignore>movedNode</var>, and a <a for=/>node</a>-or-null <var
+ignore>oldParent</var> as indicated in the <a for=/>move</a> algorithm below. Like the <span
+data-x="concept-insertion-steps-ext">insertion steps</span>, these steps must not modify the
+<a>node tree</a> that <var>insertedNode</var> <a>participates</a> in, create
+<a for=/>browsing contexts</a>, <a lt="fire an event">fire events</a>, or otherwise execute
+JavaScript. These steps may queue tasks to do these things asynchronously, however.
+
+
+<p>To <dfn export id=concept-node-move>move</dfn> a <var>node</var> into a <var>newParent</var>
+before a <var>child</var>, run these steps:
```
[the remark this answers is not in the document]
domfarolino: This was inspired by https://dom.spec.whatwg.org/#concept-node-insert, https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity, and https://dom.spec.whatwg.org/#concept-node-pre-insert. Should I still remove it?

dom.bs
```diff
@@ -2652,6 +2652,72 @@ of a <var>node</var> into a <var>parent</var> before a <var>child</var>, run the
...
+ <li><p>If <var>child</var> is non-null and its <a for=tree>parent</a> is not <var>parent</var>,
+ then <a>throw</a> a "{{NotFoundError!!exception}}" {{DOMException}}.
+
+ <li><p>If <var>node</var> is not an {{Element}} or a {{CharacterData}} <a for=/>node</a>, then
+ <a>throw</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}}.</p></li>
+
+ <li><p>If <var>node</var> is a {{Text}} <a for=/>node</a> and <var>parent</var> is a
+ <a>document</a>, then <a>throw</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}}.
+
+ <li><p>If <var>parent</var> is a {{Document}}, <var>node</var> is an {{Element}}
```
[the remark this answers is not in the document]
domfarolino: Out of curiosity, why this change? Earlier in this algorithm we do a very similar comparison of `<var>parent</var>`, but with `{{Document}}`.

dom.bs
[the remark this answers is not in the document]
annevk: Yes.

dom.bs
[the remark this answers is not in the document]
annevk: I think in that case there's multiple nodes so it's clearer to go for the longer "X node" concept as there's multiple of X. You could do "Document node" here too, but generally we shorten that where we can.

dom.bs
domfarolino: Hmm, I'm not sure it provides any more symmetry than the same-document check being enforced in the connected and disconnected case. But yeah I'll just replace this check with the shadow-inclusive root check to unblock this PR, and we'll update the tests.

domfarolino: Done in https://github.com/whatwg/dom/pull/1307/commits/65b4e13ade1039eee8e49316b3728e9f4191c414.

dom.bs
domfarolino: Done in https://github.com/whatwg/dom/pull/1307/commits/3f3e94c5beda922962dacaeb606087f57bd7f7be! Besides the exposure of the method, the semantics of that commit are not observable, right?

peasantman1:
> This PR introduces a new DOM API on the `Node` interface: `moveBefore()`. It mirrors `insertBefore()` in shape, but defers to a new DOM manipulation primitive that this PR adds in service of this new API: the "move" primitive. The move primitive contains some of the DOM tree bookkeeping steps from the remove primitive, as well as the insert primitive, and does three more interesting things:
> 
> 1. Calls the [moving steps hook](https://whatpr.org/dom/1307.html#concept-moving-steps-ext) with the moved node and the old parent (possibly null, just like the removing steps)
> 2. Queues a custom element callback reaction for the `connectedMoveCallback()`
> 3. Queues two back-to-back mutation record tasks: one for removal from the old parent; one for insertion into the new one
> 
> The power of the move primitive comes from the fact that the algorithm does not defer to the traditional insert and removal primitives, and therefore does not invoke the [removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext) and [insertion steps](https://dom.spec.whatwg.org/#concept-node-insert-ext). This allows most state to be preserved by default (i.e., we don't tear down iframes, or close dialogs). Sometimes, the insertion/removing step overrides in other specifications have steps that do need to be performed during a move anyways. These specifications are expected to override the moving steps hook and perform the necessary work accordingly. See [whatwg/html#10657](https://github.com/whatwg/html/pull/10657) for HTML.
> 
> Remaining tasks (some will be PRs in other standards):
> 
> * [x]  Custom element integration
> * [x]  Keep popovers open
> * [x]  Don't call post-connection steps if state-preserving atomic move is in progress
> * [x]  Don't call becomes connected / becomes browsing-context
> * [x]  Only disconnect subframes on removal when state-preserving atomic move is **not** in progress
> * [x]  Keep dialogs open: see [removing steps](https://html.spec.whatwg.org/C#the-dialog-element:html-element-removing-steps)
> * [x]  img/source: this shouldn't count as a [relevant mutation](https://html.spec.whatwg.org/C#relevant-mutations)
> * [x]  Preserve fullscreen
> * [x]  Preserve focus 
>   
>   * Need to resolve `focusin` event semantics
> * ~[ ] Don't reset animations / transitions. See [here](https://drafts.csswg.org/css-transitions/#starting) 
>   
>   * Maybe nothing needs to be done here. Given [how element removals are handled](https://drafts.csswg.org/css-transitions/#transition-cancel:~:text=If%20an%20element%20is%20no%20longer%20in%20the%20document%2C%20implementations%20must%20cancel%20any%20running%20transitions%20on%20it%20and%20remove%20transitions%20on%20it%20from%20the%20completed%20transitions), the spec does NOT require transitions to be removed from the UA's set of _running transitions_ for moved nodes since they are never removed from the Document.~
> * ~[ ] Preserve text-selection. See [set the selection range](https://html.spec.whatwg.org/C#set-the-selection-range). Edit: Nothing needs to be done here. Selection metadata (i.e., `selectionStart` and kin) is preserved by default in browsers, consistent with HTML (no action is taken on removal). The UI behavior of the selection not being highlighted is a side-effect of the element losing focus~
> * [ ]  Selection API: don't reset the Document's [selection](https://w3c.github.io/selection-api/#dfn-selection)
>   
>   * Updates to the selection range should happen according to how the DOM Standard primitives update ranges. The Selection API specification admits as much, by [deferring to the](https://w3c.github.io/selection-api/#responding-to-dom-mutations) insert and removal algorithms. Therefore, we should reference the move primitive from the Selection API specification, and ensure that the move primitive in _this DOM Standard PR_ updates live ranges correctly: [Reference the move primitive in DOM mutations section w3c/selection-api#341](https://github.com/w3c/selection-api/pull/341)
>   * `selectionchange` event: We've decided to allow `selectionchange` event to still fire, [since it is queued in a task](https://w3c.github.io/selection-api/#ref-for-dfn-schedule-a-selectionchange-event-1). No changes for this part are required.
> * [x]  Pointer event state reset: see [here](https://w3c.github.io/pointerevents/#suppressing-a-pointer-event-stream)
>   
>   * The specification does not reset an element's previously-set pointer capture upon DOM node insertion or removal. While this seems like a gap in that specification ([[DOM integration] No state is reset on node removal w3c/pointerevents#526](https://github.com/w3c/pointerevents/issues/526)), it appears that nothing needs to be done in that specification to satisfy this preservation semantic.
>   * Test: https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/moveBefore/tentative/pointer-events.html
> * [x]  Hide input/select picker: [here](https://html.spec.whatwg.org/C#show-the-picker%2C-if-applicable)
>   
>   * Nothing in the specification seems to hook into DOM for removal when the picker dialog is shown, so no work needs to be done in this PR. See ["Show the picker" steps never dismisses the picker upon node removal html#10743](https://github.com/whatwg/html/issues/10743) for a follow-up.
> * [x]  Preserve pointer lock: [here](https://w3c.github.io/pointerlock/#requestPointerLock)
>   
>   * When a ["pointer-lock target becomes disconnected"](https://w3c.github.io/pointerlock/#ref-for-dfn-pointer-lock-target-12), pointer lock is exited. Since the "becomes disconnected" hook is not run during an atomic move, pointer lock is preserved naturally.
> * [ ]  Containment: keep last remembered size(see [here](https://drafts.csswg.org/css-sizing-4/#last-remembered))
> 
> * [x]  At least two implementers are interested (and none opposed):
>   
>   * Chromium: https://issues.chromium.org/issues/40150299
>   * WebKit: [DOM State-preserving move (`Node.prototype.moveBefore`) WebKit/standards-positions#375](https://github.com/WebKit/standards-positions/issues/375)
>   * Gecko: [DOM State-preserving move (`Node.prototype.moveBefore`) mozilla/standards-positions#1053](https://github.com/mozilla/standards-positions/issues/1053)
> * [x]  [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
>   
>   * https://github.com/web-platform-tests/wpt/tree/master/dom/nodes/moveBefore/tentative (need to mark as non-tentative)
> * [x]  [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
>   
>   * Chromium: https://issues.chromium.org/issues/40150299
>   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1923880
>   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=281223
> * [x]  [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: I've added the impacts-documentation label, per https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests
> * [x]  The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use.
> 
> (See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)
> 
> [Preview](https://whatpr.org/dom/1307.html) | [Diff](https://whatpr.org/dom/1307/719e45b...4b54a0d.html)

 Help

peasantman1:
> This PR introduces a new DOM API on the `Node` interface: `moveBefore()`. It mirrors `insertBefore()` in shape, but defers to a new DOM manipulation primitive that this PR adds in service of this new API: the "move" primitive. The move primitive contains some of the DOM tree bookkeeping steps from the remove primitive, as well as the insert primitive, and does three more interesting things:
> 
> 1. Calls the [moving steps hook](https://whatpr.org/dom/1307.html#concept-moving-steps-ext) with the moved node and the old parent (possibly null, just like the removing steps)
> 2. Queues a custom element callback reaction for the `connectedMoveCallback()`
> 3. Queues two back-to-back mutation record tasks: one for removal from the old parent; one for insertion into the new one
> 
> The power of the move primitive comes from the fact that the algorithm does not defer to the traditional insert and removal primitives, and therefore does not invoke the [removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext) and [insertion steps](https://dom.spec.whatwg.org/#concept-node-insert-ext). This allows most state to be preserved by default (i.e., we don't tear down iframes, or close dialogs). Sometimes, the insertion/removing step overrides in other specifications have steps that do need to be performed during a move anyways. These specifications are expected to override the moving steps hook and perform the necessary work accordingly. See [whatwg/html#10657](https://github.com/whatwg/html/pull/10657) for HTML.
> 
> Remaining tasks (some will be PRs in other standards):
> 
> * [x]  Custom element integration
> * [x]  Keep popovers open
> * [x]  Don't call post-connection steps if state-preserving atomic move is in progress
> * [x]  Don't call becomes connected / becomes browsing-context
> * [x]  Only disconnect subframes on removal when state-preserving atomic move is **not** in progress
> * [x]  Keep dialogs open: see [removing steps](https://html.spec.whatwg.org/C#the-dialog-element:html-element-removing-steps)
> * [x]  img/source: this shouldn't count as a [relevant mutation](https://html.spec.whatwg.org/C#relevant-mutations)
> * [x]  Preserve fullscreen
> * [x]  Preserve focus 
>   
>   * Need to resolve `focusin` event semantics
> * ~[ ] Don't reset animations / transitions. See [here](https://drafts.csswg.org/css-transitions/#starting) 
>   
>   * Maybe nothing needs to be done here. Given [how element removals are handled](https://drafts.csswg.org/css-transitions/#transition-cancel:~:text=If%20an%20element%20is%20no%20longer%20in%20the%20document%2C%20implementations%20must%20cancel%20any%20running%20transitions%20on%20it%20and%20remove%20transitions%20on%20it%20from%20the%20completed%20transitions), the spec does NOT require transitions to be removed from the UA's set of _running transitions_ for moved nodes since they are never removed from the Document.~
> * ~[ ] Preserve text-selection. See [set the selection range](https://html.spec.whatwg.org/C#set-the-selection-range). Edit: Nothing needs to be done here. Selection metadata (i.e., `selectionStart` and kin) is preserved by default in browsers, consistent with HTML (no action is taken on removal). The UI behavior of the selection not being highlighted is a side-effect of the element losing focus~
> * [ ]  Selection API: don't reset the Document's [selection](https://w3c.github.io/selection-api/#dfn-selection)
>   
>   * Updates to the selection range should happen according to how the DOM Standard primitives update ranges. The Selection API specification admits as much, by [deferring to the](https://w3c.github.io/selection-api/#responding-to-dom-mutations) insert and removal algorithms. Therefore, we should reference the move primitive from the Selection API specification, and ensure that the move primitive in _this DOM Standard PR_ updates live ranges correctly: [Reference the move primitive in DOM mutations section w3c/selection-api#341](https://github.com/w3c/selection-api/pull/341)
>   * `selectionchange` event: We've decided to allow `selectionchange` event to still fire, [since it is queued in a task](https://w3c.github.io/selection-api/#ref-for-dfn-schedule-a-selectionchange-event-1). No changes for this part are required.
> * [x]  Pointer event state reset: see [here](https://w3c.github.io/pointerevents/#suppressing-a-pointer-event-stream)
>   
>   * The specification does not reset an element's previously-set pointer capture upon DOM node insertion or removal. While this seems like a gap in that specification ([[DOM integration] No state is reset on node removal w3c/pointerevents#526](https://github.com/w3c/pointerevents/issues/526)), it appears that nothing needs to be done in that specification to satisfy this preservation semantic.
>   * Test: https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/moveBefore/tentative/pointer-events.html
> * [x]  Hide input/select picker: [here](https://html.spec.whatwg.org/C#show-the-picker%2C-if-applicable)
>   
>   * Nothing in the specification seems to hook into DOM for removal when the picker dialog is shown, so no work needs to be done in this PR. See ["Show the picker" steps never dismisses the picker upon node removal html#10743](https://github.com/whatwg/html/issues/10743) for a follow-up.
> * [x]  Preserve pointer lock: [here](https://w3c.github.io/pointerlock/#requestPointerLock)
>   
>   * When a ["pointer-lock target becomes disconnected"](https://w3c.github.io/pointerlock/#ref-for-dfn-pointer-lock-target-12), pointer lock is exited. Since the "becomes disconnected" hook is not run during an atomic move, pointer lock is preserved naturally.
> * [ ]  Containment: keep last remembered size(see [here](https://drafts.csswg.org/css-sizing-4/#last-remembered))
> 
> * [x]  At least two implementers are interested (and none opposed):
>   
>   * Chromium: https://issues.chromium.org/issues/40150299
>   * WebKit: [DOM State-preserving move (`Node.prototype.moveBefore`) WebKit/standards-positions#375](https://github.com/WebKit/standards-positions/issues/375)
>   * Gecko: [DOM State-preserving move (`Node.prototype.moveBefore`) mozilla/standards-positions#1053](https://github.com/mozilla/standards-positions/issues/1053)
> * [x]  [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
>   
>   * https://github.com/web-platform-tests/wpt/tree/master/dom/nodes/moveBefore/tentative (need to mark as non-tentative)
> * [x]  [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
>   
>   * Chromium: https://issues.chromium.org/issues/40150299
>   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1923880
>   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=281223
> * [x]  [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: I've added the impacts-documentation label, per https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests
> * [x]  The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use.
> 
> (See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)
> 
> [Preview](https://whatpr.org/dom/1307.html) | [Diff](https://whatpr.org/dom/1307/719e45b...4b54a0d.html)

Help

dom.bs
```diff
@@ -2872,6 +2872,128 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+ <li><p>If <var>newParent</var> is a <a for=Element>shadow host</a> whose <a for=/>shadow root</a>'s
+ <a for=ShadowRoot>slot assignment</a> is "<code>named</code>" and <var>node</var> is a
+ <a>slottable</a>, then <a>assign a slot</a> for <var>node</var>.
+
+ <li><p>If <var>newParent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
+ <var>newParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> is the empty list,
+ then run <a>signal a slot change</a> for <var>newParent</var>.
+
+ <li><p>Run <a>assign slottables for a tree</a> with <var>node</var>'s <a for=tree>root</a>.
```
smaug----, line 2966: Does this happen somewhere also for the node's old parent? We need to re-assign slots also there.
(and are there tests for this?)

smaug----: I didn't immediately find any tests for moveBefore + slotchange event

smaug----:
Are there any tests for MutationObserver + moveBefore?

domfarolino:
> Are there any tests for MutationObserver + moveBefore?

I thought there were but I cannot find them, so I'll add some today.

dom.bs
domfarolino: OK I've added tests [here](https://chromium-review.googlesource.com/c/chromium/src/+/6111212) (pending) and addressed this in the spec with: https://github.com/whatwg/dom/pull/1307/commits/32a8a683e49c2772d08a622413500f69d0455f3a.

domfarolino:
OK `slotchange` and `MutationObserver` tests are added in https://github.com/web-platform-tests/wpt/pull/49810 and https://chromium-review.googlesource.com/c/chromium/src/+/6108064 (upstreamed soon), so I think everything here is resolved. I'll look at the comments on the HTML integration PR, but I think this one is ready to go.

dom.bs
```diff
@@ -2872,6 +2872,135 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
 </ol>
 
 
+<p><a lt="Other applicable specifications">Specifications</a> may define <dfn export
+id=concept-node-move-ext>moving steps</dfn> for all or some <a for=/>nodes</a>. The algorithm is
```
annevk, line 2876: The DOM standard doesn't wrap on all spaces. Phrasing-level elements need to be kept on a single line. (This applies several times in this PR.)

It also seems okay to drop the explicit ID here as we stopped doing those.

dom.bs
```diff
@@ -2872,6 +2872,135 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+passed a <a for=/>node</a> <var ignore>movedNode</var>, and a <a for=/>node</a>-or-null <var
+ignore>oldParent</var> as indicated in the <a for=/>move</a> algorithm below. Like the <span
+data-x="concept-insertion-steps-ext">insertion steps</span>, these steps must not modify the
```
annevk, line 2879: This is not how cross-references work in the DOM standard. There's no `data-x`. I'm surprised this passes CI.

annevk (commented):
I can fix these nits before merging if you want. I will merge this and the HTML PR on January 20 provided there is no further feedback.

dom.bs
```diff
@@ -2872,6 +2872,135 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+ <li><p>Let <var>oldNextSibling</var> be <var>node</var>'s <a for=tree>next sibling</a>.
+
+ <li><p><a for=set>Remove</a> <var>node</var> from <var>oldParent</var>'s <a for=tree>children</a>.
+
+ <li><p>If <var>node</var> is <a for=slottable>assigned</a>, then run <a>assign slottables</a> for
+ <var>node</var>'s <a>assigned slot</a>.
+
+ <li><p>If <var>oldParent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
+ <var>oldParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> is the empty list, then
+ run <a>signal a slot change</a> for <var>oldParent</var>.
```
smaug----, line 2940: Aren't we missing the steps needed to track the case when slot element itself if moving. In the "remove" algorithm step 14.
And I didn't see tests for moving slots.

domfarolino: Apologies for the delay on this. Yeah I think you're right. I've added the missing step in https://github.com/whatwg/dom/pull/1307/commits/87b6acbf12368302a05b22ed349d7bacd89b1507 and a test for this in https://chromium-review.googlesource.com/c/chromium/src/+/6298941 which will be upstreamed soon.

dom.bs
```diff
@@ -2881,6 +2881,145 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
 </ol>
 
 
+<p><a lt="Other applicable specifications">Specifications</a> may define
+<dfn export id=concept-node-move-ext>moving steps</dfn> for all or some <a for=/>nodes</a>. The
+algorithm is passed a <a for=/>node</a> <var ignore>movedNode</var>, and a <a for=/>node</a>-or-null
+<var ignore>oldParent</var> as indicated in the <a for=/>move</a> algorithm below. Like the
+<a>insertion steps</a>, these steps must not modify the <a>node tree</a> that
+<var>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>,
```
annevk, line 2889:
```suggestion
<var>movedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>,
```

dom.bs
```diff
@@ -2881,6 +2881,145 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+ <li><p>Let <var>oldNextSibling</var> be <var>node</var>'s <a for=tree>next sibling</a>.
+
+ <li><p><a for=set>Remove</a> <var>node</var> from <var>oldParent</var>'s <a for=tree>children</a>.
+
+ <li><p>If <var>node</var> is <a for=slottable>assigned</a>, then run <a>assign slottables</a> for
+ <var>node</var>'s <a>assigned slot</a>.
+
+ <li><p>If <var>oldParent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
+ <var>oldParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> is the empty list, then
```
annevk, line 2948:
```suggestion
 <var>oldParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> <a for=list>is empty</a>, then
```
(or something like that)

dom.bs
```diff
@@ -2881,6 +2881,145 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+ <li>
+  <p>If <var>node</var> has an <a>inclusive descendant</a> that is a <a>slot</a>:
+
+  <ol>
+   <li><p>Run <a>assign slottables for a tree</a> with <var>oldParent</var>'s <a for=tree>root</a>.
+
+   <li><p>Run <a>assign slottables for a tree</a> with <var>node</var>.
+  </ol>
+ </li>
```
annevk, line 2959:
```suggestion
```

dom.bs
```diff
@@ -2881,6 +2881,145 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+
+ <li><p>Otherwise, <a for=set>insert</a> <var>node</var> into <var>newParent</var>'s
+ <a for=tree>children</a> before <var>child</var>'s <a for=tree>index</a>.
+
+ <li><p>If <var>newParent</var> is a <a for=Element>shadow host</a> whose <a for=/>shadow root</a>'s
+ <a for=ShadowRoot>slot assignment</a> is "<code>named</code>" and <var>node</var> is a
+ <a>slottable</a>, then <a>assign a slot</a> for <var>node</var>.
+
+ <li><p>If <var>newParent</var>'s <a for=tree>root</a> is a <a for=/>shadow root</a>, and
+ <var>newParent</var> is a <a>slot</a> whose <a for=slot>assigned nodes</a> is the empty list,
```
annevk, line 2989: empty list thing again

dom.bs
```diff
@@ -3234,6 +3360,16 @@ Element includes ParentNode;
   the <a>node tree</a> are violated.
   <!-- "NotFoundError" is impossible -->
 
+ <dt><code><var>node</var> . <a method for=ParentNode lt="moveBefore()">moveBefore</a>(<var>movedNode</var>, <var>child</var>)</code>
+ <dd>
+  <p>Moves, without first removing, <var>movedNode</var> into <var>node</var> after <var>child</var>
+  if <var>child</var> is non-null; otherwise after the <a>last child</a> of <var>node</var>. This
+  methods preserves state associated with <var>movedNode</var> so that it persists after the move.
```
annevk, line 3367:
```suggestion
  method preserves state associated with <var>movedNode</var>.
```
persists seems redundant with preserves

dom.bs
```diff
@@ -3234,6 +3360,16 @@ Element includes ParentNode;
...
+
+  <p><a>Throws</a> a "{{HierarchyRequestError!!exception}}" {{DOMException}} if the constraints of
+  the <a>node tree</a> are violated.
+  <!-- "NotFoundError" is impossible -->
```
annevk, line 3371: We should probably call out that we also throw this when we cannot preserve state.

dom.bs
```diff
@@ -2881,6 +2881,145 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
+algorithm is passed a <a for=/>node</a> <var ignore>movedNode</var>, and a <a for=/>node</a>-or-null
+<var ignore>oldParent</var> as indicated in the <a for=/>move</a> algorithm below. Like the
+<a>insertion steps</a>, these steps must not modify the <a>node tree</a> that
+<var>insertedNode</var> <a>participates</a> in, create <a for=/>browsing contexts</a>,
+<a lt="fire an event">fire events</a>, or otherwise execute JavaScript. These steps may queue tasks
+to do these things asynchronously, however.
+
+
+<p>To <dfn>move</dfn> a <a for=/>node</a> <var>node</var> into a <a for=/>node</a>
+<var>newParent</var> before a <a for=/>node</a> <var>child</var>:
```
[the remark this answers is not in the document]
domfarolino: That's right, fixing. Thanks.

annevk (approved):
@smaug---- can you double check as well? Is Monday a reasonable merge day? I feel like everyone else already got sufficient time in the last couple of rounds.

smaug----:
I think this is fine.

https://issues.chromium.org/issues/400758510 is interesting. That was filed immediately when someone tried to use the API. That is more for the HTML part of the feature though. But I think opt-in behavior of the new callback is reasonable (though the callback should likely get old parent as a param, but that can be a followup)

peasantman1:
These people are using federated learning to  download information in my device

PR #1325, opened 2024-11-18
Editorial: break from the inner invoke loop rather than returning

annevk:
This is slightly cleaner and less confusing (to me anyway).

***
<a href="https://whatpr.org/dom/1325.html" title="Last updated on Nov 18, 2024, 11:00 AM UTC (08c84c0)">Preview</a> | <a href="https://whatpr.org/dom/1325/c2d5aa3...08c84c0.html" title="Last updated on Nov 18, 2024, 11:00 AM UTC (08c84c0)">Diff</a>


domenic (approved):
Although I usually like early return, I agree break is clearer here. Maybe because:

- There's a return value, so this is slight amounts of deduping.
- The return statement is immediately afterward.

PR #1326, opened 2024-11-19
Improve readability. maybe..

zhangenming:
At first I was a little confused when I read the words and it felt strange
I'm trying to make it a little more readable in the hope that it helps

before:
image

after:
image

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4656,7 +4656,7 @@ dom-Range-extractContents, dom-Range-cloneContents -->
  <i>clone children flag</i> set if <var>deep</var> is true.
 </ol>
 
-<p>A <a for=/>node</a> <var>A</var> <dfn export for=Node id=concept-node-equals>equals</dfn> a
+<p><a for=/>Node</a> <var>A</var> <dfn export for=Node id=concept-node-equals>equals</dfn>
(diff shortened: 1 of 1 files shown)


annevk:
While I appreciate the suggestion, this makes it less consistent with the language we use elsewhere.

An alternative that could maybe work is to rename the variables to _nodeA_ and _nodeB_ to distinguish them more from the leading "A". WDYT?

zhangenming:
image

zhangenming:
how about use 'the' word
image

zhangenming:
> this makes it less consistent with the language we use elsewhere.
Either solution seems to require global modification

annevk:
"The" has the same issue as removing "A". I'm not sure what you mean with global modification for my solution? I don't think there's that many variables named "A"? Or perhaps I misunderstood the problem?

zhangenming:
<var>A</var> search result : 44

zhangenming:
a node a // strange
a node nodeA // redundancy
the node a // better
node a // best

Whichever option is chosen
Will cause`this makes it less consistent with the language we use elsewhere.`
So i think the whole thing needs to be revised

PR #1334, opened 2024-12-12
Remove quadratic behavior from clone a node

### LINKED ISSUE #1219: clone with clone children flag set should append clone before cloning children
The [clone](https://dom.spec.whatwg.org/#concept-node-clone) algorithm in the DOM spec currently describes the cloning algorithm when the "clone children" flag is set as recursively cloning the children and then appending them to the cloned parent.  While, strictly, it doesn't link to [append](https://dom.spec.whatwg.org/#concept-node-append), it probably meant to, and the definition of append involves running the insertion steps on all shadow-including descendants.  This means that the algorithm described in the specification is quadratic in the tree depth, since the insertion steps are run depth times for each node.

This quadratic behavior is in fact present in current Chromium.  @mfreed7 just landed [a CL to to fix that](https://chromium-review.googlesource.com/c/4728983); it seemed like given the way insertion steps usually (although not quite always) care about whether the element is [connected](https://dom.spec.whatwg.org/#connected), the difference in ordering probably wouldn't be observable, since most of the insertion steps would only matter when the resulting tree gets appended to something that is connected.  (There are some insertion steps where this isn't true, though, but they also generally seem to care about being connected to some ancestor in particular, for example for `picture` element media selection.  I think for `picture` element selection there's a decent chance that this change is an observable behavior change in Chromium... though not observably spec-violating since the spec for `picture` stuff is quite vague about timing (through use of [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel).)

I decided to just now to look at the relevant Gecko code; `nsINode::CloneNode` calls `nsINode::Clone` which calls `nsINode::CloneAndAdopt` which appears to have the same behavior as the *new* Chromium behavior.

Given that:

* this alternative behavior is faster
* one engine has already adopted it, and another is trying to
* it seems likely that the difference is Web-observable, and that chance will increase over time

I think it seems likely to be worthwhile to change the DOM spec to describe element cloning in this new way.

The disadvantage is that I think it will make the spec a bit more complicated.  (Thus, I'm filing an issue rather than immediately writing a PR, since the PR would take a good bit more time to write!)

### LINKED ISSUE #1220: Provide a "deep clone" shorthand
The "with the clone children flag set to true" is kinda clunky in prose... it would be amazing to have "deeply clone" shorthand that does the same thing. 

For example, instead of:

> Let _clone_ be a clone of _someElementOrFragment_, with the clone children flag set to true.

Nicer: 
> Let _clone_ be a **deep clone** of _someElementOrFragment_.

annevk:
Strictly speaking this should be an editorial change, but as it's fairly significant not marking it as such. This improves integration with HTML as well.

As part of this change we move the shadow tree behavior under creating the element clone as that is a more logical fit.

Unfortunately subtree has to be passed everywhere as HTML needs it for its template element logic. That's also why clone a node's children is exported.

Corresponding HTML PR: SOON.

Fixes #1219. Closes #1220.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …


dom.bs
```diff
@@ -4573,11 +4601,31 @@ dom-Range-extractContents, dom-Range-cloneContents -->
...
+     <a for=Element>shadow root</a>'s <a for=ShadowRoot>mode</a>, true, <var>node</var>'s
+     <a for=Element>shadow root</a>'s <a for=ShadowRoot>serializable</a>, <var>node</var>'s
+     <a for=Element>shadow root</a>'s <a for=ShadowRoot>delegates focus</a>, and <var>node</var>'s
+     <a for=Element>shadow root</a>'s <a for=ShadowRoot>slot assignment</a>.
+
+     <li><p>Set <var>copy</var>'s <a for=Element>shadow root</a>'s <a for=ShadowRoot>declarative</a>
+     to <var>node</var>'s <a for=Element>shadow root</a>'s <a for=ShadowRoot>declarative</a>.
+
+     <li><p><a>Clone a node's children</a> given <var>node</var>'s <a for=Element>shadow root</a>,
+     <var>copy</var>'s <a for=Element>shadow root</a>, <var>document</var>, and <var>subtree</var>.
```
smaug----, line 4622: Don't we want shadow root cloning to happen after appending to avoid calling insertion steps again when the host is added to its parent?

dom.bs
```diff
@@ -4532,33 +4532,86 @@ each <a for=tree>descendant</a> <a>exclusive <code>Text</code> node</a> <var>nod
...
+   <a for=Element>shadow root</a>'s <a for=ShadowRoot>delegates focus</a>, and <var>node</var>'s
+   <a for=Element>shadow root</a>'s <a for=ShadowRoot>slot assignment</a>.
+
+   <li><p>Set <var>copy</var>'s <a for=Element>shadow root</a>'s <a for=ShadowRoot>declarative</a>
+   to <var>node</var>'s <a for=Element>shadow root</a>'s <a for=ShadowRoot>declarative</a>.
+
+   <li><p>For each <var>child</var> of <var>node</var>'s <a for=Element>shadow root</a>'s
+   <a for=tree>children</a>, in <a>tree order</a>: <a>clone a node and append</a> given
+   <var>child</var>, <var>copy</var>'s <a for=Element>shadow root</a>, <var>document</var>, and
+   <var>subtree</var>.
```
smaug----, line 4594: In Gecko shadow root attaching happens after going through children.
And the same seems to happen in Chrome too https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/element.cc;l=689;drc=fe99dee25a6b65dbe6c9e97099334e9c027a24a1;bpv=0;bpt=1

The order of slotchange events is the same in Firefox and Chrome (h6, div, span), Safari has - I was told - div, span, h6.

annevk: Yeah it is, but is that the result of cloning order or something else? As per the discussion in #695 there's already a difference between engines.

I can restore the original order here though. That seems better for now either way.

dom.bs
```diff
@@ -4533,33 +4533,82 @@ each <a for=tree>descendant</a> <a>exclusive <code>Text</code> node</a> <var>nod
...
+ <var>subtree</var> as parameters.
+
+ <li><p>If <var>subtree</var> is true, then for each <var>child</var> of <var>node</var>'s
+ <a for=tree>children</a>, in <a>tree order</a>: <a>clone a node</a> given <var>child</var> with
+ <a for="clone a node"><i>document</i></a> set to <var>document</var>,
+ <a for="clone a node"><i>subtree</i></a> set to <var>subtree</var>, and
+ <a for="clone a node"><i>parent</i></a> set to <var>copy</var>.
+
+ <li>
+  <p>If <var>node</var> is an <a for=/>element</a>, <a for=Element>shadow host</a>, and its
```
dbaron, line 4574: This insertion of "an element" makes the clause not really seem to parse properly (as English) anymore.  I'm not quite sure what the change relative to the old wording was intended to be, but I can't tell what the new wording means.

dbaron (approved):
Sorry for the delay getting to this.  This looks good to me except for a single sentence (see below).

dom.bs
```diff
@@ -4533,33 +4533,82 @@ each <a for=tree>descendant</a> <a>exclusive <code>Text</code> node</a> <var>nod
...
- <li><a for=/>Assert</a>: <var>node</var> is not a <a for=/>document</a> or <var>node</var> is
+ <li><p>Let <var>copy</var> be the result of <a>cloning a single node</a> given <var>node</var> and
  <var>document</var>.
 
+ <li><p>If <var>parent</var> is non-null, then <a for=/>append</a> <var>copy</var> to
+ <var>parent</var>.
+
+ <li><p>Run any <a>cloning steps</a> defined for <var>node</var> in
+ <a>other applicable specifications</a> and pass <var>node</var>, <var>copy</var>, and
+ <var>subtree</var> as parameters.
```
smaug----, line 4565: Gecko runs cloning steps as part of cloning a single node, and I think webkit does that too https://searchfox.org/wubkat/rev/72adbb9f99534551a728acb270d4f111da5e9515/Source/WebCore/dom/Element.cpp#648,655,663 
and chrome too https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/element.cc;l=755;drc=f5e280b6c11f41545154266cd2317d3c730e9b30;bpv=0;bpt=1

smaug---- (approved):
I think cloning steps should be before append

PR #1342, opened 2024-12-17
Add definition for composed live range

dizhang168:
As discussed at TPAC 2024, the specification for the new getComposedRanges() API need the introduction of the new definition "Composed live range" [1], which should react to mutations. This PR attempts to define that new concept. Once landed, we will add the spec language that uses composed live range in the Selection API. An overview of the spec changes necessary can be found at [2].

[1] https://github.com/w3c/selection-api/issues/2#issuecomment-2427784793
[2] https://github.com/dizhang168/shadow-dom-selection/blob/main/selection-api-spec-changes.md

   * Chromium: Positive
   * Webkit: Positive (https://github.com/mozilla/standards-positions/issues/1055)
   * Gecko: Positive (https://github.com/WebKit/standards-positions/issues/376)
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -22,6 +22,8 @@ urlPrefix: https://www.w3.org/TR/xml-names/#NT-
   text: QName; url: QName
 url: https://w3c.github.io/DOM-Parsing/#dom-range-createcontextualfragment
  type: method; text: createContextualFragment(); for: Range
+url: https://w3c.github.io/selection-api/#dom-selection-getrangeat
+ type: method; text: getRangeAt(); for: Range
 type: interface
  url: https://w3c.github.io/touch-events/#idl-def-touchevent
   text: TouchEvent
@@ -3027,6 +3029,14 @@ optional <i>suppress observers flag</i>, run these steps:
  <li><p>For each <a>live range</a> whose <a for=range>end node</a> is an <a>inclusive descendant</a>
  of <var>node</var>, set its <a for=range>end</a> to (<var>parent</var>, <var>index</var>).
 
+ <li><p>For each <a>composed live range</a> whose <a for=range>start node</a> is a
+ <a>shadow-including inclusive descendant</a> of <var>node</var>, set its <a for=range>start</a> to
+ (<var>parent</var>, <var>index</var>).
+
+ <li><p>For each <a>composed live range</a> whose <a for=range>end node</a> is an
+ <a>shadow-including inclusive descendant</a> of <var>node</var>, set its <a for=range>end</a> to
+ (<var>parent</var>, <var>index</var>).
+
  <li><p>For each <a>live range</a> whose <a for=range>start node</a> is <var>parent</var> and
  <a for=range>start offset</a> is greater than <var>index</var>, decrease its
  <a for=range>start offset</a> by 1.
@@ -5743,7 +5753,8 @@ are:
 <hr>
 
 <p>The <dfn method for=Document><code>createRange()</code></dfn> method steps are to return a new
-<a>live range</a> with (<a>this</a>, 0) as its <a for=range>start</a> an <a for=range>end</a>.
+{{Range}} <a>live range</a> with (<a>this</a>, 0) as its <a for=range>start</a> and <a
+for=range>end</a>.
 
 <p class=note>The {{Range/Range()}} constructor can be used instead.
 
@@ -7878,11 +7889,11 @@ range.setEnd(secondText, 4)
 <a>node tree</a> above cannot be represented by a <a>range</a>. <a>Ranges</a> are only useful for
 <a for=/>nodes</a>.
 
-<p>{{Range}} objects, unlike {{StaticRange}} objects, are affected by mutations to the
-<a>node tree</a>. Therefore they are also known as <a>live ranges</a>. Such mutations will not
-invalidate them and will try to ensure that it still represents the same piece of content.
-Necessarily, a <a>live range</a> might itself be modified as part of the mutation to the
-<a>node tree</a> when, e.g., part of the content it represents is mutated.
+<p>{{Range}} objects, unlike {{StaticRange}} objects, are affected by mutations to the <a>node
+tree</a>. Therefore they are <a>live ranges</a>. Such mutations will not invalidate them and will
+try to ensure that it still represents the same piece of content. Necessarily, a <a>live range</a>
+might itself be modified as part of the mutation to the <a>node tree</a> when, e.g., part of the
+content it represents is mutated.
 
 <p class=note>See the <a for=/>insert</a> and <a for=/>remove</a> algorithms, the
 {{Node/normalize()}} method, and the <a>replace data</a> and <a lt="split a Text node">split</a>
@@ -8109,12 +8120,22 @@ interface Range : AbstractRange {
 };
 </pre>
 
-<p>Objects implementing the {{Range}} interface are known as
-<dfn export id=concept-live-range>live ranges</dfn>.
+<p>A <dfn export id=concept-live-range>live range</dfn> is a <a>range</a> that is affected by
+mutations to the <a>node tree</a>.</p>
+
+<p>Objects implementing the {{Range}} interface are <a>live ranges</a>.
 
 <p class=note>Algorithms that modify a <a>tree</a> (in particular the <a for=/>insert</a>,
 <a for=/>remove</a>, <a>replace data</a>, and <a lt="split a Text node">split</a> algorithms) modify
-<a>live ranges</a> associated with that <a>tree</a>.
+<a>live ranges</a> associated with that <a>tree</a>.</p>
+
+<p>A <dfn export id=concept-composed-live-range>composed live range</dfn> is a <a>live range</a>
+that has one associated {{Range}} object - <dfn export
+id=concept-composed-live-range-cached-live-range for="composed live range">cached live
+range</dfn>.</p>
+
+<p class=note>The <a>cached live range</a> is used to maintain backward compatibility with the
+{{getRangeAt}} API.</p>
 
 <p>The <dfn export id=concept-range-root for="live range">root</dfn> of a <a>live range</a> is the
 <a for=tree>root</a> of its <a for=range>start node</a>.
@@ -8184,7 +8205,7 @@ but not its <a for=range>end node</a>, or vice versa.
 
 <dl class=domintro>
  <dt><code><var>range</var> = new <a constructor>Range()</a></code>
- <dd>Returns a new <a>live range</a>.
+ <dd>Returns a new {{Range}} <a>live range</a>.
 </dl>
 
 <p>The <dfn constructor for=Range lt="Range()"><code>new Range()</code></dfn> constructor steps are
@@ -8249,6 +8270,10 @@ steps:
 
      <li>Set <var>range</var>'s
      <a for=range>start</a> to <var>bp</var>.
+
+     <li>If <var>range</var> is the associated <a>cached live range</a> of a <a>composed live
+     range</a> <var>composed live range</var>, set <var>composed live range</var>’s <a
+     for="range">start</a> to <var>bp</var>.
     </ol>
    <dt>If these steps were invoked as "set the end"
    <dd>
@@ -8264,6 +8289,10 @@ steps:
 
      <li>Set <var>range</var>'s
      <a for=range>end</a> to <var>bp</var>.
+
+     <li>If <var>range</var> is the associated <a>cached live range</a> of a <a>composed live range</a>
+     <var>composed live range</var>, set <var>composed live range</var>’s <a for="range">end</a> to
+     <var>bp</var>.
     </ol>
   </dl>
 </ol>
@@ -8544,8 +8573,8 @@ method steps are:
  (<var>new node</var>, <var>new offset</var>).
 </ol>
 
-<p>To <dfn export id=concept-range-extract for="live range">extract</dfn> a <a>live range</a>
-<var>range</var>, run these steps:
+<p>To <dfn export id=concept-range-extract for="live range">extract</dfn> a {{Range}} <a>live
+range</a> <var>range</var>, run these steps:
 
 <ol>
  <li><p>Let <var>fragment</var> be a new {{DocumentFragment}} <a for=/>node</a> whose
@@ -8721,7 +8750,7 @@ method steps are:
    <li><a>Append</a> <var>clone</var>
    to <var>fragment</var>.
 
-   <li>Let <var>subrange</var> be a new <a>live range</a>
+   <li>Let <var>subrange</var> be a new {{Range}} <a>live range</a>
    whose <a for=range>start</a> is
    (<var>original start node</var>, <var>original start offset</var>) and
    whose <a for=range>end</a> is
@@ -8775,7 +8804,7 @@ method steps are:
    <li><a>Append</a> <var>clone</var>
    to <var>fragment</var>.
 
-   <li>Let <var>subrange</var> be a new <a>live range</a>
+   <li>Let <var>subrange</var> be a new {{Range}} <a>live range</a>
    whose <a for=range>start</a> is
    (<var>last partially contained child</var>, 0) and whose
    <a for=range>end</a> is
@@ -8800,7 +8829,7 @@ result of <a for="live range">extracting</a> <a>this</a>.
 
 <p>To
 <dfn export id=concept-range-clone for="live range" lt="clone the contents|cloning the contents">clone the contents</dfn>
-of a <a>live range</a> <var>range</var>, run these steps:
+of a {{Range}} <a>live range</a> <var>range</var>, run these steps:
 
 <ol>
  <li><p>Let <var>fragment</var> be a new {{DocumentFragment}} <a for=/>node</a> whose
@@ -8928,7 +8957,7 @@ of a <a>live range</a> <var>range</var>, run these steps:
    <li><a>Append</a> <var>clone</var>
    to <var>fragment</var>.
 
-   <li>Let <var>subrange</var> be a new <a>live range</a>
+   <li>Let <var>subrange</var> be a new {{Range}} <a>live range</a>
    whose <a for=range>start</a> is
    (<var>original start node</var>, <var>original start offset</var>) and
    whose <a for=range>end</a> is
@@ -8986,7 +9015,7 @@ of a <a>live range</a> <var>range</var>, run these steps:
    <li><a>Append</a> <var>clone</var>
    to <var>fragment</var>.
 
-   <li>Let <var>subrange</var> be a new <a>live range</a>
+   <li>Let <var>subrange</var> be a new {{Range}} <a>live range</a>
    whose <a for=range>start</a> is
    (<var>last partially contained child</var>, 0) and whose
    <a for=range>end</a> is
@@ -9006,7 +9035,7 @@ of a <a>live range</a> <var>range</var>, run these steps:
 result of <a for="live range">cloning the contents</a> of <a>this</a>.
 
 <p>To <dfn export id=concept-range-insert for="live range">insert</dfn> a <a for=/>node</a>
-<var>node</var> into a <a>live range</a> <var>range</var>, run these steps:
+<var>node</var> into a {{Range}} <a>live range</a> <var>range</var>, run these steps:
 
 <ol>
  <li>If <var>range</var>'s <a for=range>start node</a> is a {{ProcessingInstruction}} or {{Comment}}
@@ -9156,7 +9185,8 @@ check first thing, which matches everyone but Firefox.
 </ol>
 
 <p>The <dfn method for=Range><code>cloneRange()</code></dfn> method steps are to return a new
-<a>live range</a> with the same <a for=range>start</a> and <a for=range>end</a> as <a>this</a>.
+{{Range}} <a>live range</a> with the same <a for=range>start</a> and <a for=range>end</a> as
+<a>this</a>.
 
 <p>The <dfn method for=Range><code>detach()</code></dfn> method steps are to do nothing.
 <span class=note>Its functionality (disabling a {{Range}} object) was removed, but the method itself
@@ -10408,6 +10438,7 @@ David Håsäther,
 David Hyatt,
 Deepak Sherveghar,
 Dethe Elza,
+Di Zhang,
 Dimitri Glazkov,
 Domenic Denicola,
 Dominic Cooney,


dom.bs
```diff
@@ -3027,6 +3029,14 @@ optional <i>suppress observers flag</i>, run these steps:
  <li><p>For each <a>live range</a> whose <a for=range>end node</a> is an <a>inclusive descendant</a>
  of <var>node</var>, set its <a for=range>end</a> to (<var>parent</var>, <var>index</var>).
 
+ <li><p>For each <a>composed live range</a> whose <a for=range>start node</a> is a
+ <a>shadow-including inclusive descendant</a> of <var>node</var>, set its <a for=range>start</a> to
+ (<var>parent</var>, <var>index</var>).
+
+ <li><p>For each <a>composed live range</a> whose <a for=range>end node</a> is an
+ <a>shadow-including inclusive descendant</a> of <var>node</var>, set its <a for=range>end</a> to
+ (<var>parent</var>, <var>index</var>).
```
annevk, line 3038: This will need to be rebased once we land `moveBefore()`.

However, I also think this needs to be reconciled with the "for each live range" above as we don't want to do duplicate work.

dom.bs
```diff
@@ -5743,7 +5753,8 @@ are:
 <hr>
 
 <p>The <dfn method for=Document><code>createRange()</code></dfn> method steps are to return a new
-<a>live range</a> with (<a>this</a>, 0) as its <a for=range>start</a> an <a for=range>end</a>.
+{{Range}} <a>live range</a> with (<a>this</a>, 0) as its <a for=range>start</a> and <a
+for=range>end</a>.
```
annevk, line 5757: I would say "Range object" here. We also don't wrap in phrasing elements.

dom.bs
```diff
@@ -7878,11 +7889,11 @@ range.setEnd(secondText, 4)
 <a>node tree</a> above cannot be represented by a <a>range</a>. <a>Ranges</a> are only useful for
 <a for=/>nodes</a>.
 
-<p>{{Range}} objects, unlike {{StaticRange}} objects, are affected by mutations to the
-<a>node tree</a>. Therefore they are also known as <a>live ranges</a>. Such mutations will not
-invalidate them and will try to ensure that it still represents the same piece of content.
-Necessarily, a <a>live range</a> might itself be modified as part of the mutation to the
-<a>node tree</a> when, e.g., part of the content it represents is mutated.
+<p>{{Range}} objects, unlike {{StaticRange}} objects, are affected by mutations to the <a>node
```
annevk, line 7892: No wrapping in phrasing elements. (Please check the rest of the diff yourself for this.)

dom.bs
```diff
@@ -8109,12 +8120,22 @@ interface Range : AbstractRange {
...
+
+<p>Objects implementing the {{Range}} interface are <a>live ranges</a>.
 
 <p class=note>Algorithms that modify a <a>tree</a> (in particular the <a for=/>insert</a>,
 <a for=/>remove</a>, <a>replace data</a>, and <a lt="split a Text node">split</a> algorithms) modify
-<a>live ranges</a> associated with that <a>tree</a>.
+<a>live ranges</a> associated with that <a>tree</a>.</p>
+
+<p>A <dfn export id=concept-composed-live-range>composed live range</dfn> is a <a>live range</a>
+that has one associated {{Range}} object - <dfn export
```
annevk, line 8133: This needs to use "—".

dom.bs
```diff
@@ -8544,8 +8573,8 @@ method steps are:
  (<var>new node</var>, <var>new offset</var>).
 </ol>
 
-<p>To <dfn export id=concept-range-extract for="live range">extract</dfn> a <a>live range</a>
-<var>range</var>, run these steps:
+<p>To <dfn export id=concept-range-extract for="live range">extract</dfn> a {{Range}} <a>live
+range</a> <var>range</var>, run these steps:
```
annevk, line 8577: I would use Range object again here.

dom.bs
annevk: Shouldn't this update the cached live range as well?

dom.bs
```diff
@@ -8109,12 +8120,22 @@ interface Range : AbstractRange {
...
 
 <p class=note>Algorithms that modify a <a>tree</a> (in particular the <a for=/>insert</a>,
 <a for=/>remove</a>, <a>replace data</a>, and <a lt="split a Text node">split</a> algorithms) modify
-<a>live ranges</a> associated with that <a>tree</a>.
+<a>live ranges</a> associated with that <a>tree</a>.</p>
+
+<p>A <dfn export id=concept-composed-live-range>composed live range</dfn> is a <a>live range</a>
+that has one associated {{Range}} object - <dfn export
+id=concept-composed-live-range-cached-live-range for="composed live range">cached live
+range</dfn>.</p>
```
annevk, line 8135: Why is this called cached live range?

annevk (commented):
@smaug---- you should review this as well in due course. And @rniwa too I suspect.

dom.bs
dizhang168: The cached live range should be updated by the existing "For each live range ..." rules. The cached live range is not composed and hence shouldn't consider for shadow-including inclusive descendant.

dom.bs
dizhang168: I wanted to give it a name so it doesn't get confusion between the "composed" live range and its associated "live range". Further, it is cached within the composed live range. I am open to naming suggestions.

dom.bs
annevk: Good call on the cached live range. But don't the existing rules also update the "composed live range" (because a composed live range is a live range)?

dom.bs
annevk: Maybe "legacy selection range"?

dom.bs
dizhang168: Yes, the rule will be applied twice. I think that's the expected behavior. This will help to keep mutation results consistent, especially when we are not crossing shadow boundaries.

dizhang168:
I have opened the draft PR for Selection API to use the definition described here, to help visualize:
https://github.com/w3c/selection-api/pull/345

dom.bs
annevk: I'm not sure I understand. Can you elaborate in which case it helps? It just looks inefficient to me.

dizhang168: Hum, I think my original statement that the rule should be applied twice is wrong. Suppose the document stores a list of all live ranges, including cached live range. For each of them, it will update per mutation rule. Then if it is a cached live range and hence, attached to a composed live range, it will update the composed live range per the ["set the start or end" algorithm](https://whatpr.org/dom/1342.html#concept-range-bp-set) redefined in this PR. If so, the composed live range do not need be updated again by the mutation, as it is already up to date.

Now, to avoid inefficiency, we could change every "For each live range" to "For each {{Range}} object".

dom.bs
domfarolino: I think "legacy selection range" sounds good, to make clear that it's only maintained for behavior expected by the Selection API. I might also go further and rename "composed live range" to "composed selection range" or something, to also indicate that this concept only exists for Selection API purposes.. thoughts?

dom.bs
domfarolino: Note that the "set the start or end" algorithm is not invoked from #concept-node-remove at all, so it is unrelated. So with that out of the way, it seems that for all composed ranges that are ordinary inclusive descendants of the to-be-removed node, these composed ranges would get updated "twice" (idempotently): first by steps 4/5, second by steps 6/7. In the case where the composed range _is_ a shadow-inclusive descendant of the removed node but not an ordinary inclusive descendant, than only steps 6/7 will apply, and there's no redundancy.

So it's inefficient as @annevk points out, but not broken. Should we fix this? Does it matter?

dom.bs
```diff
@@ -8141,12 +8149,21 @@ interface Range : AbstractRange {
 };
 </pre>
 
-<p>Objects implementing the {{Range}} interface are known as
-<dfn export id=concept-live-range>live ranges</dfn>.
+<p>A <dfn export id=concept-live-range>live range</dfn> is a <a>range</a> that is affected by
+mutations to the <a>node tree</a>.</p>
+
+<p>Objects implementing the {{Range}} interface are <a>live ranges</a>.
```
domfarolino, line 8155: Before this PR, we have the following layout of definitions:

 - "range" == `AbstractRange` IDL object
 - "live range" == `Range` IDL object

After this PR, we have:

 - "range" == `AbstractRange` IDL object (no changes)
 - "live range" == any "range" (`AbstractRange`) responsive to mutations
    - "live range" includes `Range` IDL objects
    - "live range" includes "composed live range" `AbstractRange` IDL

So what we've done is make "live range" now include `AbstractRange` objects. I think you've done this to avoid making "composed live range" a full-blown `Range` IDL object, since it doesn't need to be one—as we discussed, implementations appropriately use some internal non-IDL object to represent a "composed live range". But even with your changes I think "composed live range" is still an IDL object—it's just an `AbstractRange` instead of a `Range`.

So I think the definition rejiggering in this PR is kinda half way in between two solutions. We can either:

 1. Stick to our guns that "composed live range" should NOT be an IDL object, and do what Fetch does where we have IDL objects like [`Request`](https://fetch.spec.whatwg.org/#request) and also "internal" spec concepts like [#concept-request](https://fetch.spec.whatwg.org/#concept-request). In our case we'd have `AbstractRange` and then an internal `#concept-range`-like struct. "composed live range" would be-a `#concept-range`, and not an IDL object. It would likely hold a "cached" `Range` inside.
 2. Decide we're OK with "composed live range" being an IDL object in the spec, even though it never is one in implementation, nor is it ever web-exposed. In that case, we can either keep it as an `AbstractRange` IDL object which it is in this PR. OR we can simplify this PR a bit and leave the definition of "live range" alone—then we'd make "composed live range" a `Range` IDL object that just so happens to never be exposed to script directly. That feels weird since there isn't a reason for "composed live range" to be an IDL object besides it would make this PR smaller/easier.

The details of these options are discussed in https://github.com/w3c/selection-api/issues/2#issuecomment-2599729387 and https://github.com/w3c/selection-api/issues/2#issuecomment-2428009143, and I'd love @annevk's thoughts.

dom.bs
```diff
@@ -8273,29 +8290,53 @@ steps:
...
      <a for="live range">root</a> is not equal to
      <var>node</var>'s <a for=tree>root</a>,
-     or if <var>bp</var> is
-     <a for="boundary point">after</a> the
-     <var>range</var>'s <a for=range>end</a>, set
-     <var>range</var>'s <a for=range>end</a>
+     set <var>range</var>'s <a for=range>end</a>
      to <var>bp</var>.
 
+     <li>Otherwise, if <var>bp</var> is
```
domfarolino, line 8296: This doesn't need to be tucked behind an "Otherwise", right? Like it doesn't need to be exclusive from the first condition above I think, right? Plus I think it's confusing to have "Otherwise [....]. If" on the same line, when I think they can just be separate conditions. So what do you think about the following:

1. (same as what you have)
2. If bp is after the range's end, then:
    1. Set range's end to bp
    2. If _range_ is the cached live range of a composed live range _composed live range_, then set _composed live range_'s end to bp
3. (same as what you have)

dom.bs
```diff
@@ -8273,29 +8290,53 @@ steps:
...
+     <a for="composed live range">cached live range</a> of a
+     <a>composed live range</a> <var>composed live range</var>,
+     set <var>composed live range</var>’s
+     <a for="range">end</a> to <var>bp</var>.
+
      <li>Set <var>range</var>'s
      <a for=range>start</a> to <var>bp</var>.
+     If <var>range</var> is the
+     <a for="composed live range">cached live range</a> of a
+     <a>composed live range</a> <var>composed live range</var>,
```
domfarolino, line 8310: We can pull the _composed live range_ variable out early on. How about we detect it and set this variable right after step 2 or step 3, before we branch off into the "set the start" or "set the end" step 4 branches? That should be less redundant.

dom.bs
```diff
@@ -7911,10 +7919,10 @@ range.setEnd(secondText, 4)
...
 
 <p>{{Range}} objects, unlike {{StaticRange}} objects, are affected by mutations to the
-<a>node tree</a>. Therefore they are also known as <a>live ranges</a>. Such mutations will not
-invalidate them and will try to ensure that it still represents the same piece of content.
-Necessarily, a <a>live range</a> might itself be modified as part of the mutation to the
-<a>node tree</a> when, e.g., part of the content it represents is mutated.
+<a>node tree</a>. Therefore they are <a>live ranges</a>. Such mutations will not invalidate them and
+will try to ensure that it still represents the same piece of content. Necessarily, a
+<a>live range</a> might itself be modified as part of the mutation to the <a>node tree</a> when,
+e.g., part of the content it represents is mutated.
```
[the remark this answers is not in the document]
dizhang168: The only change is to remove "also known as", as I felt that was implying only {{Range}} objects are considered live ranges. Re-reading and this change is probably not necessary.

dom.bs
dizhang168: It was to avoid redundancy in the implementation (i.e. when the roots are different and we already collapsed the range's end to bp, no need to collapse again).

I like your format suggestion, I will change to that.

dom.bs
dizhang168: I renamed them to "composed selection range" and "legacy selection range".

dom.bs
annevk: Redefining live range seems potentially problematic: https://dontcallmedom.github.io/webdex/l.html#live%20ranges%40%40dom%25%25dfn

We'd also have to go through all callers in the DOM specification to be sure this doesn't break anything.

dom.bs
annevk: I think we should fix it as it's somewhat confusing. We can exclude the composed ranges from the first set of checks.

dom.bs
dizhang168: The only references to "live ranges" are from Selection API and Highlight API. In both case, they mean to refer to Range object. We can update those references, although maybe it would be simpler to keep the current live ranges definition and change "composed live range" to:

> A **composed live range** is a [range](https://dom.spec.whatwg.org/#concept-range) that is affected by mutations to the node tree and has an associated [live range](https://dom.spec.whatwg.org/#concept-live-range) called a **legacy live range**.

dom.bs
```diff
@@ -8273,29 +8296,55 @@ steps:
...
+     <li>Otherwise, if <var>bp</var> is
+     <a for="boundary point">after</a> the
+     <var>range</var>'s <a for=range>end</a>, then:
+      <ol>
+       <li>Set <var>range</var>'s <a for=range>end</a>
+       to <var>bp</var>.
+
+       <li>If <var>selection range</var> is not null,
+       set <var>selection range</var>’s
+       <a for="range">end</a> to <var>bp</var>.
```
sefeng211, line 8311: It looks like we only collapse the `composed selection range` when the nodes are in the same tree. Shouldn't we also compare the nodes if they are in different tree (with a new algorithm like `compose-after`), and collapse the `composed selection range` if the new start is `compose-after` the existing end? This also applies to the setting the end case.

dom.bs
```diff
@@ -8141,12 +8149,21 @@ interface Range : AbstractRange {
...
+<p>Objects implementing the {{Range}} interface are <a>live ranges</a>.
 
 <p class=note>Algorithms that modify a <a>tree</a> (in particular the <a for=/>insert</a>,
 <a for=/>remove</a>, <a>replace data</a>, and <a lt="split a Text node">split</a> algorithms) modify
-<a>live ranges</a> associated with that <a>tree</a>.
+<a>live ranges</a> associated with that <a>tree</a>.</p>
+
+<p>A <dfn export id=concept-composed-selection-range>composed selection range</dfn> is a
+<a>live range</a> that has an associated {{Range}} object, a
+<dfn export id=concept-legacy-selection-range for="composed selection range">legacy selection range</dfn>.</p>
```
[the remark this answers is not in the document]
annevk: It's not the `Range` object that's legacy, it's having to keep around a pretend range for the purposes of the selection API as it was designed pre-shadow-trees that's kinda legacy.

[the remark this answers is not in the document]
smaug----: Sure, but I don't still like it. It is not legacy, in the sense that there would be some better thing to replace it or so (in case of light dom). "Legacy" doesn't explain what the range is for.

dom.bs
domfarolino: I think @dizhang168's latest comment is a good direction to go. It's basically option (2) in https://github.com/whatwg/dom/pull/1342#discussion_r1932710902 and satisfies @annevk's concerns about changing the definition of "live range". I vote we go that way.

dom.bs
```diff
@@ -8265,6 +8282,12 @@ steps:
  <a>boundary point</a>
  (<var>node</var>, <var>offset</var>).
 
+ <li>If <var>range</var> is the <a for="composed selection range">legacy selection range</a> of a
```
domfarolino, line 8285: Could you add a clarification to the algorithm declaration of "start the set|start the end" to give the "range" parameter a type? I think the intention is that this algorithm should only ever be called on "live range" objects, and never a "composed selection range", so let's give it a "live range" type if you believe that's accurate.

dom.bs
domfarolino: At first I thought we wouldn't want to unconditionally collapse the associated `composed selection range` if, while we're "setting the start", we discover that the proposed start _bp_ is actually after the end. I thought we wouldn't want to do this because we need to be able to support case like `setBaseAndExtent()` where the selection would end up with the ["backwards" direction](https://developer.mozilla.org/en-US/docs/Web/API/Selection/direction). But then I realized in that case, the Selection API spec just calls this DOM algorithm with the start/end nodes inverted so that it doesn't collapse like this—so basically it's the Selection API's responsibility for handling that.

So yeah I think I agree there's a problem here. Specifically, if we end up calling `getSelectionRange.getRangeAt(0).setStart(nodeAfterCurrentEnd, 0)`, then we will correctly collapse the `composed selection range`'s "legacy live range", but not the `composed selection range` itself.

Maybe we should be collapsing the composed selection range in this case, unless the current spec prose is intentional to support some use case I'm not thinking of. WDYT @dizhang168?

dom.bs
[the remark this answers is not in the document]
sefeng211: maybe the `legacy selection range` could be `same-tree selection range`, and `composed selection range` could be `composed-tree selection range`?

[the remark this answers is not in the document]
domfarolino: same-tree selection range sounds kinda nice. I personally think "composed selection range" is fine as-is.

dom.bs
dizhang168: Sounds good. I will update the PR accordingly.

dom.bs
dizhang168: Changed to:
`<dfn export id=concept-range-bp-set lt="set the start|set the end" for="live range"> ... of a <a>live range</a> <var>range</var>.`

dom.bs
dizhang168: Good catch, I should update this algorithm to collapse the selection range's endpoints if we are crossing documents.

I don't think we need to add a new "composed after/before" concept though. I have updated the algorithm such that the first condition checks if we are in the same tree. If bp and range's end are in different documents, it would match step 1, collapse and will not reach the "otherwise" step 2.

1. If range’s root is not equal to node’s root, then:
  1.1. Set range’s end to bp.
  1.2. If selection range is not null and range’s start node’s node document is not node’s node document, set selection range’s end to bp.
2. Otherwise, if bp is after the range’s end, then:
  2. 1Set range’s end to bp.
  2.2 If selection range is not null, set selection range’s end to bp.
3. Set range’s start to bp.
4. If selection range is not null, set selection range’s start to bp.

For the second point about how we need to be careful calling "before/after" in Selection API spec, I agree. Let's make sure we only compare "before/after" with shadow hosts if the node is in a shadow tree.

dom.bs
```diff
@@ -8273,29 +8296,55 @@ steps:
...
+       set <var>selection range</var>’s
+       <a for="range">end</a> to <var>bp</var>.
+      </ol>
+
      <li>Set <var>range</var>'s
      <a for=range>start</a> to <var>bp</var>.
+
+     <li>If <var>selection range</var> is not null,
+     set <var>selection range</var>’s
+     <a for="range">start</a> to <var>bp</var>.
```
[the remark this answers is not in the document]
dizhang168: Yes, the fact that we have a collapsed legacy live range within a shadow DOM is a shortcoming of the existing Selection API. We are keeping it this way for backward compatibility, but hopefully the new changes will feel more natural to the users and they will depend on getComposedRanges() to get the right selection endpoints within the document.

dom.bs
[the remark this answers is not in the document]
dizhang168: Listing the naming options:
1. "composed live range" and "cached live range"
2. "composed selection range" and "legacy selection range"
3. "composed-tree selection range" and "same-tree selection range"
4. "composed selection range" and "same-tree selection range"
5. "composed selection range" and "selection range"

All these names seem ok to me, I will let the majority decide.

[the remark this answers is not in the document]
annevk: Based on an email conversation I'm a bit less sure what the semantics of this range are. Say you have an element with a shadow root that contains the word "Test". The end user selects "es". `getSelection()` should not provide access to this shadow root, but apparently the set of changes proposed here do not enforce that. Or am I misunderstanding something?

dom.bs
sefeng211: I read the updated algorithm won't collapse the example given Dominic (`getSelectionRange.getRangeAt(0).setStart(nodeInDifferentRootAfterCurrentEnd, 0)`) because the nodes are still in the same document, am I missing something? So now the `start node` of the `composed selection range` is after the `end node` of it?

sefeng211: What about disconnected nodes, it also won't collapse the composed selection range?

dizhang168: disconnected means it would be in a different document. It would match 1.2 and the selection composed range will be collapsed to {nodeInDifferentTree, 0}.

sefeng211: I thought disconnected could still mean the nodes are in the same document, like a new node, but it's not added to three yet?

dizhang168: That is a good point, to take care of disconnected nodes, I will change the above condition 1.2 to check for shadow-inclusive root instead of document.

dom.bs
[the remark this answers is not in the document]
dizhang168: I created 2 issues to track the above conversation:
https://github.com/whatwg/dom/issues/1363 for the naming choices
https://github.com/whatwg/dom/issues/1362 for getSelection() still giving access to the shadow root
Lets discuss these topics at the upcoming Editing WG or WHATNOT meeting.

domfarolino (approved):
I think this PR looks good; I'll let Anne comment on editorial/wrapping stuff, as the wrapping seems odd, but it was odd before the PR anyways. But I want to just highlight two interesting implications of this PR, for implementers like @sefeng211 to weigh in on.

The meat of this PR is the modification to "set the start"/"set the end". There are no semantic changes to legacy live ranges returned by `getRangeAt(0)`—the only change is that when "legacy uncomposed range"s are updated, we propagate the changes up to its "composed selection range", maintained by the Selection API. For example, when we "set the start" of a "legacy uncomposed range", we always set its "composed selection range" start too; the same goes for end. That's not too interesting. But there are two **interesting implications around collapsing** that we should agree on.

1. When we "set the start" of a "legacy uncomposed range" across shadow boundaries, the "legacy uncomposed range" collapses to the new start boundary point—that's all fine as it predates this PR. However, this PR DOES NOT collapse its associated "composed selection range", as long as everything is same-document. This is exercised by [this line in the spec](https://whatpr.org/dom/1342.html#:~:text=If%20selection%20range%20is%20not%20null%20and%20range%E2%80%99s%20start%20node%E2%80%99s%20shadow%2Dincluding%20root%20is%20not%20node%E2%80%99s%20shadow%2Dincluding%20root%2C%20set%20selection%20range%E2%80%99s%20end%20to%20bp). Firefox currently collapses the composed selection range in this case, going against this PR, so we should confirm with @sefeng211 if this PR is acceptable and Firefox can change their behavior. @dizhang168 is working on a WPT for this case now, but you can see an example of this scenario in https://gist.github.com/domfarolino/92d04f5e517da13971f69171476043d0.

2. When we "set the start" of a "legacy uncomposed range" in the same root, but in a way that inverts its start/end, the "legacy uncomposed range" collapses to the new start—that behavior predates this PR too. But per this PR, we ALSO collapse its associated "composed selection range" at the new start boundary point. This is done by [this line in the spec](https://whatpr.org/dom/1342.html#:~:text=Otherwise%2C%20if%20bp%20is%20after%20the%20range,selection%20range%E2%80%99s%20end%20to%20bp). Implications:
    ```js
    // The composed range is not collapsed, but getRangeAt(0) *is* (inside shadow).
    getSelection().setBaseAndExtent(lightDom, 0, shadowText, 5);

    // This inverts the legacy uncomposed range, re-collapsing it at offset `10`.
    // It ALSO collapses the *composed* selection range.
    getSelection().getRangeAt(0).setStart(shadowText, 10);
    ```

All of this is fine to me, but since these are the tricker parts of this PR I just want to highlight them here. @sefeng211 + @annevk, how do Firefox / WebKit  feel about these implications? Again, this is not about how `getRangeAt(0)` behaves after these options—it's about how changes to that range collapse or do not collapse the composed selection range. If there are strong preferences on either of the two cases here, we'd love to know.

sefeng211:
> However, this PR DOES NOT collapse its associated "composed selection range", as long as everything is same-document 

Di changed the PR to use shadow-including root rather than comparing the document, I think we now also collapse the composed selection range which is correct IMO. 

> When we "set the start" of a "legacy uncomposed range" in the same root, but in a way that inverts its start/end, the "legacy uncomposed range" collapses to the new start—that behavior predates this PR too. But per this PR, we ALSO collapse its associated "composed selection range" at the new start boundary point

This also looks fine to me.

domfarolino:
> Di changed the PR to use shadow-including root rather than comparing the document, I think we now also collapse the composed selection range which is correct IMO.

But shadow-including root will climb up to the document, if the node is connected. So I don't think there's a difference here. I believe the spec does not collapse in this case, because it is basically doing a same-document comparison check.

sefeng211:
> But shadow-including root will climb up to the document, if the node is connected. So I don't think there's a difference here. I believe the spec does not collapse in this case, because it is basically doing a same-document comparison check

Yeah, I mean with this we collapse if one of the nodes is disconnected which is something the PR didn't do before.

domfarolino:
> Yeah, I mean with this we collapse if one of the nodes is disconnected which is something the PR didn't do before.

OK, got it. So just to be clear, the behavior of NOT collapsing the composed selection range in the first case I mention is fine with you? I ask only because it looks like Firefox DOES collapse the composed selection range in that case, which I think would fail the incoming WPT being written.

sefeng211:
> OK, got it. So just to be clear, the behavior of NOT collapsing the composed selection range in the first case I mention is fine with you? I ask only because it looks like Firefox DOES collapse the composed selection range in that case, which I think would fail the incoming WPT being written.

I think we are on the same page but let me reiterate. You mean that we don't collapse the composed selection range as long as both nodes are connected right? I am fine with this. If Firefox collapses in case then I think it's a Firefox bug...

domfarolino:
> You mean that we don't collapse the composed selection range as long as both nodes are connected right?

Right. The spec collapses the live range (the `getRangeAt(0)` range) but NOT the composed selection range.

> If Firefox collapses in case then I think it's a Firefox bug...

Great to know! Yeah Di is adding a test for this in https://crrev.com/c/6354926 to assert the spec behavior, so hopefully that helps. Thanks a lot!

dizhang168:
Hi @domfarolino @sefeng211! I have updated the [tests](https://github.com/web-platform-tests/wpt/blob/master/selection/shadow-dom/tentative/Selection-getComposedRanges-range-update.html) to test out the "set the start/end" algorithm, as discussed in the conversation above.
I have also updated this PR to simplify the algorithm by having the composed selection range's checks in a scoped if condition. Part of it includes adding a new **composed position** definition. PTAL, thank you.

dom.bs
```diff
@@ -8493,74 +8527,60 @@ set the start or end</dfn> of a <a>live range</a> <var>range</var> to a <a>bound
...
+     to <var>bp</var>.
+
+     <li>Set <var>range</var>'s
+     <a for=range>end</a> to <var>bp</var>.
 
-       <li>If <var>composed range</var> is not null and
-       <var>range</var>'s <a for=range>start node</a>'s
+     <li>f <var>composed range</var> is not null, then:
+      <ol>
+       <li>If <var>range</var>'s <a for=range>end node</a>'s
```
sefeng211, line 8572: Should this be `composed range`?

dizhang168: Right, thanks for spotting that.

sefeng211:
Thanks, LGTM!

PR #1347, opened 2025-01-06
Editorial: initialize clearTargets in event dispatch correctly

### LINKED ISSUE #1344: Event::dispatch uses possibly uninitialized `clearTargets`

Step 10. of [Event::dispatch](https://dom.spec.whatwg.org/#concept-event-dispatch) (`If clearTargets is true: `) uses a variable `clearTargets` that is defined in Step 5.11 (`Let clearTargets be true if [...]`), but if the if-condition in Step 5. (`If target is not relatedTarget or target is event’s relatedTarget:`) is false then 5.11 is never executed and `clearTargets` is uninitialized.

It is unusual for spec-pseudocode to use a variable outside of its declared scope anyways, so i think this is an accident.

annevk:
Also clean up "for each" wording while here.

Fixes #1344.

***
<a href="https://whatpr.org/dom/1347.html" title="Last updated on Jan 6, 2025, 1:37 PM UTC (20ee860)">Preview</a> | <a href="https://whatpr.org/dom/1347/7cd51e7...20ee860.html" title="Last updated on Jan 6, 2025, 1:37 PM UTC (20ee860)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -146,7 +146,7 @@ first <a>following</a> <a for=tree>sibling</a> or null if it has no <a for=tree>
 
  <li><p>Let <var>tokens</var> be a new <a>ordered set</a>.
 
- <li><p><a for=list>For each</a> <var>token</var> in <var>inputTokens</var>, <a for=set>append</a>
+ <li><p><a for=list>For each</a> <var>token</var> of <var>inputTokens</var>, <a for=set>append</a>
  <var>token</var> to <var>tokens</var>.
 
  <li>Return <var>tokens</var>.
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -1405,18 +1407,17 @@ property of the event being dispatched.
...
 
-   <li><p>Let <var>clearTargets</var> be true if <var>clearTargetsStruct</var>'s
-   <a for=Event/path>shadow-adjusted target</a>, <var>clearTargetsStruct</var>'s
-   <a for=Event/path>relatedTarget</a>, or an {{EventTarget}} object in
-   <var>clearTargetsStruct</var>'s <a for=Event/path>touch target list</a> is a <a for=/>node</a>
-   and its <a for=tree>root</a> is a <a for=/>shadow root</a>; otherwise false.
+   <li><p>If <var>clearTargetsStruct</var>'s <a for=Event/path>shadow-adjusted target</a>,
+   <var>clearTargetsStruct</var>'s <a for=Event/path>relatedTarget</a>, or an {{EventTarget}} object
+   in <var>clearTargetsStruct</var>'s <a for=Event/path>touch target list</a> is a <a for=/>node</a>
+   whose <a for=tree>root</a> is a <a for=/>shadow root</a>: set <var>clearTargets</var> to true.
```
domenic, line 1413:
```suggestion
   whose <a for=tree>root</a> is a <a for=/>shadow root</a>, then set <var>clearTargets</var> to true.
```

domenic (approved):
Optional nit

dom.bs
annevk: I did this deliberately as I thought using a colon here would make it clearer. In particular because the clause following "or" is rather long. WDYT?

PR #1353, opened 2025-02-04
Add reference target to shadow root

dandclark:
Reference Target is a feature to enable using IDREF attributes such as `for` and `aria-labelledby` to refer to elements inside a component's shadow DOM, while maintaining encapsulation of the internal details of the shadow DOM. The main goal of this feature is to enable ARIA to work across shadow root boundaries.

In this change, add the `referenceTarget` property to `ShadowRoot` and add a definition of `reference target` that's exported for use in other specs.

See the [reference target explainer](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/reference-target-explainer.md).

  * Chromium is positive (I'd like to ship it there)
   * https://github.com/mozilla/standards-positions/issues/1035 (Positive)
   * https://github.com/WebKit/standards-positions/issues/356 (no position stated yet)
   * https://wpt.fyi/results/shadow-dom/reference-target/tentative
   * PR to make non-tentative: https://github.com/web-platform-tests/wpt/pull/51213
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: https://issues.chromium.org/issues/346835896/
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1952585
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=285634
- [X] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/mdn/issues/832

***
<a href="https://whatpr.org/dom/1353.html" title="Last updated on Aug 24, 2026, 11:12 PM UTC (e923e68)">Preview</a> | <a href="https://whatpr.org/dom/1353/94d41b3...e923e68.html" title="Last updated on Aug 24, 2026, 11:12 PM UTC (e923e68)">Diff</a>


dom.bs
```diff
@@ -6057,6 +6058,7 @@ interface ShadowRoot : DocumentFragment {
   readonly attribute boolean clonable;
   readonly attribute boolean serializable;
   readonly attribute Element host;
+  attribute DOMString? referenceTarget;
```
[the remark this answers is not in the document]
smaug----: Or whether this API should look a bit different so that adding v2 would feel a bit less weird.

Just as an example, not proposing this... what if this was more like 
attribute ReferenceTarget referenceTarget;

and then 
interface ReferenceTarget {
  attribute DOMString? default;
  // v2 would add more attributes
}

dom.bs
```diff
@@ -6092,6 +6094,9 @@ It is initially set to false.</p>
 <p><a for=/>Shadow roots</a> have an associated <dfn for=ShadowRoot>serializable</dfn> (a boolean).
 It is initially set to false.</p>
 
+<p><a for=/>Shadow roots</a> have an associated <dfn export for=ShadowRoot>reference target</dfn>.
```
annevk, line 6097: This is missing a type declaration.

dom.bs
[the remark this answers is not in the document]
annevk: Where is the sketch for v2? We should at least have an idea on what this transition will look like. Did someone write that out?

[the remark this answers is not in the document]
dandclark: Sorry for the slow follow up here -- v2 is sketched out at https://github.com/WICG/webcomponents/blob/gh-pages/proposals/reference-target-explainer.md#-phase-2-shadowroot-referencetargetmap-attribute.

That examples in that proposal imply an IDL like this:

```
interface ShadowRoot : DocumentFragment {
  // ...
  attribute DOMString? referenceTarget;
  attribute ReferenceTargetMap referenceTargetMap; // added in Phase 2.
}

// Added in Phase 2:
interface ReferenceTargetMap {
  DOMString? ariaControls;
  DOMString? ariaActiveDescendant;
  DOMString? htmlFor;
  // etc...
}
```

So the behavior would be that `shadowRoot.referenceTarget` is the default, and is overridden by any non-null values in the `ReferenceTargetMap`.

I can see an argument for grouping everything together into a single `ReferenceTarget` interface, but I do think it'd be nice to keep the simple case simple and allow a direct `shadowRoot.referenceTarget = "target"` rather than requiring the additional indirection `shadowRoot.referenceTarget.default = "target"`. Either design seems basically reasonable though.

dom.bs
dandclark: Fixed, thanks!

dom.bs
```diff
@@ -6348,6 +6349,7 @@ interface ShadowRoot : DocumentFragment {
   readonly attribute boolean clonable;
   readonly attribute boolean serializable;
   readonly attribute Element host;
+  attribute DOMString? referenceTarget;
```
annevk, line 6634: Unrelated to the above: this getter doesn't appear to be defined. The PR also needs rebasing unfortunately.

dandclark: Thanks, added getter/setter definition for this and updated the PR to resolve the conflict.

dandclark:
@annevk , @smaug---- , this is ready for another review along with https://github.com/whatwg/html/pull/10995. For context on the event propagation change, see https://github.com/WICG/webcomponents/issues/1098. Thanks!

smaug----:
What is the current testing situation? alice added several tests, but this is a rather complicated feature (at least from implementation point of view) so we need to ensure testing coverage is particularly good.
Gecko doesn't fail currently in too many ([when run in Mozilla CI](https://searchfox.org/firefox-main/search?q=FAIL&path=reference-target%2Ftentative%2F*ini&case=false&regexp=false)).
Chrome fails only some on [wpt.fyi](https://wpt.fyi/results/shadow-dom/reference-target/tentative?label=experimental&label=master&aligned).

But are we testing all the complicated cases?

dandclark:
I've tried to ensure that the WPTs for this are thorough; I and other contributors added quite a few during work on the Chromium prototype.

E.g. https://wpt.fyi/results/shadow-dom/reference-target/tentative/property-reflection.html and https://wpt.fyi/results/shadow-dom/reference-target/tentative/property-reflection-idl-setters.html test reflection behavior for all the properties affected by reference target, with references pointing to and from every HTML element type.

https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/reference-target/tentative/event-path.html covers all the interesting event propagation scenarios discussed in https://github.com/WICG/webcomponents/issues/1098, plus some others I thought of.

All of the platform features testable in WPTs that can be affected by reference target (form submission, `commandfor`, `interestfor`, `<label>` (implicit association and with `for`)), `popovertarget`, and `aria-labelledby`, have WPTs under https://wpt.fyi/results/shadow-dom/reference-target/tentative.

There are a few other ARIA properties affected by reference target that aren't yet testable in WPTs; there are tests for those in the Chromium source tree at https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/shadow-dom/reference-target/tentative/.

For some reason it doesn't seem like the [Mozilla CI results](https://searchfox.org/firefox-main/search?q=FAIL&path=reference-target%2Ftentative%2F*ini&case=false&regexp=false) are including the full set of tests. There are a larger number of tests in https://wpt.fyi/results/shadow-dom/reference-target/tentative that Gecko is shown as failing that don't appear in the linked [Mozilla CI results](https://searchfox.org/firefox-main/search?q=FAIL&path=reference-target%2Ftentative%2F*ini&case=false&regexp=false).

smaug----:
The pref in Gecko isn't enabled in Nightly, so wpt.fyi still shows many tests as failing. The pref is enabled in Mozilla CI.

dom.bs
```diff
@@ -6603,10 +6611,22 @@ is null.
...
 
 <p>A <a for=/>shadow root</a>'s <a>get the parent</a> algorithm, given an <var>event</var>, returns
-null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow root</a> is the
-<a for=tree>root</a> of <var>event</var>'s <a for=Event>path</a>'s first struct's
-<a for=Event/path>invocation target</a>; otherwise <a for=/>shadow root</a>'s
-<a for=DocumentFragment>host</a>.
+the result of the following steps:</p>
+
+<ol>
+ <li><p>If <var>event</var>’s <a>composed flag</a> is set, return <a for=/>shadow root</a>’s
```
annevk: Now that we're making this a proper algorithm we should make _shadowRoot_ a variable in the introduction as I don't think we can use "this" here.

dom.bs
```diff
@@ -6603,10 +6611,22 @@ is null.
...
+ <a for=DocumentFragment>host</a>.</p></li>
+ <li><p>Let <var>target</var> be <var>event</var>’s <a for=Event>path</a>'s first struct's
```
annevk: The `<li>`s are separated by newlines.

dom.bs
annevk: then return* (applies multiple times)

dom.bs
annevk: Can we use path[0] here?

dom.bs
```diff
@@ -6628,6 +6648,10 @@ null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow roo
 <p>The <dfn attribute for=ShadowRoot><code>host</code></dfn> getter steps are to return
 <a>this</a>'s <a for=DocumentFragment>host</a>.
 
+<p>The <dfn attribute for=ShadowRoot><code>referenceTarget</code></dfn> getter steps are to return
+<a>this</a>'s <a for=ShadowRoot>reference target</a>. The {{ShadowRoot/referenceTarget}} setter
+steps are to set <a>this</a>'s <a for=ShadowRoot>reference target</a> to the given value.
```
annevk, line 6721: These should be separate paragraphs.

dom.bs
```diff
@@ -6603,10 +6611,22 @@ is null.
...
+ <a for=DocumentFragment>host</a>.</p></li>
+ <li><p>Let <var>target</var> be <var>event</var>’s <a for=Event>path</a>'s first struct's
+ <a for=Event/path>invocation target</a>.</p></li>
+ <li><p>Let <var>source</var> be <var>event</var>'s <a for=Event>source</a>.</p></li>
+ <li><p>If <a for=/>shadow root</a> is not the <a for=tree>root</a> of <var>target</var>, and either
+ <var>source</var> is null or <a for=/>shadow root</a> is not the <a for=tree>root</a> of
+ <var>source</var>, return <a for=/>shadow root</a>'s <a for=DocumentFragment>host</a>.</p></li>
+ <li><p>If <var>source</var> is not null, and <var>source</var>'s <a for=tree>root</a> is a
+ <a>shadow-including ancestor</a> of <a for=/>shadow root</a>, return the result of
+ <a>retargeting</a> <var>target</var> against <var>source</var>.</p></li>
```
annevk: I wonder if we should add one or two examples here because this is getting rather involved. It seems really weird that as part of calculating the dispatch tree we are retargeting.

Also just to return the event target as the parent of the shadow tree. I can't really wrap my head around this.

annevk: This in particular I would expect to have test coverage for every branch in detail. Including capture and bubble phases, the works. Does that exist?

dom.bs
```diff
@@ -7738,7 +7763,8 @@ are:
  <var>init</var>["{{ShadowRootInit/mode}}"], <var>init</var>["{{ShadowRootInit/clonable}}"],
  <var>init</var>["{{ShadowRootInit/serializable}}"],
  <var>init</var>["{{ShadowRootInit/delegatesFocus}}"],
- <var>init</var>["{{ShadowRootInit/slotAssignment}}"], and <var>registry</var>.
+ <var>init</var>["{{ShadowRootInit/slotAssignment}}"], and
+ <var>init</var>["{{ShadowRootInit/referenceTarget}}"], and <var>registry</var>.
```
annevk, line 7880: double and

dom.bs
annevk: We don't use smart quotes. (This happens three times in this PR I believe.)

dom.bs
```diff
@@ -6603,10 +6611,22 @@ is null.
 <hr>
 
 <p>A <a for=/>shadow root</a>'s <a>get the parent</a> algorithm, given an <var>event</var>, returns
-null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow root</a> is the
-<a for=tree>root</a> of <var>event</var>'s <a for=Event>path</a>'s first struct's
-<a for=Event/path>invocation target</a>; otherwise <a for=/>shadow root</a>'s
-<a for=DocumentFragment>host</a>.
+the result of the following steps:</p>
```
annevk: We should add `<div algorithm="shadow root's get the parent">` around this.

dom.bs
```diff
@@ -479,6 +479,9 @@ signaling that something has occurred, e.g., that an image has completed downloa
 <p class=note>Other specifications use <a for=Event>relatedTarget</a> to define a
 <code>relatedTarget</code> attribute. [[UIEVENTS]]
 
+<p>An <a>event</a> has an associated <dfn export for=Event>source</dfn> (a
+<a>potential event target</a>). Unless stated otherwise it is null.
```
annevk: This should probably have a note explaining where this is used. "Other specifications ..."

dom.bs
```diff
@@ -6603,10 +6611,22 @@ is null.
...
 <p>A <a for=/>shadow root</a>'s <a>get the parent</a> algorithm, given an <var>event</var>, returns
-null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow root</a> is the
-<a for=tree>root</a> of <var>event</var>'s <a for=Event>path</a>'s first struct's
-<a for=Event/path>invocation target</a>; otherwise <a for=/>shadow root</a>'s
-<a for=DocumentFragment>host</a>.
+the result of the following steps:</p>
+
+<ol>
+ <li><p>If <var>event</var>’s <a>composed flag</a> is set, return <a for=/>shadow root</a>’s
+ <a for=DocumentFragment>host</a>.</p></li>
```
annevk: We don't have these optional closing tags in this specification. (Applies quite a few times.)

dandclark: Done, but this makes line 6599 inconsistent with the surrounding blocks, so I opened https://github.com/whatwg/dom/pull/1466 to remove these everywhere.

dom.bs
```diff
@@ -6602,11 +6610,31 @@ is null.
...
+ <li><p>Let <var>source</var> be <var>event</var>'s <a for=Event>source</a>.
+
+ <li><p>If <var>shadowRoot</var> is not the <a for=tree>root</a> of <var>target</var>, and either
+ <var>source</var> is null or <var>shadowRoot</var> is not the <a for=tree>root</a> of
+ <var>source</var>, then return <var>shadowRoot</var>'s <a
+ for=DocumentFragment>host</a>.
+
+ <li><p>If <var>source</var> is not null, and <var>source</var>'s <a for=tree>root</a> is a
+ <a>shadow-including ancestor</a> of <var>shadowRoot</var>, then return the result of
+ <a>retargeting</a> <var>target</var> against <var>source</var>.
```
smaug----: As I said in https://github.com/WICG/webcomponents/issues/1098#issuecomment-4511356663 I think we should still investigate relatedTarget model for this. Basically, the relevant events would be composed but propagate only up to the level where .target and .source are different.

dom.bs
```diff
@@ -1743,7 +1743,8 @@ and an optional <var>legacyOutputDidListenersThrowFlag</var>, run these steps:
 <div algorithm>
 <p>To <dfn export id=concept-event-fire>fire an event</dfn> named <var>e</var> at <var>target</var>,
 optionally using an <var>eventConstructor</var>, with a description of how IDL attributes are to be
-initialized, and a <var>legacy target override flag</var>, run these steps:
+initialized, and a <var>legacy target override flag</var> and an optional <var>relatedTarget</var>,
+run these steps:
```
annevk: Too many "and"s and a missing Oxford comma. And if we are going to repeat optional (which is probably a good idea) we should add it in front of the flag too. But that also means we need to make this a named parameter since we probably want to pass this without passing the constructor or the flag. https://infra.spec.whatwg.org/#algorithm-params details conventions.

We can also drop "run these steps" to modernize.

dom.bs
```diff
@@ -1754,6 +1755,9 @@ initialized, and a <var>legacy target override flag</var>, run these steps:
 
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>e</var>.
 
+ <li><p>If <var>relatedTarget</var> is given, set <var>event</var>'s <a for=Event>relatedTarget</a>
```
annevk: then set

smaug---- (approved):
Looks reasonable, with the nits annevk has mentioned

dom.bs
dandclark: Fixed the grammar and made these both named optional parameters.

dom.bs
dandclark: These changes to shadow root's "get the parent" have been reverted and this will now use `relatedTarget` as per https://github.com/whatwg/html/pull/10995#issuecomment-4572681552.

dom.bs
dandclark: This was removed.

dom.bs
dandclark: These changes to shadow root's "get the parent" were removed.

dom.bs
dandclark: These changes to shadow root's "get the parent" were removed.

annevk:
@dandclark it looks like this PR and https://github.com/whatwg/html/pull/10995 need rebasing (again?). I wasn't able to get to these before my break, but there should be time now. Please leave a comment when you think everything is in order as just rebasing might not trigger a notification.

dandclark:
Thanks for looking @annevk! I resolved conflicts for both PRs.

dom.bs
```diff
@@ -1375,15 +1375,15 @@ property of the event being dispatched.
...
+{{EventTarget}} object <var>target</var>, with an optional <var>legacy target override flag</var>
+(default false) and an optional <var>legacyOutputDidListenersThrowFlag</var>:
 
 <ol>
  <li><p>Set <var>event</var>'s <a>dispatch flag</a>.
 
  <li>
   <p>Let <var>targetOverride</var> be <var>target</var>, if <var>legacy target override flag</var>
-  is not given, and <var>target</var>'s <a>associated <code>Document</code></a> otherwise.
+  is true, and <var>target</var>'s <a>associated <code>Document</code></a> otherwise.
```
annevk, line 1386: This should be false. But if we're going to change this we should also change it away from being a flag. Hmm.

It also seems you are referencing these flags from HTML but haven't turned them into `<dfn>`s here.

dom.bs
```diff
@@ -1765,7 +1765,8 @@ string <var>phase</var>, a boolean <var>invocationTargetInShadowTree</var>, and
 <div algorithm>
 <p>To <dfn export id=concept-event-fire>fire an event</dfn> named <var>e</var> at <var>target</var>,
 optionally using an <var>eventConstructor</var>, with a description of how IDL attributes are to be
-initialized, and a <var>legacy target override flag</var>, run these steps:
+initialized, an optional boolean <var>legacy target override flag</var> (default false), and an
+optional {{EventTarget}} <var>relatedTarget</var>:
```
annevk, line 1769:
```suggestion
optional {{EventTarget}} object <var>relatedTarget</var>:
```

dom.bs
```diff
@@ -1776,14 +1777,17 @@ initialized, and a <var>legacy target override flag</var>, run these steps:
 
  <li><p>Initialize <var>event</var>'s {{Event/type}} attribute to <var>e</var>.
 
+ <li><p>If <var>relatedTarget</var> is given, then set <var>event</var>'s <a
+ for=Event>relatedTarget</a> to <var>relatedTarget</var>.
```
annevk, line 1781: Might be easier to use (default null) above and then always set here?

dom.bs
```diff
@@ -6705,6 +6714,12 @@ null if <var>event</var>'s <a>composed flag</a> is unset and <a for=/>shadow roo
 <p>The <dfn attribute for=ShadowRoot><code>host</code></dfn> getter steps are to return
 <a>this</a>'s <a for=DocumentFragment>host</a>.
 
+<p>The <dfn attribute for=ShadowRoot><code>referenceTarget</code></dfn> getter steps are to return
+<a>this</a>'s <a for=ShadowRoot>reference target</a>.
```
annevk, line 6718: I think this getter and setter should appear before `host`. Also in the IDL. We should keep the order consistent across all the places we introduce this concept.

dom.bs
dandclark: This is fixed. Changed it to be passed as an EventTarget rather than a flag, and exported the `<dfn>` for use in HTML.

dom.bs
dandclark: Done.

dom.bs
dandclark: Done.

PR #1354, opened 2025-02-06
Meta: export "signal abort"

domenic:
I want to use this for a spec which can abort for both internal reasons and from a developer- provided AbortSignal. I create a dependent abort signal from the internal abort signal and the developer-provided one, and "signal abort" on the internal one appropriately. Then the rest of the spec can just worry about the dependent one.

***
<a href="https://whatpr.org/dom/1354.html" title="Last updated on Feb 6, 2025, 6:30 AM UTC (cce2430)">Preview</a> | <a href="https://whatpr.org/dom/1354/e6bb175...cce2430.html" title="Last updated on Feb 6, 2025, 6:30 AM UTC (cce2430)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2024,8 +2024,8 @@ object <var>signal</var>:
 </div>
 
 <div algorithm>
-<p>To <dfn for=AbortSignal>signal abort</dfn>, given an {{AbortSignal}} object <var>signal</var> and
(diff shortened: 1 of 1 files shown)


domenic:
I forgot, I'm supposed to use the controller now!

PR #1372, opened 2025-05-18
Add part attribute and note that part and exportparts are superglobal

wlib:
`part` and `exportparts` were the (I hope) only two attributes that I missed in https://github.com/wlib/html-info due to them not appearing in [the spec](https://dom.spec.whatwg.org/#:~:text=id%2C%20class%2C%20and%20slot%20are%20effectively%20superglobal,.). There was [an issue](https://github.com/w3c/csswg-drafts/issues/3424) filed in 2018 asking to add `part` as a superglobal attribute to the DOM spec. This PR also notes that `exportparts` is a superglobal attribute, but does not make any references to its functionality. I want to ask whether this PR is desirable in the first place because other attributes like those defined in ARIA (e.g. `role` and `aria-label`) also are superglobal with corresponding IDL attributes defined on `Element`. I think that having a table that indexes all of the attributes in one place would be desirable, but it's unclear which spec such an index should live in.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6514,6 +6514,7 @@ interface Element : Node {
   [CEReactions] attribute DOMString className;
   [SameObject, PutForwards=value] readonly attribute DOMTokenList classList;
   [CEReactions, Unscopable] attribute DOMString slot;
+  [SameObject, PutForwards=value] readonly attribute DOMTokenList part;
 
   boolean hasAttributes();
   [SameObject] readonly attribute NamedNodeMap attributes;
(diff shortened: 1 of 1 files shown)


annevk (commented):
Doing this makes sense to me. However:

1. We need to remove these requirements from the original specification in a PR linked from the commit message.
2. We need to mention these attributes in HTML in the same place we mention id, class, and slot today (can also be a PR linked from the commit message).
3. Do these attributes have no steps to perform when they are manipulated?
4. We should consider, as a follow-up, to introduce a dedicated section on super global attributes to make the concept more formal and also to give some descriptions on the semantics of these attributes as that's currently missing.

PR #1386, opened 2025-07-01
Allow passing null for customElementRegistry

annevk:
Apparently people are passing a ShadowRoot instance as the argument to attachShadow().

Context: https://bugs.webkit.org/show_bug.cgi?id=295174. We'll add tests as part of fixing this in WebKit.

   * WebKit
   * …
   * See above.
   * Chromium: Doesn't have an implementation yet.
   * Gecko: Doesn't have an implementation yet.
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=295174

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6598,7 +6598,7 @@ dictionary ShadowRootInit {
   SlotAssignmentMode slotAssignment = "named";
   boolean clonable = false;
   boolean serializable = false;
-  CustomElementRegistry customElementRegistry;
+  CustomElementRegistry customElementRegistry? = null;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -6598,7 +6598,7 @@ dictionary ShadowRootInit {
   SlotAssignmentMode slotAssignment = "named";
   boolean clonable = false;
   boolean serializable = false;
-  CustomElementRegistry customElementRegistry;
+  CustomElementRegistry? customElementRegistry = null;
```
domenic, line 6601: Maybe add a comment here explaining why we're departing from our usual preference to only have one way of signaling absence.

domenic (approved):

dom.bs
smaug----: Yeah, some comment would good, since this is a rather odd case. Wouldn't have guessed anyone using attachShadow that way, but don't see reason why not.

keithamus (approved):

Issue #952, opened 2021-02-25
document.createEvent() should maybe not be conditional upon exposure


annevk:
Apparently `TouchEvent` is always exposed in Chrome, but `document.createEvent("touchevent")` depends on the "touch events flag" (see https://github.com/w3c/touch-events/issues/64). Having said that, neither Firefox nor Safari expose `TouchEvent` so maybe Chrome should change?

cc @patrickhlauke @RByers

patrickhlauke:
I vaguely remember asking before somewhere, but: philosophically, does it not make sense to expose `TouchEvent` (as an API that the browser understands/supports) regardless of presence of an actual touchscreen? Are there other APIs that are simply not exposed (as if the browser didn't understand them / support them at all) in the absence of certain conditions?

And yes, Chrome's selective support for certain properties/calls based on the actual presence of a touchscreen is a hacky workaround to naive "mobile-only" detection that assumed "if there's a touchscreen, then i can just listen to touch events as it's mobile ... no need to use mouse events or keyboard / can just switch the the 'mobile' version of the site / etc"

annevk:
It makes sense, but it might well break code that does
```js
if ("TouchEvent" in self) { ... }
```
and the precedent was set by WebKit as I understand it so legacy code would assume WebKit's model.

patrickhlauke:
> so legacy code would assume WebKit's model.

I seem to remember (but Googlers may have more detail) that the internal touch events flag thing and deciding when to be truthful (yes, of course we know what Touch Events are, even if you don't have an actual touchscreen) and when to lie came mostly from looking at (then already) legacy code to see how sites were doing naive "isMobile?" type sniffing and then essentially only listening for touch events and not mouse/keyboard - breaking sites on touch-enabled laptops/desktops. This may have been based on data analysis of "what's the most prevalent pattern that sites use for this sort of naive detection", but I'm not sure).

RByers:
Mobile Safari exposes TouchEvent, right? Desktop Safari doesn't support touch at all, so no surprise not to have it there. Anyway further alignment on any differences between browsers SGTM. Anything in mobile Safari or Firefox for Android is probably sufficiently web compatible for chromium to try.

Over to @flackr who leads input on Chrome these days (I'm not sure who on the team might be working in this area).

patrickhlauke:
sorry, I realised just now that I almost repeated everything I said a few days ago already in my second comment. Serves me right to try and "catch up" with threads without seeing what I might have previously said.

[...]

PR #1391, opened 2025-07-19
Link exposed to the WebIDL definition, remove potentially confusing note

shannonbooth:
I recently implemented this, and noticed that this could be linked, so this does that. However, on deeper investigation, I did notice that browsers didn't behave as I expected for [DeviceMotionEvent](https://w3c.github.io/deviceorientation/spec-source-orientation.html#devicemotion) since that is only exposed in secure contexts.

I expected browsers to throw on `document.createEvent('devicemotionevent')` like they do for `new DeviceMotionEvent('foo')` in non secure contexts, but from some limited testing they seem to not do that.

Since the intent of the spec reads to throw in this situation, I'm just removing the spec note here which seems to indicate it only applies for TouchEvent and will write a corresponding WPT test if this change is fine.

Sidenote: [TouchEvent](https://w3c.github.io/touch-events/#idl-def-touchevent) by strict reading of the IDL is always exposed on window in the IDL definition and the spec instead uses a [separate implementation defined boolean](https://w3c.github.io/touch-events/#conditionally-exposing-legacy-touch-event-apis) which reads to only apply for the mixin, so this wouldn't apply here. But that seems to not reflect with the intent of the touch events spec, and now that specification is unmaintained.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/dom/1391.html" title="Last updated on Jul 19, 2025, 5:11 AM UTC (1165398)">Preview</a> | <a href="https://whatpr.org/dom/1391/1b4bd48...1165398.html" title="Last updated on Jul 19, 2025, 5:11 AM UTC (1165398)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6090,12 +6090,9 @@ are:
   {{DOMException}}.
 
  <li>
-  <p>If the interface indicated by <var>constructor</var> is not exposed on the
-  <a>relevant global object</a> of <a>this</a>, then <a>throw</a> a
+  <p>If the interface indicated by <var>constructor</var> is not <a>exposed</a> on the
+  <a>relevant realm</a> of <a>this</a>, then <a>throw</a> a
   "{{NotSupportedError!!exception}}" {{DOMException}}.
-
-  <p class=note>Typically user agents disable support for touch events in some configurations, in
-  which case this clause would be triggered for the interface {{TouchEvent}}.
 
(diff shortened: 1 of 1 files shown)


shannonbooth:
Ah, apologies, I completely missed https://github.com/whatwg/dom/issues/952 which mentions TouchEvent since I was looking at the device motion event instead of the touch event, now I am not sure about this merge request at all.

PR #1404, opened 2025-09-12
Add OpaqueRange interface

### LINKED ISSUE #1468: OpaqueRange: cross-post from whatwg/html#11478 for DOM community input

Cross-posting from https://github.com/whatwg/html/issues/11478 to get DOM community input on `OpaqueRange`, a proposed new `AbstractRange` subclass for encapsulated content.

`OpaqueRange` represents a live span of content within a host element, so developers can work with that content using range-like APIs. For `<textarea>` and text-based `<input>`, that content is the element's value text. The design also has a path to custom elements, so authors can expose encapsulated ranges over their own internal content in the future.

It enables operations such as `getBoundingClientRect()`, `getClientRects()`, and integration with the CSS Custom Highlight API for UI such as inline suggestions, highlights, and anchored popovers. It preserves encapsulation by exposing only offsets while not exposing the underlying nodes, so the host element's internal structure stays hidden.

Relevant links:
- Explainer: https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/OpaqueRange/explainer.md
- WHATWG Discussion Thread: https://github.com/whatwg/html/issues/11478
- DOM Spec PR: https://github.com/whatwg/dom/pull/1404
- HTML Spec PR: https://github.com/whatwg/html/pull/11741

Posting here so DOM reviewers can weigh in. **Please keep discussion to https://github.com/whatwg/html/issues/11478. This issue is mainly a pointer for visibility.**

stephanieyzhang:
OpaqueRange is a specialized, live AbstractRange subtype whose boundary points reference internal nodes within host-defined elements (e.g., `<input>`/`<textarea>` today, with a path to custom elements in the future). It enables range-based operations over encapsulated content while avoiding exposure of internal DOM nodes. This PR also updates AbstractRange so `startContainer`/`endContainer` are nullable (`Node?`), which allows OpaqueRange to return null for those getters while Range/StaticRange continue returning nodes.

- Explainer: https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/OpaqueRange/explainer.md
- WHATWG Discussion: https://github.com/whatwg/html/issues/11478
- Corresponding HTML Spec PR: https://github.com/whatwg/html/pull/11741
- Corresponding CSSOM PRs: 
     - https://github.com/w3c/csswg-drafts/pull/12904
     - https://github.com/w3c/csswg-drafts/pull/13929

   * Microsoft Edge
   * WebKit: https://github.com/WebKit/standards-positions/issues/541
   * Mozilla: https://github.com/mozilla/standards-positions/issues/1289
   * https://wpt.fyi/results/dom/ranges/tentative?label=master&label=experimental&aligned
   * PR to make tests non-tentative: Coming soon
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: https://issues.chromium.org/issues/421421332
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1990256
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=299375

***

### :boom: Error: 422 Unprocessable Entity :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 18, 2026, 4:25 PM UTC)_.

<details>
<summary>More</summary>

:rotating_light: [Spec Generator](https://www.w3.org/publications/spec-generator/) - Spec Generator is the web service used to build bikeshed/ReSpec specs

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=bikeshed-spec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fstephanieyzhang%2Fdom%2Fef8b1eb60fa9843670a38a0d6598ca122d1216a6%2Fdom.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%201404)

**Error output:**

```json
[
    {
        "lineNum": null,
        "messageType": "warning",
        "text": "Multiple elements have the same id 'concept-range-is-opaque':\n  <dfn> on line 8691:37, <dfn> on line 8707:37\nDeduping, but this ID may not be stable across revisions."
    },
    {
        "lineNum": "8707:37",
        "messageType": "fatal",
        "text": "Multiple local 'dfn' <dfn>s for 'range' have the same linking text 'is opaque'."
    },
    {
        "lineNum": null,
        "messageType": "failure",
        "text": "Did not generate, due to errors exceeding the allowed error level."
    }
]
```

_This seems to be an issue with the [Spec Generator](https://www.w3.org/publications/spec-generator/) service. PR Preview doesn't manage this service and so has no control over it. If you've identified an issue with it, you can [report the issue to the maintainers of Spec Generator](https://github.com/w3c/spec-generator/issues/new) directly. Please be courteous. Thank you!_

_If you don't have enough information above to solve the error by yourself or if the issue doesn't seem related to Spec Generator, you can [file an issue with PR Preview](https://github.com/tobie/pr-preview/issues/new?title=Unidentified%20Error&body=See%20whatwg/dom%231404.)._

</details>


dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
 
 
+<h3 id=interface-formcontrolrange>Interface {{FormControlRange}}</h3>
+
+<pre class=idl>
+[Exposed=Window]
+interface FormControlRange : AbstractRange {
+  constructor();
+
+  undefined setFormControlRange((HTMLInputElement or HTMLTextAreaElement) element,
```
dandclark: This is probably the wrong place to bikeshed this name but the more I look at it the more I don't like that we say "FormControlRange" both in the name of the class and the name of this method. At some point we should make space for a proper bikeshedding discussion on this method name, considering other things like `setInFormControl`, `setFormControl`, `setControl`, `setStartAndEnd` or simply `set`.

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+                                unsigned long endOffset);
+
+  DOMRectList getClientRects();
+  DOMRect getBoundingClientRect();
+  stringifier;
+};
+</pre>
+
+<p>Objects implementing the {{FormControlRange}} interface are known as
+<dfn export id=concept-live-form-control-range>live FormControlRanges</dfn>.</p>
```
dandclark: I don't think it's necessary to use "live" in the name like this.

Yes, they are 'live', but it doesn't need to be part of the name, we can make sure that's understood from the algorithms and maybe from an editorial note.

I know that at https://dom.spec.whatwg.org/#concept-live-range "live" is used, but that's necessary only because the "live" range's interface name is just `Range`, so there's a need to put an extra word in there to disambiguate it from AbstractRange and StaticRange. FormControlRange doesn't have that problem.

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+"<code>password</code>".</p>
+
+<p>For a supported host element <var>el</var>, its <i>value string</i> is the same string exposed by
+its IDL attribute {{HTMLTextAreaElement/value}}/{{HTMLInputElement/value}}. Offsets for
+{{FormControlRange}} are indices into this string in the inclusive range
+[0, <code>value.length</code>], matching
+{{HTMLTextAreaElement/selectionStart}}/{{HTMLTextAreaElement/selectionEnd}} and
+{{HTMLInputElement/selectionStart}}/{{HTMLInputElement/selectionEnd}} units.</p>
+
+<p>The <dfn constructor for=FormControlRange lt="FormControlRange()">
```
dandclark: could this be put first to match the order of appearance in the IDL?

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+<dl class=domintro>
+ <dt><code><var ignore>formControlRange</var> = new
+ <a constructor lt="FormControlRange()">FormControlRange()</a></code>
+ <dd>Returns a new <a>live FormControlRange</a> that tracks text offsets in a text control’s
+     value as it changes.</dd>
+
+ <dt><code><var ignore>formControlRange</var> .
+ {{FormControlRange/setFormControlRange()}} (<var ignore>element</var>,
+ <var ignore>startOffset</var>, <var ignore>endOffset</var>)</code>
+ <dd>Sets the endpoints to <code>[start, end)</code> within <var>element</var>’s
```
dandclark: Should `start` and `end` here be `startOffset` and `endOffset`?
I'm also not sure about the use of the `[` and `)`. The endOffset can be equal to the length of the value, not just strictly less than.

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+<ul>
+ <li><dfn export for=FormControlRange id=formcontrolrange-control>control</dfn>
+     (null or an {{HTMLInputElement}}/{{HTMLTextAreaElement}})</li>
+ <li><dfn export for=FormControlRange id=formcontrolrange-start>start offset</dfn>
+     (a non-negative integer)</li>
+ <li><dfn export for=FormControlRange id=formcontrolrange-end>end offset</dfn>
+     (a non-negative integer)</li>
+</ul>
+
+<p>The {{AbstractRange/startContainer}} and {{AbstractRange/endContainer}} getters of a
```
dandclark: I think it is clearer to define these things as a series of getter steps. See the examples from AbstractRange:

```
The startContainer getter steps are to return this’s start node.

The startOffset getter steps are to return this’s start offset.

The endContainer getter steps are to return this’s end node.

The endOffset getter steps are to return this’s end offset.

The collapsed getter steps are to return true if this is collapsed; otherwise false.
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+{{FormControlRange}} return its <a for=FormControlRange>control</a>. The
+{{AbstractRange/startOffset}} and {{AbstractRange/endOffset}} getters return its
+<a for=FormControlRange>start offset</a> and <a for=FormControlRange>end offset</a>.
+{{AbstractRange/collapsed}} is true if and only if the two offsets are equal.</p>
+
+<p>An {{Element}} <var>el</var> supports form control ranges if it is an {{HTMLTextAreaElement}},
```
dandclark:
```suggestion
<p>An {{Element}} <var>el</var> <dfn>supports form control range</dfn> if it is an {{HTMLTextAreaElement}},
```
Let's make this a formal linkable definition.

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+or an {{HTMLInputElement}} whose type
+<a href="https://html.spec.whatwg.org/multipage/input.html#do-not-apply">supports the selection APIs</a>:
+"<code>text</code>", "<code>search</code>", "<code>tel</code>", "<code>url</code>", or
+"<code>password</code>".</p>
+
+<p>For a supported host element <var>el</var>, its <i>value string</i> is the same string exposed by
```
dandclark:
```suggestion
<p>For a <a lt="supports form control range">supported</a>host element <var>el</var>, its <i>value string</i> is the same string exposed by
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+<code>new FormControlRange()</code></dfn> constructor steps are to set <a>this</a>’s
+<a for=FormControlRange>control</a> to null and its
+<a for=FormControlRange>start offset</a> and <a for=FormControlRange>end offset</a> to 0.</p>
+
+<p>The <dfn method for=FormControlRange>
+<code>setFormControlRange(<var>element</var>, <var>start</var>, <var>end</var>)</code></dfn>
+method steps are:</p>
+
+<ol>
+ <li>If <var>element</var> does not support form control ranges, then <a>throw</a> a
```
dandclark:
```suggestion
 <li>If <var>element</var> does not <a>support form control range</a>, then <a>throw</a> a
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+};
+</pre>
+
+<p>Objects implementing the {{FormControlRange}} interface are known as
+<dfn export id=concept-live-form-control-range>live FormControlRanges</dfn>.</p>
+
+<dl class=domintro>
+ <dt><code><var ignore>formControlRange</var> = new
+ <a constructor lt="FormControlRange()">FormControlRange()</a></code>
+ <dd>Returns a new <a>live FormControlRange</a> that tracks text offsets in a text control’s
```
dandclark: nit: please replace instances of `’` with `'`.

Edge's find-on-page treats these as the same character so you might need to look in your code editor to see what I mean.

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+method steps are:</p>
+
+<ol>
+ <li>If <var>element</var> does not support form control ranges, then <a>throw</a> a
+     "{{NotSupportedError!!exception}}" {{DOMException}}.</li>
+ <li>Let <var>len</var> be the length of <var>element</var>’s <i>value string</i>.</li>
+ <li>If <var>start</var> &gt; <var>len</var> or <var>end</var> &gt; <var>len</var>, then
+     <a>throw</a> an "{{IndexSizeError!!exception}}" {{DOMException}}.</li>
+ <li>If <var>start</var> &gt; <var>end</var>, then set <var>end</var> to <var>start</var>.</li>
+ <li>Set <a for=FormControlRange>control</a> to <var>element</var>,
```
dandclark:
```suggestion
 <li>Set <a>this</a>'s <a for=FormControlRange>control</a> to <var>element</var>,
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+     <a for=FormControlRange>start offset</a> to <var>start</var>, and
```
dandclark:
```suggestion
     <a>this</a>'s <a for=FormControlRange>start offset</a> to <var>start</var>, and
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+     <a for=FormControlRange>end offset</a> to <var>end</var>.</li>
```
dandclark:
```suggestion
     <a>this</a>'s <a for=FormControlRange>end offset</a> to <var>end</var>.</li>
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+     the new text. A collapsed range moves after the insertion.</li>
+ <li>Clamping and collapse: Offsets are clamped to the current value length. If the
+     {{AbstractRange/startOffset}} would exceed the {{AbstractRange/endOffset}}, set the end to the
+     start.</li>
+</ul>
+
+<p>The <dfn export for=FormControlRange id=dom-formcontrolrange-stringifier>stringification behavior</dfn> must run these steps:</p>
+
+<ol>
+ <li><p>If <a for=FormControlRange>control</a> is null, then return the empty string.</p>
```
dandclark:
```suggestion
 <li><p>If <a>this</a>'s <a for=FormControlRange>control</a> is null, then return the empty string.</p>
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+ <li><p>Let <var>value</var> be <a for=FormControlRange>control</a>'s <i>value string</i>.</p>
```
dandclark:
```suggestion
 <li><p>Let <var>value</var> be <a>this</a>'s <a for=FormControlRange>control</a>'s <i>value string</i>.</p>
```

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+ <li><p>Let <var>start</var> be <a for=FormControlRange>start offset</a>, and let <var>end</var> be min(<a for=FormControlRange>end offset</a>, <var>value</var>.<code>length</code>).</p>
```
dandclark:
```suggestion
 <li><p>Let <var>start</var> be <a>this</a>'s <a for=FormControlRange>start offset</a>, and let <var>end</var> be min(<a>this</a>'s <a for=FormControlRange>end offset</a>, <var>value</var>.<code>length</code>).</p>
```

dandclark: Elsewhere we ensure that `start offset` <= `end offset` and `end offset` <= `value.length`, so is it necessary to check for these things in the stringifier?

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+ <li>Let <var>len</var> be the length of <var>element</var>’s <i>value string</i>.</li>
+ <li>If <var>start</var> &gt; <var>len</var> or <var>end</var> &gt; <var>len</var>, then
+     <a>throw</a> an "{{IndexSizeError!!exception}}" {{DOMException}}.</li>
+ <li>If <var>start</var> &gt; <var>end</var>, then set <var>end</var> to <var>start</var>.</li>
+ <li>Set <a for=FormControlRange>control</a> to <var>element</var>,
+     <a for=FormControlRange>start offset</a> to <var>start</var>, and
+     <a for=FormControlRange>end offset</a> to <var>end</var>.</li>
+</ol>
+
+<p>A {{FormControlRange}} is <em>live</em>: when the control’s value string changes, the range’s
```
dandclark: An interesting case I thought of is what happens if sets a FormControlRange on an `<input>` and then the `<input>`'s `type` is changed to something that doesn't support `FormControlRange`, like `date`?

Maybe the FormControlRange's `control` should get set to null when that happens?

dom.bs
```diff
@@ -9892,6 +9892,126 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+{{AbstractRange/startOffset}} and {{AbstractRange/endOffset}} are updated automatically
+to preserve the same logical content. These behaviors mirror {{Range}} boundary adjustments
+in the DOM, but are applied to the UTF-16 code units of a form control’s value.</p>
+
+<ul>
```
dandclark: This is good as a placeholder and/or as a note to explain how the range gets updated when the `value` changes.
To really be complete, we probably need to make changes against the HTML spec to define this behavior. That would probably look like defining an algorithm taking parameters describing the value change that handles the boundary adjustments for each FormControlRange associated with the control, and then finding all the places in the HTML spec where these controls change their values and calling that algorithm in each of those places.

dandclark (commented):
Great to see a spec for this coming together!

dom.bs
dandclark: See how for `Range`, the liveness is implemented by handling range updates anywhere the node tree is changed, e.g. in Step 5 of https://dom.spec.whatwg.org/#concept-node-insert.

rniwa:
FWIW, I'd rather extend `Range` so that it can work with shadow host in general instead of creating a new interface specifically for form controls.

dom.bs
stephanieyzhang: Sounds good. Added a TODO to remind myself in the future

dom.bs
stephanieyzhang: Sounds good - added!

dom.bs
stephanieyzhang: Added a TODO to update this once HTML spec changes are made

stephanieyzhang:
> FWIW, I'd rather extend `Range` so that it can work with shadow host in general instead of creating a new interface specifically for form controls.

We explored extending Range, including shadow host support, but found compatibility and encapsulation issues that made it risky to change existing Range semantics. Based on feedback across discussions, we went with a dedicated FormControlRange instead. The trade-offs and alternatives are covered in the [explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/FormControlRange/explainer.md#extending-the-range-api-1st-approach).

rniwa:
> > FWIW, I'd rather extend `Range` so that it can work with shadow host in general instead of creating a new interface specifically for form controls.
> 
> We explored extending Range, including shadow host support, but found compatibility and encapsulation issues that made it risky to change existing Range semantics. Based on feedback across discussions, we went with a dedicated FormControlRange instead. The trade-offs and alternatives are covered in the [explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/FormControlRange/explainer.md#extending-the-range-api-1st-approach).

Changing the behavior of `setStart`/`setEnd` wouldn't be an only way to solve this problem. I'd instead add a new argument or a new function which specifies "offset" within element's shadow tree.

dom.bs
```diff
@@ -9892,6 +9892,167 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+APIs</a>: "<code>text</code>", "<code>search</code>", "<code>tel</code>", "<code>url</code>", or
+"<code>password</code>".
+
+<p>The <dfn constructor for=FormControlRange lt="FormControlRange()">
+<code>new FormControlRange()</code></dfn> constructor steps are to set <a>this</a>'s
+<a for=FormControlRange>control</a> to null and its <a for=FormControlRange>start offset</a> and
+<a for=FormControlRange>end offset</a> to 0.
+
+<p>For a <a lt="supports form control range">supported</a> host element <var>el</var>, its
+<i>value string</i> is the same string exposed by its IDL attribute
```
dandclark: Let's also make `value string` a `<dfn>` that the other instances link to.

dom.bs
```diff
@@ -9892,6 +9892,167 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+
+<p>The endOffset getter steps are to return <a>this</a>'s
+<a for=FormControlRange>end offset</a>.
+
+<p>The collapsed getter steps are to return true if <a>this</a>'s
+<a for=FormControlRange>start offset</a> equals <a>this</a>'s
+<a for=FormControlRange>end offset</a>; otherwise false.
+
+<p>An {{Element}} <var>el</var> <dfn>supports form control range</dfn> if it is an
+{{HTMLInputElement}}, or an {{HTMLTextAreaElement}} whose type
```
dandclark:
```suggestion
{{HTMLTextAreaElement}}, or an {{HTMLInputElement}} whose type
```

dom.bs
```diff
@@ -9892,6 +9892,167 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+<a href="https://html.spec.whatwg.org/multipage/input.html#do-not-apply">supports the selection
```
dandclark: It might be better to use the terminology of the HTML spec about this, and say something like "An element `el` supports form control range if it is an HTMLTextAreaElement, or an HTMLInputElement with a type to which the selection APIs apply..."

Actually, it might be better still to just say this:

"An element `el` supports form control range if it is an HTMLTextAreaElement, or an HTMLInputElement whose type is one of text, search, tel, url, or password", and then in a note mention that these types were chosen based on the fact that they're the ones to which the related selection APIs apply.

dom.bs
```diff
@@ -9892,6 +9892,167 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+<pre class=idl>
+[Exposed=Window]
+interface FormControlRange : AbstractRange {
+  constructor();
+
+  undefined setFormControlRange((HTMLInputElement or HTMLTextAreaElement) element,
+      unsigned long startOffset,
+      unsigned long endOffset
+  );
+  <!-- TODO: Bikeshed method name. Alternatives suggested include setInFormControl,
```
dandclark: This could perhaps also be tracked as an Open Question in the explainer.

dom.bs
```diff
@@ -9892,6 +9892,167 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+
+ <li><p>If <var>startOffset</var> &gt; <var>endOffset</var>, then set <var>endOffset</var> to
+ <var>startOffset</var>.
+
+ <li><p>Set <a>this</a>'s <a for=FormControlRange>control</a> to <var>element</var>,
+ <a>this</a>'s <a for=FormControlRange>start offset</a> to <var>startOffset</var>, and
+ <a>this</a>'s <a for=FormControlRange>end offset</a> to <var>endOffset</var>.
+</ol>
+
+<p>If an {{HTMLInputElement}}'s <code>type</code> changes to a type that does not
```
dandclark: (this, too, will likely need to be something that we'll need to put in the HTML spec)

dom.bs
```diff
@@ -9892,6 +9892,167 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+ <li><p>If <a>this</a>'s <a for=FormControlRange>control</a> is null, then return the empty
+ string.
+
+ <li><p>Let <var>value</var> be <a>this</a>'s <a for=FormControlRange>control</a>'s <i>value
+ string</i>.
+
+ <li><p>Let <var>start</var> be <a>this</a>'s <a for=FormControlRange>start offset</a>, and let
+ <var>end</var> be <a>this</a>'s <a for=FormControlRange>end offset</a>.
+
+ <li><p>If <var>start</var> ≥ <var>end</var>, then return the empty string.
```
dandclark: Not sure we need this step. I don't think we can ever have `start` > `end`, and the case where `start` == `end` returns the empty string should be implied by the following step returning the substring from `start` to `end`.

dandclark (commented):
This is looking good! Next big steps will be defining the right integration points from the HTML spec, and working towards consensus on the API shape.

dandclark:
> Changing the behavior of setStart/setEnd wouldn't be an only way to solve this problem. I'd instead add a new argument or a new function which specifies "offset" within element's shadow tree.

@rniwa Thanks for the feedback! Yeah, there are a lot of ways this could be done. This suggestion sounds closer to the 2nd considered approach [here](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/FormControlRange/explainer.md#extending-the-range-api-2nd-approach). There are enough differences in how a Range inside a builtin element would work vs a "normal" range that we believe it will be overall cleaner and less confusing to split the functionality into a different type. For example the behavior we propose here is that a FormControlRange can't have one boundary point inside a `<textarea>`/`<input>` and another boundary point outside, which would be a clumsier invariant to express with the Range API surface. Adding this to Range would also add confusion about what things like [`Range.selectNodeContents`](https://developer.mozilla.org/en-US/docs/Web/API/Range/selectNodeContents) should do; should it move the range to select an `<input>`'s value string, or not? We could add a parameter to all the setter functions to switch the behavior, but again it seems cleaner to relegate the behavior to a different interface vs adding parameters to a bunch of Range functions.

Anyway, https://github.com/whatwg/html/issues/11478 is a better place to have this discussion -- more folks are following that issue vs this draft PR. @stephanieyzhang it might be helpful if you could update the PR description for this to point to that issue, as well as a direct link to the explainer.

dom.bs
```diff
@@ -4,6 +4,7 @@ H1: DOM
 Shortname: dom
 Text Macro: TWITTER thedomstandard
 Text Macro: LATESTRD 2025-06
+Text Macro: COMMIT-SHA 0000000
```
stephanieyzhang, line 7: Remove

dom.bs
```diff
@@ -9892,6 +9892,144 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
 [[CSSOM-VIEW]]
 
 
+<h3 id=interface-plaintext-range>Interface {{PlainTextRange}}</h3>
+
+<pre class=idl>
+[Exposed=Window]
+interface PlainTextRange : AbstractRange {
```
annevk: I think this is the wrong name now the idea is to no longer tie it to text.

dom.bs
```diff
@@ -9892,6 +9892,144 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+host-defined <a href="#plaintext-range-string">plain text range string</a> rather than in the
+<a>node tree</a>. It has no associated container node; its {{AbstractRange/startContainer}} and
+{{AbstractRange/endContainer}} getters return null, and its {{AbstractRange/startOffset}} and
+{{AbstractRange/endOffset}} getters return indices into that string.</p>
+
+<p>A {{PlainTextRange}} has associated state:</p>
+
+<ul>
+ <li><p><dfn export for=PlainTextRange id=plaintext-range-host>host</dfn> (null or an
+ {{Element}}).</p></li>
```
annevk: Why do we need this? The host should be in charge of allocating these at which point these don't really need to have a host owner.

dom.bs
```diff
@@ -9892,6 +9892,144 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+ <li><p><dfn export for=PlainTextRange id=plaintext-range-start>start offset</dfn> (a non-negative
+ integer).</p></li>
+ <li><p><dfn export for=PlainTextRange id=plaintext-range-end>end offset</dfn> (a non-negative
+ integer).</p></li>
+</ul>
+
+<p>The {{AbstractRange/startContainer}} getter steps for {{PlainTextRange}} objects are to return
+null.</p>
```
annevk: This doesn't work. You need to adjust `AbstractRange` getter (including IDL) to allow for returning null here. And we do still want to store the actual start/end container so `AbstractRange` will need to support some kind of "is opaque" switch.

dom.bs
```diff
@@ -9892,6 +9892,144 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+ start.</p></li>
+</ul>
+
+<p class=note>Specifications that define elements which <a>support plain text ranges</a> (such as
+the HTML Standard for text controls) must normatively define when a {{PlainTextRange}} is created,
+how it is associated with a host, when its host becomes null, and when its boundary points are
+updated using the rules above.</p>
+
+<p>The <dfn export for=PlainTextRange id=dom-plaintext-range-stringifier>stringification
+behavior</dfn> must run these steps:</p>
```
annevk: As I mentioned in https://github.com/whatwg/html/issues/11478#issuecomment-3531228837 I don't think exposing `toString()` is correct at this level.

dom.bs
stephanieyzhang: Moved allocation/creation to the host element API in HTML (getValueRange() on input/textarea)

dom.bs
stephanieyzhang: Thanks -- updated the AbstractRange IDL/getters. I omitted an explicit “opaque switch” since I assumed it’s an implementation detail, but I can add spec text if you think it’s needed.

dom.bs
```diff
@@ -9981,6 +9988,61 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
 
 
+<h3 id=interface-opaquerange>Interface {{OpaqueRange}}</h3>
+
+<pre class=idl>
+[Exposed=Window]
+interface OpaqueRange : AbstractRange {
+  DOMRectList getClientRects();
+  DOMRect getBoundingClientRect();
+};
```
smaug----, line 10040: Is there a reason why OpaqueRange offsets are readonly? Or there isn't any method to update them?

smaug----, line 10040: I wonder if the interface should have some kind of "disconnect" method to stop observing whatever changes it might observe internally.  Only disconnecting when removing the relevant element from DOM isn't enough, I think

stephanieyzhang: The offsets are readonly since they're inherited from AbstractRange. Currently the [HTML PR](https://github.com/whatwg/html/pull/11741) defines `createValueRange(start, end)` on the element for creation. We haven't yet added an API for updating, so that's an open question. Some options could be an element-side API (e.g. `updateValueRange(range, start, end))` to keep OpaqueRange generic, or setters directly on OpaqueRange.

Would appreciate your thoughts @annevk @smaug----

dom.bs
stephanieyzhang: Updated - added an `is opaque` flag for the `startContainer`/`endContainer` getters check.

dom.bs
stephanieyzhang: Good idea. Added `disconnect()` to explicitly disconnect the range, stop live updates, reset offsets, and return empty geometry results.

dom.bs
```diff
@@ -8694,6 +8694,11 @@ interface AbstractRange {
 <dfn export id=concept-range-end-offset for=range>end offset</dfn> is its <a for=range>end</a>'s
 <a for="boundary point">offset</a>.
 
+<p>A <a>range</a> has an associated <dfn export for=range id=concept-range-is-opaque>is opaque</dfn>
```
annevk, line 8707: associated boolean*

We don't want flags anymore.

dom.bs
```diff
@@ -8694,6 +8694,11 @@ interface AbstractRange {
...
+flag, initially false. When true, the {{AbstractRange/startContainer}} and
+{{AbstractRange/endContainer}} getters return null. Only {{OpaqueRange}} objects have this flag set
+to true.
```
annevk, line 8710: This second bit should probably be in a note.

dom.bs
```diff
@@ -8669,9 +8669,9 @@ be between 0 and the <a>boundary point</a>'s <a for="boundary point">node</a>'s
 <pre class=idl>
 [Exposed=Window]
 interface AbstractRange {
-  readonly attribute Node startContainer;
+  readonly attribute Node? startContainer;
   readonly attribute unsigned long startOffset;
```
annevk, line 8677: Does it make sense to expose startOffset and endOffset if this range potentially spans multiple nodes? They will not have meaningful values I think for opaque ranges.

dom.bs
```diff
@@ -9981,6 +9988,96 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+[Exposed=Window]
+interface OpaqueRange : AbstractRange {
+  undefined disconnect();
+  DOMRectList getClientRects();
+  DOMRect getBoundingClientRect();
+};
+</pre>
+
+<p>Objects implementing the {{OpaqueRange}} interface are known as {{OpaqueRange}} objects.
+{{OpaqueRange}} objects cannot be constructed directly; they are created by specifications defining
```
annevk, line 10010: I thought we would have a constructor so people can create these for custom elements?

dom.bs
annevk: I think we actually don't want to expose the offsets as per my other comment. I'm not sure if we want to expose collapsed. Perhaps it should just be separate from `AbstractRange`.

dom.bs
```diff
@@ -9981,6 +9988,96 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+elements that <a>support opaque ranges</a>.
+
+<p>An {{OpaqueRange}} has an
+<dfn export for=OpaqueRange>associated element</dfn> (an {{Element}} or null), initially null. It is
+set by the specification that creates the {{OpaqueRange}}.
```
annevk, line 10015: Why do we need this and `disconnect()`? Can't that be with the APIs that vend `OpaqueRange` objects? It's also a bit unclear what disconnecting means beyond that, maybe in part because this doesn't actually integrate with the tree mutation algorithms yet.

dom.bs
```diff
@@ -9981,6 +9988,96 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+</div>
+
+<p>The {{OpaqueRange/getClientRects()}} and {{OpaqueRange/getBoundingClientRect()}} methods are
+defined in [[CSSOM-VIEW]].
+
+<p>An {{OpaqueRange}} is a <a>range</a> whose <a for=range>is opaque</a> flag is true.
+
+<p>An {{Element}}
+<dfn export id=supports-opaque-range>supports opaque ranges</dfn> if its specification defines that
+it does. In HTML, this includes certain {{HTMLInputElement}} types and {{HTMLTextAreaElement}}.
```
annevk, line 10067: I'm not sure why we need this concept.

dom.bs
```diff
@@ -9981,6 +9988,96 @@ and {{Range/getBoundingClientRect()}} methods are defined in other specification
...
+
+<p class=note>Other specifications can designate additional elements, including custom elements.
+
+<p>An {{OpaqueRange}} is live, meaning its offsets are automatically updated when the underlying
```
annevk, line 10071: I think we should define this very differently. We should define opaque ranges the same way we define live ranges. And then in HTML we should say that these elements have some kind of underlying `Text` node that this opaque range "exposes".

The reason for that is because we want opaque ranges to be a generic container. And the way you have specified them they are still very much restricted to specific elements.

annevk:
It also seems this needs some rebasing.

dom.bs
stephanieyzhang: Moved into a note.

dom.bs
stephanieyzhang: Renamed to "associated boolean". Let me know if you actually wanted the concept removed entirely rather than just reworded.

dom.bs
stephanieyzhang: For form controls, they're indices into the value string, even when the range spans multiple nodes. I'm unsure of how this would look for custom elements though.

I've gotten feedback from people testing OpaqueRange on Chromium (e.g. Microsoft Editor SDK) that they rely on the offsets, using `value.substring(range.startOffset, range.endOffset)` in place of `toString()`, and comparing the caret index against the offsets to test inside/outside. Dropping them would require a workaround.

dom.bs
stephanieyzhang: For custom elements, I was thinking something like `ElementInternals.createOpaqueRange(...)` instead of a constructor. It ties creation to the element that owns the value (mirroring `createValueRange()` on form controls) and avoids having a "host-less" OpaqueRange.

rniwa: We can have a constructor with a mandatory argument?

stephanieyzhang: To make sure I'm understanding things correctly and that we're aligned:

1. Are we keeping `createValueRange()` on `<input>` and `<textarea>` as the vending API for form controls?
2. And then adding `new OpaqueRange({startContainer: node, ...})` as a separate constructor for custom elements?
3. Where does `ElementInternals.createOpaqueRange(...)` fit, if at all? Or do authors just use the constructor directly for custom-elements?
4. For `<input>`/`<textarea>` specifically, which API would devs be expected to use: `createValueRange()` or the constructor?

dom.bs
annevk: I think that means we need a `valueFromRange()` API or some such on `input` and `textarea`. Since the `input` element or `textarea` element allocated the `OpaqueRange` they will know what it contains and can thus answer such a question (if they want to). That would correctly preserve encapsulation and keep the design generic.

And as this range will have to point to the conceptual `Text` node of the element, the element can naturally only answer the API call for ranges that it created itself.

dom.bs
annevk: 1. Yes, that makes sense to me.
2. Yes.
3. We should not need anything on `ElementInternals`. Web developers can use the constructor and expose APIs such as `createValueRange()` on their elements as they see fit. Including the more complicated ranges that expose more than just a (conceptual) `Text` node.
4. They can't use the constructor because they don't have access to the conceptual `Text` node. Only `input` and `textarea` have access to that information and can thus create the range. As such web developers will have to go through `createValueRange()`.

smaug----: What does creating a `new OpaqueRange(...) `really mean? How would a custom element indicate what kind of value getClientRects as an example should return.

(Hmm, is OpaqueRange kind of a proxy object,  and then there should be the other side of the thing which affects how it behaves. In case of input and textarea, that other side is native and with Custom Elements it would need to be JS based, maybe. Just mumbling aloud.)

annevk: Yeah perhaps we need to go having an `OpaqueRangeController` (or `OpaqueRangeInternals`) directly if we offer a constructor. Basically an object that allows you to manipulate the internal values. Either using the revealing constructor pattern or doing something similar to `AbortController`. I was initially thinking that it might be sufficient if you can just indicate start and end and have the remainder fall out of it being live, but maybe not.

dom.bs
stephanieyzhang: Removed

dom.bs
stephanieyzhang: Added wording that separates the hidden boundary-point offsets from the exposed offsets. The creating specification defines and maintains the exposed offsets

dom.bs
stephanieyzhang: Removed the `associated-element` concept and `disconnect()` from this PR. If we do decide to add `disconnect()` after further discussion (or explicit lifecycle control), it'll be done in a followup PR.

dom.bs
stephanieyzhang: Removed custom-element construction from this PR. It will be addressed in a followup PR after further discussion and local prototyping

PR #1423, opened 2025-11-17
Set registry on null registry element on adopt instead of on insert

### LINKED ISSUE #1413: Allow elements with a null customElementRegistry without Shadow DOM

Custom element upgrade is currently not easily controllable. When `define` is called on a registry, all elements in document are synchronously upgraded. This can cause issues if upgrades should occur at a specific time and/or not cause a significant performance impact. 

The order of upgrades can also be problematic. While it's brittle to rely on upgrade order, normalizing order between the upgrade case and parser creation case can help improve performance. For example, events are often [used to communicate](https://github.com/webcomponents-cg/community-protocols) between elements and sometimes as a handshake to manage subscriptions / data transfer. Ensuring upgrades occur in tree order rather than define order facilitates this simply.

The  `<template>` element can be used to prevent element upgrades. However, it does not render elements at all, which is limiting.

Elements can be defined with parser blocking script so that they customize in tree order. However, blocking script is typically discouraged since it can delay time to interaction.

With the addition of scoped custom elements, a declarative Shadow DOM with a null registry can be used to control upgrade timing and ordering (see [example](https://codepen.io/Steve-Orvell/pen/GgomEVp?editors=1000)). `<div><template shadowrootmode="open" shadowrootcustomelementregistry>...`. Authors can initialize/upgrade the shadowRoot when ready. However, this requires using Shadow DOM, which can be limiting.

### How would you solve it?

1. Allow elements with a `null` `customElementRegistry` to exist in the document scope.
2. Add `customelementregistry` attribute to elements which can have a shadowRoot. This would do the same as `shadowrootcustomelementregistry` on template: assign a null registry.

This would allow authors to control the timing and ordering of custom element upgrade without relying  on declarative Shadow DOM, `<template>`, or blocking scripts.

### Anything else?

_No response_

ja-y-son:
In order to support scenarios for elements with a null custom element registry without shadow DOM, we decided to modify the behavior when it comes to setting registry on null registry element. By setting the registry on adopt instead of on insert allows us to have null registry element in light DOM while being backwards-compatible.

Fixes #1413 partially. There are two more spec changes expected to add element attribute and allow null registry option during element creation and shadow attachment as mentioned in https://github.com/whatwg/dom/issues/1413#issuecomment-3530198733

   * WebKit
   * Chromium
   * https://github.com/web-platform-tests/wpt/commit/cc1c056521abd0ff97456455dbebe113617a2bc1
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=302970

***
<a href="https://whatpr.org/dom/1423.html" title="Last updated on Nov 25, 2025, 10:29 AM UTC (9821f8a)">Preview</a> | <a href="https://whatpr.org/dom/1423/3fbec78...9821f8a.html" title="Last updated on Nov 25, 2025, 10:29 AM UTC (9821f8a)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -6016,8 +6016,8 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
 </div>
 
 <div algorithm>
-<p>To <dfn export id=concept-node-adopt>adopt</dfn> a <var>node</var> into a <var>document</var>, run
-these steps:
+<p>To <dfn export id=concept-node-adopt>adopt</dfn> a <a for=/>node</a> <var>node</var> into a
+<a for=/>document</a> <var>document</var>:
 
 <ol>
  <li><p>Let <var>oldDocument</var> be <var>node</var>'s <a for=Node>node document</a>.
@@ -6030,8 +6030,8 @@ these steps:
 
   <ol>
    <li>
-    <p>For each <var>inclusiveDescendant</var> in <var>node</var>'s
-    <a>shadow-including inclusive descendants</a>:
+    <p>For each <var>inclusiveDescendant</var> of <var>node</var>'s
+    <a>shadow-including inclusive descendants</a>, in <a>shadow-including tree order</a>:
 
     <ol>
      <li><p>Set <var>inclusiveDescendant</var>'s <a for=Node>node document</a> to <var>document</var>.
@@ -6067,14 +6067,15 @@ these steps:
       </ol>
     </ol>
 
-   <li><p>For each <var>inclusiveDescendant</var> in <var>node</var>'s
-   <a>shadow-including inclusive descendants</a> that is <a for=Element>custom</a>,
-   <a>enqueue a custom element callback reaction</a> with <var>inclusiveDescendant</var>, callback
-   name "<code>adoptedCallback</code>", and « <var>oldDocument</var>, <var>document</var> ».
+   <li><p>For each <var>inclusiveDescendant</var> of <var>node</var>'s
+   <a>shadow-including inclusive descendants</a> that is <a for=Element>custom</a>, in
+   <a>shadow-including tree order</a>: <a>enqueue a custom element callback reaction</a> with
+   <var>inclusiveDescendant</var>, callback name "<code>adoptedCallback</code>", and
+   « <var>oldDocument</var>, <var>document</var> ».
    <!-- attributeChangedCallback is also old, then new -->
 
-   <li><p>For each <var>inclusiveDescendant</var> in <var>node</var>'s
-   <a>shadow-including inclusive descendants</a>, in <a>shadow-including tree order</a>, run the
+   <li><p>For each <var>inclusiveDescendant</var> of <var>node</var>'s
+   <a>shadow-including inclusive descendants</a>, in <a>shadow-including tree order</a>: run the
    <a>adopting steps</a> with <var>inclusiveDescendant</var> and <var>oldDocument</var>.
   </ol>
 </ol>


ja-y-son:
@rniwa @annevk PTAL, thanks!

Ryosuke, I saw that you have a [PR](https://github.com/WebKit/WebKit/pull/54004) out to support null registry element scenario. Thank you so much for putting those tests together! I'll also go through the current WPTs on my end to see if there are anything else we need to update for these spec changes.

dom.bs
```diff
@@ -2902,12 +2902,7 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
 
       <ol>
-       <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is
-       null, then set <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> to
-       the result of <a for=/>looking up a custom element registry</a> given
-       <var>inclusiveDescendant</var>'s <a for=tree>parent</a>.
-
-       <li><p>Otherwise, if <var>inclusiveDescendant</var>'s
+       <li><p>If <var>inclusiveDescendant</var>'s
        <a for=Element>custom element registry</a>'s <a for=CustomElementRegistry>is scoped</a> is
```
annevk: If the registry is null we have UB here, no?

dom.bs
```diff
@@ -6018,12 +6006,14 @@ these steps:
     <ol>
      <li><p>Set <var>inclusiveDescendant</var>'s <a for=Node>node document</a> to <var>document</var>.
 
-     <li><p>If <var>inclusiveDescendant</var> is a <a for=/>shadow root</a> and
-     <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>
-     <a>is a global custom element registry</a>, then set <var>inclusiveDescendant</var>'s
-     <a for=ShadowRoot>custom element registry</a> to <var>document</var>'s
-     <a>effective global custom element registry</a>.
-
+     <li><p>If <var>inclusiveDescendant</var> is a <a for=/>shadow root</a>:
```
annevk: Wrapping here doesn't follow the guidelines.

dom.bs
```diff
@@ -6018,12 +6006,14 @@ these steps:
...
+      <ol>
+        <li>If <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>
+        <a>is a global custom element registry</a> or null, then set <var>inclusiveDescendant</var>'s
+        <a for=ShadowRoot>custom element registry</a> to <var>document</var>'s
+        <a>effective global custom element registry</a>.
```
annevk: Where exactly are we accounting for "keep custom element registry null" now?

dom.bs
```diff
@@ -6032,7 +6022,7 @@ these steps:
        <var>inclusiveDescendant</var>'s <a for=Element>attribute list</a> to <var>document</var>.
 
        <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>
-       <a>is a global custom element registry</a>, then set <var>inclusiveDescendant</var>'s
+       <a>is a global custom element registry</a> or null, then set <var>inclusiveDescendant</var>'s
```
annevk: I think this is weird now. "is a global custom element registry" also tries to account for null, but now you are invalidating that. You should change the concepts around.

dom.bs
ja-y-son: Updated and put the `<p>` tag in a new line.

dom.bs
ja-y-son: Added condition to make sure only element with non-null custom element registry will run these couple steps.

dom.bs
ja-y-son: We do not assign custom element registry during insertion steps so that will account for "keep custom element registry null", but then we need to assign the global registry at adopt steps to ensure the backwards compatibility.

dom.bs
ja-y-son: Updated and split up the logic here.

sorvell (commented):
This matches my understanding of the first set of changes proposed in https://github.com/whatwg/dom/issues/1413#issuecomment-3530198733 and agreed to at TPAC. Good job!

ja-y-son:
@rniwa One thing I'd like to clarify from your [PR](https://github.com/WebKit/WebKit/pull/54004): One of the test files added is phrased as "append does not upgrade". However, is it correct that we still want to run upgrade during append/insertion steps? It is possible that a null registry element gets a registry when it's appended cross document (which involves adopt steps). In that case we'll want to run an upgrade on the element right? Want to make sure I got the correct understanding here.

rniwa:
> @rniwa One thing I'd like to clarify from your [PR](https://github.com/WebKit/WebKit/pull/54004): One of the test files added is phrased as "append does not upgrade". However, is it correct that we still want to run upgrade during append/insertion steps?

So regular append should upgrade elements that have a non-null custom element registry. An element with null registry should retain its null registry state and not get upgraded upon insertion.

> It is possible that a null registry element gets a registry when it's appended cross document (which involves adopt steps). In that case we'll want to run an upgrade on the element right? Want to make sure I got the correct understanding here.

Yes, in the cross-document append case, we'd first adopt the element to the destination document at which point the null registry will be replaced with the global registry of the destination document.

dom.bs
annevk: Does that mean we can remove "keep custom element registry null" completely? I think it's only used for this.

dom.bs
```diff
@@ -2899,15 +2899,11 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
      <a for=iteration>continue</a>.
 
      <li>
-      <p>If <var>inclusiveDescendant</var> is an <a for=/>element</a>:
+      <p>If <var>inclusiveDescendant</var> is an <a for=/>element</a> and
+      <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is not null:
```
```suggestion
      <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is non-null:
```

dom.bs
```diff
@@ -6018,23 +6007,35 @@ these steps:
...
-     <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>
-     <a>is a global custom element registry</a>, then set <var>inclusiveDescendant</var>'s
-     <a for=ShadowRoot>custom element registry</a> to <var>document</var>'s
-     <a>effective global custom element registry</a>.
+     <li>
+      <p>If <var>inclusiveDescendant</var> is a <a for=/>shadow root</a>:
 
+      <ol>
+        <li><p>If <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>
+        is null, set <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>
```
annevk: then set

annevk: I think overall these two steps here and below would be better as follows, without using "is a global custom element registry" as that would do a redundant null check:

> If _node_'s custom element registry is null or _node_'s custom element registry's is scoped is false, then set _node_'s custom element registry to _document_'s effective global custom element registry.

annevk: Also, it appears you are changing the indentation here and below. This document uses single-space indentation.

ja-y-son: Updated!

dom.bs
ja-y-son: Removed.

ja-y-son:
> One minor point: it seems like since insert 7.1 does adopt, perhaps insert 7.7.3.1 and 7.7.4.1 are no longer needed?

@sorvell My understanding is that the scoped document set is used for searching for upgrade candidates in the case of new definition added, so we'll need to keep those two steps around.

dom.bs
```diff
@@ -6018,11 +6007,16 @@ these steps:
...
-     <a>effective global custom element registry</a>.
+     <li>
+      <p>If <var>inclusiveDescendant</var> is a <a for=/>shadow root</a>:
+
+      <ol>
+       <li><p>if <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>
+       is null or <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>'s
+       <a for=CustomElementRegistry>is scoped</a> is false, then set
+       <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a> to
+       <var>document</var>'s <a>effective global custom element registry</a>.
```
```suggestion
     <li><p>If <var>inclusiveDescendant</var> is a <a for=/>shadow root</a> and either
     <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a> is null or
     <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>'s
     <a for=CustomElementRegistry>is scoped</a> is false, then set <var>inclusiveDescendant</var>'s
     <a for=ShadowRoot>custom element registry</a> to <var>document</var>'s
     <a>effective global custom element registry</a>.
```

annevk: But I think this is incorrect and we still need "keep custom element registry null" here. Have you looked at how the HTML Standard sets "keep custom element registry null"? How is the `shadowrootcustomelementregistry` content attribute supported in this new world?

ja-y-son: In [HTML spec](https://html.spec.whatwg.org/#parsing-main-inhead) where we define the parsing logic for "**A start tag whose tag name is "template**", we have two repetitive lines describing how `shadowrootcustomelementregistry` attribute is used:
- If templateStartTag has a [shadowrootcustomelementregistry](https://html.spec.whatwg.org/#attr-template-shadowrootcustomelementregistry) attribute, then set registry to null.
- If templateStartTag has a [shadowrootcustomelementregistry](https://html.spec.whatwg.org/#attr-template-shadowrootcustomelementregistry) attribute, then set shadow's [keep custom element registry null](https://dom.spec.whatwg.org/#shadowroot-keep-custom-element-registry-null) to true.

I think we can remove that second line and completely remove the usage of "keep custom element registry null" and we can simply rely on setting the shadowroot's registry to null to support `shadowrootcustomelementregistry` content attribute behavior.

annevk: How do we distinguish the case of a shadow root from a null document registry without the attribute from one with the attribute?

ja-y-son: Is it possible to have a DSD with null registry due to `shadowrootcustomelementregistry` in a null registry document though? Assuming my understanding is correct on "DSD is only created during document parsing and we're not able to get a null registry document during document parsing (has to through JS code like `document.implementation.createHTMLDocument();`)"

In that case, we can safely assign a registry to null registry shadow root on adopt and remove the need of "keep custom element registry" as we don't assign registry on append regardless.

sorvell: > DSD is only created during document parsing

It can be created with `setHTMLUnsafe`.

sorvell: @annevk Is this the reason or related to it that you're suggesting we still need the check? To distinguish these cases?

```html
<template id="template">
  <div id="host1">
    <template shadowrootmode="open" shadowrootclonable><x-foo></x-foo></template>
  </div>
  <div id="host2">
    <template shadowrootmode="open" shadowrootclonable shadowrootcustomelementregistry><x-foo></x-foo></template>
  </div>
</template>
```

```js
customElements.define('x-foo', class extends HTMLElement {});
document.body.append(template.content.cloneNode(true));
console.assert(host1.shadowRoot.querySelector('x-foo:defined') === true);
console.assert(host2.shadowRoot.querySelector('x-foo:defined') === false);
```

annevk: Yeah, except I think you also need `shadowrootclonable` attributes.

sorvell: yeah, makes sense. Edited example above for clarity.

ja-y-son: Makes sense, I put "keep custom element registry" back and check it during node adoption. It seems like we're missing some WPT coverage on the behavior of "keep custom element registry". Will add it.

dom.bs
```diff
@@ -2899,17 +2899,13 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
      <a for=iteration>continue</a>.
 
      <li>
-      <p>If <var>inclusiveDescendant</var> is an <a for=/>element</a>:
+      <p>If <var>inclusiveDescendant</var> is an <a for=/>element</a> and
+      <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is non-null:
```
annevk, line 2914: I suspect this is not a change in behavior, as we can't upgrade elements with a null registry, but was this change intentional?

dom.bs
```diff
@@ -2899,17 +2899,13 @@ before a <var>child</var>, with an optional <i>suppress observers flag</i>, run
...
-       null, then set <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> to
-       the result of <a for=/>looking up a custom element registry</a> given
-       <var>inclusiveDescendant</var>'s <a for=tree>parent</a>.
-
-       <li><p>Otherwise, if <var>inclusiveDescendant</var>'s
-       <a for=Element>custom element registry</a>'s <a for=CustomElementRegistry>is scoped</a> is
-       true, <a for=set>append</a> <var>inclusiveDescendant</var>'s <a for=Node>node document</a> to
+       <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
+       <a for=CustomElementRegistry>is scoped</a> is true, then <a for=set>append</a>
+       <var>inclusiveDescendant</var>'s <a for=Node>node document</a> to
```
annevk, line 2919: To be sure, we continue to do this during insert instead of adopt to minimize the number of documents a registry has to look through, right?

dom.bs
```diff
@@ -6023,11 +6007,21 @@ these steps:
...
+      <p>If <var>inclusiveDescendant</var> is a <a for=/>shadow root</a> and if any of the following
+      are true:
+
+      <ul>
+       <li><p><var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a> is null
+       and <var>inclusiveDescendant</var>'s <a for=ShadowRoot>keep custom element registry null</a>
+       is false; or
+
+       <li><p><var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a>
+       <a>is a global custom element registry</a>,
```
annevk, line 6049: I changed this to "is a global custom element registry" as this was lacking a null check before.

annevk (approved):
I rebased and pushed some nits. This looks good to me now but if people could have another look and answer my questions I'd appreciate it.

dom.bs
ja-y-son: Yes that is also my understanding.

dom.bs
ja-y-son: I see what you meant. Given that we can't upgrade element nor getting a connectedCallback with null registry, I think we can leave this null check as is.

dom.bs
ja-y-son: @annevk @sorvell 
One thing I was thinking today about the adoption steps:
From the above example, based on the new spec, it's clear that shadowroot should be getting effective global registry or null depending on "keep custom element registry null" and the original factors (null or is using another global registry) during adoption.
But what about `x-foo`? I think the intended behavior is that `x-foo`' s registry will remain to be null. However, based on the new spec, the element has null registry originally, so during adoption in above case, it will gain the effective global registry of document, which does not align with the expected behavior.
When we set registry for null registry element during insertion in previous spec, we assign the registry based off the inclusiveDescendant's parent's registry. Do we need to do something similar when we're setting registry for null registry element during adoption? Perhaps something like "set inclusiveDescendant's parent's (or inclusiveDescendant's tree scope's) registry if it is an effective global registry in the new document, otherwise set null"

sorvell: For the x-foo in host2, yeah, think you’re right if not otherwise covered. Case:
1. Adopted element is null
2. Adopting document is global
3. But adopted element’s root node is null with keep registry null is true.

ja-y-son:
@annevk A case I’m not sure if we covered in the updated spec and would like to know your thoughts:
https://github.com/whatwg/dom/pull/1423#discussion_r2563096391

annevk:
I filed #1429 to track that. It does seem problematic. Unfortunately I missed it before I hit merge.

PR #1424, opened 2025-11-19
Allow the creation of null registry element and shadow root

### LINKED ISSUE #1413: Allow elements with a null customElementRegistry without Shadow DOM

ja-y-son:
As part of the effort to support null custom element registry elements without shadow DOM, we're updating the spec to allow the creation of element and shadowroot to use null as a valid argument and create null registry element and shadowroot.

Fixes #1413 partially. We need another PR to support element attribute for null custom element registry in addition to #1423.

   * webkit
   * chromium
   * https://github.com/web-platform-tests/wpt/pull/56108
   * https://github.com/web-platform-tests/wpt/pull/56176
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=302583

***
<a href="https://whatpr.org/dom/1424.html" title="Last updated on Nov 24, 2025, 8:37 AM UTC (48ff4a0)">Preview</a> | <a href="https://whatpr.org/dom/1424/918ebc0...48ff4a0.html" title="Last updated on Nov 24, 2025, 8:37 AM UTC (48ff4a0)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5856,18 +5856,24 @@ dictionary <var>options</var> and a <a for=/>document</a> <var>document</var>:
   <p>If <var>options</var> is a dictionary:
 
   <ol>
-   <li><p>If <var>options</var>["{{ElementCreationOptions/customElementRegistry}}"]
-   <a for=map>exists</a>, then set <var>registry</var> to it.
-
-   <li><p>If <var>registry</var> is non-null, and <var>registry</var>'s
-   <a for=CustomElementRegistry>is scoped</a> is false, and <var>registry</var> is not
-   <var>document</var>'s <a for=Document>custom element registry</a>, then <a>throw</a> a
-   "{{NotSupportedError!!exception}}" {{DOMException}}.
-
    <li><p>If <var>options</var>["{{ElementCreationOptions/is}}"] <a for=map>exists</a>, then set
    <var>is</var> to it.
 
-   <li><p>If <var>registry</var> is non-null and <var>is</var> is non-null, then <a>throw</a> a
+   <li>
+    <p>If <var>options</var>["{{ElementCreationOptions/customElementRegistry}}"]
+    <a for=map>exists</a>:
+
+    <ol>
+     <li><p>If <var>is</var> is non-null, then <a>throw</a> a "{{NotSupportedError!!exception}}"
+     {{DOMException}}.
+
+     <li><p>Set <var>registry</var> to
+     <var>options</var>["{{ElementCreationOptions/customElementRegistry}}"].
+    </ol>
+
+   <li><p>If <var>registry</var> is non-null, <var>registry</var>'s
+   <a for=CustomElementRegistry>is scoped</a> is false, and <var>registry</var> is not
+   <var>document</var>'s <a for=Document>custom element registry</a>, then <a>throw</a> a
    "{{NotSupportedError!!exception}}" {{DOMException}}.
   </ol>
 
@@ -7606,7 +7612,7 @@ are:
  <li><p>If <var>init</var>["{{ShadowRootInit/customElementRegistry}}"] <a for=map>exists</a>, then
  set <var>registry</var> to it.
 
- <li><p>If <var>registry</var> is non-null, and <var>registry</var>'s
+ <li><p>If <var>registry</var> is non-null, <var>registry</var>'s
  <a for=CustomElementRegistry>is scoped</a> is false, and <var>registry</var> is not <a>this</a>'s
  <a>node document</a>'s <a for=Document>custom element registry</a>, then <a>throw</a> a
  "{{NotSupportedError!!exception}}" {{DOMException}}.


ja-y-son:
@rniwa @annevk here's another PR from what we discussed.

May be a dumb question: with this change, we probably no longer want to use null as the default vaule for `customElementRegistry` in `ShadowRootInit`. In that case are we still accounting for the use case where developer passing in a ShadowRoot instance as the argument to `attachShadow`?

rniwa:
You mean someone doing: `host.attachShadow(anotherShadowRoot)`? Given `customElementRegistry` property on `ShadowRoot` continue to return the custom element registry of the shadow root, I don't see how that'll be broken?

sorvell:
This matches my understanding of the first set of changes proposed in https://github.com/whatwg/dom/issues/1413#issuecomment-3530198733 and agreed to at TPAC. Good job!

dom.bs
```diff
@@ -7604,17 +7603,12 @@ are:
  <li><p>Let <var>registry</var> be <a>this</a>'s <a>node document</a>'s
```
sorvell, line 7609: Minor nit: this is correct but maybe some emphasis should be added as it's easy to gloss over the word "document" and it's critical that the default is the document (global) registry and *not* the node's registry.

annevk: I don't think we want to emphasize node document throughout this document. That would just end up looking busy.

dom.bs
```diff
@@ -5858,9 +5859,10 @@ dictionary <var>options</var> and a <a for=/>document</a> <var>document</var>:
    <li><p>If <var>options</var>["{{ElementCreationOptions/customElementRegistry}}"]
    <a for=map>exists</a>, then set <var>registry</var> to it.
 
-   <li><p>If <var>registry</var>'s <a for=CustomElementRegistry>is scoped</a> is false and
-   <var>registry</var> is not <var>document</var>'s <a for=Document>custom element registry</a>,
-   then <a>throw</a> a "{{NotSupportedError!!exception}}" {{DOMException}}.
+   <li><p>If <var>registry</var> is not null and <var>registry</var>'s
+   <a for=CustomElementRegistry>is scoped</a> is false and <var>registry</var> is not
+   <var>document</var>'s <a for=Document>custom element registry</a>, then <a>throw</a> a
+   "{{NotSupportedError!!exception}}" {{DOMException}}.
```
```suggestion
   <li><p>If <var>registry</var> is non-null, <var>registry</var>'s
   <a for=CustomElementRegistry>is scoped</a> is false, and <var>registry</var> is not
   <var>document</var>'s <a for=Document>custom element registry</a>, then <a>throw</a> a
   "{{NotSupportedError!!exception}}" {{DOMException}}.
```

dom.bs
```diff
@@ -7604,17 +7603,12 @@ are:
...
-
-   <li><p>If <var>registry</var>'s <a for=CustomElementRegistry>is scoped</a> is false and
-   <var>registry</var> is not <a>this</a>'s <a>node document</a>'s
-   <a for=Document>custom element registry</a>, then <a>throw</a> a
-   "{{NotSupportedError!!exception}}" {{DOMException}}.
-  </ol>
+ <li><p>If <var>registry</var> is not null and <var>registry</var>'s
+ <a for=CustomElementRegistry>is scoped</a> is false and <var>registry</var> is not <a>this</a>'s
+ <a>node document</a>'s <a for=Document>custom element registry</a>, then <a>throw</a> a
+ "{{NotSupportedError!!exception}}" {{DOMException}}.
```
```suggestion
 <li><p>If <var>registry</var> is non-null, <var>registry</var>'s
 <a for=CustomElementRegistry>is scoped</a> is false, and <var>registry</var> is not <a>this</a>'s
 <a>node document</a>'s <a for=Document>custom element registry</a>, then <a>throw</a> a
 "{{NotSupportedError!!exception}}" {{DOMException}}.
```

dom.bs
```diff
@@ -7604,17 +7603,12 @@ are:
  <li><p>Let <var>registry</var> be <a>this</a>'s <a>node document</a>'s
  <a for=Document>custom element registry</a>.
 
- <li>
-  <p>If <var>init</var>["{{ShadowRootInit/customElementRegistry}}"] is non-null:
+ <li><p>Set <var>registry</var> to <var>init</var>["{{ShadowRootInit/customElementRegistry}}"].
```
annevk: Actually this doesn't work. It needs to be conditional on customElementRegistry existing.

ja-y-son: Nice catch, added the condition.

annevk:
OP is just lacking implementation bugs at this point, unless you want to add more test coverage still?

ja-y-son:
Added the implementation bugs in OP and the test coverage should be good for now. This PR should be ready to land :)

annevk:
While doing a final check here I noticed that the new logic would end up throwing for the combination of "is" and "customElementRegistry" way too often. While WebKit would not object to that, I doubt that was the intent. I pushed a commit to fix that (and also fix some remaining nits).

Did you not yet implement this change perhaps or do we not have good test coverage for this?

ja-y-son:
Thanks for catching that "is" and "customElementRegistry" combination issue and fixing the nits I missed before! We've yet to implement this change and that was indeed not intended.

Issue #1389, opened 2025-07-14
Ability to wait for async abort algorithms?


saschanaz:
### What is the issue with the DOM Standard?

Currently it's not straightforward to track whether all abort algorithms are finished. You would call `AbortController.abort` and then would hope it that things are aborted, before starting anything new. It's okay to do so when the new job never clashes with the previous supposed-to-be-aborted job (like some simple `fetch()` calls), but might clash if the abort algorithm includes closing a file handle, etc.

Would be nice if abort event had something like `ExtendableEvent.waitUntil`.

annevk:
Isn't this tracked by the thing you are aborting? It should return an AbortError at some point.

Also, if this is proposing something new, any reason to not follow the template?

saschanaz:
> Isn't this tracked by the thing you are aborting?

Perhaps? But that means "the thing" can't purely use AbortSignal internally but need to implement something else, if "the thing" is also aborting its downstream.

> any reason to not follow the template?

Selected a wrong template 🤦🏻‍♀️

annevk:
I suppose. If you define `async getData` that takes a signal. And internally it uses `async getDataInternal` that also takes a signal. It could pass the outer signal to `getDataInternal` and await `getDataInternal` to see if it gets an AbortError or some kind of success and based on that decide to return or throw for `getData`. It's a choice of course whether it wants to wait for `getDataInternal`, but there's nothing we can really do about that I think.

saschanaz:
Hmm, I was thinking more about addEventListener-ish "keep working until you get the abort signal" situation. Suppose something like:

```js
while (!signal.aborted) {
  await write(serialize(data), { signal })
  await flush();
  sleep(10000)
}
```

And then somewhere else you do:

1. Abort the file write repeat
2. Upload the file

At step 2 you don't know whether the last write is flushed or not, because abort() doesn't wait for anything. (in this specific case you can probably call `flush()` again to make sure, though.)

What would be nicer (while the API addition is arbitrary and probably not nice enough)

```js
while (!signal.aborted) {
  signal.cleanup(() => flush());
  await write(serialize(data), { signal })
  sleep(10000)
}

// And then somewhere else:
// makes sure everything is aborted
await controller.abort()
```

PR #1425, opened 2025-11-20
Add abortSignal.addAbortCallback

jakearchibald:
This PR is a discussion starter. The goals:

- Make progress on allowing TC39 able to adopt/use `AbortController` and `AbortSignal` without requiring `Event` and `EventTarget`.
- Provide a system where abort 'listeners' can be GC'd as soon as abort has happened.

The behaviour in this PR means that abort reactions in userland and the web platform interleave. @bakkot is keen on this. I'm a little unsure, so I'm interested in more opinions.

I'm a little worried about:

```js
const signal = controller.signal;

let thing;

signal.addAbortCallback(() => {
  console.log(thing.running);
  // The above is unexpectedly true,
  // because the platform's steps to set it to false haven't run yet.
});

thing = doThing(signal);
controller.abort();
```

But maybe this already happens with promises?

---

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

***
<a href="https://whatpr.org/dom/1425.html" title="Last updated on Nov 20, 2025, 5:22 AM UTC (3acd735)">Preview</a> | <a href="https://whatpr.org/dom/1425/918ebc0...3acd735.html" title="Last updated on Nov 20, 2025, 5:22 AM UTC (3acd735)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1926,6 +1926,7 @@ interface AbortSignal : EventTarget {
   readonly attribute boolean aborted;
   readonly attribute any reason;
   undefined throwIfAborted();
+  undefined addAbortCallback(VoidFunction callback);
 
   attribute EventHandler onabort;
 };</pre>
@@ -2036,6 +2037,26 @@ is [=AbortSignal/aborted=]; otherwise false.
 
 <p>The <dfn method for=AbortSignal>throwIfAborted()</dfn> method steps are to throw <a>this</a>'s
 <a for=AbortSignal>abort reason</a>, if <a>this</a> is [=AbortSignal/aborted=].
+
+<div algorithm>
+<p>The <dfn method for=AbortSignal><code>addAbortCallback(<var>callback</var>)</code></dfn> method steps are:
+
+ <ol>
+  <li><p>If <a>this</a> is [=AbortSignal/aborted=], then:
+   <ol>
+    <li><p><a spec=webidl>invoke</a> <var>callback</var> with « », and "<code>report</code>".
+
+    <li><p>Return.
+   </ol>
+
+
+  <li>
+   <p>[=AbortSignal/Add|Add the following abort steps=] to [=this=]:</p>
+   <ol>
+    <li><p><a spec=webidl>invoke</a> <var>callback</var> with « », and "<code>report</code>".
+   </ol>
+ </ol>
+</div>
 
 <div class=example id=example-throwifaborted>
  <p>This method is primarily useful for when functions accepting {{AbortSignal}}s want to throw (or


bakkot:
With regards to the example, the way a user is actually likely to run into this is more like

```js
import { operation } from 'some-framework';

async function whatever(signal) {
  await Promise.all([
    builtin1({ signal }),
    operation({ signal }),
    builtin2({ signal }),
  ]);
}
```

From the user's perspective, it is very strange if these are aborted in anything other than LIFO or FIFO order. Users are not (and should not be) reasoning about whether the operations they're calling are defined in their framework or in the platform, and there is no way to understand the `builtin1, builtin2, operation` order without such reasoning.

jakearchibald:
That example is pretty compelling.

saschanaz:
I'd like the design to allow #1389 in the future 👀 (doesn't seem to block it right now, which is cool)

jasnell:
From a language-level perspective, I think it's better if we defined the mechanism as a cancelation/abort protocol in the language... of which `AbortSignal` could be update to be / considered an implementation.

Specifically, I've been imagining something along the lines of...

```js
const genericAbortSignal = {
  [Symbol.onabort]() {
    // Register and return the abort handle
    return {
      cancel() { /* Unregister the abort handler */ },
      [Symbol.dispose]() { this.cancel(); }
    };
  },
};

doSomething(genericAbortSignal);

async function doSomething(signal) {
  using abortHandle = signal[Symbol.onabort]();
  abortHandle.onabort = (reason) => { /* ... */ };
  // Do async stuff that can be aborted...
}
```

Then, the actual `AbortSignal` class here could just add `[Symbol.addAbortHandler]` such that it defers to `addEventListener`.

This type of approach would allow any object to become an AbortSignal. It would allow TC39 to define abort semantic without relying on `AbortSignal`, `AbortController`, `Event`, or `EventTarget`.

The requirement of the protocol would be that a call to `Symbol.onabort` must return an object that minimally must expose a `cancel()` method that would unregister the handler from the signal.

`AbortSignal` could easily be adapted to support this:

```js
class AbortSignal extends EventTarget {
  // ...

  [Symbol.onabort]() {
    let callback;
    const self = this;
    const handler = {
      cancel() {
        self.removeEventListener('abort', callback);
      };
    };
    callback = (event) => handler.onabort?.(event.reason);
    this.addEventListener('abort', callback, { once: true });
    return handler;
  }
}
```

bakkot:
@jasnell There's no need to define a symbol-based protocol. The protocol for consumers can literally be "call `.addAbortCallback(yourCleanupFunction)`". That requires no other changes from WHATWG.

(Though I agree that a `removeAbortCallback(yourCleanupFunction)`, or similar mechanism, would also be nice.)

---

Edit: also, your implementation ends up still using the `addEventListener`; the primary purpose of this PR is to give us a path to avoid that machinery, because of [the various problems with it](https://matrixlogs.bakkot.com/TC39_Delegates/2025-11-19#L229).

jasnell:
> ... also, your implementation ends up still using the addEventListener

Only within a hypothetical impl of the abort protocol in the existing `AbortSignal` machinery. That would not be an actual requirement tho. I showed that only for illustration.

I think the nice thing about protocol approach is that it would not require use of the `AbortSignal` class at all. From the languages point of view, an "abort signal" can be any object that happens to implement `@@Symbol.onabort` (or whatever we want to call it). Exactly how that is implemented is left entirely open.

bakkot:
Calling `.addAbortCallback(yourCleanupFunction)` works just as well as using a symbol; arbitrary objects can have a method named `"addAbortCallback"`. There's no need to introduce a new symbol here.

The main reasons to use a symbol-based rather than string-based protocol are when this is a protocol that arbitrary objects might implement (like Iterator), or when you want to switch on the presence of the protocol (like Promise). Neither holds here.

jasnell:
Symbol vs Not-a-symbol isn't that important to me. I'm good with either. More focused on the overall pattern than the particular color of the bikeshed ;-)

bakkot:
So the only difference between what you're proposing and what's in this PR already is just that instead of the new method returning `undefined` it would return an object with a `cancel` method which removes the listener?

jasnell:
That and we'd be sure to define it (in the language) as a protocol that is not dependent specifically on using `AbortSignal`/`AbortController`. I want to make certain that it is allowable that `!(signal instanceof AbortSignal` but it would still work...

that is, I want us to say, "An abort signal is any object that implements [...] function" ... as opposed to "An abort signal is always an `AbortSignal` as defined by WHATWG"... hopefully that make sense.

bakkot:
Yes, the idea is that ECMAScript would literally use the new method added in this PR (i.e., it will just attempt to call `addAbortCallback` on whatever object is passed).

If ECMAScript was going to reach into the internals of AbortSignal there would be no need for the new method.

dom.bs
```diff
@@ -2037,6 +2038,26 @@ is [=AbortSignal/aborted=]; otherwise false.
...
+    <li><p><a spec=webidl>invoke</a> <var>callback</var> with « », and "<code>report</code>".
+
+    <li><p>Return.
+   </ol>
+
+
+  <li>
+   <p>[=AbortSignal/Add|Add the following abort steps=] to [=this=]:</p>
+   <ol>
+    <li><p><a spec=webidl>invoke</a> <var>callback</var> with « », and "<code>report</code>".
```
acutmore, line 2056: Noob question: is it implicit that the callback is no longer held once it's called?

Jamesernator: The abort algorithms list is [cleared when abort signals are run](https://dom.spec.whatwg.org/#run-the-abort-steps).

acutmore: Wonderful! Thanks @Jamesernator

acutmore:
`addAbortCallback` returning a disposable (`Symbol.dispose`) sounds useful to me - or are there concerns about always allocating an object that might only sometimes be used?

michaelficarra:
> or are there concerns about always allocating an object that might only sometimes be used?

It's pretty easy for an engine to know when this expression is in a position where the return value is not used.

bakkot:
Was reminded of prior discussion by @littledan in https://gist.github.com/littledan/47b4fe9cf9196abdcd53abee940e92df (which lands in pretty much the same place as this PR, but also suggests returning a disposable "unregister" capability). Also cc @domfarolino and @rbuckton as people who participated in that discussion.

PR #1431, opened 2025-11-28
Make DOMTokenList's supported token concept also depend on element

annevk:
It makes no sense for it to just depend on the attribute name.

Also modernize this section while here.

(PR template doesn't apply as this is mostly a clarification. The creation steps for this object seem incorrect to me by the way, as I doubt we want to make that result in a mutation record, but that's best solved separately.)

***
<a href="https://whatpr.org/dom/1431.html" title="Last updated on Nov 28, 2025, 8:31 AM UTC (d9dd565)">Preview</a> | <a href="https://whatpr.org/dom/1431/9b0f03a...d9dd565.html" title="Last updated on Nov 28, 2025, 8:31 AM UTC (d9dd565)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -10501,7 +10501,7 @@ running <a>get an attribute value</a> given <var>set</var>'s <a for=DOMTokenList
 
 <hr>
 
-<div algorithm>
+<div algorithm="DOMTokenList/attribute change steps">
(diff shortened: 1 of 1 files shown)


noamr (approved):
LGTM, seems pretty straightforward. Thanks!

PR #1433, opened 2025-11-28
Address infinite loop in TreeWalker's nextNode()

### LINKED ISSUE #1102: TreeWalker.nextNode loops forever if last node of a detached node tree is not accepted by the filter
For example, given:
```js
var el = document.createElement("div");
var walker = document.createTreeWalker(document, 0, null);
walker.currentNode = el;
var next = walker.nextNode()
```

Following these steps: https://dom.spec.whatwg.org/#dom-treewalker-nextnode
```
1. Let node be this's current.
```

This is `el` above.

```
2. Let result be FILTER_ACCEPT.
3. While true:
   1. While result is not FILTER_REJECT and node has a child:
```

`result` is `FILTER_ACCEPT` but `el` has no children, so we skip this.

```
    2. Let sibling be null.
    3. Let temporary be node.
    4. While temporary is non-null:
       1. If temporary is this’s root, then return null.
```
`temporary` is `el` and `this's root` is `document`, so we skip this.

```
       2. Set sibling to temporary’s next sibling.
       3. If sibling is non-null, then set node to sibling and break.
```

`el` has no siblings, so we skip this.

```
        4. Set temporary to temporary’s parent.
```
`el` has no parent, so we set temporary to `null` and thus exit the loop.

```
    5. Set result to the result of filtering node within this.
    6. If result is FILTER_ACCEPT, then set this’s current to node and return node.
```

We passed `0` for `whatToShow` in `createTreeWalker`, meaning result is `FILTER_SKIP`, so we skip this and return to the top of the `While true:`

Note that on all future iterations, none of the conditions have changed as the only variable that's changed is `result`, which has changed from `FILTER_ACCEPT` to `FILTER_SKIP`. This only affects step 3.1 which still fails to meet the condition.

It seems that `If temporary is this’s root, then return null.` was meant to protect against this, but it was not considered that the root may not be reachable, for example when `currentNode` is in a detached node tree where `root` is the document.

WebKit, Gecko and Blink do not loop forever in this case and simply return `null`.

This was seen in <https://archive.org/includes/build/js/ia-topnav.min.js?v=a0bdccf6>, which does something more like:
```js
var walker = document.createTreeWalker(document, NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT, null, false);

var el = document.createElement("template");
el.innerHTML = "text<div></div>text";

walker.currentNode = el.content;

var node;

for (; (node = walker.nextNode()) !== null;) {
    // Do something with node.
}
```

The call to nextNode in question is located at line 2, column 24518 of the given JavaScript.

annevk:
Tests: https://github.com/web-platform-tests/wpt/pull/60053.

Fixes #1102.

@Lubrsi @awesomekling @smaug---- care to review? (Specification is based on Gecko's implementation, but clearly missed the mark here.)

***
<a href="https://whatpr.org/dom/1433.html" title="Last updated on May 21, 2026, 1:42 PM UTC (ea93637)">Preview</a> | <a href="https://whatpr.org/dom/1433/2a4363c...ea93637.html" title="Last updated on May 21, 2026, 1:42 PM UTC (ea93637)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -10484,11 +10484,14 @@ object <var>walker</var> and "<code>next</code>" or "<code>previous</code>" <var
 
      <li><p>Set <var>sibling</var> to <var>temporary</var>'s <a for=tree>next sibling</a>.
 
-     <li><p>If <var>sibling</var> is non-null, then set <var>node</var> to <var>sibling</var> and
-     <a for=iteration>break</a>.
+     <li><p>If <var>sibling</var> is non-null, then <a for=iteration>break</a>.
 
      <li><p>Set <var>temporary</var> to <var>temporary</var>'s <a for=tree>parent</a>.
     </ol>
+
+   <li><p>If <var>sibling</var> is null, then <a for=iteration>break</a>.
+
+   <li><p>Set <var>node</var> to <var>sibling</var>.
 
    <li><p>Set <var>result</var> to the result of <a for=/>filtering</a> <var>node</var> within
    <a>this</a>.
@@ -10496,6 +10499,8 @@ object <var>walker</var> and "<code>next</code>" or "<code>previous</code>" <var
    <li><p>If <var>result</var> is {{NodeFilter/FILTER_ACCEPT}}, then set <a>this</a>'s
    <a for=TreeWalker>current</a> to <var>node</var> and return <var>node</var>.
   </ol>
+
+ <li><p>Return null.
 </ol>
 </div>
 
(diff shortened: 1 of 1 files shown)


abrhimaiiye-dev:
aliyeebrhim6@gmail.com

abrhimaiiye-dev (commented):
aliyeebrhim6@gmail.com

smaug---- (approved):

AtkinsSJ:
Sorry for the wait! I've just implemented this instead of the hack that Ladybird had previously, and I'm not seeing any regressions so I think this is correct. :+1:

Issue #1429, opened 2025-11-27
Adopting children of an intentional null shadow root goes wrong


annevk:
@annevk @sorvell 
One thing I was thinking today about the adoption steps:
From the above example, based on the new spec, it's clear that shadowroot should be getting effective global registry or null depending on "keep custom element registry null" and the original factors (null or is using another global registry) during adoption.
But what about `x-foo`? I think the intended behavior is that `x-foo`' s registry will remain to be null. However, based on the new spec, the element has null registry originally, so during adoption in above case, it will gain the effective global registry of document, which does not align with the expected behavior.
When we set registry for null registry element during insertion in previous spec, we assign the registry based off the inclusiveDescendant's parent's registry. Do we need to do something similar when we're setting registry for null registry element during adoption? Perhaps something like "set inclusiveDescendant's parent's (or inclusiveDescendant's tree scope's) registry if it is an effective global registry in the new document, otherwise set null"

_Originally posted by @ja-y-son in https://github.com/whatwg/dom/pull/1423#discussion_r2563096391_

annevk:
It's tricky as what you adopt could have a mix of null and scoped registry nodes. If you look at the root (tree scope is an implementation-specific term) it doesn't work for `<a-b customelementregistry>`. And if you look at the parent, children of scoped registry nodes will remain null. The latter seems acceptable though as that's a novel case you have to go somewhat out of your way to construct.

So that would mean the logic we want is this:

- If _inclusiveDescendant_'s parent is null or _inclusiveDescendant_'s parent is a DocumentFragent but not a ShadowRoot node, then do the current logic.
- Otherwise, if _inclusiveDescendant_ is an element, _inclusiveDescendant_'s custom element registry is null or _inclusiveDescendant_'s custom element registry's is scoped is false, and _inclusiveDescendant_'s parent is non-null, then set _inclusiveDescendant_'s custom element registry to _inclusiveDescendant_'s parent's custom element registry's effective global custom element registry.

(This would need some further changes as "effective global custom element registry" is only defined for documents right now and "custom element registry" is not generic across elements and shadow roots, but hopefully the idea is clear.)

cc @rniwa

ja-y-son:
That makes sense to me. Just to be clear, in the first point you mentioned above we're also assuming the *inclusiveDescendant* being an element right? So that would take us to the following logic:

- If *inclusvieDescendant* is a shadow root, do the current logic
- if *inclusiveDescenant* is an element,
[...]

PR #1437, opened 2025-12-02
Update element adoption logic for scoped registry

ja-y-son:
Fixing #1429 to make sure element is getting the correct scoped custom element registry upon adoption.

   * Webkit
   * Chromium
   * https://github.com/web-platform-tests/wpt/pull/56423
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=302970

***
<a href="https://whatpr.org/dom/1437.html" title="Last updated on Jul 3, 2026, 9:07 AM UTC (e3c3983)">Preview</a> | <a href="https://whatpr.org/dom/1437/995b4c3...e3c3983.html" title="Last updated on Jul 3, 2026, 9:07 AM UTC (e3c3983)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4944,7 +4944,7 @@ and an optional <a for=/>document</a> <dfn export for="clone a node"><var>docume
 
    <li><p>If <var>shadowRootRegistry</var> <a>is a global custom element registry</a>, then set
    <var>shadowRootRegistry</var> to <var>document</var>'s
-   <a>effective global custom element registry</a>.
+   <a for=Document>custom element registry</a>'s <a>effective global custom element registry</a>.
 
    <li><p><a>Attach a shadow root</a> with <var>copy</var>, <var>node</var>'s
    <a for=Element>shadow root</a>'s <a for=ShadowRoot>mode</a>, true, <var>node</var>'s
@@ -4993,7 +4993,8 @@ and an optional <a for=/>document</a> <dfn export for="clone a node"><var>docume
    <var>fallbackRegistry</var>.
 
    <li><p>If <var>registry</var> <a>is a global custom element registry</a>, then set
-   <var>registry</var> to <var>document</var>'s <a>effective global custom element registry</a>.
+   <var>registry</var> to <var>document</var>'s <a for=Document>custom element registry</a>'s
+   <a>effective global custom element registry</a>.
 
    <li><p>Set <var>copy</var> to the result of <a>creating an element</a>, given
    <var>document</var>, <var>node</var>'s <a for=Element>local name</a>, <var>node</var>'s
@@ -6119,7 +6120,8 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
     </ul>
 
     <p>then set <var>inclusiveDescendant</var>'s <a for=ShadowRoot>custom element registry</a> to
-    <var>document</var>'s <a>effective global custom element registry</a>.
+    <var>document</var>'s <a for=Document>custom element registry</a>'s
+    <a>effective global custom element registry</a>.
 
    <li>
     <p>Otherwise, if <var>inclusiveDescendant</var> is an <a for=/>element</a>:
@@ -6128,11 +6130,27 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
      <li><p>Set the <a for=Node>node document</a> of each <a>attribute</a> in
      <var>inclusiveDescendant</var>'s <a for=Element>attribute list</a> to <var>document</var>.
 
-     <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is null
-     or <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
-     <a for=CustomElementRegistry>is scoped</a> is false, then set
-     <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> to
-     <var>document</var>'s <a>effective global custom element registry</a>.
+     <li>
+      <p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is null or
+      <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
+      <a for=CustomElementRegistry>is scoped</a> is false:
+
+      <ol>
+       <li><p>Let <var>registry</var> be null.
+
+       <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is
+       non-null, <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is null, or
+       <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is an
+       <a>exclusive <code>DocumentFragment</code> node</a>, then set <var>registry</var> to
+       <var>document</var>'s <a for=Document>custom element registry</a>.
+
+       <li><p>Otherwise, set <var>registry</var> to the result of
+       <a for=/>looking up a custom element registry</a> given <var>inclusiveDescendant</var>'s
+       <a for=tree>parent</a>.
+
+       <li><p>Set <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> to
+       <var>registry</var>'s <a>effective global custom element registry</a>.
+      </ol>
 
      <li><p>If <var>inclusiveDescendant</var> is <a for=Element>custom</a>, then
      <a>enqueue a custom element callback reaction</a> with <var>inclusiveDescendant</var>, callback
@@ -6171,13 +6189,12 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
 <dfn>is a global custom element registry</dfn> if <var>registry</var> is non-null and
 <var>registry</var>'s <a for=CustomElementRegistry>is scoped</a> is false.
 
-<p>A <a for=/>document</a> <var>document</var>'s <dfn>effective global custom element registry</dfn>
-is:
+<p>Null or a {{CustomElementRegistry}} object <var>registry</var>'s
+<dfn>effective global custom element registry</dfn> is:
 
 <ol>
- <li><p>If <var>document</var>'s <a for=Document>custom element registry</a>
- <a>is a global custom element registry</a>, then return <var>document</var>'s
- <a for=Document>custom element registry</a>.
+ <li><p>If <var>registry</var> <a>is a global custom element registry</a>, then return
+ <var>registry</var>.
 
  <li><p>Return null.
 </ol>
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -6083,8 +6083,26 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
        <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is null
        or <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
        <a for=CustomElementRegistry>is scoped</a> is false, then set
-       <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> to
-       <var>document</var>'s <a>effective global custom element registry</a>.
+       <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> based on the
+       following conditions:
```
annevk: We can just have a ":" after "is false". No need to explain what will happen.

dom.bs
```diff
@@ -6083,8 +6083,26 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
+
+        <ol>
+         <li><p>If <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is null or
+         <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is a {{DocumentFragment}} but not
+         a {{ShadowRoot}}, then set <var>inclusiveDescendant</var>'s
+         <a for=Element>custom element registry</a> to <var>document</var>'s
+         <a>effective global custom element registry</a>.
+
+         <li><p>If <var>inclusiveDescendant</var>'s <a for=tree>parent</a>'s
+         <a for=Element>custom element registry</a> is null or is equal to <var>document</var>'s
```
annevk: This only works if parent is an element.

dom.bs
```diff
@@ -6083,8 +6083,26 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
+         <a>effective global custom element registry</a>, then set <var>inclusiveDescendant</var>'s
+         <a for=Element>custom element registry</a> to <var>inclusiveDescendant</var>'s
+         <a for=tree>parent</a>'s <a for=Element>custom element registry</a>.
```
annevk: Why is this not using the "effective global custom element registry" strategy discussed? It's not clear to me this is the same or why we need to be looking at _document_ at all.

dom.bs
ja-y-son: Updated to use "look up custom element registry" with a given node

dom.bs
ja-y-son: Updated to use "effective global custom element registry" strategy as discussed. Was trying to see if I could get the point across with another approach but it seems like "effective global custom element registry" is still the better way to go. Lmk if I'm missing anything on this idea.

dom.bs
```diff
@@ -6082,9 +6084,22 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
+         a {{ShadowRoot}}, then set <var>inclusiveDescendant</var>'s
+         <a for=Element>custom element registry</a> to <var>document</var>'s
+         <a for=Document>custom element registry</a>'s
+         <a>effective global custom element registry</a>.
+
+         <li><p>Otherwise, let <var>registry</var> be the result of
+         <a for=/>looking up a custom element registry</a> given <var>inclusiveDescendant</var>'s
+         <a for=tree>parent</a>; set <var>inclusiveDescendant</var>'s
+         <a for=Element>custom element registry</a> to <var>registry</var>'s
+         <a>effective global custom element registry</a>
```
annevk: This is an improvement and I like the usage of "looking up a custom ..." here, that's nice! (And we should probably move that definition to DOM at some point.) But instead of these two steps, I suggest we do this up as follows:

1. Let _registry_ be null.
2. If [first condition], then set _registry_ to _document_'s ...
3. Otherwise, set _registry_ to the result of looking up ...
4. Set _inclusiveDescendant_ ...

(And don't forget the trailing dot.)

dom.bs
```diff
@@ -6127,13 +6142,12 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
-is:
+<p>Null or a {{CustomElementRegistry}} object <var>registry</var>'s
+<dfn>effective global custom element registry</dfn> is:
 
 <ol>
- <li><p>If <var>document</var>'s <a for=Document>custom element registry</a>
- <a>is a global custom element registry</a>, then return <var>document</var>'s
- <a for=Document>custom element registry</a>.
+ <li><p>If <var>registry</var> <a>is a global custom element registry</a>, then return
+ <var>registry</var>
```
```suggestion
 <var>registry</var>.
```

dom.bs
ja-y-son: Updated to break the steps down following the above example.

dom.bs
```diff
@@ -6082,9 +6084,23 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
 
        <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is null
        or <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
-       <a for=CustomElementRegistry>is scoped</a> is false, then set
-       <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> to
-       <var>document</var>'s <a>effective global custom element registry</a>.
+       <a for=CustomElementRegistry>is scoped</a> is false:
```
annevk: You need a newline after the `<li>` here because the `<p>` is not the sole child.

annevk (approved):
This looks good to me. We'll need tests next.

For the tests I'd like us to exercise some of the more complicated scenarios with some nodes having been assigned registries and others not, etc.

ja-y-son:
Just added some tests but I haven't fully verified them yet. Feel free to take a look.

dom.bs
```diff
@@ -6080,11 +6082,26 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
-       or <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
-       <a for=CustomElementRegistry>is scoped</a> is false, then set
-       <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> to
-       <var>document</var>'s <a>effective global custom element registry</a>.
+       <li>
+        <p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is null
+        or <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
+        <a for=CustomElementRegistry>is scoped</a> is false:
+
+         <ol>
```
annevk: This is indented too much. It needs to be aligned with the `<p>` as it's a sibling.

ja-y-son: Updated!

ja-y-son:
@annevk Want to verify that this is the expected behavior from this change on effective gloabl registry. Consider this test case from `adoption.window.js`
```
test(t => {
  const contentDocument = document.body.appendChild(document.createElement('iframe')).contentDocument;
  t.add_cleanup(() => contentDocument.defaultView.frameElement.remove());
  const element = document.createElement('div');
  assert_equals(element.customElementRegistry, customElements);

  const scoped = new CustomElementRegistry();
  const scopedElement = contentDocument.createElement('div', { customElementRegistry: scoped });
  contentDocument.body.appendChild(scopedElement);
  assert_equals(scopedElement.customElementRegistry, scoped);
  scopedElement.appendChild(element);
  assert_equals(element.customElementRegistry, contentDocument.customElementRegistry);
}, "Adoption with global registry into a scoped registry");
```

Given that now we look at the parent's registry's effective global registry, the last assert of the test should probably be 
```
assert_equals(element.customElementRegistry, null);
```
given that the parent is using a scoped registry.
Is this expected? Or we want it to keep getting global registry for this element? It feels a bit odd that it originally has a global registry but after adoption it goes to null registry. Perhaps we should make sure that non-null registry element shouldn't get null registry after adoption (unless there are no global registry available)?

annevk:
Yeah, I suppose distinguishing those cases would be better. I doubt anyone is going to run into this in practice, but we might as well make it nice.

dom.bs
```diff
@@ -6080,11 +6082,36 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
+          <a>effective global custom element registry</a> of the result of
+          <a for=/>looking up a custom element registry</a> given <var>inclusiveDescendant</var>'s
+          <a for=tree>parent</a>.
+
+          <ol>
+           <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is
+           non-null and <var>registry</var> is null, set <var>registry</var> to
+           <var>document</var>'s <a for=Document>custom element registry</a>'s
+           <a>effective global custom element registry</a>.
+          </ol>
```
annevk: I don't understand this. If you mean for this to only apply in the Otherwise branch you need to nest the first bit of the Otherwise branch as well.

ja-y-son:
I think I made it unnecessarily complicated in the previous iteration. Given that our main goal is that global registry element shouldn't become null registry during adoption, I think we can simply include this category (global registry) to use document's registry's effective global registry, so it wouldn't be getting null from parent. Does this make sense? Or perhaps I'm missing something here.

dom.bs
```diff
@@ -6080,11 +6082,27 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
+        <a for=CustomElementRegistry>is scoped</a> is false:
+
+        <ol>
+         <li><p>Let <var>registry</var> be null.
+
+         <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>
+         is non-null or <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is null or
+         <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is a {{DocumentFragment}} but not
+         a {{ShadowRoot}}, then set <var>registry</var> to <var>document</var>'s
+         <a for=Document>custom element registry</a>.
```
```suggestion
         <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is
         non-null, <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is null, or
         <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is a {{DocumentFragment}} but not
         a {{ShadowRoot}}, then set <var>registry</var> to <var>document</var>'s
         <a for=Document>custom element registry</a>.
```

annevk (approved):
Yeah I think this works. I'll re-review on Monday to be sure. If @sorvell, @rniwa, and maybe @keithamus could have a look as well that'd be great.

(Checking _inclusiveDescendant_'s custom element registry again bothers me a little bit, but I don't see a good way out of that. I also want to double check the "DocumentFragment  but not ShadowRoot" language.)

ja-y-son:
@annevk Went through this PR again and I think it still works. Do we want to merge this PR first while wrapping up https://github.com/web-platform-tests/wpt/pull/56423

dom.bs
```diff
@@ -4936,7 +4936,8 @@ and an optional <a for=/>document</a> <dfn export for="clone a node"><var>docume
    <var>fallbackRegistry</var>.
 
    <li><p>If <var>registry</var> <a>is a global custom element registry</a>, then set
-   <var>registry</var> to <var>document</var>'s <a>effective global custom element registry</a>.
+   <var>registry</var> to <var>document</var>'s <a for=Document>custom element registry</a>
```
```suggestion
   <var>registry</var> to <var>document</var>'s <a for=Document>custom element registry</a>'s
```
(might have to rewrap)

dom.bs
```diff
@@ -6080,11 +6082,28 @@ algorithm is passed <var ignore>node</var> and <var ignore>oldDocument</var>, as
...
+        is false; or <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a>'s
+        <a for=CustomElementRegistry>is scoped</a> is false:
+
+        <ol>
+         <li><p>Let <var>registry</var> be null.
+
+         <li><p>If <var>inclusiveDescendant</var>'s <a for=Element>custom element registry</a> is
+         non-null, <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is null, or
+         <var>inclusiveDescendant</var>'s <a for=tree>parent</a> is a {{DocumentFragment}} but not
+         a {{ShadowRoot}}, then set <var>registry</var> to <var>document</var>'s
```
```suggestion
         a {{ShadowRoot}} <a for=/>node</a>, then set <var>registry</var> to <var>document</var>'s
```
I might actually want to define exclusive DocumentFragment for this. Similar to what we have for Text nodes.

dom.bs
```diff
@@ -6745,6 +6763,9 @@ developers to pass a {{ShadowRoot}} node instead of a dictionary to {{Element/at
  <dt><dfn export for=Element>custom element registry</dfn>
  <dd>Null or a {{CustomElementRegistry}} object.
 
+ <dt><dfn export id=concept-element-keep-custom-elemenet-registry-null for=Element>keep custom element registry null</dfn>
```
```suggestion
 <dt><dfn export for=Element>keep custom element registry null</dfn>
```
We don't do hardcoded IDs for new concepts. Use the generated ID.

annevk:
Also, now that "keep custom element registry null" is part of this PR, it depends on https://github.com/whatwg/html/pull/12000 so we should mention that in the commit message and get that PR finished as well.

dom.bs
ja-y-son: Could you elaborate on what you meant by define exclusive DocumentFragment?

ja-y-son:
I updated the commit message so it mentions the new concept that will be defined in https://github.com/whatwg/html/pull/12000. I'll prioritize getting this adoption logic PR landed and will for sure get the attribute PR finished next. Let me know if you prefer me to take "keep custom elemetn registry null" idea out of this PR for now; I can also land this PR without the concept and put it together with attribute PR.

annevk:
Let's keep it out of this PR then.

dom.bs
annevk: I'm happy to take care of that but if you want to take a stab at it you'd have to define something equivalent to https://dom.spec.whatwg.org/#exclusive-text-node for DocumentFragment.

ja-y-son: I added the definition for exclusive document fragment. Let me know if I captured what you thought correctly.

annevk (approved):

ja-y-son:
@annevk Thanks for the review and the nit! I don't seem to have the permission to merge the PR. Mind landing this PR for me?

annevk:
Ryosuke said he'd make time for the tests this week and I'd like to wait for that as we land tests and specification changes around the same time.

Issue #1007, opened 2021-08-12
Is is supposed to say 'SHADOW ROOT' and not DOCUMENT FRAGMENT?


DougRoyer:
https://dom.spec.whatwg.org/commit-snapshots/f2a2dedd8831cbae9afeab87b915e1234c49cba1/#concept-node-adopt

It says: 
(3) If node is a DocumentFragment node whose host is non-null, then return. 

Problem:
A DocumentFragment node does not have a 'host', a SHADOW_ROOT_NODE has one.

rniwa:
No, that host refers to the [template element](https://html.spec.whatwg.org/multipage/scripting.html#template-contents).

PR #1447, opened 2026-01-22
Add Marker interface

foolip:
This node is intended for the `<!start>`, `<!end>` and `<!marker>`
syntax, to be created by the HTML parser.

   * …
   * …
   * … 
   * Chromium: …
   * Gecko: …
   * WebKit: …

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5045,6 +5045,10 @@ and an optional <a for=/>document</a> <dfn export for="clone a node"><var>docume
    <dd><p>Set <var>copy</var>'s <a for=ProcessingInstruction>target</a> and
    <a for=CharacterData>data</a> to those of <var>node</var>.
 
+   <dt>{{Marker}}
+   <dd><p>Set <var>copy</var>'s <a for=Marker>type</a> and <a for=Marker>name</a> to those of
+   <var>node</var>.
+
    <dt>Otherwise
    <dd><p>Do nothing.
   </dl>


dom.bs
```diff
@@ -4421,6 +4421,7 @@ interface Node : EventTarget {
   const unsigned short DOCUMENT_TYPE_NODE = 10;
   const unsigned short DOCUMENT_FRAGMENT_NODE = 11;
   const unsigned short NOTATION_NODE = 12; // legacy
+  const unsigned short MARKER_NODE = 13;
```
WebReflection, line 4425: love it, happy to provide a polyfill around this too, in case nobody else is already working on it

foolip:
@annevk this isn't complete, but I wanted to send it out because I'm not totally sure about a few choices.

- I think a single interface with a read-only type (marker/end/start) makes sense, but unlike elements you can't create a node with any string. I don't think we can allow `new Marker("doctype", "html")`, but am not sure if DOM should allow `new Marker("whatever")` even where HTML wouldn't parse `<!whatever>` into such a node.
- Should we have `document.createMarker()` in addition to the constructor?

annevk:
We probably need to investigate why `ShadowRoot` didn't get a new node type. I think a constructor probably suffices, but we should restrict it to match what HTML can do.

dom.bs
WebReflection: example:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Document</title>
  <script type="module">
    class Marker extends CharacterData {
      get nodeType() {
        return 13;
      }
      hasAttribute(name) {
        return new RegExp(` ${name}(=(['"])?(.*)?\\2)?`).test(this.data);
      }
      getAttribute(name) {
        return this.hasAttribute(name) ? RegExp.$3 : null;
      }
    }

    const promoted = new WeakSet;
    const tw = document.createTreeWalker(document.body, NodeFilter.SHOW_COMMENT);
    let node;
    while (node = tw.nextNode()) {
      if (!promoted.has(node) && /^(start|end|marker)(?:$|\s+)/.test(node.data)) {
        promoted.add(node);
        const kind = RegExp.$1;
        Object.setPrototypeOf(node, Marker.prototype);
        console.log({ name: node.getAttribute('name') });
      }
    }
  </script>
</head>
<body>
  <!start>
  O
  <!marker name=unquoted>
  <!marker name="double quoted">
  <!marker name='single quoted'>
  <!marker-nope name="value">
  <!nope name="value">
  K
  <!end>
</body>
</html>
```

now, I've no idea how people out there are supposed to distinguish `<!marker>` from `<!--marker-->`, as example, but the polyfill would take those starting content for granted and produce a *Marker* node already.

WebReflection: forgot the devtools result:

image

WebReflection:
`document.createMarker(kind = 'marker', attributes = {})` would be a great API to me, it creates a `nodeType` `13` and it will throw if `kind` is not `marker`, `start` or `end` + it will add "*escaped marker attributes*" out of the box, with quoting and everything.

foolip:
> We probably need to investigate why `ShadowRoot` didn't get a new node type.

I wondered about this too, and guess it's because it inherits from `DocumentFragment` which made it possible to use `DOCUMENT_FRAGMENT_NODE`.

`ShadowRoot` was added in b2ee400a6542dc76d41bed5d2edb8c85ba241bfc, but I don't see any clues from that.

Is there some place you suspect the discussion might have happened that I can go investigate?

annevk:
Maybe in WebApps archives? I think there was also some aversion to introducing a new node type and having to revisit existing algorithms. But I'm not sure that ended up being good thing as we still have some technical debt to resolve from that.

foolip:
I found https://github.com/whatwg/dom/issues/1007 via [list archives](https://www.w3.org/Search/Mail/Public/search?keywords=SHADOW_ROOT_NODE&indexes=Public), but it's not a discussion on whether constant should be added or not.

From a search of Chromium commit messages and WebKit bugs, I found some decisions:

- In https://bugs.webkit.org/show_bug.cgi?id=58704, `SHADOW_ROOT_NODE` (14) was added to Node.h but seemingly not as a web-exposed constant on `Node`. (At the time there was also a `XPATH_NAMESPACE_NODE` as number 13.)
- In https://bugs.webkit.org/show_bug.cgi?id=76693, `ShadowRoot` was made to inherit from `DocumentFragment` instead of `Node`
- In https://bugs.webkit.org/show_bug.cgi?id=77514, `SHADOW_ROOT_NODE` was removed again, referring to what the spec said.

I haven't looked at the spec history, but from this I gather the reasoning was that the new constant should only be added if the new interface inherits directly from `Node`, as it cannot be avoided in that case. Since we're suggesting that `Marker` inherit from `Node`, it will need a new type and we'll need to consider it in places like `NodeIterator` and `TreeWalker` as well.

We will also need a new "#marker" return value for `node.nodeName`, and probably a bunch of other little changes.

foolip:
The original spec edit to use `DOCUMENT_FRAGMENT_NODE` for `ShadowRoot` seems to be https://github.com/WICG/webcomponents/commit/0700c45cd804336f1f498454db040e1325e7ce22, and I also found it discussed in https://github.com/WICG/webcomponents/issues/324, with nobody arguing for a different type than document fragments, so it was left that way.

WebReflection:
to whom it might concern, I've read the proposed specs and created a polyfill:
https://gist.github.com/WebReflection/291357aa6bbbd97d54a971ce5f5aae4e

I took a chance to add attributes but it's unclear in this document how those would be handled.

WebReflection:
> We will also need a new "#marker" return value for node.nodeName, and probably a bunch of other little changes.

I missed that, updated [the gist](https://gist.github.com/WebReflection/291357aa6bbbd97d54a971ce5f5aae4e?permalink_comment_id=5952616#gistcomment-5952616) to reflect that too ... it's looking good to me as polyfill:

  * the class extends `Node`, not *comment*
  * both `type` and `name` are reflected
  * all nodes are updated in the living DOM and in each ShadowRoot
  * literally nothing else happens

You can `new Marker(type, name = '')` out of the box but comments found live or appended will be automatically promoted, no "*node substitution*" needed, just like custom elements would.

I hope this will help shaping the specs.

**edit** P.S. as every polyfill in this world kinda requires, that module should be the very first one imported on top of everything else, yet it will play nicely with other polyfills, as it shouldn't bother any other related logic.

dom.bs
```diff
@@ -8550,6 +8557,42 @@ constructor steps are to set <a>this</a>'s <a for=CharacterData>data</a> to <var
...
+enum MarkerType { "marker", "start", "end" };
+
+[Exposed=Window]
+interface Marker : Node {
+  constructor(MarkerType type, optional DOMString name = "");
+
+  readonly attribute MarkerType type;
+  readonly attribute DOMString name;
+};
+</pre>
```
noamr, line 8580: Should also include mixin [`NonDocumentTypeChildNode`](https://dom.spec.whatwg.org/#interface-nondocumenttypechildnode) I think.

WebReflection: [polyfill updated](https://gist.github.com/WebReflection/291357aa6bbbd97d54a971ce5f5aae4e?permalink_comment_id=5952616#gistcomment-5952616)

foolip:
Closing in favor of https://github.com/whatwg/dom/pull/1454, we're not adding a new `Marker` interface any longer but instead using the existing `ProcessingInstruction`.

PR #1450, opened 2026-02-09
Pass better parent/root information to (re)moving steps.

dbaron:
For removing steps and moving steps, instead of passing only oldParent (which is null when the removed node is not the root of the removed subtree), pass both oldAncestor (never null) and isSubtreeRoot.

This is the DOM spec part of the fix for https://github.com/whatwg/html/issues/12098
For the HTML spec change see https://github.com/whatwg/html/pull/12144

   * Chromium
   * Gecko (assuming https://github.com/whatwg/html/issues/12098#issuecomment-3872487261 is sufficient)
   * I think this *probably* doesn't need tests since previous implementors and test authors assumed that the spec said what this pair of PRs is updating it to actually say.
   * N/A
- [X] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: Not needed.

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -3273,9 +3273,9 @@ from a <a for=/>node</a> <var>parent</var>:
 <div algorithm>
 <p><a lt="Other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-remove-ext>removing steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed a <a for=/>node</a> <var ignore>removedNode</var>, a <a for=/>node</a>
-<var ignore>oldAncestor</var>, and a boolean <var ignore>isSubtreeRoot</var> as indicated in the
-<a for=/>remove</a> algorithm below.
+algorithm is passed a <a for=/>node</a> <var ignore>removedNode</var>, a boolean
+<var ignore>isSubtreeRoot</var>, and a <a for=/>node</a> <var ignore>oldAncestor</var> as indicated
+in the <a for=/>remove</a> algorithm below.
 </div>
 
 <div algorithm=remove>
@@ -3316,7 +3316,7 @@ optional boolean <dfn for=remove><var>suppressObservers</var></dfn> (default fal
    <li><p>Run <a>assign slottables for a tree</a> with <var>node</var>.
   </ol>
 
- <li><p>Run the <a>removing steps</a> with <var>node</var>, <var>parent</var>, and true.
+ <li><p>Run the <a>removing steps</a> with <var>node</var>, true, and <var>parent</var>.
 
  <li><p>Let <var>isParentConnected</var> be <var>parent</var>'s <a>connected</a>.
 
@@ -3333,7 +3333,7 @@ optional boolean <dfn for=remove><var>suppressObservers</var></dfn> (default fal
   <a>shadow-including tree order</a>:
 
   <ol>
-   <li><p>Run the <a>removing steps</a> with <var>descendant</var>, <var>parent</var>, and false.
+   <li><p>Run the <a>removing steps</a> with <var>descendant</var>, false, and <var>parent</var>.
 
    <li><p>If <var>descendant</var> is <a for=Element>custom</a> and <var>isParentConnected</var>
    is true, then <a>enqueue a custom element callback reaction</a> with <var>descendant</var>,


dom.bs
```diff
@@ -3108,9 +3108,10 @@ to do these things asynchronously, however.
 
   <ol>
    <li>
-    <p>If <var>inclusiveDescendant</var> is <var>node</var>, then run the <a>moving steps</a> with
-    <var>inclusiveDescendant</var> and <var>oldParent</var>. Otherwise, run the <a>moving steps</a>
-    with <var>inclusiveDescendant</var> and null.
+    <p>Let <var>isSubtreeRoot</var> be whether <var>inclusiveDescendant</var> is <var>node</var>.
```
annevk: Do we need this in standards? It seems you can determine this by checking this's parent as well?

dbaron:
> Thank you for fixing this. I think we should rename _oldParent_ then to something that reflects it can be an ancestor as well. _oldAncestor_ might be good enough? Or maybe _oldSubtreeRoot_ if we wanted a little more precision. WDYT?

I think either _oldAncestor_ or _oldSubtreeParent_ would be fine, but not _oldSubtreeRoot_.

dom.bs
dbaron: I think we do, at least for the `source` and `img` elements' *moving* steps (since we can't check the parent for the moving steps).

annevk:
Hmm fair, let's go with _oldAncestor_ then.

dom.bs
annevk: I think we should only add it to the moving steps for now then? Not entirely sure though, perhaps it's nicer if the relevant moving and removing steps can be similar. Bit unfortunate we already need that much context, but so be it.

dbaron: I think it's better to keep the parallel structure, particularly given the trouble spec editors have already had with getting these steps correct.

dbaron:
I've updated both PRs to use `oldAncestor`.  As noted above I think I prefer keeping `isSubtreeRoot` on both algorithms to keep the structure parallel and to make things clearer to people adding new (re)moval steps in the future.

dom.bs
```diff
@@ -2994,8 +2994,9 @@ null.
 <div algorithm>
 <p><a lt="Other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-move-ext>moving steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed a <a for=/>node</a> <var>movedNode</var>, and a <a for=/>node</a>-or-null
-<var ignore>oldParent</var> as indicated in the <a for=/>move</a> algorithm below. Like the
+algorithm is passed a <a for=/>node</a> <var>movedNode</var>, a <a for=/>node</a>
+<var ignore>oldAncestor</var>, and a boolean <var ignore>isSubtreeRoot</var>
+as indicated in the <a for=/>move</a> algorithm below. Like the
```
annevk, line 2999: Shall we make _isSubtreeRoot_ the second argument for clarity?

dom.bs
```diff
@@ -3108,9 +3109,10 @@ to do these things asynchronously, however.
 
   <ol>
    <li>
-    <p>If <var>inclusiveDescendant</var> is <var>node</var>, then run the <a>moving steps</a> with
-    <var>inclusiveDescendant</var> and <var>oldParent</var>. Otherwise, run the <a>moving steps</a>
-    with <var>inclusiveDescendant</var> and null.
+    <p>Let <var>isSubtreeRoot</var> be whether <var>inclusiveDescendant</var> is <var>node</var>.
```
annevk: We are always very explicit with booleans in standards. So I'd prefer

> Let _isSubtreeRoot_ be true if _inclusiveDescendant_ is _node_; otherwise false.

annevk: It should also be its own step, not just its own paragraph.

dom.bs
```diff
@@ -3108,9 +3109,10 @@ to do these things asynchronously, however.
...
+
+    <p>Run the <a>moving steps</a> with <var>inclusiveDescendant</var>, <var>oldParent</var>, and
+    <var>isSubtreeRoot</var>.
 
     <p class="note">Because the <a>move</a> algorithm is a separate primitive from
     <a for=/>insert</a> and <a for=/>remove</a>, it does not invoke the traditional
```
annevk: Aside: I think we should remove "traditional" here.

dom.bs
dbaron: Done

dom.bs
dbaron: Done.

dom.bs
dbaron: Done.

dom.bs
```diff
@@ -3270,9 +3273,9 @@ from a <a for=/>node</a> <var>parent</var>:
 <div algorithm>
 <p><a lt="Other applicable specifications">Specifications</a> may define
 <dfn export id=concept-node-remove-ext>removing steps</dfn> for all or some <a for=/>nodes</a>. The
-algorithm is passed a <a for=/>node</a> <var ignore>removedNode</var> and a
-<a for=/>node</a>-or-null <var ignore>oldParent</var>, as indicated in the <a for=/>remove</a>
-algorithm below.
+algorithm is passed a <a for=/>node</a> <var ignore>removedNode</var>, a <a for=/>node</a>
+<var ignore>oldAncestor</var>, and a boolean <var ignore>isSubtreeRoot</var> as indicated in the
+<a for=/>remove</a> algorithm below.
```
annevk: Let's change the order here as well for consistency. We'll have to audit & update callers anyway.

dbaron: Oops, I meant to do that but somehow I missed it here and only did it in the HTML part!  Done now.

PR #1452, opened 2026-02-24
Collapse range before mutations in deleteContents and extract

### LINKED ISSUE #1446: https://dom.spec.whatwg.org/#dom-range-deletecontents may collapse the Range to an invalid position

Step 9: _For each node of nodesToRemove, in [tree order](https://dom.spec.whatwg.org/#concept-tree-order): [remove](https://dom.spec.whatwg.org/#concept-node-remove) node._
may modify DOM
and then step 11 does just _Set [start](https://dom.spec.whatwg.org/#concept-range-start) and [end](https://dom.spec.whatwg.org/#concept-range-end) to (newNode, newOffset)._
and those newNode and newOffset were calculated before removal.

stevenobiajulu:
Collapse range before mutations in deleteContents and extract

Move the step that sets the range's start and end to (newNode, newOffset) from after DOM mutations to before them, in both `deleteContents()` and the `extract` algorithm. The DOM's built-in live range maintenance mechanisms (`live range pre-remove steps` and `replace data` range adjustments) keep the range valid through subsequent operations. Collapsing after mutations is unreliable because script can run during removal (via removing steps) and modify the DOM, invalidating the pre-computed values.

Fixes #1446.

---

   * Mozilla (smaug----, Mozilla-affiliated GitHub account, opened #1446)
   * Chromium (shipped fix in [CL 7615530](https://chromium-review.googlesource.com/c/chromium/src/+/7615530), WPT tests merged in web-platform-tests/wpt#58132)
   * https://github.com/web-platform-tests/wpt/pull/58008
   * Chromium: https://issues.chromium.org/issues/486922855
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2018839
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=308549
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/43230

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -9283,9 +9283,8 @@ method steps are:
     point.
   </ol>
 
- <li>
-  <p>Set <a>this</a>'s <a for=range>start</a> and <a for=range>end</a> to
-  (<var>newNode</var>, <var>newOffset</var>).
+ <li><p>Set <a>this</a>'s <a for=range>start</a> and <a for=range>end</a> to
+ (<var>newNode</var>, <var>newOffset</var>).
 
  <li><p>If <var>originalStartNode</var> is a {{CharacterData}} <a for=/>node</a>, then
  <a>replace data</a> of <var>originalStartNode</var> with <var>originalStartOffset</var>,
@@ -9415,9 +9414,8 @@ method steps are:
     point.
   </ol>
 
- <li>
-  <p>Set <var>range</var>'s <a for=range>start</a> and <a for=range>end</a> to
-  (<var>newNode</var>, <var>newOffset</var>).
+ <li><p>Set <var>range</var>'s <a for=range>start</a> and <a for=range>end</a> to
+ (<var>newNode</var>, <var>newOffset</var>).
 
  <li>
   <p>If <var>firstPartiallyContainedChild</var> is a {{CharacterData}} <a for=/>node</a>:


dom.bs
```diff
@@ -9283,6 +9283,17 @@ method steps are:
...
+ <li>
+  <p>Set <a>this</a>'s <a for=range>start</a> and <a for=range>end</a> to
+  (<var>newNode</var>, <var>newOffset</var>).
+
+  <p class=note>This is done before performing mutations, because the DOM's built-in
+  <a>live range</a> maintenance mechanisms
+  (<a>live range pre-remove steps</a> and <a>replace data</a> range adjustments)
+  will automatically keep the range valid through the subsequent operations. Collapsing the
+  range after mutations would be unreliable, since script can run during
+  <a for=/>removal</a> and modify the DOM.
```
zcorpan: "removal" is not a defined term. However I'm not sure this note is needed. @annevk WDYT?

```suggestion
  <a for=/ lt=remove>removal</a> and modify the DOM.
```

annevk: I don't think we need the notes either.

stevenobiajulu:
@annevk Updated per your review — removed the explanatory notes from both algorithms. The rationale is in the commit message. Also updated the implementer interest checklist.

stevenobiajulu:
@annevk Thanks for the review\! It looks like the Participation check is pending because our organization (UseJunior) hasn't been verified yet. Happy to provide anything that's needed on our end to move that along.

annevk:
I would like @smaug---- to double check as well since he filed the original issue.

I saw that you filed browser bugs but I think browsers already handle this correctly. You can execute script during removal by using a `pagehide` listener in an `<iframe>` element's child document and making sure the `<iframe>` element gets removed if you want to double check. Only if that demonstrates something I think new tests would be warranted.

stevenobiajulu:
I ran the empirical test. @annevk you're right — all three browsers already implement the corrected behavior.

**Test structure**

```
container > [p#before(0), trigger(1), p#inside(2), p#sentinel(3)]
Range: (container, 1) → (container, 3)
```

The mutation callback removes `p#before` (at index 0, before the range's start boundary) while the operation is executing. Under the old spec, `collapse(container, 1)` would be applied after this removal, overwriting the live-range adjustment with a stale offset. Under the new spec the range collapses first, so the live-range machinery tracks the removal and produces `startOffset: 0`. The observable difference is `startOffset: 0` (new spec) vs `startOffset: 1` (old spec).

**Results — Chrome 133, Firefox 135, Safari 18**

| Track | Trigger mechanism | Operation | `startOffset` | Verdict |
|---|---|---|---|---|
| A | `disconnectedCallback` (custom element) | `deleteContents` | 0 (all three) | ✅ new-spec behavior |
| A | `disconnectedCallback` (custom element) | `extractContents` | 0 (all three) | ✅ new-spec behavior |
| B | `pagehide`/`unload` in iframe `srcdoc` | `deleteContents` | 0 (all three) | ✅ new-spec behavior |
| B | `pagehide`/`unload` in iframe `srcdoc` | `extractContents` | 0 (all three) | ✅ new-spec behavior |

Track A: `disconnectedCallback` fired synchronously mid-operation in all three browsers (not deferred past the `[CEReactions]` boundary on `deleteContents()` as I'd initially worried — reactions flush when each internal `remove()` completes).

Track B: `pagehide` and `unload` both fired synchronously in all three browsers including Firefox, which surprised me given the HTML spec's "without any unload events firing" language and Bugzilla #1850228. That bug may describe a different scenario.

Screenshots attached. 
**Chrome**:
Screenshot 2026-03-06 at 11 25 28 AM

**Safari**:
Screenshot 2026-03-06 at 11 27 11 AM

**Firefox**:
Screenshot 2026-03-06 at 12 00 38 PM

**What this means for the open items**

- The spec change is a correctness clarification, not a fix for a bug in shipping browsers. I'll update the filed browser bugs to note that.
- On the WPT tests in #58008: given that there's nothing for browsers to fix, I'm happy to close that PR. Alternatively the static collapse-position tests could stay as regression guards for the spec ordering itself — up to you.

annevk:
Thanks again @stevenobiajulu!

PR #1454, opened 2026-03-04
Define processing instruction attributes

foolip:
Processing instruction attributes are parsed from the underlying `data`. When they are mutated, the underlying `data` reflects the change as well. The parsing is XML-like, and case sensitive.

   * Chromium (@foolip + @noamr)
   * WebKit (See https://github.com/whatwg/dom/pull/1454#pullrequestreview-4546510558)
   * https://github.com/web-platform-tests/wpt/pull/60815
   * Chromium: https://issues.chromium.org/issues/481087638
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2049621
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=317640
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: tracked via chromium release process.

***
<a href="https://whatpr.org/dom/1454.html" title="Last updated on Jun 25, 2026, 7:19 AM UTC (4df4a48)">Preview</a> | <a href="https://whatpr.org/dom/1454/cd961d3...4df4a48.html" title="Last updated on Jun 25, 2026, 7:19 AM UTC (4df4a48)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8553,6 +8553,36 @@ interface ProcessingInstruction : CharacterData {
  <dd>Returns a new {{ProcessingInstruction}} <a for=/>node</a> whose
  <a for=ProcessingInstruction>target</a> is <var>target</var> and
  <a for=CharacterData>data</a> is <var>data</var>.
+
+ <dt><code><var>pi</var> . {{ProcessingInstruction/target}}</code>
+ <dd>Returns the <a for=ProcessingInstruction>target</a>.
+
+ <dt><code><var>pi</var> . <a method for=ProcessingInstruction lt=hasAttributes()>hasAttributes</a>()</code>
+ <dd><p>Returns true if <var>pi</var> has attributes; otherwise false.
+
+ <dt><code><var>pi</var> . <a method for=ProcessingInstruction lt=getAttributeNames()>getAttributeNames</a>()</code>
+ <dd><p>Returns the names of all of <var>pi</var>'s attributes. Cannot contain duplicates.
+
+ <dt><code><var>pi</var> . <a method for=ProcessingInstruction lt=getAttribute()>getAttribute</a>(<var>name</var>)</code>
+ <dd><p>Returns the value of <var>pi</var>'s attribute named <var>name</var>, and null if there is
+ no such attribute.
+
+ <dt><code><var>pi</var> . <a method for=ProcessingInstruction lt=setAttribute()>setAttribute</a>(<var>name</var>, <var>value</var>)</code>
+ <dd><p>Sets <var>pi</var>'s attribute named <var>name</var> to <var>value</var>.
+
+ <dt><code><var>pi</var> . <a method for=ProcessingInstruction lt=removeAttribute()>removeAttribute</a>(<var>name</var>)</code>
+ <dd><p>Removes <var>pi</var>'s attribute named <var>name</var>.
+
+ <dt><code><var>pi</var> . <a method for=ProcessingInstruction lt=toggleAttribute()>toggleAttribute</a>(<var>name</var> [, <var>force</var>])</code>
+ <dd>
+  <p>If <var>force</var> is not given, "toggles" <var>name</var>, removing it if it is
+  present and adding it if it is not present. If <var>force</var> is true, adds
+  <var>name</var>. If <var>force</var> is false, removes <var>name</var>.
+
+  <p>Returns true if <var>name</var> is now present; otherwise false.
+
+ <dt><code><var>pi</var> . <a method for=ProcessingInstruction lt=hasAttribute()>hasAttribute</a>(<var>name</var>)</code>
+ <dd><p>Returns true if <var>pi</var> has an attribute named <var>name</var>; otherwise false.
 </dl>
 
 <div algorithm>
--- a/dom.bs
+++ b/dom.bs
@@ -8721,7 +8721,8 @@ false.
  "<code>&gt;&lt;/attrs&gt;</code>".
 
  <li><p>Let <var>fragment</var> be the result of invoking the
- <a>fragment parsing algorithm steps</a> with <var>context</var> and <var>markup</var>.
+ <a>fragment parsing algorithm steps</a> with <var>context</var> and <var>markup</var>. If this
+ throws an exception, then return.
 
  <li>
   <p>If <var>fragment</var>'s <a>first child</a> is an <a for=/>element</a>:
--- a/dom.bs
+++ b/dom.bs
@@ -8601,8 +8601,6 @@ constructor steps are:
 
  <li><p><a for=ProcessingInstruction>Initialize</a> <a>this</a> with <var>target</var> and
  <var>data</var>.
-
- <li><p><a>Update attributes from data</a> given <a>this</a>.
 </ol>
 </div>
 
--- a/dom.bs
+++ b/dom.bs
@@ -16,6 +16,13 @@ Indent: 1
 urlPrefix: https://www.w3.org/TR/xml/#NT-; spec: XML
  type: dfn
   text: Name; url: Name; for: XML
+urlPrefix: https://www.w3.org/TR/xml-stylesheet/#; spec: XML-STYLESHEET
+ type: dfn
+  text: rules for parsing pseudo-attributes from a string; url: dt-parsing
+  text: parsing result; url: dt-parsing-result
+  text: pseudo-attribute; url: dt-pseudo-attribute
+  text: name; url: dt-name; for: pseudo-attribute
+  text: value; url: dt-value; for: pseudo-attribute
 url: https://w3c.github.io/DOM-Parsing/#dom-range-createcontextualfragment
  type: method; text: createContextualFragment(); for: Range
 type: interface
@@ -50,14 +57,15 @@ spec:html; type:element
 <p>This specification depends on the Infra Standard. [[!INFRA]]
 
 <p>Some of the terms used in this specification are defined in <cite>Encoding</cite>,
-<cite>Selectors</cite>, <cite>Trusted Types</cite>, <cite>Web IDL</cite>, <cite>XML</cite>, and
-<cite>Namespaces in XML</cite>.
+<cite>Selectors</cite>, <cite>Trusted Types</cite>, <cite>Web IDL</cite>, <cite>XML</cite>,
+<cite>Namespaces in XML</cite>, and <cite>Associating Style Sheets with XML documents</cite>.
 [[!ENCODING]]
 [[!SELECTORS4]]
 [[!TRUSTED-TYPES]]
 [[!WEBIDL]]
 [[!XML]]
 [[!XML-NAMES]]
+[[!XML-STYLESHEET]]
 
 <p>When extensions are needed, the DOM Standard can be updated accordingly, or a new standard
 can be written that hooks into the provided extensibility hooks for
@@ -8729,19 +8737,16 @@ method steps are:
 <ol>
  <li><p><a for=map>Clear</a> <var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>.
 
- <li><p>Let <var>document</var> be a new {{Document}} node whose <a for=Document>type</a> is
- "<code>html</code>".
+ <li><p>Let <var>result</var> be the <a>parsing result</a> of invoking the
+ <a>rules for parsing pseudo-attributes from a string</a> given <var>pi</var>'s
+ <a for=CharacterData>data</a>.
 
- <li><p>Let <var>html</var> be the concatenation of "<code>&lt;html </code>",
- <var>pi</var>'s <a for=CharacterData>data</a>, and "<code>&gt;</code>".
+ <li><p>If <var>result</var> is an error, then return.
 
- <li><p><a href="https://html.spec.whatwg.org/#parse-html-from-a-string">Parse HTML from a string</a>
- given <var>document</var> and <var>html</var>.
-
- <li><p><a for=list>For each</a> <var>attribute</var> of <var>document</var>'s
- <a>document element</a>'s <a for=Element>attribute list</a>, set <var>pi</var>'s
- <a for=ProcessingInstruction>attribute map</a>[<var>attribute</var>'s <a for=Attr>local name</a>]
- to <var>attribute</var>'s <a for=Attr>value</a>.
+ <li><p>For each <a>pseudo-attribute</a> <var>pseudoAttr</var> of <var>result</var>, set
+ <var>pi</var>'s
+ <a for=ProcessingInstruction>attribute map</a>[<var>pseudoAttr</var>'s [=pseudo-attribute/name=]]
+ to <var>pseudoAttr</var>'s [=pseudo-attribute/value=].
 </ol>
 </div>
 
--- a/dom.bs
+++ b/dom.bs
@@ -8617,8 +8617,14 @@ are to return the result of <a for=map>getting the keys</a> of <a>this</a>'s
 
 <div algorithm>
 <p>The <dfn method for=ProcessingInstruction><code>getAttribute(<var>name</var>)</code></dfn> method
-steps are to return <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
-<a for=map>with default</a> null.
+steps are:
+
+<ol>
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
+
+ <li><p>Return <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
+ <a for=map>with default</a> null.
+</ol>
 </div>
 
 <div algorithm>
@@ -8629,6 +8635,8 @@ method steps are:
 <ol>
  <li><p>If <var>name</var> is not a <a>valid attribute local name</a>, then <a>throw</a> an
  "{{InvalidCharacterError!!exception}}" {{DOMException}}.
+
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
 
  <li><p><a for=map>Set</a> <a>this</a>'s
  <a for=ProcessingInstruction>attribute map</a>[<var>name</var>] to <var>value</var>.
@@ -8642,6 +8650,8 @@ method steps are:
 method steps are:
 
 <ol>
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
+
  <li><p><a for=map>Remove</a> <a>this</a>'s
  <a for=ProcessingInstruction>attribute map</a>[<var>name</var>].
 
@@ -8657,6 +8667,8 @@ method steps are:
 <ol>
  <li><p>If <var>name</var> is not a <a>valid attribute local name</a>, then <a>throw</a> an
  "{{InvalidCharacterError!!exception}}" {{DOMException}}.
+
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
 
  <li><p>Let <var>attributes</var> be <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>.
 
@@ -8680,9 +8692,14 @@ method steps are:
 
 <div algorithm>
 <p>The <dfn method for=ProcessingInstruction><code>hasAttribute(<var>name</var>)</code></dfn>
-method steps are to return true if <a>this</a>'s
-<a for=ProcessingInstruction>attribute map</a>[<var>name</var>] <a for=set>exists</a>; otherwise
-false.
+method steps are:
+
+<ol>
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
+
+ <li><p>Return true if <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
+ <a for=set>exists</a>; otherwise false.
+</ol>
 </div>
 
 <div algorithm>
--- a/dom.bs
+++ b/dom.bs
@@ -8591,6 +8591,91 @@ interface ProcessingInstruction : CharacterData {
 <a for=/>map</a>, initially empty.
 
 <div algorithm>
+<p>To <dfn export for=ProcessingInstruction>initialize</dfn> a {{ProcessingInstruction}}
+<a for=/>node</a> <var>pi</var>, with <var>target</var> and <var>data</var>:
+
+<ol>
+ <li><p>If <var>target</var> does not match the <code>[=XML/Name=]</code> production, then
+ <a>throw</a> an "{{InvalidCharacterError!!exception}}" {{DOMException}}.
+
+ <li><p>If <var>data</var> contains the string "<code>?></code>", then <a>throw</a> an
+ "{{InvalidCharacterError!!exception}}" {{DOMException}}.
+
+ <li><p>Set <var>pi</var>'s <a for=ProcessingInstruction>target</a> to <var>target</var>.
+
+ <li><p>Set <var>pi</var>'s <a for=CharacterData>data</a> to <var>data</var>.
+
+ <li><p><a>Update attributes from data</a> given <var>pi</var>.
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn>update attributes from data</dfn>, given a {{ProcessingInstruction}} <a for=/>node</a>
+<var>pi</var>:
+
+<ol>
+ <li><p><a for=map>Clear</a> <var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>.
+
+ <li><p>Let <var>result</var> be the <a>parsing result</a> of invoking the
+ <a>rules for parsing pseudo-attributes from a string</a> given <var>pi</var>'s
+ <a for=CharacterData>data</a>.
+
+ <li><p>If <var>result</var> is an error, then return.
+
+ <li><p>For each <a>pseudo-attribute</a> <var>pseudoAttr</var> of <var>result</var>:
+ <a for=map>set</a> <var>pi</var>'s
+ <a for=ProcessingInstruction>attribute map</a>[<var>pseudoAttr</var>'s [=pseudo-attribute/name=]]
+ to <var>pseudoAttr</var>'s [=pseudo-attribute/value=].
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn>update data from attributes</dfn>, given a {{ProcessingInstruction}} <a for=/>node</a>
+<var>pi</var>:
+
+<ol>
+ <li><p>Let <var>data</var> be the empty string.
+
+ <li>
+  <p><a for=map>For each</a> <var>name</var> → <var>value</var> of <var>pi</var>'s
+  <a for=ProcessingInstruction>attribute map</a>:
+
+  <ol>
+   <li><p>If <var>data</var> is not the empty string, then append U+0020 SPACE to <var>data</var>.
+
+   <li><p>Append <var>name</var> to <var>data</var>.
+
+   <li><p>Append U+003D (=) to <var>data</var>.
+
+   <li><p>Append U+0022 (") to <var>data</var>.
+
+   <li><p>Replace any U+0026 (&amp;) in <var>value</var> with "<code>&amp;amp;</code>".
+
+   <li><p>Replace any U+003C (&lt;) in <var>value</var> with "<code>&amp;lt;</code>".
+
+   <li><p>Replace any U+003D (&gt;) in <var>value</var> with "<code>&amp;gt;</code>".
+
+   <li><p>Replace any U+0022 (") in <var>value</var> with "<code>&amp;quot;</code>".
+
+   <li><p>Append <var>value</var> to <var>data</var>.
+
+   <li><p>Append U+0022 (") to <var>data</var>.
+  </ol>
+
+ <li><p><a>Replace data</a> of <var>pi</var> with 0, <var>pi</var>'s <a for=Node>length</a>,
+ <var>data</var>, and true.
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn export>get a processing instruction attribute</dfn>, given a {{ProcessingInstruction}}
+<a for=/>node</a> <var>pi</var> and a string <var>name</var>, return <var>pi</var>'s
+<a for=ProcessingInstruction>attribute map</a>[<var>name</var>] <a for=map>with default</a> null.
+</div>
+
+<hr>
+
+<div algorithm>
 <p>The
 <dfn constructor for=ProcessingInstruction lt=ProcessingInstruction(target)><code>new ProcessingInstruction(<var>target</var>, <var>data</var>)</code></dfn>
 constructor steps are:
@@ -8690,90 +8775,6 @@ method steps are:
 method steps are to return true if <a>this</a>'s
 <a for=ProcessingInstruction>attribute map</a>[<var>name</var>] <a for=set>exists</a>; otherwise
 false.
-</div>
-
-<div algorithm>
-<p>To <dfn export for=ProcessingInstruction>initialize</dfn> a {{ProcessingInstruction}}
-<a for=/>node</a> <var>pi</var>, with <var>target</var> and <var>data</var>:
-
-<ol>
- <li><p>If <var>target</var> does not match the <code>[=XML/Name=]</code> production, then
- <a>throw</a> an "{{InvalidCharacterError!!exception}}" {{DOMException}}.
-
- <li><p>If <var>data</var> contains the string "<code>?></code>", then <a>throw</a> an
- "{{InvalidCharacterError!!exception}}" {{DOMException}}.
-
- <li><p>Set <var>pi</var>'s <a for=ProcessingInstruction>target</a> to <var>target</var>.
-
- <li><p>Set <var>pi</var>'s <a for=CharacterData>data</a> to <var>data</var>.
-
- <li><p><a>Update attributes from data</a> given <var>pi</var>.
-</ol>
-</div>
-
-<div algorithm>
-<p>To <dfn>update attributes from data</dfn>, given a {{ProcessingInstruction}} <a for=/>node</a>
-<var>pi</var>:
-
-<ol>
- <li><p><a for=map>Clear</a> <var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>.
-
- <li><p>Let <var>result</var> be the <a>parsing result</a> of invoking the
- <a>rules for parsing pseudo-attributes from a string</a> given <var>pi</var>'s
- <a for=CharacterData>data</a>.
-
- <li><p>If <var>result</var> is an error, then return.
-
- <li><p>For each <a>pseudo-attribute</a> <var>pseudoAttr</var> of <var>result</var>:
- <a for=map>set</a> <var>pi</var>'s
- <a for=ProcessingInstruction>attribute map</a>[<var>pseudoAttr</var>'s [=pseudo-attribute/name=]]
- to <var>pseudoAttr</var>'s [=pseudo-attribute/value=].
-</ol>
-</div>
-
-<div algorithm>
-<p>To <dfn>update data from attributes</dfn>, given a {{ProcessingInstruction}} <a for=/>node</a>
-<var>pi</var>:
-
-<ol>
- <li><p>Let <var>data</var> be the empty string.
-
- <li>
-  <p><a for=map>For each</a> <var>name</var> → <var>value</var> of <var>pi</var>'s
-  <a for=ProcessingInstruction>attribute map</a>:
-
-  <ol>
-   <li><p>If <var>data</var> is not the empty string, then append U+0020 SPACE to <var>data</var>.
-
-   <li><p>Append <var>name</var> to <var>data</var>.
-
-   <li><p>Append U+003D (=) to <var>data</var>.
-
-   <li><p>Append U+0022 (") to <var>data</var>.
-
-   <li><p>Replace any U+0026 (&amp;) in <var>value</var> with "<code>&amp;amp;</code>".
-
-   <li><p>Replace any U+003C (&lt;) in <var>value</var> with "<code>&amp;lt;</code>".
-
-   <li><p>Replace any U+003D (&gt;) in <var>value</var> with "<code>&amp;gt;</code>".
-
-   <li><p>Replace any U+0022 (") in <var>value</var> with "<code>&amp;quot;</code>".
-
-   <li><p>Append <var>value</var> to <var>data</var>.
-
-   <li><p>Append U+0022 (") to <var>data</var>.
-  </ol>
-
- <li><p><a>Replace data</a> of <var>pi</var> with 0, <var>pi</var>'s <a for=Node>length</a>,
- <var>data</var>, and true.
-</ol>
-</div>
-
-<div algorithm>
-<p>To <dfn export id=concept-pi-get-attribute>get a processing instruction attribute</dfn>, given a
-{{ProcessingInstruction}} <a for=/>node</a> <var>pi</var> and a string <var>name</var>, return
-<var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
-<a for=map>with default</a> null.
 </div>
 
 
--- a/dom.bs
+++ b/dom.bs
@@ -8738,7 +8738,7 @@ method steps are:
 
  <li><p>If <var>result</var> is an error, then return.
 
- <li><p>For each <a>pseudo-attribute</a> <var>pseudoAttr</var> of <var>result</var>,
+ <li><p>For each <a>pseudo-attribute</a> <var>pseudoAttr</var> of <var>result</var>:
  <a for=map>set</a> <var>pi</var>'s
  <a for=ProcessingInstruction>attribute map</a>[<var>pseudoAttr</var>'s [=pseudo-attribute/name=]]
  to <var>pseudoAttr</var>'s [=pseudo-attribute/value=].
--- a/dom.bs
+++ b/dom.bs
@@ -8636,8 +8636,6 @@ method steps are:
  <li><p>If <var>name</var> is not a <a>valid attribute local name</a>, then <a>throw</a> an
  "{{InvalidCharacterError!!exception}}" {{DOMException}}.
 
- <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
-
  <li><p><a for=map>Set</a> <a>this</a>'s
  <a for=ProcessingInstruction>attribute map</a>[<var>name</var>] to <var>value</var>.
 
@@ -8650,8 +8648,6 @@ method steps are:
 method steps are:
 
 <ol>
- <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
-
  <li><p><a for=map>Remove</a> <a>this</a>'s
  <a for=ProcessingInstruction>attribute map</a>[<var>name</var>].
 
@@ -8667,8 +8663,6 @@ method steps are:
 <ol>
  <li><p>If <var>name</var> is not a <a>valid attribute local name</a>, then <a>throw</a> an
  "{{InvalidCharacterError!!exception}}" {{DOMException}}.
-
- <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
 
  <li><p>Let <var>attributes</var> be <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>.
 
@@ -8693,14 +8687,9 @@ method steps are:
 
 <div algorithm>
 <p>The <dfn method for=ProcessingInstruction><code>hasAttribute(<var>name</var>)</code></dfn>
-method steps are:
-
-<ol>
- <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
-
- <li><p>Return true if <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
- <a for=set>exists</a>; otherwise false.
-</ol>
+method steps are to return true if <a>this</a>'s
+<a for=ProcessingInstruction>attribute map</a>[<var>name</var>] <a for=set>exists</a>; otherwise
+false.
 </div>
 
 <div algorithm>
@@ -8782,15 +8771,9 @@ method steps are:
 
 <div algorithm>
 <p>To <dfn export id=concept-pi-get-attribute>get a processing instruction attribute</dfn>, given a
-{{ProcessingInstruction}} <a for=/>node</a> <var>pi</var> and a string <var>name</var>:
-
-<ol>
- <li><p>Let <var>lowercaseName</var> be <var>name</var> in <a>ASCII lowercase</a>.
-
- <li><p>Return <var>pi</var>'s
- <a for=ProcessingInstruction>attribute map</a>[<var>lowercaseName</var>]
- <a for=map>with default</a> null.
-</ol>
+{{ProcessingInstruction}} <a for=/>node</a> <var>pi</var> and a string <var>name</var>, return
+<var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
+<a for=map>with default</a> null.
 </div>
 
 
--- a/dom.bs
+++ b/dom.bs
@@ -8733,7 +8733,7 @@ method steps are:
  <li><p>Let <var>document</var> be a new {{Document}} node whose <a for=Document>type</a> is
  "<code>html</code>".
 
- <li><p>Let <var>html</var> be the concatentation of "<code>&lt;html </code>",
+ <li><p>Let <var>html</var> be the concatenation of "<code>&lt;html </code>",
  <var>pi</var>'s <a for=CharacterData>data</a>, and "<code>&gt;</code>".
 
  <li><p><a href="https://html.spec.whatwg.org/#parse-html-from-a-string">Parse HTML from a string</a>
--- a/dom.bs
+++ b/dom.bs
@@ -8713,31 +8713,19 @@ false.
 <ol>
  <li><p><a for=map>Clear</a> <var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>.
 
- <li><p>Let <var>context</var> be the result of <a>creating an element</a> given
- <var>pi</var>'s <a>node document</a>, "<code>html</code>", and the <a>HTML namespace</a>.
+ <li><p>Let <var>document</var> be a new {{Document}} node whose <a for=Document>type</a> is
+ "<code>html</code>".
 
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -8518,17 +8527,222 @@ interface CDATASection : Text {
...
+  boolean toggleAttribute(DOMString name, optional boolean force);
+  boolean hasAttribute(DOMString name);
 };</pre>
 
+<dl class=domintro>
+ <dt><code><var ignore>pi</var> = new <a constructor lt="ProcessingInstruction()">ProcessingInstruction(<var>target</var> [, <var>data</var> = ""])</a></code>
+ <dd>Returns a new {{ProcessingInstruction}} <a for=/>node</a> whose
+ <a for=ProcessingInstruction>target</a> is <var>target</var> and
+ <a for=CharacterData>data</a> is <var>data</var>.
+</dl>
```
annevk, line 8584: This should be made complete.

dom.bs
```diff
@@ -8518,17 +8527,222 @@ interface CDATASection : Text {
 <pre class=idl>
 [Exposed=Window]
 interface ProcessingInstruction : CharacterData {
+  constructor(DOMString target, optional DOMString data = "");
```
annevk, line 8536: I initially thought this would be a bit ugly compared to using IDL records or some such so you can set the attributes directly, but I think overall it's probably okay if we have to keep around the dual-data-structure anyway.

dom.bs
```diff
@@ -8518,17 +8527,222 @@ interface CDATASection : Text {
...
+ <li><p><a for=map>Clear</a> <var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>.
+
+ <li><p>Let <var>context</var> be the result of <a>creating an element</a> given
+ <var>pi</var>'s <a>node document</a>, "<code>html</code>", and the <a>HTML namespace</a>.
+
+ <li><p>Let <var>markup</var> be the concatentation of "<code>&lt;attrs </code>",
+ <var>pi</var>'s <a for=CharacterData>data</a>, and
+ "<code>&gt;&lt;/attrs&gt;</code>".
+
+ <li><p>Let <var>fragment</var> be the result of invoking the <a>fragment parsing algorithm steps</a> with <var>context</var> and <var>markup</var>.
```
annevk: This looks like it exceeds a 100 columns.

dom.bs
```diff
@@ -8518,17 +8527,222 @@ interface CDATASection : Text {
...
+
+<ol>
+ <li><p>Let <var>data</var> be the empty string.
+
+ <li>
+  <p><a for=map>For each</a> <var>name</var> → <var>value</var> of <var>pi</var>'s
+  <a for=ProcessingInstruction>attribute map</a>:
+
+  <ol>
+   <li><p>If <var>data</var> is not the empty string, append U+0020 SPACE to <var>data</var>.
```
annevk: then append*

dom.bs
```diff
@@ -8518,17 +8527,222 @@ interface CDATASection : Text {
...
+ <li><p>If <a>this</a>'s <a for=Node>node document</a> is an <a>HTML document</a>, then set
+ <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
+
+ <li><p>Return true if <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
+ <a for=set>exists</a>; otherwise false.
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn>update attributes from data</dfn> for a {{ProcessingInstruction}} <a for=/>node</a> <var>pi</var>:
```
annevk: Are we sure this parser is compatible with how xsl-stylesheet is parsed today?

dom.bs
foolip: Done!

dom.bs
foolip: Yes, it was based on how this works in Chromium:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/processing_instruction.cc;l=120-146;drc=bc971aee7d328d1e886ba76adc540805ea480981

In WebKit it is similar:

https://github.com/WebKit/WebKit/blob/9af8ffff919a1a1478b8a43cef821adb49a4e6a1/Source/WebCore/dom/ProcessingInstruction.cpp#L99
https://github.com/WebKit/WebKit/blob/26d7991c939a41bbab5d01263900598b67220208/Source/WebCore/xml/parser/XMLDocumentParserLibxml2.cpp#L1556-L1574

In Gecko it's a little bit different, every attribute lookup uses an attribute parser and I can't tell if it exactly matches XML or not:
https://github.com/mozilla-firefox/firefox/blob/22d04b52b0eb8d9fa11bf8ede5ccc0243a07c5ba/dom/xml/XMLStylesheetProcessingInstruction.cpp#L92
https://github.com/mozilla-firefox/firefox/blob/afe51e48c9deecccbaf64a74d6f207456433df98/dom/xml/ProcessingInstruction.cpp#L77
https://github.com/mozilla-firefox/firefox/blob/22d04b52b0eb8d9fa11bf8ede5ccc0243a07c5ba/dom/base/nsContentUtils.cpp#L1842-L1943

The difference from what's already implemented is that we'd use the HTML parser in HTML documents, instead of unconditionally using the XML parser.

For compat, this difference means that existing content that creates a `<?xml-stylesheet href="..."?>` PI and moves it to an HTML document would take the HTML parser path instead of XML. For that to make a difference, I think that content would need to be invalid XML attribute syntax (like `<?xml-stylesheet href='...'?>`) and depends on it being broken (doing nothing) when being moved to HTML.

noamr: The main difference is what happens if `pi.data` is set after the PI is moved to an HTML document (or constructed with `createProcessingInstruction`). In the new scenario invalid XML yet valid HTML can be used in that `data` for the `xml-stylesheet` attributes.

foolip:
The approach here is to parse attributes from the PI data, and to keep data and attributes in sync whenever one of them changes. Alternatives considered:

**Attribute mode:** Treat `data` as legacy and switch to "attribute mode" whenever attributes are mutated, replacing `data` with the empty string and optionally making it readonly. Serialization would need to change to use the attributes in this mode. Not pursued because `setAttribute()` would have the non-obvious side effect of making future `data` mutations throw or (alternatively) have no effect on serialization. This option is only attractive if it could be the default, but I can't imagine that's web compatible.

**Just `data`:** Parse `data` whenever getting an attribute. This seems close to how Gecko handled attributes for `<?xml-stylesheet?>`. But updating `data` in `setAttribute()` requires a parsing all attributes. There would be no observable difference to the current keep-everything-in-sync approach.

**Attributes for HTML, `data` for XML:** Since PIs are new in HTML, we could have "attribute mode" in HTML and keep `data` behavior intact for XML. But this would make the APIs significantly different in HTML and XML documents, and it would leave `<?xml-stylesheet?>` unexplained.

dom.bs
```diff
@@ -8518,17 +8527,260 @@ interface CDATASection : Text {
...
+   <li><p>If <var>data</var> is not the empty string, then append U+0020 SPACE to <var>data</var>.
+
+   <li><p>Append <var>name</var> to <var>data</var>.
+
+   <li><p>Append U+003D (=) to <var>data</var>.
+
+   <li><p>Append U+0022 (") to <var>data</var>.
+
+   <li><p>Append the result of
+   <a href="https://html.spec.whatwg.org/#escapingString">escaping a string</a> given
```
foolip: This is because https://html.spec.whatwg.org/#escapingString isn't exported and obviously this needs to be fixed. I'm just not sure yet if the escaping should be the same for HTML and XML, or if it should be split like parsing.

dom.bs
```diff
@@ -6005,11 +6005,15 @@ method steps are:
  "<code>?></code>", then <a>throw</a> an
  "{{InvalidCharacterError!!exception}}" {{DOMException}}. <!-- Gecko does this. -->
 
- <li>Return a new {{ProcessingInstruction}}
+ <li>Let <var>pi</var> be a new {{ProcessingInstruction}}
```
annevk: Let's fix this algorithm to use `<li><p>` as well.

annevk: But even better. We should make this algorithm invoke "initialize a ProcessingInstruction node" that essentially does what this algorithm does except for allocating a new node. Then we can reuse it for the constructor and keep them synchronized indefinitely.

dom.bs
```diff
@@ -8518,17 +8527,260 @@ interface CDATASection : Text {
...
+<a for=ProcessingInstruction>attribute map</a>.
+</div>
+
+<div algorithm>
+<p>The <dfn method for=ProcessingInstruction><code>getAttribute(<var>name</var>)</code></dfn> method
+steps are:
+
+<ol>
+ <li><p>If <a>this</a>'s <a for=Node>node document</a> is an <a>HTML document</a>, then set
+ <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
```
annevk: Do we really need to do this? Even for elements we only do this for HTML elements, not all of them.

dom.bs
```diff
@@ -8518,17 +8527,260 @@ interface CDATASection : Text {
...
+
+ <li><p>Let <var>context</var> be the result of <a>creating an element</a> given
+ <var>pi</var>'s <a>node document</a>, "<code>html</code>", and the <a>HTML namespace</a>.
+
+ <li><p>Let <var>markup</var> be the concatentation of "<code>&lt;attrs </code>",
+ <var>pi</var>'s <a for=CharacterData>data</a>, and
+ "<code>&gt;&lt;/attrs&gt;</code>".
+
+ <li><p>Let <var>fragment</var> be the result of invoking the
+ <a>fragment parsing algorithm steps</a> with <var>context</var> and <var>markup</var>.
```
annevk: I don't think it's great that the parsing rules depend on the context document. I also don't see how this is handling errors in the XML case.

dom.bs
foolip: Done in the style of event's "initialize" which required disambiguation invocations of that.

Not setting target and data on creation does leave the two undefined until "initialize" is called since https://dom.spec.whatwg.org/#concept-pi-target and https://dom.spec.whatwg.org/#concept-cd-data don't have defaults. Saying that they default to the empty string would be weird in the case of `target` since that cannot be the empty string after creation+initialization.

If you think this is ugly we could instead make it a "validate target and data" algorithm that only has the exception-throwing bits.

dom.bs
foolip: We're not constrained by web compat, only by what we think would be most consistent and least surprising to developers.

This was based on [get an attribute by name](https://dom.spec.whatwg.org/#concept-element-attributes-get-by-name) and would make the behavior very close to element attributes, assuming that we use the HTML parser to parse attributes in HTML documents. That behavior is:
- In HTML documents, the parser lowercases attribute names, and `getAttribute()` etc. all lowercase the input before looking for an exact match.
- With a PI imported from an XML document into an HTML document, `getAttribute()` would be unable to get the value of a camel-cased attribute.

However, we don't have the *NS variants that avoid the lowercasing, which would make camel-cased attributes from an XML parser inaccessible if moved to an HTML document. That's not great.

The simplest fix I can see is to just never lowercase the input, at the expense of `pi.getAttribute('NAME')` working differently from `element.getAttribute('NAME')` in an HTML document.

We also have the option of always lowercasing attribute names. That's tied to the discussion of which parser to use, in the next comment.

dom.bs
foolip: XML errors would result in a "SyntaxError" exception per https://html.spec.whatwg.org/#xml-fragment-parsing-algorithm, and thus leave the PI without any attributes.

If you don't want this to depend on the document, we could use the HTML parser unconditionally, but that would mean that `<?xml-stylesheet href=style.css?>` starts working in XML, and might force non-browsers that process XML documents with such PIs to parse those attribute as HTML.

I think using a different parser depending on the document is the most consistent with existing APIs like `innerHTML`, and avoids affecting existing content as much as possible.

dom.bs
annevk: My preference is that we don't touch casing at all and that these thus become case-sensitive just as they would on SVG elements inside an HTML document.

dom.bs
annevk: But this is not a context where we can rethrow the exception in so we'd have to handle it. But I don't think we need to use the XML parser.

foolip: Oh, I see what you're saying.

If we don't use the XML parser, do you mean always use the HTML parser, or something else?

dom.bs
foolip: OK, I've pushed a change for that.

dom.bs
annevk: I was thinking that, but it always lowercases attributes so I don’t think we can without an additional flag or some such. Not sure what would be best.

noamr: Doesn't using the XML parser when it's an XML document solve it?
Sure it doesn't throw an XML parsing error, but it's not different from today when setting the `data` in an `xml-stylesheet` PI. If data doesn't resolve to attributes, it should remove all the pseudo-attributes in the XML case.

Using the context document for this makes attributes work for all practical purposes without changing current behaviour of XML PIs too much.

I don't entirely understand the reasoning behind opposing the "by context document" idea.

foolip: I've handled the exception now with "If this throws an exception, then return."

I'm definitely open to using a single parser or switching based on some other condition, but the dual constraints of preserving current XML behavior and allowing for HTML syntax in HTML led to the current approach.

dom.bs
```diff
@@ -8518,17 +8517,254 @@ interface CDATASection : Text {
...
+
+  <ol>
+   <li><p>Let <var>element</var> be the <a>first child</a> of <var>fragment</var>.
+
+   <li><p>Assert: <var>element</var> is an <a for=/>element</a> with <a for=Element>local name</a>
+   "<code>attrs</code>".
+
+   <li><p><a for=list>For each</a> <var>attribute</var> of <var>element</var>'s
+   <a for=Element>attribute list</a>, set <var>pi</var>'s
+   <a for=ProcessingInstruction>attribute map</a>[<var>attribute</var>'s <a for=Attr>local name</a>]
```
foolip: Using the local name here means ignoring the namespace, but with input like `<?target attr="value" xmlns:bla="..." bla:attr="blavalue"?>` in XML, the second attribute would clobber the first and the result would be a single attribute with value "blavalue". That seems worse than for example erroring if the namespace is not null.

This seems like something @zcorpan might have opinions on :)

dom.bs
annevk: I'm not sure why we need to preserve the error conditions for the XML case.

I don't think it's acceptable that the behavior of `new ProcessingInstruction(..., ...)` ends up depending on the global in which you invoke it. That's not good design.

dom.bs
annevk: What happens for these cases today with `xml-stylesheet`? We also have to think about what happens here when you serialize and parse again. This strikes me as another reason not to rely on the XML parser.

dom.bs
noamr: > I'm not sure why we need to preserve the error conditions for the XML case.
> 
> I don't think it's acceptable that the behavior of `new ProcessingInstruction(..., ...)` ends up depending on the global in which you invoke it. That's not good design.

But that is a the case for all the DOM constructors, no? It's a side effect of associating the node to the document associated with the global object.

I am not opposed to an alternative, but not sure how to deal with attribute name casing.

annevk: I'm not sure which constructor has such global-depending side effects on one of its arguments?

noamr: > I'm not sure which constructor has such global-depending side effects on one of its arguments?

One of the arguments? None. But perhaps we don't need a constructor or it doesn't have to have a data argument if the constructor is the main concern?

noamr: ... I guess the alternative is "Use HTML fragment parser with a case-sensitive attribute flag regardless of owner document". I'm also OK with that.

foolip: I guess the option that would be most consistent with elements is that we don't have a constructor at all and switch the parser based on the node document.

If we do want a parser (seems nice) then we could also consider an "XML parser-created" flag that we set and use the XML parser only in that case.

zcorpan: I think it's similar to `document.createElement()` case-folding in HTML but not in XML. Whether to support a constructor or not impacts DX a bit, but imo doesn't need to affect the logic for HTML vs XML parser.

Using XML parser in XML docs and HTML parser in HTML docs seems reasonable to me, if we want to keep `xml-stylesheet` behavior as-is. I assume some folks would be upset if browsers start to use an HTML parser to parse the `xml-stylesheet` pseudo-attributes in XML documents.

dom.bs
zcorpan: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/14573

I'd expect the qualified name, so that `bla:href` is not treated as the `href` pseudo-attribute.

`element.getAttribute()` also uses qualified name.

dom.bs
annevk: Well we are extending what processing instructions can do so I think we should do that consistently across the syntaxes. Having subtle differences depending on how you got your processing instruction or what document it is currently in is not great, even if there is some precedent with other nodes.

I don't think people will be upset that browsers support some new ways to spell `xml-stylesheet` pseudo-attributes. XSLT removal is the significant change there.

And `document.createElement()` is rather terrible for its lowercasing as well by the way. That behavior has costed us all a lot of time trying to get it right and nail it down.

dom.bs
annevk: I don't actually understand how this works with the XML parser. If you have `x:href` and `href` as pseudo-attributes in whatever order and no `xmlns` attribute, `x:href` should be an irrecoverable error, no?

foolip: Just `x:href` would be an error, yes, but if `xmlns:x="..."` is also used, it would parse.

@zcorpan thanks for the test case, that shows that the line I've commented on does need to change. We could use the qualified name, but we could just as well skip any attributes that aren't in the null namespace, possibly also making use of namespaces here non-conforming.

foolip: Huh, https://software.hixie.ch/utilities/js/live-dom-viewer/saved/14573 uses single quotes, which made me wonder. I put `<?xml-stylesheet href='style.css'?>` into an XHTML document and that already works in Chrome, Firefox and Safari.

Not sure why that works.

annevk: But `<?xml-stylesheet x:href="" href="..."?>` "works" fine today in browsers.

foolip: Ignore the "huh" about single quotes, that's valid XML syntax and I just never knew...

> But `<?xml-stylesheet x:href="" href="..."?>` "works" fine today in browsers.

Oh, I didn't realize that worked. It looks like the reason in Chromium is that this code path uses a SAX parser with a "start element" handler that just ignores namespaces. That's not possible to specify using the XML fragment parser like I attempted here.

One option is to measure how common this is and then change implementations to be stricter about this.

But this could all be moot if we don't use the XML parser at all, so let me try to summarize that discussion first.

foolip:
We need to decide which parser is used to parse attributes from `data`, and I think these are all of the options:

1. HTML or XML parser depending on the node document, similar to `innerHTML`. This was the initial state of this PR, via [fragment parsing algorithm steps](https://html.spec.whatwg.org/#fragment-parsing-algorithm-steps)
2. HTML parser everywhere except for `<?xml-stylesheet?>`, which would use the XML parser.
3. Always use the HTML parser
4. Always use the XML parser
5. A bespoke parser with whatever properties we'd like, possibly being more strict for `<?xml-stylesheet?>`.

Option 2 seems like the most plausible to me right now, if we decide to make `<?xml-stylesheet x:href="" href="..."?>` no longer work.

@annevk WDYT?

annevk:
I'd prefer 4-5 over 1-2 personally. And between 3-5 it's between 3 and 5. It will have to be somewhat custom I think whatever we pick because of casing and `:`.

I don't think the differences of 1 have served us well for other features. I don't see a real need to carry them over. 2 seems very weird.

I don't think we should care about the exact requirements of https://www.w3.org/TR/xml-stylesheet/ but it does expect `x:href="" href=""` to work. I think it would be somewhat odd to regress on that as that level of strictness is weird.

foolip:
Thanks @annevk!

Always using the HTML parser is straightforward in spec and implementation, the only thing that's not ideal about is that the more lenient parsing of `<?xml-namespace?>` may affect non-browser tooling. Basically what @zcorpan said, "I assume some folks would be upset if browsers start to use an HTML parser to parse the `xml-stylesheet` pseudo-attributes in XML documents."

With a bespoke parser, that's mostly straightforward with "[collect a sequence of code points](https://infra.spec.whatwg.org/#collect-a-sequence-of-code-points)", but is there a way to decode character entities in attribute values that doesn't duplicate all of the tokenizer states character entities? Or could we do something simpler that ends up diverging slightly from how regular attribute values are parsed?

annevk:
So the concrete worry is that people creating dedicated XML content today would start relying on more leniency in browsers and this would have downstream repercussions? I think that's an acceptable risk. We have evolved XML (with `<template>` and `<script>` for instance) and will continue to do so (with new processing instructions). Tooling will adapt as needed.

If we can't reuse the HTML parser directly or want to create our own, following what https://w3c.github.io/webvtt/#webvtt-cue-text-tokenizer did seems somewhat reasonable, although it would be better with a proper hook in HTML.

foolip:
I've changed this to always use the HTML parser now. The constraints of wanting to support `<?marker name=x>` (without quotes) and not wanting the APIs to depend on context only leaves option 3 and 5 from https://github.com/whatwg/dom/pull/1454#issuecomment-4034420660, and reusing the HTML parser is the simplest of the two.

On lowercasing, I think this behavior might still be surprising:

```javascript
pi = new ProcessingInstruction('target', 'FOO=v1');
pi.getAttribute('foo') == 'v1'; // normal in HTML, no surprise
pi.setAttribute('FOO', 'v2');
pi.getAttribute('foo') == 'v1' && pi.getAttribute('FOO') == 'v2'; // unlike HTML, but OK
pi.data += ' ';
pi.getAttribute('foo') == 'v1' && !pi.hasAttribute('FOO'); // collapsed after roundtrip, surprise?
```

We could (1) live with it (2) lowercase in all APIs or (3) use a bespoke parser that preserves attribute case. Any of these would be OK for me.

If we stick with the HMTL parser, there's also a layering question remaining. Both "Parse HTML from a string" and "escaping a string" would need to be exported in HTML for this to work as-is, but I think it would be cleaner to add and export "parse attributes" and "serialize attributes" to HTML for this purpose.

noamr:
> I've changed this to always use the HTML parser now. The constraints of wanting to support `<?marker name=x>` (without quotes) and not wanting the APIs to depend on context only leaves option 3 and 5 from [#1454 (comment)](https://github.com/whatwg/dom/pull/1454#issuecomment-4034420660), and reusing the HTML parser is the simplest of the two.
> 
> On lowercasing, I think this behavior might still be surprising:
> 
> ```
> pi = new ProcessingInstruction('target', 'FOO=v1');
> pi.getAttribute('foo') == 'v1'; // normal in HTML, no surprise
> pi.setAttribute('FOO', 'v2');
> pi.getAttribute('foo') == 'v1' && pi.getAttribute('FOO') == 'v2'; // unlike HTML, but OK
> pi.data += ' ';
> pi.getAttribute('foo') == 'v1' && !pi.hasAttribute('FOO'); // collapsed after roundtrip, surprise?
> ```
> 
> We could (1) live with it (2) lowercase in all APIs or (3) use a bespoke parser that preserves attribute case. Any of these would be OK for me.
> 
> If we stick with the HMTL parser, there's also a layering question remaining. Both "Parse HTML from a string" and "escaping a string" would need to be exported in HTML for this to work as-is, but I think it would be cleaner to add and export "parse attributes" and "serialize attributes" to HTML for this purpose.

Always lowercasing in the PI attribute API makes sense to me.

annevk:
Either everything needs to canonicalize to lowercase or we pass a flag to the HTML parser that tells it to preserve case (and don't do anything with casing in the API of course). It feels odd for it to lowercase, but I could live with it and it would certainly be easier. Whether it's surprising probably depends on the kind of web developer.

foolip:
I agree that the sensible options are to either lowercase everywhere or preserve case everywhere. I've gone with the former for now, but could live with the parser change too.

WebReflection:
> It feels odd for it to lowercase,

that's what any other attribute on HTML space does though, right?

```js
document.body.setAttribute('WUT', 123);
document.body.getAttribute('wut'); // "123"
```

**edit** I am not saying for developers it's OK to abuse lower-casing, I am just saying for "*some*" developer that breaks expectations

dom.bs
zcorpan: Hmmmm. Yeah I guess it would work to always use the HTML parser, even for `xml-stylesheet` in XML documents. There can still be a document conformance rule that `xml-stylesheet` PIs in XML documents conform to the XML syntax.

Without XSLT support, `xml-stylesheet` will only be useful for CSS, which is probably less interesting for non-browser XML tools.

zcorpan: If we only use the HTML parser, we can consider starting the parser in the right state and stopping parsing when it would switch to the data state.

zcorpan:
I'm OK with using the HTML parser and lowercasing in the API.

dom.bs
foolip: It's now using the HTML parser.

foolip:
I looked through what remains to be done here, and I think its's just to  export "Parse HTML from a string" and "escaping a string" in HTML or add appropriate hooks to invert the layering. If anyone knows of more, please comment.

dom.bs
```diff
@@ -8518,17 +8517,259 @@ interface CDATASection : Text {
...
+</div>
+
+<div algorithm>
+<p>The
+<dfn method for=ProcessingInstruction><code>setAttribute(<var>name</var>, <var>value</var>)</code></dfn>
+method steps are:
+
+<ol>
+ <li><p>If <var>name</var> is not a <a>valid attribute local name</a>, then <a>throw</a> an
+ "{{InvalidCharacterError!!exception}}" {{DOMException}}.
```
annevk, line 8722: I think for other maps (e.g., `DOMTokenList`) we validate this for each accessor. Should we do that here as well?

dom.bs
```diff
@@ -8518,17 +8517,259 @@ interface CDATASection : Text {
...
+  <p>If <var>attributes</var>[<var>name</var>] does not <a for=set>exist</a>:
+
+  <ol>
+   <li><p>If <var>force</var> is not given or is true, set <var>attributes</var>[<var>name</var>]
+   to the empty string, <a>update data from attributes</a> for <a>this</a>, and then return true.
+
+   <li><p>Return false.
+  </ol>
+
+ <li><p>Otherwise, if <var>force</var> is not given or is false, <a for=map>remove</a>
```
annevk: No need for Otherwise as we use early returns.

dom.bs
```diff
@@ -8518,17 +8517,259 @@ interface CDATASection : Text {
...
+
+<div algorithm>
+<p>To <dfn export for=ProcessingInstruction id=concept-pi-initialize>initialize</dfn> a
+{{ProcessingInstruction}} <a for=/>node</a> <var>pi</var> with <var>target</var> and
+<var>data</var>:
+
+<ol>
+ <li><p>If <var>target</var> does not match the <code>[=XML/Name=]</code> production, then
+ <a>throw</a> an "{{InvalidCharacterError!!exception}}" {{DOMException}}.
+ <!-- DOM3 does not check for "xml" -->
```
annevk: I think at this point we can remove the DOM3 comment.

dom.bs
```diff
@@ -8518,17 +8517,259 @@ interface CDATASection : Text {
...
+
+ <li><p>If <var>data</var> contains the string "<code>?></code>", then <a>throw</a> an
+ "{{InvalidCharacterError!!exception}}" {{DOMException}}. <!-- Gecko does this. -->
```
annevk: We can also drop the Gecko comment.

dom.bs
hsivonen: Producing `&nbsp;`, which isn't in https://www.w3.org/TR/xml-stylesheet/#NT-PredefEntityRef seems like a bad idea in the XML case at least.

dom.bs
noamr: I think it would be less surprising if we align with element attributes? (throw only on setters)

foolip: I've tried to match how `setAttribute()` works on element more than `DOMTokenList`, but can you explain what validation you're thinking of? `DOMTokenList`'s `contains()` doesn't do any validation, and the new `pi.hasAttribute()` also doesn't.

My intention is that no invalid data be able to be set on the map, so that the getters don't need any validation.

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+<ol>
+ <li><p>If <var>name</var> is not a <a>valid attribute local name</a>, then <a>throw</a> an
+ "{{InvalidCharacterError!!exception}}" {{DOMException}}.
+
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
+
+ <li><p>Let <var>attributes</var> be <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>.
+
+ <li>
+  <p>If <var>attributes</var>[<var>name</var>] does not <a for=set>exist</a>:
```
annevk: This is the wrong exist. You want for=map.

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+constructor steps are:
+
+<ol>
+ <li><p>Set <a>this</a>'s <a for=Node>node document</a> to <a>current global object</a>'s
+ <a>associated <code>Document</code></a>.
+
+ <li><p><a for=ProcessingInstruction>Initialize</a> <a>this</a> with <var>target</var> and
+ <var>data</var>.
+
+ <li><p><a>Update attributes from data</a> for <a>this</a>.
```
annevk: This is redundant with Initialize above.

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+<ol>
+ <li><p><a for=map>Clear</a> <var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>.
+
+ <li><p>Let <var>document</var> be a new {{Document}} node whose <a for=Document>type</a> is
+ "<code>html</code>".
+
+ <li><p>Let <var>html</var> be the concatenation of "<code>&lt;html </code>",
+ <var>pi</var>'s <a for=CharacterData>data</a>, and "<code>&gt;</code>".
+
+ <li><p><a href="https://html.spec.whatwg.org/#parse-html-from-a-string">Parse HTML from a string</a>
```
annevk: I'm worried that we'll forget about fixing this.

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+<ol>
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
+
+ <li><p>Return true if <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>[<var>name</var>]
+ <a for=set>exists</a>; otherwise false.
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn export for=ProcessingInstruction id=concept-pi-initialize>initialize</dfn> a
```
```suggestion
<p>To <dfn export for=ProcessingInstruction>initialize</dfn> a
```

annevk: Why does this need exporting? If HTML also needs to create processing instructions we should probably have a dedicated create, I think.

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+ <li><p><a>Update attributes from data</a> for <a>this</a>.
+</ol>
+</div>
+
 <p>{{ProcessingInstruction}} <a for=/>nodes</a> have an associated
 <dfn export id=concept-pi-target for=ProcessingInstruction>target</dfn>.
 
+<p>{{ProcessingInstruction}} <a for=/>nodes</a> have an associated
+<dfn export id=concept-pi-attribute-map for=ProcessingInstruction>attribute map</dfn>, which is a
+<a for=/>map</a>, initially empty.
```
annevk: We should move these two associated concepts to before the constructor.

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+
+ <li><p>Set <var>name</var> to <var>name</var> in <a>ASCII lowercase</a>.
+
+ <li><p>Let <var>attributes</var> be <a>this</a>'s <a for=ProcessingInstruction>attribute map</a>.
+
+ <li>
+  <p>If <var>attributes</var>[<var>name</var>] does not <a for=set>exist</a>:
+
+  <ol>
+   <li><p>If <var>force</var> is not given or is true, set <var>attributes</var>[<var>name</var>]
```
annevk: We should use `<a for=map>set</a>` here as well.

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+<p>To <dfn>update attributes from data</dfn> for a {{ProcessingInstruction}} <a for=/>node</a>
+<var>pi</var>:
+
+<ol>
+ <li><p><a for=map>Clear</a> <var>pi</var>'s <a for=ProcessingInstruction>attribute map</a>.
+
+ <li><p>Let <var>document</var> be a new {{Document}} node whose <a for=Document>type</a> is
+ "<code>html</code>".
+
+ <li><p>Let <var>html</var> be the concatenation of "<code>&lt;html </code>",
```
annevk: `<a for=string>concatenation</a>`

dom.bs
```diff
@@ -8518,17 +8517,258 @@ interface CDATASection : Text {
...
+ <var>pi</var>'s <a for=CharacterData>data</a>, and "<code>&gt;</code>".
+
+ <li><p><a href="https://html.spec.whatwg.org/#parse-html-from-a-string">Parse HTML from a string</a>
+ given <var>document</var> and <var>html</var>.
+
+ <li><p><a for=list>For each</a> <var>attribute</var> of <var>document</var>'s
+ <a>document element</a>'s <a for=Element>attribute list</a>, set <var>pi</var>'s
```
```suggestion
 <a>document element</a>'s <a for=Element>attribute list</a>: set <var>pi</var>'s
```

annevk: And I guess we should link set here as well? Might be good to do another audit.

dom.bs
annevk: You can consider this resolved. It seems good to be consistent with elements.

hsivonen:
So far, when we've made DOM APIs change behavior based on the HTMLness flag on the `Document`, it has caused issues.

I'm well aware that concerns about XML are unfashionable, but I think it's bad to risk creating a browser-XML dialect that would be incompatible with other XML just to be consistent with the rest of HTML parsing on the HTML side.

That is, I'm uneasy about broadering the pseudo-attribute syntax from https://www.w3.org/TR/xml-stylesheet/ on the XML side and I'm uneasy about making the HTML and XML sides differ, which logically results in considering the use of https://www.w3.org/TR/xml-stylesheet/ syntax on the HTML side as well.

How bad would that be? WebKit and Blink seem to delegate to an XML tokenizer. Gecko has a special-purpose tokenizer.

Looking at [Google HTML/CSS Style Guide](https://google.github.io/styleguide/htmlcssguide.html) and use cases, it seems to me that using HTML attribute tokenization instead of https://www.w3.org/TR/xml-stylesheet/ isn't really something that's needed from use cases at hand and would add support for things that are frowned upon anyway these days when writing HTML, so this seems to be a matter of "for consistency" with the rest of HTML.

If we believe that we need to re-use HTML attribute syntax "for consistency" with the full entity set and everything in HTML, I think it would be then less bad to make the behavior depend on HTMLness flag of the `Document` in order not to introduce a browser dialect of `xml-stylesheet` on the XML side than to introduce the use the HTML tokenizer to `xml-stylesheet` in XML, which would introduce a browser dialect of XML (for CSS even if XSLT goes away).

As for @annevk 's remark about WebVTT: In Gecko, the WebVTT tokenizer is implemented in JS and calls into HTML parsing in an expensive way to support the full HTML entity set. Reusing the WebVTT tokenizer on the spec level would not result in any implementation ease in Gecko.

Supporting the full HTML entity set and the way semicolon omission works is both in spec and implementation the hardest part. WebVTT delegates to HTML. If we want to support the full HTML entity set, we should delegate to the HTML parser outright (like the PR currently does) instead of doing the WebVTT thing of supposedly defining a simplified thing but then jumping into the hard part of the more complex thing both spec and implementation-wise anyway.

I think it's sad to have that complexity "for consistency". Perhaps "for consistency" is the right thing, but I still find it sad to have the complexity, especially for something that's ["Do not use" in the Google style guide](https://google.github.io/styleguide/htmlcssguide.html#Entity_References) and spec-wise a backward-compat feature in the sense that we're not extending the entity set.

noamr:
> So far, when we've made DOM APIs change behavior based on the HTMLness flag on the `Document`, it has caused issues.
> 
> I'm well aware that concerns about XML are unfashionable, but I think it's bad to risk creating a browser-XML dialect that would be incompatible with other XML just to be consistent with the rest of HTML parsing on the HTML side.
> 
> That is, I'm uneasy about broadering the pseudo-attribute syntax from https://www.w3.org/TR/xml-stylesheet/ on the XML side and I'm uneasy about making the HTML and XML sides differ, which logically results in considering the use of https://www.w3.org/TR/xml-stylesheet/ syntax on the HTML side as well.
> 
> How bad would that be? WebKit and Blink seem to delegate to an XML tokenizer. Gecko has a special-purpose tokenizer.
> 
> Looking at [Google HTML/CSS Style Guide](https://google.github.io/styleguide/htmlcssguide.html) and use cases, it seems to me that using HTML attribute tokenization instead of https://www.w3.org/TR/xml-stylesheet/ isn't really something that's needed from use cases at hand and would add support for things that are frowned upon anyway these days when writing HTML, so this seems to be a matter of "for consistency" with the rest of HTML.

Do you mean things like `<?start name=foo>`, which would be HTML-compliant but not XML-compliant?
I think it's more than consistency... it's keeping the HTML fault tolerance.

> 
> If we believe that we need to re-use HTML attribute syntax "for consistency" with the full entity set and everything in HTML, I think it would be then less bad to make the behavior depend on HTMLness flag of the `Document` in order not to introduce a browser dialect of `xml-stylesheet` on the XML side than to introduce the use the HTML tokenizer to `xml-stylesheet` in XML, which would introduce a browser dialect of XML (for CSS even if XSLT goes away).

I'm fine with that and that was the original PR. @annevk opposed this though?

> As for @annevk 's remark about WebVTT: In Gecko, the WebVTT tokenizer is implemented in JS and calls into HTML parsing in an expensive way to support the full HTML entity set. Reusing the WebVTT tokenizer on the spec level would not result in any implementation ease in Gecko.
> 
> Supporting the full HTML entity set and the way semicolon omission works is both in spec and implementation the hardest part. WebVTT delegates to HTML. If we want to support the full HTML entity set, we should delegate to the HTML parser outright (like the PR currently does) instead of doing the WebVTT thing of supposedly defining a simplified thing but then jumping into the hard part of the more complex thing both spec and implementation-wise anyway.
> 
> I think it's sad to have that complexity "for consistency". Perhaps "for consistency" is the right thing, but I still find it sad to have the complexity, especially for something that's ["Do not use" in the Google style guide](https://google.github.io/styleguide/htmlcssguide.html#Entity_References) and spec-wise a backward-compat feature in the sense that we're not extending the entity set.

I think people would find it surprising that parsing pseudo-attributes would work different from parsing attributes.
I personally prefer to use the document-flag thing.
But all that being said, this whole thing is not something I feel strongly about. Developers would adapt to whatever we decide here.
If you and @annevk can reach a consensus on either of theh 3 options, I'm OK to align.

annevk:
See https://github.com/whatwg/dom/pull/1454#issuecomment-4037041799 for my opinion on the various options.

noamr:
> So far, when we've made DOM APIs change behavior based on the HTMLness flag on the `Document`, it has caused issues.
> 
> I'm well aware that concerns about XML are unfashionable, but I think it's bad to risk creating a browser-XML dialect that would be incompatible with other XML just to be consistent with the rest of HTML parsing on the HTML side.
> 
> That is, I'm uneasy about broadering the pseudo-attribute syntax from https://www.w3.org/TR/xml-stylesheet/ on the XML side and I'm uneasy about making the HTML and XML sides differ, which logically results in considering the use of https://www.w3.org/TR/xml-stylesheet/ syntax on the HTML side as well.
> 
> How bad would that be? WebKit and Blink seem to delegate to an XML tokenizer. Gecko has a special-purpose tokenizer.
> 
> Looking at [Google HTML/CSS Style Guide](https://google.github.io/styleguide/htmlcssguide.html) and use cases, it seems to me that using HTML attribute tokenization instead of https://www.w3.org/TR/xml-stylesheet/ isn't really something that's needed from use cases at hand and would add support for things that are frowned upon anyway these days when writing HTML, so this seems to be a matter of "for consistency" with the rest of HTML.
> 
> If we believe that we need to re-use HTML attribute syntax "for consistency" with the full entity set and everything in HTML, I think it would be then less bad to make the behavior depend on HTMLness flag of the `Document` in order not to introduce a browser dialect of `xml-stylesheet` on the XML side than to introduce the use the HTML tokenizer to `xml-stylesheet` in XML, which would introduce a browser dialect of XML (for CSS even if XSLT goes away).
> 
> As for @annevk 's remark about WebVTT: In Gecko, the WebVTT tokenizer is implemented in JS and calls into HTML parsing in an expensive way to support the full HTML entity set. Reusing the WebVTT tokenizer on the spec level would not result in any implementation ease in Gecko.
> 
> Supporting the full HTML entity set and the way semicolon omission works is both in spec and implementation the hardest part. WebVTT delegates to HTML. If we want to support the full HTML entity set, we should delegate to the HTML parser outright (like the PR currently does) instead of doing the WebVTT thing of supposedly defining a simplified thing but then jumping into the hard part of the more complex thing both spec and implementation-wise anyway.
> 
> I think it's sad to have that complexity "for consistency". Perhaps "for consistency" is the right thing, but I still find it sad to have the complexity, especially for something that's ["Do not use" in the Google style guide](https://google.github.io/styleguide/htmlcssguide.html#Entity_References) and spec-wise a backward-compat feature in the sense that we're not extending the entity set.

I've discussed this with @foolip and @tunetheweb and I think the best course of action is to go back to using the XML parser here also for HTML (option 4 [here](https://github.com/whatwg/dom/pull/1454#issuecomment-4034420660). This is the strictest options and we can loosen things if there is a pushback later from web developers, which I think there won't be.

dom.bs
foolip: Obsoleted by 5b58de39413fcb5da1041f7ed9fe44526f538494 since we now use https://www.w3.org/TR/xml-stylesheet/#dt-parsing

dom.bs
foolip: fixed in d95cb3094aac4f0114313304617a3a455b67bb65 + ba3f8ac6d9c5eb45f9f2c85a5526ad9347cf9c21, and no other "set" in the diff should be `<a for=map>set</a>`.

dom.bs
foolip: Fixed in d276f20ea6b6b83877d409bb8b4bc19c400f2ada

dom.bs
foolip: Removed in 5b58de39413fcb5da1041f7ed9fe44526f538494.

dom.bs
foolip: fixed in f312bfd3d824191770745507949b1da1d93ce71a

dom.bs
foolip: Done in bbc6cbf52ec0e8a086a3bf4518cea6f3cdf88e78

dom.bs
foolip: Fixed in 4f28c0ab9a95d3aece5f6465b3adaec9979443dc.

dom.bs
foolip: Now we always use https://www.w3.org/TR/xml-stylesheet/#dt-parsing.

dom.bs
foolip: This was fixed in be2f1fae27ff217184bd4e61c82a7e7850ef6530 with bespoke escaping. Escaping `>` is important because of the HTML parser.

dom.bs
foolip: Made obsolete by 5b58de39413fcb5da1041f7ed9fe44526f538494.

dom.bs
```diff
@@ -8526,17 +8533,266 @@ interface CDATASection : Text {
...
+ <var>data</var>, and true.
+</ol>
+</div>
+
+<div algorithm>
+<p>To <dfn export id=concept-pi-get-attribute>get a processing instruction attribute</dfn>, given a
+{{ProcessingInstruction}} <a for=/>node</a> <var>pi</var> and a string <var>name</var>:
+
+<ol>
+ <li><p>Let <var>lowercaseName</var> be <var>name</var> in <a>ASCII lowercase</a>.
```
annevk: Why do we still lowercase?

foolip: We don't have to, but lowercasing in HTML nicely handles the "xml" target in any case, aligning with https://www.w3.org/TR/xml/#dt-pi.

But we could preserve case and do case-insensitive comparisons for "xml" and "xml-stylesheet". Would you prefer that?

noamr: Sorry this has been discussed in this thread, see https://github.com/whatwg/dom/pull/1454#issuecomment-4050736658. Not sure it's resolved. I'm personally OK with keeping it case sensitive but it seems a bit inconsistent with elements.

annevk: Clearly we need more testing. Currently `data:text/xml,<?xml-STYLESHEET href="data:text/css,test{background:lime}"?><test/>` is not green in browsers. But `data:text/xml,<?xml-stylesheet href="data:text/css,test{background:lime}"?><test/>` is.

Similarly, `data:text/xml,<?xml-stylesheet HREF="data:text/css,test{background:lime}"?><test/>` is not green.

So I think case-sensitive and case-sensitive comparisons are the way to go, except perhaps for `<?xml` if tests prove otherwise there.

(I only tested Firefox and Safari.)

foolip: `<?xml` would turn into a comment in the HTML parser, so that case should be fine.

foolip: Dropped in 96de387fe0fc89cda9d0e48426d7e666886ca9fb

dom.bs
```diff
@@ -8526,17 +8533,249 @@ interface CDATASection : Text {
...
+
+<div algorithm>
+<p>The <dfn method for=ProcessingInstruction><code>getAttributeNames()</code></dfn> method steps
+are to return the result of <a for=map>getting the keys</a> of <a>this</a>'s
+<a for=ProcessingInstruction>attribute map</a>.
+</div>
+
+<div algorithm>
+<p>The <dfn method for=ProcessingInstruction><code>getAttribute(<var>name</var>)</code></dfn> method
+steps are to return the result of <a lt="get a processing instruction attribute">getting a processing
```
annevk: No newlines inside phrasing-level elements. You also don't need `lt` here. "get" and "getting" will both work.

annevk (approved):
This looks good to me. You can consider WebKit supportive. I would like to see OP filled out though and in particular I'd like to be able to review the tests.

I'm still a bit on the fence about the `Name` production, but that can be changed in the future.

dom.bs
noamr: Fixed

noamr:
> This looks good to me. You can consider WebKit supportive. I would like to see OP filled out though and in particular I'd like to be able to review the tests.

OP is filled out and tests are ready for review.

zcorpan (approved):

WebReflection:
to whom it might concern, a quick polyfill for this is here https://github.com/ungap/processing-instruction/blob/main/index.js but I wonder just one thing: are attributes observed at the MutationObserver level too or it's just a side effect for the fact that the `data` changes?

I don't want to go too greedy at the MutationObserver level because between setting explicitly data and changing attributes is basically indistinguishable from a user-land prospective unless tons of slow boilerplate is implemented + I wonder if, separately, anyone would be interested in a quick TreeWalker that finds comment nodes and replaces these with processedInstructions before further usage (most notably the template one, but also `innerHTML` or similar operations), thank you and congrats for the merge and tests and whatnot.

noamr:
> to whom it might concern, a quick polyfill for this is here https://github.com/ungap/processing-instruction/blob/main/index.js but I wonder just one thing: are attributes observed at the MutationObserver level too or it's just a side effect for the fact that the `data` changes?

It's just data. PI attributes are not observed like element attributes. We can revise this in the future if there is a use case. 

> 
> I don't want to go too greedy at the MutationObserver level because between setting explicitly data and changing attributes is basically indistinguishable from a user-land prospective unless tons of slow boilerplate is implemented + I wonder if, separately, anyone would be interested in a quick TreeWalker that finds comment nodes and replaces these with processedInstructions before further usage (most notably the template one, but also `innerHTML` or similar operations), thank you and congrats for the merge and tests and whatnot.

WebReflection:
@noamr thanks! then the polyfill patches *if needed* and it provides as default export a `patch(node)` that transform all awkward comments added as ProcessInstruction via `innerHTML` and similar into proper pi nodes with attributes methods. Happy to improve/change/fix whatever you think is super off, considering `@ungap` philosophy is "*faster and covering 99% of use cases is better than bloat for 1:1 feature-parity*" but obviously if there's something *really wrong* I might revisit it.

WebReflection:
@noamr apologies for this last question: I've checked tests to shape the poly and what I haven't found in there is any test for `pi.attributes` as read-only accessor ... has this been discussed at all, would this ever land or ... *YAGNI* ? Thanks, that proxy/map/iterable thing would add bloat but I was curious to understand if it's needed (or will be needed) at all ... I am all for the *YAGNI* approach though, for what is worth it.

noamr:
> @noamr apologies for this last question: 

No apology needed and feel free for it to not be your last!

> I've checked tests to shape the poly and what I haven't found in there is any test for `pi.attributes` as read-only accessor ... has this been discussed at all, would this ever land or ... _YAGNI_ ? Thanks, that proxy/map/iterable thing would add bloat but I was curious to understand if it's needed (or will be needed) at all ... I am all for the _YAGNI_ approach though, for what is worth it.

I'm not opposed to it and I think it wasn't discussed. It can't be a `NamedNodeMap` because it doesn't have an element backing, so it would be slightly different in any case.
It's easy enough to add if there is popular demand.

WebReflection:
> It's easy enough to add if there is popular demand.

well ... it's a proxy on JS side that needs to be there due identity checks + it behaves like an Array but it's effectively more like an `arguments` reference so, in case it'll ever land (I hope no!) I'd be more than curious to see how easier/shaped that would be, so far so good though, less things to maintain/think about, thank you.

noamr:
> > It's easy enough to add if there is popular demand.
> 
> well ... it's a proxy on JS side that needs to be there due identity checks + it behaves like an Array but it's effectively more like an `arguments` reference so, in case it'll ever land (I hope no!) I'd be more than curious to see how easier/shaped that would be, so far so good though, less things to maintain/think about, thank you.

There are actually multiple examples of this in the web platform, it's effectively a [Map-like object](https://webidl.spec.whatwg.org/#idl-maplike)

WebReflection:
I have that in LinkeDOM, sure, and I've proposed one to Pyodide but it's a Proxy after all and requires "bloat" as polyfill ... I never said I cannot, I rather said "*happy to hear I don't need to*" 😉

PR #1388, opened 2025-07-02
Avoid passing null through to 'valid name prefix'

shannonbooth:
Fixes https://github.com/whatwg/dom/issues/1387

- [ ] At least two implementers are interested (and none opposed):
   * …
   * …
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/pull/53533
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: …
   * Gecko: [bugzilla.mozilla.org/show_bug.cgi?id=1773312](https://bugzilla.mozilla.org/show_bug.cgi?id=1773312)
   * WebKit: [bugs.webkit.org/show_bug.cgi?id=241419](https://bugs.webkit.org/show_bug.cgi?id=241419)
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

***
<a href="https://whatpr.org/dom/1388.html" title="Last updated on Jul 2, 2025, 1:37 PM UTC (ba9fc4b)">Preview</a> | <a href="https://whatpr.org/dom/1388/780b8dd...ba9fc4b.html" title="Last updated on Jul 2, 2025, 1:37 PM UTC (ba9fc4b)">Diff</a>

Issue #1387, opened 2025-07-02
Misalignment between spec and WPT for "validate and extract"


shannonbooth:
### What is the issue with the DOM Standard?

WPT has tests such as the following: https://github.com/web-platform-tests/wpt/blob/856c33e87e61f513ddc81e082f78b6d24027e83d/dom/nodes/Document-createElementNS.js#L69

Where qualified name in [validate and extract](https://dom.spec.whatwg.org/#validate-and-extract) (in this case, `":foo"`) ends up being something that begins with a colon.

As I understand, this means that when we strictly split that prefix on `":"`, `splitResult[0]` ends up being the empty string, and `localName` is `foo`.

Therefore when we hit step 5.

> 5. If prefix is not a [valid namespace prefix](https://dom.spec.whatwg.org/#valid-namespace-prefix), then [throw](https://webidl.spec.whatwg.org/#dfn-throw) an "[InvalidCharacterError](https://webidl.spec.whatwg.org/#invalidcharactererror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

Since prefix is the empty string, a DOMException is thrown as the empty string is not a valid namespace. Sidenote: This step also can result in "null" being passed into the AO which expects a string.

I can get all WPT tests passing if:

>  5. If prefix is not a valid namespace prefix, then throw an "InvalidCharacterError" DOMException.

Is changed to:

>  5. If prefix non-null and not the empty and and is not a valid namespace prefix, then throw an "InvalidCharacterError" DOMException.

And:

> 8. If prefix is non-null and namespace is null, then throw a "NamespaceError" DOMException.

Is changed to:

> 8. If prefix is non-null and prefix is not the empty string and namespace is null, then throw a "NamespaceError" DOMException.

shannonbooth:
I'm not 100% if my suggested change is correct or not from the divergence in browsers on WPT and what spec intent is, or if I should make a PR to change the corresponding WPTs instead

annevk:
I think we could make _prefix_ always be a string if we make step 5 a substep of 4. Then we don't need the non-null check, just a not the empty string check in 8.

cc @josepharhar

shannonbooth:
Ah, right!

[...]

PR #1135, opened 2022-12-12
Editorial: clarify namespace validate and extract

dlrobertson:
Clarify the usage of strictly split a string in the algorithm validate and extract defined for namespaces.

Fixes: https://github.com/whatwg/dom/issues/1134

***
<a href="https://whatpr.org/dom/1135.html" title="Last updated on Dec 13, 2022, 2:48 PM UTC (0242fae)">Preview</a> | <a href="https://whatpr.org/dom/1135/1aa6fb0...0242fae.html" title="Last updated on Dec 13, 2022, 2:48 PM UTC (0242fae)">Diff</a>

Issue #1134, opened 2022-12-12
namespaces: validate and extract uses the strictly split a string in a vague manner


dlrobertson:
Namespaces [validate and extract](https://dom.spec.whatwg.org/#validate-and-extract) uses strictly split a string in the following step:

> 5. If qualifiedName contains a U+003A (:), then [strictly split](https://infra.spec.whatwg.org/#strictly-split) the string on it and set prefix to the part before and localName to the part after.

[strictly-split](https://infra.spec.whatwg.org/#strictly-split) defines the result as a [list](https://infra.spec.whatwg.org/#lists), so "the part before" and "the part after" feels a bit vague. Due to the validation in step two and the branch condition the list will have two elements, but it would be a bit clearer to use list indexing.

dlrobertson:
To what extent is it helpful to be explicit about things? When thinking about a step like this, I tend to think of something like:


  5. If _qualifiedName_ contains a U+003A (:), then run the following steps:
    1.  Let _splitResult_ be the result of [strictly split](https://infra.spec.whatwg.org/#strictly-split) givent _qualifiedName_.
    2. Set _prefix_ to _splitResult_[0].
    3. Set _localName_ to _splitResult_[1].

Unwrapping things like this feels a bit verbose. Is there a better way?

annevk:
You can leave out " run the following steps".

@domenic did we start using

> Let [_prefix_, _localName_] be the result of ...

already?

Even without that though your version is clearer and I'd definitely support moving to that.

domenic:
We've been doing it as "Let (_prefix_, _localName_) be the result of..." but only for tuples, not lists. (And it's not documented anywhere.)

PR #1455, opened 2026-03-06
Stop using “strictly split” in “validate and extract”

### LINKED ISSUE #1453: Issue with validate and extract (foo:bar:baz inputs)

In #1387 we discussed the `foo::bar` case and how that would throw, but we did not discuss that for `foo:bar:baz` you end up with a prefix `foo` and local name `bar` and an ignored part `:baz`. That seems really bad?

I think if we always split on the first `:` things are mostly okay. You could still get some re-parsing weirdness in XML, but I'm pretty sure that already exists with HTML and XML nodes mixed.

This came up in https://github.com/WICG/sanitizer-api/issues/373.

cc @josepharhar @domenic @shannonbooth @evilpie @sideshowbarker @otherdaniel @zcorpan

sideshowbarker:
The _“validate and extract”_ algorithm currently uses _“strictly split”_ to split the qualified name on U+003A (:), which splits on every colon into an array, and takes only the first and second elements from the array. For a qualified name like `foo:bar:baz`, that produces the prefix `foo` and the local name `bar` — silently discarding the `:baz` part.

This change splits on the first colon only instead, so the local name is `bar:baz`.

This came up in https://github.com/WICG/sanitizer-api/issues/373. Fixes #1453.

   * WebKit ✅
   * Gecko
   * Chromium
   * https://github.com/web-platform-tests/wpt/pull/58331

   | Engine            |  Bug |
   | ------------- | ----------------------------------------------- |
   | Chromium     | https://issues.chromium.org/issues/490251709   |
   | Gecko            | https://bugzilla.mozilla.org/show_bug.cgi?id=2021558 |
   | WebKit          | [already implemented in `https://github.com/WebKit/WebKit/pull/56000`] |
   
### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -254,12 +254,9 @@ U+0000 NULL, or U+003E (>).
   <p>If <var>qualifiedName</var> contains a U+003A (:):
 
   <ol>
-   <li><p>Let <var>splitResult</var> be the result of running <a>strictly split</a> given
-   <var>qualifiedName</var> and U+003A (:).
+   <li><p>Set <var>prefix</var> to the part of <var>qualifiedName</var> before the first U+003A (:).
 
-   <li><p>Set <var>prefix</var> to <var>splitResult</var>[0].
-
-   <li><p>Set <var>localName</var> to <var>splitResult</var>[1].
+   <li><p>Set <var>localName</var> to the part of <var>qualifiedName</var> after the first U+003A (:).
 
    <li><p>If <var>prefix</var> is not a [=valid namespace prefix=], then [=throw=] an
    "{{InvalidCharacterError}}" {{DOMException}}.


annevk:
This looks good, but we should use the PR template for this change given that we know Chromium and Gecko are doing this incorrectly, right?

sideshowbarker:
> we should use the PR template for this change given that we know Chromium and Gecko are doing this incorrectly, right?

Yup — updated

josepharhar:
Looks good to me, hopefully there are no compat issues we run into. @otherdaniel

annevk:
@josepharhar are we all going to try and deploy this and report back if it breaks? @smaug---- does that work for you? If so, we might not need agenda+.

sideshowbarker:
For the record here, what _“validate and extract”_ said before the patch associated with https://github.com/whatwg/dom/issues/1134 was this:

> If _qualifiedName_ contains a U+003A (:), then [strictly split](https://infra.spec.whatwg.org/#strictly-split) the string on it and set _prefix_ to the part before and _localName_ to the part after.

So, it was ambiguous — as https://github.com/whatwg/dom/issues/1134 pointed out. But, one of the two ways it _could_ have been read as intending is:

> If _qualifiedName_ contains a U+003A (:), then <strike>[strictly split](https://infra.spec.whatwg.org/#strictly-split) the string on it and</strike> set _prefix_ to the part before and _localName_ to the part after.

In other words, it _could_ have been read as actually intending what the change in this PR does. In other-other words, the change here is just reverting the spec back to what it had actually  originally been intended to specify.

And nobody in the https://github.com/whatwg/dom/issues/1134 discussion ever explicitly instead said anything like, _“**Throw away everything else in the name** after splitting it into a list and taking just the first two list items is clearly what we had meant to require here.”_

Instead, that’s just… effectively what it had ended up getting changed to state — without anybody ever having explicitly said there was some good reason for it needing to require throwing away everything else in the name.

josepharhar:
I thought that using strictly split meant to throw everything after the second colon away because of this discussion: https://github.com/whatwg/dom/issues/1387#issuecomment-3029579206

sideshowbarker:
> I thought that using strictly split meant to throw everything after the second colon away because of this discussion: https://github.com/whatwg/dom/issues/1387#issuecomment-3029579206

Dunno — but the PR which resulted in the local-name-computation part of the spec ending up as it is now, https://github.com/whatwg/dom/pull/1135, is from 2022. So it predates that https://github.com/whatwg/dom/issues/1387#issuecomment-3029579206 discussion (~8 months ago) by more than 2 years. The only actual spec change made in response to that https://github.com/whatwg/dom/issues/1387#issuecomment-3029579206 discussion was https://github.com/whatwg/dom/pull/1388/changes — which was, **“If _prefix_ is not a _valid namespace prefix_, then throw”** (no new effect on the requirements about the local name).

smaug----:
> @smaug---- does that work for you? If so, we might not need agenda+.

Sounds good to me

PR #1462, opened 2026-04-16
Record task-like timing in fire an event

noamr:
This allows clustering a "fire an event" timeframe in terms of frame timing monitoring, as opposed to relying on an "implicit document", which is a deprecated and non-interoperable term.

The implication of this is that if this event firing schedules a render that ends up as a long animation frame, the whole event firing duration would be counted towards this performance entry, regardless of "tasks".

This is already how long animation frames behave in the chromium implementation and should be unobservable from the previous spec wording that relied on "implicit document".

See w3c/long-animation-frames#36

   * chromium (matches current implementation)
   * Gecko (see https://github.com/whatwg/dom/pull/1462#pullrequestreview-5100389918)
   * This doesn't change current behavior, but rather tightens the spec. current tests are in https://wpt.fyi/results/long-animation-frames?
   * N/A (no behavior change)

***
<a href="https://whatpr.org/dom/1462.html" title="Last updated on May 21, 2026, 3:36 PM UTC (525c226)">Preview</a> | <a href="https://whatpr.org/dom/1462/2a4363c...525c226.html" title="Last updated on May 21, 2026, 3:36 PM UTC (525c226)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -1748,6 +1748,7 @@ initialized, and a <var>legacy target override flag</var>, run these steps:
 <ol>
  <li><p>Let <var>recordingTimingAsTask</var> be the result of calling <a>start recording task time if
  needed</a> given <var>target</var>'s <a>relevant settings object</a>.
+
  <li><p>If <var>eventConstructor</var> is not given, then let <var>eventConstructor</var> be
  {{Event}}.
 
--- a/dom.bs
+++ b/dom.bs
@@ -1770,7 +1770,7 @@ initialized, and a <var>legacy target override flag</var>, run these steps:
   <p>If <var>recordingTimingAsTask</var> is true:
 
   <ol>
-   <li><p>Assert</a>: <var>target</var>'s <a>relevant settings object</a> is a {{Document}}.
+   <li><a>Assert</a>: <var>target</var>'s <a>relevant settings object</a> is a {{Document}}.
 
    <li><p><a>Record task end time</a> given the <a>unsafe shared current time</a> and <var>target</var>'s
    <a>relevant settings object</a>.
--- a/dom.bs
+++ b/dom.bs
@@ -1746,6 +1746,8 @@ optionally using an <var>eventConstructor</var>, with a description of how IDL a
 initialized, and a <var>legacy target override flag</var>, run these steps:
 
 <ol>
+ <li><p>Let <var>recordingTimingAsTask</var> be the result of calling <a>start recording task time if
+ needed</a> given <var>target</var>'s <a>relevant settings object</a>.
  <li><p>If <var>eventConstructor</var> is not given, then let <var>eventConstructor</var> be
  {{Event}}.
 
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -1760,8 +1762,19 @@ initialized, and a <var>legacy target override flag</var>, run these steps:
 
   <p class=note>This also allows for the {{Event/isTrusted}} attribute to be set to false.
 
- <li><p>Return the result of <a>dispatching</a> <var>event</var> at <var>target</var>, with
+ <li><p>Let <var>result</var> be the result of <a>dispatching</a> <var>event</var> at <var>target</var>, with
  <var>legacy target override flag</var> set if set.
+
+ <li><p>If <var>recordingTimingAsTask</var> is true, then:
```
```suggestion
 <li>
  <p>If <var>recordingTimingAsTask</var> is true:
```

dom.bs
```diff
@@ -1746,6 +1746,8 @@ optionally using an <var>eventConstructor</var>, with a description of how IDL a
 initialized, and a <var>legacy target override flag</var>, run these steps:
 
 <ol>
+ <li><p>Let <var>recordingTimingAsTask</var> be the result of calling <a>start recording task time if
+ needed</a> given <var>target</var>'s <a>relevant settings object</a>.
```
annevk: Newline after this.

annevk (commented):
OP isn't filled out.

dom.bs
noamr: Done

WebReflection: puzzled by the fact there is no linter around these kind of exchange that would make things obvious and failing for the proposer of the PR ... curious to know why is that in these days of AI capabilities, yet I don't need an answer, just suggesting nobody should waste time over these kind of gotchas in here, imho.

noamr: Agreed, we just talked about improving the tooling for these PRs, especially for the tedious parts. There were many efforts but all of them seemed to have reached some 80-90% marked and never made it to general use.

WebReflection: I guess my broader concern is that anyone following with interest a feature ends up reading exchanges around a missed new line ... which is, well, something 😁

noamr:
@smaug---- can I enlist mozilla as interested in this? This is done after our conversation to enable some bits of long animation frames for gecko.

dom.bs
```diff
@@ -1746,6 +1746,9 @@ optionally using an <var>eventConstructor</var>, with a description of how IDL a
...
```
annevk: Oh also this contains a newline inside phrasing

annevk (approved):
Looks good modulo nit. Can land once OP is done.

dom.bs
smaug----: How does this work when the called algorithm does https://w3c.github.io/long-animation-frames/#start-recording-task-time-if-needed:~:text=If%20settingsObject%20is%20not%20a%20Document

noamr: I opened https://github.com/w3c/long-animation-frames/pull/39 to fix that - the idea there is to skip recording this time if we're not in a window.

smaug---- (approved):
ok, thanks. With that Loaf fix, this should work. I'm sure implementations do then optimizations to not take extra timestamps, but that is implementation detail.

dom.bs
noamr:
```suggestion
 <li><p>Let <var>recordingTimingAsTask</var> be the result of calling
 <a>start recording task time if needed</a> given <var>target</var>'s
 <a>relevant settings object</a>.
```

noamr: Fixed

noamr:
> Looks good modulo nit. Can land once OP is done.

Let's land this? Thanks!

PR #1469, opened 2026-05-29
Add a BoundaryPointsRange mixin

annevk:
This allows AbstractRange to be empty so it can be the superclass of OpaqueRange (which won't expose its boundary points).

***
<a href="https://whatpr.org/dom/1469.html" title="Last updated on May 29, 2026, 11:37 AM UTC (98bee63)">Preview</a> | <a href="https://whatpr.org/dom/1469/0300825...98bee63.html" title="Last updated on May 29, 2026, 11:37 AM UTC (98bee63)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8683,13 +8683,7 @@ be between 0 and the <a>boundary point</a>'s <a for="boundary point">node</a>'s
 
 <pre class=idl>
 [Exposed=Window]
-interface AbstractRange {
-  readonly attribute Node startContainer;
-  readonly attribute unsigned long startOffset;
-  readonly attribute Node endContainer;
-  readonly attribute unsigned long endOffset;
-  readonly attribute boolean collapsed;
-};
+interface AbstractRange {};
 </pre>
 
 <p>Objects implementing the {{AbstractRange}} interface are known as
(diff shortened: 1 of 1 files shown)


stephanieyzhang:
lgtm - thanks for creating this! I'll rebase [#1404](https://github.com/whatwg/dom/pull/1404) on top once this lands.

fwiw, I grepped the WHATWG HTML source and csswg-drafts for `{{AbstractRange/...}}` refs that might break due to this rename and didn't find anything, but flagging just in case I missed something.

annevk:
Oh good call. Per https://dontcallmedom.github.io/webdex/a.html#AbstractRange%40%40%40%40interface and in particular https://dontcallmedom.github.io/webdex/c.html#collapsed@@AbstractRange@attribute this would break https://w3c.github.io/editing/docs/execCommand/#effectively-contained. Although arguably that is in violation of the rule that you cannot define an algorithm in terms of public API. cc @zcorpan 

I can create a patch for that tomorrow. I think the observable differences will be covered by the generated IDL tests, but I should also mention that in the eventual commit message for clarity.

annevk:
https://github.com/whatwg/dom/pull/1470

PR #1470, opened 2026-06-07
Prepare for OpaqueRange

annevk:
Introduce NodeRange between AbstractRange and StaticRange/Range and move *Container members there.

Also make it possible for NodeRange and a future OpaqueRange to return different values for *Offset and collapsed members by making them return the result of an internal algorithm.

We keep start/end/start node/etc. on AbstractRange as OpaqueRange will need those to be the same (it will be manipulated just like live ranges are). Technically OpaqueRange is a node range as well, this is just not exposed to the outside world.

***
<a href="https://whatpr.org/dom/1470.html" title="Last updated on Jun 7, 2026, 9:31 AM UTC (6d094cd)">Preview</a> | <a href="https://whatpr.org/dom/1470/0300825...6d094cd.html" title="Last updated on Jun 7, 2026, 9:31 AM UTC (6d094cd)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -8684,9 +8684,7 @@ be between 0 and the <a>boundary point</a>'s <a for="boundary point">node</a>'s
 <pre class=idl>
 [Exposed=Window]
 interface AbstractRange {
-  readonly attribute Node startContainer;
   readonly attribute unsigned long startOffset;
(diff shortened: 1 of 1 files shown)


Chanchaichoochuay007:
collection-find-named-element

PR #1474, opened 2026-06-24
Treat fragment parser documents as not connected

shannonbooth:
Matching the behaviour of engines. For: https://github.com/whatwg/html/issues/11023, but needs a small patch in HTML spec also, which I will raise if directionally this is okay.

   * Chromium (already implements)
   * Webkit (already implements)
   * Firefox (already implements)
   * https://github.com/web-platform-tests/wpt/blob/0d0249113dc429fbbb4fce7778caeceae0466739/domparsing/createContextualFragment.html#L85
   * https://github.com/web-platform-tests/wpt/blob/0d0249113dc429fbbb4fce7778caeceae0466739/html/semantics/forms/the-input-element/radio-disconnected-group-owner.html#L186
   * Chromium: N/A
   * Gecko: N/A
   * WebKit: N/A

***
<a href="https://whatpr.org/dom/1474.html" title="Last updated on Jun 24, 2026, 3:21 PM UTC (ea3c622)">Preview</a> | <a href="https://whatpr.org/dom/1474/cd961d3...ea3c622.html" title="Last updated on Jun 24, 2026, 3:21 PM UTC (ea3c622)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2445,7 +2445,8 @@ referred to as the <dfn export id=concept-light-tree>light tree</dfn>.
 itself.
 
 <p id=in-a-shadow-including-document>A <a for=/>node</a> is <dfn export>connected</dfn> if its
-<a>shadow-including root</a> is a <a for=/>document</a>.
+<a>shadow-including root</a> is a <a for=/>document</a> whose
+<a for=Document>temporary document for fragment parsing</a> is false.
 
 <h5 id=shadow-tree-slots>Slots</h5>
 
@@ -5576,8 +5577,9 @@ known as <dfn export id=concept-document lt="document">documents</dfn>.
 <dfn export for=Document id=concept-document-origin>origin</dfn> (an <a for=/>origin</a>),
 <dfn export for=Document id=concept-document-type>type</dfn> ("<code>xml</code>" or "<code>html</code>"),
 <dfn export for=Document id=concept-document-mode>mode</dfn> ("<code>no-quirks</code>", "<code>quirks</code>", or "<code>limited-quirks</code>"),
-<dfn export for=Document>allow declarative shadow roots</dfn> (a boolean), and
-<dfn export for=Document>custom element registry</dfn> (null or a {{CustomElementRegistry}} object).
+<dfn export for=Document>allow declarative shadow roots</dfn> (a boolean),
+<dfn export for=Document>custom element registry</dfn> (null or a {{CustomElementRegistry}} object), and
+<dfn export for=Document>temporary document for fragment parsing</dfn> (a boolean).
 [[!ENCODING]]
 [[!URL]]
 [[!HTML]]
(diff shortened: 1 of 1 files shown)


annevk:
I don't like changing the meaning of connected for this. It feels a bit like a layering violation. Also, as discovered in https://github.com/whatwg/html/issues/11669#issuecomment-3589383629 I don't think this will actually do the trick long term as not all browsers create an intermediate document.

I wonder if @noamr and @foolip have ideas on how to address this properly.

noamr:
> I don't like changing the meaning of connected for this. It feels a bit like a layering violation. Also, as discovered in [whatwg/html#11669 (comment)](https://github.com/whatwg/html/issues/11669#issuecomment-3589383629) I don't think this will actually do the trick long term as not all browsers create an intermediate document.
> 
> I wonder if @noamr and @foolip have ideas on how to address this properly.

I think a more consistent approach would be to insert elements directly to the `DocumentFragment` rather than inserting them to the inert document and then moving them to the fragment in bulk.
Then they wouldn't be "connected" in the first place and we don't need a DOM change.

shannonbooth:
Indeed it feels a bit yuck, though it doesn't seem possible to change in a nice way as long as connected is defined in terms of document and fragment parsing requires a document. If we can wire up fragment parsing without a document that indeed seems much nicer, I can try and investigate that approach (unless this is already being worked on based on what is mentioned above?)

noamr:
> Indeed it feels a bit yuck, though it doesn't seem possible to change in a nice way as long as connected is defined in terms of document and fragment parsing requires a document. If we can wire up fragment parsing without a document that indeed seems much nicer, I can try and investigate that approach (unless this is already being worked on based on what is mentioned above?)

An element inserted to a `DocumentFragment` is never `connected` so I think it would solve this.

Note that multiple in-flight and upcoming changes like https://github.com/whatwg/html/pull/11818 are doing this work through redirecting the insertion target (see https://html.spec.whatwg.org/multipage/parsing.html#appropriate-place-for-inserting-a-node). This would probably be the place to fix it as well. I would prefer if we do this after these in-flight changes are complete though.

WebReflection:
> An element inserted to a DocumentFragment is never connected so I think it would solve this.

if persistent document fragment will ever exist, the element is connected "*if the fragment is*" so please consider the fact in the future there could be live/persistent document fragments, thank you!

noamr:
> > An element inserted to a DocumentFragment is never connected so I think it would solve this.
> 
> if persistent document fragment will ever exist, the element is connected "_if the fragment is_" so please consider the fact in the future there could be live/persistent document fragments, thank you!

Sure, but in this case it wouldn't be that future type of fragment, but a normal disconnected fragment.

shannonbooth:
> > Indeed it feels a bit yuck, though it doesn't seem possible to change in a nice way as long as connected is defined in terms of document and fragment parsing requires a document. If we can wire up fragment parsing without a document that indeed seems much nicer, I can try and investigate that approach (unless this is already being worked on based on what is mentioned above?)
> 
> An element inserted to a `DocumentFragment` is never `connected` so I think it would solve this.
> 
> Note that multiple in-flight and upcoming changes like [whatwg/html#11818](https://github.com/whatwg/html/pull/11818) are doing this work through redirecting the insertion target (see https://html.spec.whatwg.org/multipage/parsing.html#appropriate-place-for-inserting-a-node). This would probably be the place to fix it as well. I would prefer if we do this after these in-flight changes are complete though.

Sounds good, I'll hold off!

PR #1477, opened 2026-06-30
Fix NodeIterator pre-remove steps for ancestors of root

### LINKED ISSUE #907: NodeIterator's pre-removing steps does not match browsers
The following is a reduced example from the [wpt](https://github.com/web-platform-tests/wpt/blob/master/dom/traversal/NodeIterator-removal.html), where "paras[0]" is the node being removed and "paras[0].firstChild" is the node used to create the NodeIterator.
```js
const div = document.createElement('div');
const p = document.createElement('p');
const t = p.appendChild(document.createTextNode('Hello'));
div.appendChild(p);

const iter = document.createNodeIterator(t);
div.removeChild(p);
```

The html structure is:
```html
<div>
  <p>Hello</p>
</div>
```

The NodeIterator's pre-removing steps:
1. The node being removed (`p`) is an inclusive ancestor of node iterator's reference (`t`), so this passes.
2. The node iterator's `pointer before reference` starts out as `true`, so this passes.
    1. There is no following node that is an inclusive descendant of node iterator's `root` (`t`) that is also not an inclusive descendant of the node being removed (`p`), which means next is null.
    2. Next is null, so we move to the next step.
    3. **Because next is null, node iterator's `pointer before reference` gets set to false.**
3. **The node being removed (`p`) has no previous sibling, so node iterator's `reference` gets set to the removed node's (`p`) parent, which is `div`.**

Following the spec, the NodeIterator gets changed to the following:
```js
NodeIterator {
  root: t,
  referenceNode: div,
  pointerBeforeReferenceNode: false
}
```

Browsers, however, seem to leave the NodeIterator unchanged, which matches the test expectations:
```js
NodeIterator {
  root: t,
  referenceNode: t,
  pointerBeforeReferenceNode: true
}
```

annevk:
Generalize the first early-return clause of the `NodeIterator` pre-remove steps so it bails when *toBeRemovedNode* is an inclusive ancestor of *nodeIterator*'s root, not only when it equals the root.

Implementations (Gecko, Blink, WebKit) and the existing [`NodeIterator-removal.html`](https://github.com/web-platform-tests/wpt/blob/master/dom/traversal/NodeIterator-removal.html) WPT test already do this.

Corresponding WPT PR: https://github.com/web-platform-tests/wpt/pull/60997.

Closes #907.

***
<a href="https://whatpr.org/dom/1477.html" title="Last updated on Jun 30, 2026, 3:12 PM UTC (4ece054)">Preview</a> | <a href="https://whatpr.org/dom/1477/729983a...4ece054.html" title="Last updated on Jun 30, 2026, 3:12 PM UTC (4ece054)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -10320,8 +10320,9 @@ given a {{NodeIterator}} object <var>nodeIterator</var> and <a for=/>node</a>
 
 <ol>
  <li><p>If <var>toBeRemovedNode</var> is not an <a for=tree>inclusive ancestor</a> of
- <var>nodeIterator</var>'s <a for=NodeIterator>reference</a>, or <var>toBeRemovedNode</var> is
- <var>nodeIterator</var>'s <a for=traversal>root</a>, then return.
+ <var>nodeIterator</var>'s <a for=NodeIterator>reference</a>, or <var>toBeRemovedNode</var> is an
+ <a for=tree>inclusive ancestor</a> of <var>nodeIterator</var>'s <a for=traversal>root</a>, then
+ return.
 
  <li>
   <p>If <var>nodeIterator</var>'s <a for=NodeIterator>pointer before reference</a> is true:


shannonbooth:
This does make the meat of the NodeIterator implementation match, but I believe we have one extra deviation in place to match the behaviour of other engines. Apparently this is actually done by Acid3 where it was originally found: https://github.com/web-platform-tests/wpt/blob/b62d7567fd2df9b5a40788a2e9a42f9de756fc21/acid/acid3/test.html#L297

Claude says this so I could try and get a summary:

> Under the literal spec, the filter callback in step 3.2 runs user JS that may remove nodes, and those pre-removing steps adjust the iterator's stored reference. But traverse is working on its local copy node, which the pre-removing steps can't touch — and then steps 4–5 overwrite the stored reference with that local anyway. So a literal implementation would (a) keep advancing from a node that may now be detached, and (b) clobber any retargeting the pre-removing steps performed.

> Fix: Have traverse operate on the iterator's live state instead of a snapshot, while still returning the node handed to the filter.

Here's a test I got it to make (for some reason it doesn't look like there is one in our tree):

```html
<script>
    test(() => {
        // Tree: root > [a, b, c]
        const root = document.createElement("div");
        const a = document.createElement("a-el");
        const b = document.createElement("b-el");
        const c = document.createElement("c-el");
        root.append(a, b, c);

        let removedDuringFilter = false;
        const it = document.createNodeIterator(root, NodeFilter.SHOW_ELEMENT, {
            acceptNode(node) {
                // Remove the node from the tree while it is being filtered, i.e. while
                // a traverse() is in flight. This fires the NodeIterator pre-removing
                // steps against the in-progress traversal pointer.
                if (node === b) {
                    b.remove();
                    removedDuringFilter = true;
                }
                return NodeFilter.FILTER_ACCEPT;
            }
        });

        // Walk down to `a`.
        println("1: " + it.nextNode().nodeName);
        println("2: " + it.nextNode().nodeName);

        // This advances to `b` and filters it; the filter removes `b`.
        const returned = it.nextNode();
        println("removed during filter: " + removedDuringFilter);

        // The value returned must be the node that was passed to the filter (`b`),
        // even though the traversal pointer was retargeted away from it.
        println("3 returned: " + (returned ? returned.nodeName : "null"));

        // The reference node must be retargeted to a still-attached node (`a`),
        // not left pointing at the removed `b`.
        println("3 referenceNode: " + it.referenceNode.nodeName);

        // Traversal continues sanely from the retargeted reference, reaching `c`
        // rather than getting stuck on the detached `b`.
        const next = it.nextNode();
        println("4 returned: " + (next ? next.nodeName : "null"));
    });
</script>

```

annevk:
Thanks, I'll put up a PR for that later today. That requires a bit more invasive change.

annevk:
https://github.com/whatwg/dom/pull/1479

PR #1478, opened 2026-07-01
Remove "allow declarative shadow roots"

noamr:
This is now a parser flag rather than a document associated boolean.

See whatwg/html#12624, also for checkboxes.

***
<a href="https://whatpr.org/dom/1478.html" title="Last updated on Jul 2, 2026, 12:26 PM UTC (5d9c5aa)">Preview</a> | <a href="https://whatpr.org/dom/1478/729983a...5d9c5aa.html" title="Last updated on Jul 2, 2026, 12:26 PM UTC (5d9c5aa)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -5017,8 +5017,7 @@ and an optional <a for=/>document</a> <dfn export for="clone a node"><var>docume
     <ol>
      <li><p>Set <var>copy</var>'s <a for=Document>encoding</a>, <a for=Document>content type</a>,
      <a for=Document>URL</a>, <a for=Document>origin</a>, <a for=Document>type</a>,
-     and <a for=Document>mode</a>, to those of
-     <var>node</var>.
+     and <a for=Document>mode</a>, to those of <var>node</var>.
 
      <li><p>If <var>node</var>'s <a for=Document>custom element registry</a>'s
      <a for=CustomElementRegistry>is scoped</a> is true, then set <var>copy</var>'s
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -5017,7 +5017,7 @@ and an optional <a for=/>document</a> <dfn export for="clone a node"><var>docume
     <ol>
      <li><p>Set <var>copy</var>'s <a for=Document>encoding</a>, <a for=Document>content type</a>,
      <a for=Document>URL</a>, <a for=Document>origin</a>, <a for=Document>type</a>,
-     <a for=Document>mode</a>, and <a for=Document>allow declarative shadow roots</a>, to those of
+     and <a for=Document>mode</a>, to those of
      <var>node</var>.
```
annevk: wrapping

I guess this also means there's a change in behavior here?

noamr: I don't think so because "allow declarative shadow roots" is only false for inert documents created for the purpose of fragment-parsing, and those are never cloned AFAICT.

In blink "allow declarative shadow roots" is a fragment parser config rather than something that stays on the document.

annevk: What about XMLHttpRequest? It's not setting that flag today and I think that was intentional. Do we have tests?

noamr: https://whatpr.org/html/12624/parsing.html#allow-declarative-shadow-roots is false by default, and https://xhr.spec.whatwg.org/#document-response uses the default settings, so I think spec-wise this is OK.

There is a test for this [here](https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/declarative/declarative-shadow-dom-opt-in.html#L121-L131).

noamr:
We still need this because of `document.open()`. Closing.

PR #1479, opened 2026-07-02
Handle NodeIterator node removal during filtering

annevk:
The `NodeIterator` `traverse` algorithm read the reference into local variables, ran the filter (which can execute author code), and then wrote those locals back. Filtering can remove nodes and run the `NodeIterator` pre-remove steps, but those adjusted only the stored reference, not the in-progress traversal position held in `traverse`'s locals. As a result a node removed during its own filtering could leave the reference pointing at a detached node, and traversal could continue from there.

This models a reference as a *node pointer* (a tuple of a node and a pointer before) and gives each `NodeIterator` a *candidate reference* node pointer that holds the in-progress traversal position. `traverse` advances the candidate reference, sets the reference to it only once a node is accepted, and returns the node passed to the filter. The pre-remove steps adjust both the reference and, while traversing, the candidate reference.

`referenceNode` continues to report the last accepted node during filtering.

Matches Gecko, Blink, and WebKit (originally found via Acid3).

Tests: https://github.com/web-platform-tests/wpt/pull/61035

***
<a href="https://whatpr.org/dom/1479.html" title="Last updated on Jul 3, 2026, 5:30 AM UTC (4f36f63)">Preview</a> | <a href="https://whatpr.org/dom/1479/c7beb2c...4f36f63.html" title="Last updated on Jul 3, 2026, 5:30 AM UTC (4f36f63)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -2366,7 +2366,7 @@ can be used to explore this matter in more detail.
 
  <p>Objects that implement {{Text}} sometimes implement {{CDATASection}}.
 
- <p>Thus, every [=node=]'s [=primary interface=] is one of: {{Document}}, {{DocumentType}},
+ <p>Thus, every [=/node=]'s [=primary interface=] is one of: {{Document}}, {{DocumentType}},
  {{DocumentFragment}}, {{ShadowRoot}}, {{Element}} or an inherited interface of {{Element}},
  {{Attr}}, {{Text}}, {{CDATASection}}, {{ProcessingInstruction}}, or {{Comment}}.
 </div>
@@ -2970,7 +2970,7 @@ optional boolean <dfn for="insert"><var>suppressObservers</var></dfn> (default f
   <a>post-connection steps</a> <em>while</em> we're traversing the <a>node tree</a>. This is because
   the <a>post-connection steps</a> can modify the tree's structure, making live traversal unsafe,
   possibly leading to the <a>post-connection steps</a> being called multiple times on the same
-  <a>node</a>.
+  <a for=/>node</a>.
  </li>
 
  <li>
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -10318,31 +10322,49 @@ filter matches any <a for=/>node</a>.
...
+
+<div algorithm>
+<p>To <dfn for=NodeIterator>adjust</dfn> a <a>node pointer</a> <var>nodePointer</var>, given a
+{{NodeIterator}} object <var>nodeIterator</var> and <a for=/>node</a> <var>toBeRemovedNode</var>,
+perform the following steps. They return a <a>node pointer</a>.
+
 <ol>
  <li><p>If <var>toBeRemovedNode</var> is not an <a for=tree>inclusive ancestor</a> of
- <var>nodeIterator</var>'s <a for=NodeIterator>reference</a>, or <var>toBeRemovedNode</var> is an
+ <var>nodePointer</var>'s <a for="node pointer">node</a>, or <var>toBeRemovedNode</var> is an
```
shannonbooth, line 10346: This is the same approach ladybird does to fix this scenario (factored slightly differently, but I think this is nicer). I don't have any ideas for a less invasive approach.

Though the thing I noticed adjusting the implementation to match this exactly is if I understand correctly, I think nodePointer needs to be defined as a variable here from `nodeIterator` since it's not passed through in "given"

annevk: I'm not sure I fully understand what you meant, but I made it more explicitly a variable of adjust. Does that help?

shannonbooth: Yup, sorry for being unclear thats what I meant

PR #1272, opened 2024-04-03
A clonable shadow tree is cloned in its entirety

annevk:
Tests: https://github.com/web-platform-tests/wpt/pull/44247.

Fixes #1249.

- [x] At least two implementers are interested (and none opposed):
   * Chromium
   * Gecko
   * WebKit
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * See above.
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: N/A
   * Gecko: ?
   * WebKit: ?
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: N/A
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

Issue #1249, opened 2024-01-29
How should `clonable` work in `cloneNode(false)`?


mfreed7:
This issue was [raised as part of the Chromium review](https://chromium-review.googlesource.com/c/chromium/src/+/5239277/comment/3d66eb39_1ccfcf33/) of the implementation of `clonable`. The question is what should happen here:

```javascript
const shadow = host.attachShadow({mode: 'open', clonable: true});
shadow.innerHTML = '<div><span>what about me?</span></div>`;
const clone = host.cloneNode(false);
```

What should happen to the shadow root? Since it is clonable, I'd expect `clone` to have a shadow root, with parameters that match `shadow`. But should the *content* of that shadow root be deep-cloned from `host`? Or should `clone.shadowRoot` be empty? Both of those feel slightly odd. I think I slightly prefer deep-cloning the shadow content, but I'm not sure.

@dbaron @annevk @saschanaz

saschanaz:
cc @avandolder

saschanaz:
We should probably follow the expectation when cloning UA-provided elements with shadow roots, which would be the option 1; deep cloning the shadow root. (If I'm not misunderstanding)

emilio:
They do not get cloned in Gecko (except for printing as a special case). A new shadow root is attached when the clone gets inserted into the document.

annevk:
Cloning the shadow tree completely makes sense to me. It does make me wonder if we at some point need to provide more cloning options.

cc @rniwa

mfreed7:
Sounds like 3 votes (including myself) for deep-cloning the shadow tree, and one for not cloning it. @emilio would you be ok changing to a deep clone behavior?

emilio:
Yeah, I don't mind either way, I'm just saying that that's not the built-in shadow tree behavior at least in Gecko.

Might be worth checking with @justinfagnani about the ergonomics of this for authors too.

E.g., code that does:

```js
connectedCallback() {
  this.attachShadow({ ... });
}
```

Or so will throw after cloning if we clone the shadow tree. Maybe fine? You kinda need to deal with that for moves inside the document.

So over-all I think given ^ I'm fine with cloning the shadow tree. But hopefully I'm not missing something.

saschanaz:
> Yeah, I don't mind either way, I'm just saying that that's not the built-in shadow tree behavior at least in Gecko.

Does anyone know what Blink and WebKit do?

mfreed7:
[...]

PR #1480, opened 2026-07-04
Restore deep cloning of shadow root contents.

pwombwell:
This restores the behaviour introduced by #1272.

After #1334, the shadow-root cloning step says to clone each child of the shadow root with `subtree` set to `subtree`. When the original call is `host.cloneNode(false)`, that causes each direct child of a clonable shadow root to be cloned shallowly.

That contradicts the behaviour agreed in #1249 / #1272 and the WPT `shadow-dom/shadow-root-clonable.html`, which expects a clonable shadow root to be cloned in its entirety even when the shadow host is shallow-cloned.

***
<a href="https://whatpr.org/dom/1480.html" title="Last updated on Jul 4, 2026, 6:12 PM UTC (27a12ac)">Preview</a> | <a href="https://whatpr.org/dom/1480/995b4c3...27a12ac.html" title="Last updated on Jul 4, 2026, 6:12 PM UTC (27a12ac)">Diff</a>


annevk (approved):
Thank you!

PR #1487, opened 2026-07-08
Let replaceChildren() replace a document's children

### LINKED ISSUE #1045: replaceChildren can not replace a Documents documentElement
Given a `document` that contains an element, and the following code: 

```
const fragment = document.createDocumentFragment();
const element = fragment.appendChild(document.createElement('element'));
document.replaceChildren(fragment);
```

Expected result: `document` now only contains `element`
Actual result: HierarchyRequestError

This is caused by replaceChildren calling "ensure pre-insertion validity without taking into consideration that the existing children of the document will be removed. This then fails on the following rule:

> Otherwise, if node (the fragment) has one element child and either parent (the document) has an element child, child is a doctype, or child is non-null and a doctype is following child.

Perhaps an argument should be added to make this algorithm ignore pre-existing children when called from replaceChild, or perhaps replaceChildren should use a modified copy of these steps, as was done in replaceChild?

annevk:
replaceChildren() removes all of the parent's children before inserting, but "ensure pre-insert validity" still counted them. So replacing a document's children, e.g. document.replaceChildren(element), threw a "HierarchyRequestError" DOMException even though the result would be a valid document.

Generalize the childToExclude concept (added for replace()) into a list of children to exclude from the validity checks, and have replaceChildren() exclude all of the parent's children. As the child being replaced is now handled uniformly, the isReplace boolean is gone.

Fixes #1045.

   * Gecko
   * WebKit
   * https://github.com/web-platform-tests/wpt/pull/61154 
   * Chromium: Captcha :-( 
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=2053747
   * WebKit: https://github.com/WebKit/WebKit/pull/68990

***
<a href="https://whatpr.org/dom/1487.html" title="Last updated on Jul 9, 2026, 12:14 PM UTC (72633f0)">Preview</a> | <a href="https://whatpr.org/dom/1487/c7f757c...72633f0.html" title="Last updated on Jul 9, 2026, 12:14 PM UTC (72633f0)">Diff</a>


smaug----:
Yeah, the change seems reasonable.

PR #1492, opened 2026-07-11
Copy a document's base URL when cloning

### LINKED ISSUE #454: Document's base URL needs to be cloned
Per https://github.com/whatwg/html/issues/421 we need to clone the base URL. #258 made this a little harder. Might want to revert that after all...

annevk:
Copy a document's base URL when cloning.

The baseURI getter now defers to a new "get the base URL" algorithm, which returns null by default and falls back to the node document's URL. HTML overrides this algorithm (to return the document base URL), so DOM no longer reaches into HTML for a value returned by a DOM API.

Cloning a document now copies its base URL into a new "base URL override", so that a shallow clone, which has no base element, still reports the base URL of the document it was cloned from.

HTML PR: https://github.com/whatwg/html/pull/12674

Fixes #454.

Tests: https://github.com/web-platform-tests/wpt/pull/61232

   * Gecko (passes all tests)
   * WebKit
   * https://github.com/web-platform-tests/wpt/pull/61232 
   * Chromium: Captcha :-( 
   * Gecko: N/A
   * WebKit: …

***
<a href="https://whatpr.org/dom/1492.html" title="Last updated on Aug 27, 2026, 8:50 AM UTC (e0bd201)">Preview</a> | <a href="https://whatpr.org/dom/1492/8a5f57c...e0bd201.html" title="Last updated on Aug 27, 2026, 8:50 AM UTC (e0bd201)">Diff</a>

### DIFF
--- a/dom.bs
+++ b/dom.bs
@@ -4646,11 +4646,20 @@ statement, switching on the interface <a>this</a> <a>implements</a>:
 
 <dl class=domintro>
  <dt><code><var>node</var> . {{Node/baseURI}}</code>
- <dd>Returns <var>node</var>'s <a for=Node>node document</a>'s <a>document base URL</a>.
+ <dd>Returns <var>node</var>'s <a for=Node>node document</a>'s base URL.
 </dl>
 
-<p>The <dfn attribute for=Node><code>baseURI</code></dfn> getter steps are to return <a>this</a>'s
-<a for=Node>node document</a>'s <a>document base URL</a>, <a lt="URL serializer" spec=url>serialized</a>.
+<p>The <dfn attribute for=Node><code>baseURI</code></dfn> getter steps are:
+
+<ol>
+ <li><p>Let <var>baseURL</var> be the result of running <a>this</a>'s <a for=Node>node document</a>'s
+ <a>get the base URL</a> algorithm.
+
+ <li><p>If <var>baseURL</var> is null, then set <var>baseURL</var> to <a>this</a>'s
+ <a for=Node>node document</a>'s <a for=Document>URL</a>.
+
+ <li><p>Return <var>baseURL</var>, <a lt="URL serializer" spec=url>serialized</a>.
+</ol>
 
 <hr>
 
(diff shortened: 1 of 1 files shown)


dom.bs
```diff
@@ -5638,6 +5650,16 @@ known as <dfn export id=concept-document lt="document">documents</dfn>.
 Whether a <a for=/>document</a> is an <a>HTML document</a> or an <a>XML document</a> affects the
 behavior of certain APIs.
 
+<p>Each <a for=/>document</a> has an associated <dfn export>get the base URL</dfn> algorithm, which
+returns a <a for=/>URL</a> or null. Unless specified otherwise it returns null.
```
domfarolino, line 5654: Should we make the overridability of these steps more explicit, like https://dom.spec.whatwg.org/#concept-node-insert-ext does?

dom.bs
```diff
@@ -5638,6 +5650,16 @@ known as <dfn export id=concept-document lt="document">documents</dfn>.
...
+
+<p>Each <a for=/>document</a> has a <dfn export>base URL override</dfn> (a <a for=/>URL</a> or
+null), initially null.
+
+<p class=note>HTML overrides the <a>get the base URL</a> algorithm to account for the
+<code>base</code> element, and sets the <a>base URL override</a> for "<code>about:</code>"-schemed
+<a for=/>documents</a>. [[!HTML]]
```
domfarolino: Maybe let's also mention that this is also set (by this spec of course, not HTML) during document cloning. I was initially confused why we didn't retain its original descriptive name, until I realized that's because this PR assigns it in new, non-`about:` cases.

domfarolino (commented):
Please link to https://github.com/whatwg/html/pull/12674 from the OP.

dom.bs
annevk: I prefer this style slightly. This is also what we do for "get the parent". In general I think we have to redo the insertion steps at some point due to ordering issues.

PR #1400, opened 2025-08-19
Remove mentions of XSLT from the DOM spec

mfreed7:
This goes along with https://github.com/whatwg/html/pull/11563.

Closes https://github.com/whatwg/dom/issues/181.

- [X] At least two implementers are interested (and none opposed):
   * Chromium - supportive
   * Gecko - supportive
   * WebKit - supportive
- [X] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://chromium-review.googlesource.com/c/chromium/src/+/7027778
- [X] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: https://crbug.com/435623334
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1990759
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=300785
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …
- [X] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

Corresponding HTML spec PR: https://github.com/whatwg/html/pull/11563

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

***

### :boom: Error: 422 Unprocessable Entity :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 3, 2026, 8:46 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/publications/spec-generator/) - Spec Generator is the web service used to build bikeshed/ReSpec specs

:link: [Related URL](https://www.w3.org/publications/spec-generator/?type=bikeshed-spec&output=html&url=https%3A%2F%2Fraw.githubusercontent.com%2Fmfreed7%2Fdom%2F1ffe24652d958826688f45b7caa614127a2395f5%2Fdom.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%201400)

**Error output:**

```json
[
    {
        "lineNum": null,
        "messageType": "warning",
[...]

PR #1499, opened 2026-08-20
Mark XSLT deprecated

mfreed7:
   * Chromium (https://developer.chrome.com/docs/web-platform/deprecating-xslt)
   * Gecko (https://github.com/mozilla/standards-positions/issues/1287)
   * WebKit (https://github.com/whatwg/html/issues/11523#issuecomment-3149280766)
   * N/A (this is for deprecation, not removal)
   * Chromium: https://crbug.com/435623334
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1990759
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=300785
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/42190

See also:
- https://github.com/whatwg/html/pull/12805
- https://github.com/whatwg/html/pull/11563
- https://github.com/whatwg/dom/pull/1400

***
<a href="https://whatpr.org/dom/1499.html" title="Last updated on Aug 24, 2026, 11:43 PM UTC (4c1abff)">Preview</a> | <a href="https://whatpr.org/dom/1499/94d41b3...4c1abff.html" title="Last updated on Aug 24, 2026, 11:43 PM UTC (4c1abff)">Diff</a>


keithamus (approved):

dom.bs
```diff
@@ -11449,6 +11449,9 @@ methods on {{Document}}.
 
 <h2 id=xslt>XSLT</h2>
 
+<p class=XXX><strong>Note:</strong> XSLT is <strong>deprecated</strong>, and its usage is
```
annevk: Why "Note:" and why not use the same warning class as you did in HTML?

mfreed7: Yeah, not sure really. I was trying to make it stand out a bit more, like the `class="warning"` version in HTML did.

Anyway, I've copy/pasted over the (new) version of this text, changed to `class="critical"`, and removed "Note".

dom.bs
```diff
@@ -11449,6 +11449,12 @@ methods on {{Document}}.
 
 <h2 id=xslt>XSLT</h2>
 
+<p class="critical">Avoid using XSLT. Its implementation in engines has fallen decades behind
```
annevk: I don't think it has fallen decades behind? XSLT 1.0 is still a standard.

mfreed7: Well, it's a [standard from 1999](https://www.w3.org/TR/xslt-10/), which has been superseded by [two new major releases](https://www.w3.org/TR/xslt/) ([2.0 in 2009](https://www.w3.org/TR/2009/PER-xslt20-20090421/), and [3.0 in 2017](https://www.w3.org/TR/xslt-30/)) with a [4.0 version in the works since 2020](https://archive.xmlprague.cz/2020/files/xmlprague-2020-proceedings.pdf#page=121). Browsers, to my knowledge, only implement version 1.0, from 27 years ago. So at least in my personal view, we're decades behind the latest standard.

Having said all of that, I really don't mind changing the language completely - just let me know what you'd be comfortable saying. I can even go back to just saying "it's deprecated" without the reason. But the other example @zcorpan [pointed me to](https://github.com/whatwg/html/pull/12805#issuecomment-5368161294) included a motivation, so I was just trying to follow that pattern.

mfreed7: Here's a proposed replacement (thanks @zcorpan). @annevk let me know what you think:

```
<p class="critical">Avoid using XSLT. The browser implementations of XSLT are highly susceptible
to memory safety vulnerabilities, and the web platform has evolved toward JavaScript as the safe
replacement. For these reasons, this feature is in the process of being removed from the web
platform. (This is a long process that takes many years.)</p>
```

mfreed7: (I went ahead and just pushed that to both PRs, in case it works.)

dom.bs
```diff
@@ -11449,6 +11449,11 @@ methods on {{Document}}.
 
 <h2 id=xslt>XSLT</h2>
 
+<p class="critical">Avoid using XSLT. The browser implementations of XSLT are highly susceptible
```
sideshowbarker:
```suggestion
<p class="critical">Avoid using XSLT in frontend JavaScript running in browsers. The browser implementations of XSLT are highly susceptible
```
…or some other language to qualify that admonition. Otherwise, without qualifying that in some way, it could be misread as us admonishing people to avoid using XSLT _everywhere_ — including all the other off-web places outside browsers where people use it. (And given previous reactions we’ve experienced in this area, we can pretty much be guaranteed some folks are gonna misread it that way.)

zcorpan: `xml-stylesheet` is without JS. Maybe "client-side XSLT in browsers" or similar.

sideshowbarker: > `xml-stylesheet` is without JS. Maybe "client-side XSLT in browsers" or similar.

I think "client-side XSLT in browsers" would be fine — or any other similar formulation. I think just having _some_ kind of language there that qualifies the scope of the admonition is way more important than the accuracy/precision of whatever particular wording we end up with.

mfreed7: Thanks for the comment. I've changed to `client-side XSLT in browsers`. Once we've agreed on the language, I'll push the same text over to the HTML PR.

annevk (approved):

dom.bs
annevk: XSLT 1.0 is not superseded. At least there's nothing at https://www.w3.org/TR/xslt-10/ that says so. It's true that there are newer versions, but they didn't obsolete the old version. (It's unlike most web technologies in that respect.)

mfreed7: Oh interesting. Yeah, I just assumed that the newer versions "automatically" supersede the old ones, but I guess not. I notice that v2 just says "XSLT 2.0 is a revised version of the XSLT 1.0", and both 2.0 and 3.0 are made to be backwards compatible. Thanks for pointing that out.

Sounds like the new language will work. Thanks for merging this.

